PR #4, opened 2016-11-28
Don't write SCKDIV when flashing

mwachs5:
The target may have already configured its clock to run at a higher frequency and would have set SCKDIV and other dividers at that time. Don't restore the SCKDIV to its default or the flash interface may run too fast and programming will fail.

Otherwise, the default value is fine and there is no need to write SCKDIV.


mwachs5:
OK to merge?

timsifive:
As long as I can still flash my board, go ahead. I can't test it right now.

Tim

PR #9, opened 2016-12-18
riscv: Increase the number of Algorithm Steps

mwachs5:
I kept getting assertion failures that I was adding more steps than I had allocated for. I wasn't sure where the "/16" came from in your code... "/4" made more sense to me if you're writing one word at a time but not sure.

This fixes my immediate issue but not sure it's the correct fix.


timsifive:
/16 was just a guess that seemed to work for me. /4 probably ends up allocating more memory than necessary, but that is the only downside.

PR #11, opened 2016-12-24
Correct off by 1 in malloc

mwachs5:
which causes this to fail on macOS (and in theory on any platform).

This manifests itself as corrupted data while Flashing


timsifive:
Thanks, Megan! Sorry for leaving that in there... I wonder if it would help to add some valgrind coverage to the test suite. I know I've run it before and there are a number of warnings in non-riscv code.

PR #13, opened 2017-01-25
riscv: Globally disable interrupts when running algorithms.

mwachs5:
This works for flashing the SPI Flash when the running program has interrupts. I don't restore interrupts afterwards, but our SPI flashing code is completely trashing the program space anyway as it is now.

However, not sure this is the most general solution.

 Does it make more sense to disable and re-enable interrupts at a higher level? Like, can the fespi SPI Flashing code write the CSR somehow (I couldn't figure out how I can do something like "target_write_csr" from that code). Then it could re-enable interrupts after the code has been fully flashed.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2704,10 +2704,10 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 	LOG_DEBUG("Disabling Interrupts");
 	register_get(&target->reg_cache->reg_list[REG_MSTATUS]);
 	current_mstatus = buf_get_u64(target->reg_cache->reg_list[REG_MSTATUS].value, 0, info->xlen);
-	buf_set_u64(mstatus_bytes, 0, info->xlen, (current_mstatus & (~((uint64_t)MSTATUS_MIE))) );
+	uint64_t ie_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
+	buf_set_u64(mstatus_bytes, 0, info->xlen, set_field(current_mstatus, ie_mask, 0));
 
 	register_set(&target->reg_cache->reg_list[REG_MSTATUS], mstatus_bytes);
-	info->mstatus_actual = current_mstatus & ~MSTATUS_MIE;
 
 	/// Run algorithm
 	LOG_DEBUG("resume at 0x%x", entry_point);
@@ -2746,7 +2746,6 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 	LOG_DEBUG("Restoring Interrupts");
 	buf_set_u64(mstatus_bytes, 0, info->xlen, current_mstatus);
 	register_set(&target->reg_cache->reg_list[REG_MSTATUS], mstatus_bytes);
-	info->mstatus_actual = current_mstatus;
 
 	/// Restore registers
 	uint8_t buf[8];
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2696,6 +2696,18 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 		}
 	}
 
+
+	// Disable Interrupts before attempting to run the algorithm.
+	// Is it possible/desirable to do this in the calling code instead?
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv.c
```diff
@@ -2696,6 +2696,18 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
...
 
+
+	// Disable Interrupts before attempting to run the algorithm.
+	// Is it possible/desirable to do this in the calling code instead?
+	uint64_t current_mstatus;
+
+	register_get(&target->reg_cache->reg_list[REG_MSTATUS]);
+	current_mstatus = info->mstatus_actual;
+	current_mstatus = current_mstatus & ~((uint64_t) 0x8);
+	register_set((&target->reg_cache->reg_list[REG_MSTATUS]), (uint8_t*) &current_mstatus);
```
timsifive: I don't think you can assume that host endianness is the same as what OpenOCD's buf functions use.

src/target/riscv/riscv.c
```diff
@@ -2696,6 +2696,18 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
...
 	}
 
+
+	// Disable Interrupts before attempting to run the algorithm.
+	// Is it possible/desirable to do this in the calling code instead?
+	uint64_t current_mstatus;
+
+	register_get(&target->reg_cache->reg_list[REG_MSTATUS]);
+	current_mstatus = info->mstatus_actual;
+	current_mstatus = current_mstatus & ~((uint64_t) 0x8);
```
timsifive: You've got access to MSTATUS_MIE etc. constants here.

timsifive:
Sorry for the review mess. Haven't quite figured out github's interface.

To your broader point of disabling interrupts at a higher level: riscv_run_algorithm() should run the algorithm without any side effects besides what the algorithm does. It might be used in cases where being destructive is not OK, eg. from riscv_checksum_memory(). It's more straightforward to disable/reenable interrupts inside run_algorithm() than it is to require each caller to know about this.

mwachs5:
Ok. Trying out these suggestions. So far I've just managed to break even the no-interrupt case :)

src/target/riscv/riscv.c
```diff
@@ -2741,6 +2742,12 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 		return ERROR_FAIL;
 	}
 
+	// Restore Interrupts
+	LOG_DEBUG("Restoring Interrupts");
+	buf_set_u64(mstatus_bytes, 0, info->xlen, current_mstatus);
+	register_set(&target->reg_cache->reg_list[REG_MSTATUS], mstatus_bytes);
+	info->mstatus_actual = current_mstatus;
```
mwachs5: @timsifive I wasn't sure if I should be updating mstatus_actual or not

src/target/riscv/riscv.c
```diff
@@ -2696,6 +2696,19 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
...
 
+
+	// Disable Interrupts before attempting to run the algorithm.
+	uint64_t current_mstatus;
+	uint8_t mstatus_bytes[8];
+
+	LOG_DEBUG("Disabling Interrupts");
+	register_get(&target->reg_cache->reg_list[REG_MSTATUS]);
+	current_mstatus = buf_get_u64(target->reg_cache->reg_list[REG_MSTATUS].value, 0, info->xlen);
+	buf_set_u64(mstatus_bytes, 0, info->xlen, (current_mstatus & (~((uint64_t)MSTATUS_MIE))) );
```
timsifive: I realize I got the argument order wrong in my previous comment. Sorry.

`set_field(current_mstatus, MSTATUS_MIE, 0)` reads better than the typecast you had to do.
Is it necessary to disable user etc. interrupts as well?

src/target/riscv/riscv.c
timsifive: I don't think you should. register_set/register_get are the entry points when a user from gdb wants to change registers. mstatus_actual is an internal detail that's not exposed to the outside world, which this code is part of.

mwachs5:
let me know when you're OK with merging this, would like to bump a bunch of repos which depend on it

PR #16, opened 2017-02-06
Refactor code to support multiple debug spec versions.

timsifive:
0.11 still works just as before.
Now there's a framework for adding 0.13 support.
There is probably still some cleanup left to do.


timsifive:
This runs gdbserver.py as well as before. Flashing code to a HiFive1 board still works. Megan also confirms it still works for her use case (flashing from Arduino UI, I think).

PR #18, opened 2017-02-28
riscv: Ensure that hart is halted before attempting to examine it.

mwachs5:
There were a few typos in 0.13 examine() which prevented it from working on a hart which wasn't halted yet.


timsifive:
Wow, those were pretty blatant. Thanks for fixing them.

PR #23, opened 2017-03-30
riscv: Use write-1-to-clear for CMDERR, not write 0 to clear.

mwachs5:
Also since we moved ABSTRACTAUTO into a seperate register, there were also some unnecessary writes to ABSTRACTCS.

I'm fine with leaving this this unmerged until the Spike implementation matches again


palmer-dabbelt:
Thanks

PR #28, opened 2017-04-04
Readmem autoexec

mwachs5:
The autoexec setting and clearing in readmem was still not right. This was manifesting by failing MemTest64 on a RV32 core.


mwachs5:
I think this is still sort of wrong, or writemem is wrong. I believe the write mem is setting ABSTRACTAUTO while busy is true which isn't legal.

mwachs5:
Actually no there is no known current issue with write mem. Simulations are just slow.

PR #30, opened 2017-04-06
Add 64-bit and multihart support

palmer-dabbelt:


palmer-dabbelt:
This fixes https://github.com/sifive/openocd/issues/27 by saving registers that need to be saved for correctness outside the register cache.

palmer-dabbelt:
@mwachs5 This still flashes my HiFive1 correctly, so I don't think I've regressed too much.

src/flash/nor/fespi.c
```diff
@@ -767,12 +768,13 @@ static int steps_execute(struct algorithm_steps *as,
...
+	int xlen = riscv_xlen(target);
 
 	struct reg_param reg_params[2];
-	init_reg_param(&reg_params[0], "x10", 32, PARAM_OUT);
-	init_reg_param(&reg_params[1], "x11", 32, PARAM_OUT);
-	buf_set_u32(reg_params[0].value, 0, 32, ctrl_base);
-	buf_set_u32(reg_params[1].value, 0, 32, data_wa->address);
+	init_reg_param(&reg_params[0], "x10", xlen, PARAM_OUT);
+	init_reg_param(&reg_params[1], "x11", xlen, PARAM_OUT);
+	buf_set_u64(reg_params[0].value, 0, xlen, ctrl_base);
```
mwachs5: Sounds like this change is now backwards compatible with RV32? Cool.

src/target/riscv/riscv-013.c
```diff
@@ -592,6 +496,7 @@ static program_t *program_new(void)
 		program->write = false;
 		program->regno = 0x1000;
 	}
+	program_add32(program, fence_i());
```
mwachs5: Do you need this? The Debug ROM has a fence.i in it before executing the program buffer

src/target/riscv/riscv-013.c
```diff
@@ -1521,300 +1217,38 @@ static int examine(struct target *target)
 
 	// Check that abstract data registers are accessible.
 	uint32_t abstractcs = dmi_read(target, DMI_ABSTRACTCS);
-	LOG_DEBUG("abstractcs=0x%x", abstractcs);
 	info->datacount = get_field(abstractcs, DMI_ABSTRACTCS_DATACOUNT);
 	info->progsize = get_field(abstractcs, DMI_ABSTRACTCS_PROGSIZE);
 
-	uint32_t value = 0x53467665;
```
mwachs5: this was probably left over from something debugging in the HW

src/target/riscv/riscv-013.c
```diff
@@ -1521,300 +1217,38 @@ static int examine(struct target *target)
...
 	}
 
-	info->never_halted = true;
-
-	int result = riscv013_poll(target);
-	if (result != ERROR_OK) {
-		return result;
-	}
+	/* FIXME: Are there 2 triggers? */
+	info->trigger_count = 2;
```
mwachs5: you should be able to figure this out dynamically. The default is 2.

src/target/riscv/riscv-013.c
```diff
@@ -1521,300 +1217,38 @@ static int examine(struct target *target)
...
 
+	/* Resumes all the harts, so the debugger can later pause them. */
+	riscv_resume_all_harts(target);
 	target_set_examined(target);
-	LOG_INFO("Examined RISCV core; XLEN=%d, misa=0x%" PRIx64, xlen(target), info->misa);
```
mwachs5: I'm going to miss this printf. Actually don't some of the tests look for this string to pass?

src/target/riscv/riscv-013.c
```diff
@@ -1941,14 +1276,93 @@ static int check_dmi_error(struct target *target)
...
+		LOG_INFO("read 0x%08x from 0x%08x", value, t_addr);
+
+		if (check_dmi_error(target)) {
+			LOG_ERROR("DMI error");
+			return ERROR_FAIL;
+		}
+	}
+
+	program_t *program = program_new();
+	program_add32(program, ebreak());
```
mwachs5: why all these ebreaks?

src/target/riscv/riscv-013.c
```diff
@@ -1964,65 +1378,50 @@ static int write_memory(struct target *target, uint32_t address,
 				LOG_ERROR("Unsupported size: %d", size);
 				return ERROR_FAIL;
 		}
-		program_add32(program, addi(S0, S0, size));
+		program_add32(program, fence());
```
mwachs5: There is already a fence in the ROM, do you need this?

src/target/riscv/riscv-013.c
```diff
@@ -1964,65 +1378,50 @@ static int write_memory(struct target *target, uint32_t address,
...
+			case 4:
+				value = t_buffer[0] |
+					((uint32_t) t_buffer[1] << 8) |
+					((uint32_t) t_buffer[2] << 16) |
+					((uint32_t) t_buffer[3] << 24);
+				break;
+			default:
+				return ERROR_FAIL;
 		}
-		int result = scans_execute(scans);
```
mwachs5: I assume this is super slow now without the scans stuff? Is the plan to resurrect it at some point?

src/target/riscv/riscv-013.c
```diff
@@ -1964,65 +1378,50 @@ static int write_memory(struct target *target, uint32_t address,
...
+		abstract_write_register(target, S1, riscv_xlen(target), value);
+		program_set_write(program, S1, value);
+
+		LOG_INFO("writing 0x%08x to 0x%08x", value, t_addr);
+
+		write_program(target, program);
+		execute_program(target, program);
+		uint32_t abstractcs;
+		wait_for_idle(target, &abstractcs);
+		program_delete(program);
```
mwachs5: shouldn't you check for cmderr here?

src/target/riscv/riscv-013.c
```diff
@@ -1964,65 +1378,50 @@ static int write_memory(struct target *target, uint32_t address,
...
+		if (check_dmi_error(target)) {
+			LOG_ERROR("DMI error");
 			return ERROR_FAIL;
-		} else if (!dmi_error) {
-			break;
 		}
 	}
 
+	program_t *program = program_new();
+	program_add32(program, ebreak());
```
mwachs5: again why all the ebreaks?

src/target/riscv/riscv-013.c
```diff
@@ -2060,3 +1457,178 @@ struct target_type riscv013_target =
...
+	uint64_t dmcontrol = dmi_read(target, DMI_DMCONTROL);
+	dmcontrol = set_field(dmcontrol, DMI_DMCONTROL_HARTSEL, r->current_hartid);
+	dmi_write(target, DMI_DMCONTROL, dmcontrol);
+}
+
+static void riscv013_halt_current_hart(struct target *target)
+{
+	RISCV_INFO(r);
+	LOG_DEBUG("halting hart %d", r->current_hartid);
+	assert(!riscv_is_halted(target));
```
mwachs5: what happens if the hart halts right here due to breakpoint

src/target/riscv/riscv.c
```diff
@@ -599,3 +608,400 @@ struct target_type riscv_target =
...
+void riscv_info_init(riscv_info_t *r)
+{
+	memset(r, 0, sizeof(*r));
+	r->dtm_version = 1;
+
+	/* FIXME: The RTOS gets enabled before the target gets initialized. */
+	r->rtos_enabled = true;
+
+	for (size_t h = 0; h < RISCV_MAX_HARTS; ++h) {
+		/* FIXME: I need to rip out Tim's probing sequence, as it
```
mwachs5: What would you replace it with?

src/target/riscv/riscv.h
```diff
@@ -2,66 +2,176 @@
 #define RISCV_H
 
 #include "opcodes.h"
+#include "gdb_regs.h"
+
+/* The register cache is staticly allocated. */
+#define RISCV_MAX_HARTS 32
```
mwachs5: Can we make this 1024 to match the actual spec? Or is this just impractical

src/target/riscv/riscv.h
```diff
@@ -2,66 +2,176 @@
...
+typedef uint64_t riscv_reg_t;
+
+enum riscv_hart_state {
+	RISCV_HART_UNKNOWN,
+	RISCV_HART_HALTED,
+	RISCV_HART_RUNNING,
+};
+
+enum riscv_halt_reason {
+	RISCV_HALT_INTERRUPT,
```
mwachs5: we should probably call this 'haltreq' instead of interrupt, since interrupt is a sifive-ism

mwachs5: Also isn't there a 'got an ebreak' reason?

src/target/riscv/riscv.h
```diff
@@ -2,66 +2,176 @@
...
+
+	/* The hart that is currently being debugged.  Note that this is
+	 * different than the hartid that the RTOS is expected to use.  This
+	 * one will change all the time, it's more of a global argument to
+	 * every function than an actual */
+	int current_hartid;
+
+	/* Enough space to store all the registers we might need to save. */
+	/* FIXME: This should probably be a bunch of register caches. */
+	uint64_t saved_registers[RISCV_MAX_HARTS][RISCV_MAX_REGISTERS];
```
mwachs5: ah now i see why this is 32

mwachs5:
Do the openocd.cfg files need to change for this version?

mwachs5:
Something seems wrong with the halt/resume tracking. There is a variable hart_state, which seems to be updated BEFORE you do a halt or resume, but not after you actually do the halt/resume. This leads to lots of messages like "hart 0 requested halt, but hart was already halted" 

Is this just an error in the messaging / my understanding of what the printf means?

https://github.com/sifive/openocd/blob/3c5cb81a0953b3e460d3461659f51763a2a71676/src/target/riscv/riscv.c#L827

Anyway this code is currently segfaulting for me during initialization, was trying to figure out what was going on and noticed this behavior.

mwachs5:
Oh. The reason the code was segfaulting was because I didn't specify '-rtos riscv' in my Openocd.cfg.

So it seems that rtos_enabled is true even if target->rtos isn't initialized. There is a FIXME on the former so maybe that is expected.

PR #33, opened 2017-04-11
Add Back test flag

mwachs5:
This string is needed for the gdbserver tests to work (I'm working on those as well and actually this string isn't quite right, but we do need something here)


mwachs5:
Argh I don't want to do this against master.

PR #43, opened 2017-05-01
Read from 0

palmer-dabbelt:


mwachs5:
Actually can you hold off on merging this, still doing a bit of testing

mwachs5:
ok now it works on 64-bit as well as 32-bit.

PR #57, opened 2017-06-08
riscv: v13 -- dmi_write must still check for the OP result

mwachs5:
The first part of this is cosmetic -- I see no reason not to print out the result of the scan (especially the error code), even if we don't care about the actual returned data.

The second part is the actual fix, making sure that 'in' is updated even in dmi_write, because the OP result is needed for dmi_write to properly detect busy.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -318,12 +318,8 @@ static dmi_status_t dmi_scan(struct target *target, uint16_t *address_in,
 	struct scan_field field = {
 		.num_bits = info->abits + DTM_DMI_OP_LENGTH + DTM_DMI_DATA_LENGTH,
 		.out_value = out,
+                .in_value = in
 	};
-
-        // We need to always do this so that we can
-        // get the OP result, even if we don't care about
-        // the data itself.
-        field.in_value = in;
 
 	assert(info->abits != 0);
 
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -325,9 +320,10 @@ static dmi_status_t dmi_scan(struct target *target, uint16_t *address_in,
 		.out_value = out,
 	};
 
-	if (address_in || data_in) {
-		field.in_value = in;
-	}
+        // We need to always do this so that we can
+        // get the OP result, even if we don't care about
+        // the data itself.
+        field.in_value = in;
```
mwachs5: I guess we could do this within the struct initialization as well

mwachs5:
Testing on FPGA now. Will let you know.

timsifive:
ISTR the reason for not printing out values we don't care about is performance. IIRC it's slower to do a scan and get results than to just do the scan. But I can't think right now of the exact mechanism that makes that happen, so maybe it's not true. I'd run the download speed test to see if it's affected.

mwachs5:
The v11 code always got the result, it just didn't print it out. It has to at least populate the result because it needs the OP in order to know whether it succeeded or not. 

The v13 code didn't always populate the result, which was a bug. I made it act the same as v11.

As for the printing, I would prefer to at least see the error response (?, +, b). Rather than muck around with figuring that out I just made it print the same for both Reads and Writes. I didn't update the v11 because I wasn't currently debugging it, but probably makes sense to make the format the same for v11 and v13. I don't think the extra few characters in the printf is going to make a big perf. difference. It must always populate the actual scan results either way.

PR #58, opened 2017-06-08
Merge GNU MCU Eclipse OpenOCD v0.10.0-1-20170607-2132-dev

ilg-ul:
This patch updates to the latest OpenOCD commit (June 2) and to the current GNU MCU Eclipse OpenOCD used to release the binaries.

The RISC-V changes to OpenOCD are conditionally compiled with 

```c
#if defined(GNU_MCU_ECLIPSE_RISCV)
...
#endif
```

and require `-DGNU_MCU_ECLIPSE_RISCV` added to the build. A `--enable-riscv` option is planned for a future version.


timsifive:
Did you run the 32- and 64-bit tests in riscv-tests/debug?

ilg-ul:
nope. do these tests run on macOS?

if you'll take a look at the changes, you'll notice that all RISC-V patches are exactly the ones in your fork, but braced with #ifdef/#endif.

so, if your fork passed the tests, the integration into GNU MCU Eclipse is likely to pass them too.

palmer-dabbelt:
Thanks, but we have to merge openocd/master due to a customer requirement so this is going to have to be low priority for a bit.

Sorry!

ilg-ul:
> we have to merge openocd/master

openocd/master from June 2 already included

ilg-ul:
the longer you delay and the more changes you make to the current branch, the more conflicts you'll have to resolve.

palmer-dabbelt:
Ah, OK -- if master is in there then this is good, I just didn't want to move to the 0.10 branch.  The feature we need is 64-bit physical addresses, is the support in the rest of OpenOCD part of this merge?

ilg-ul:
sorry, I don't get your question.

if the comment in the top of the page states that the latest openocd commit from June 2 is included, and I repeated the same fact a few minutes ago, what exactly is the question?

by the time I did the merge, it was the absolutely latest openocd commit available.

ilg-ul:
> 64-bit physical addresses

I don't know exactly what you need, but I solved lots of conflicts due to the change to target_addr_t which can be a 64-bits address.

if you want to do it again, please be my guest.

palmer-dabbelt:
The target_addr_t change was the change we were looking for.  I was told that change happened after the 0.10.0 release, but it looks like I'm just crazy.  This appears to be the commit we want, which is really old 47b8cf84202bf792cf66fbfa01169e9592236b8a.

Thanks for doing this.  @timsifive: can you test this and merge it in ASAP?

ilg-ul:
the plan with GNU MCU Eclipse OpenOCD was to periodically merge the latest commits from https://git.code.sf.net/p/openocd/code and from https://github.com/riscv/riscv-openocd.git, apply some cosmetics to your changes, if needed (like bracing with #if/#endif), and generating all kind of binaries for all platforms (for example https://github.com/gnu-mcu-eclipse/openocd/releases/tag/v0.10.0-1-20170607-2132-dev)

is this ok for you?

palmer-dabbelt:
We generally handle releases like this:
* Before we're trying to upstream our code, we base our port on an old release.  This is essentially what we're doing for OpenOCD, but we didn't pick an actual release just some arbitrary commit.  This was just an oversight.
* While we're trying to get the code upstream, we begin to regularly merge in from upstream/master.
* Once we have been merged into upstream and upstream has released a stable version with out port in it, we base releases off that stable version.  Often times there are a handful of patches we need to backport.

So I think that's a good plan until we get the RISC-V OpenOCD port merged into upstream/master and then released as a stable version, at which point we should be using that stable version for releases.

ilg-ul:
openocd does not have real releases. once a year they bump the version number at an arbitrary moment. their official position is that the latest commit is the latest release and encourage users to compile their binaries themselves.

for my users this is not realistic, and every few months I provide binaries, with whatever is their latest commit. if you look at the github statistics, the binaries have thousands of downloads, so I would say are more or less ok.

palmer-dabbelt:
Ah, OK -- I guess that was my confusion (it's why I thought you were on a release branch).  In that case I agree that we should just ship arbitrary hashes.  We're going to try to do monthly releases at SiFive, so maybe it'd be best to just sync up our tags.  In this case we can try to just have pretty much the same codebases.

We're trying to get another OpenOCD release out today, but it's already tagged.  That's why things are a bit busy here and I wasn't really looking at the PR.  Hopefully we can get this all sorted out and we'll keep our trees pretty close in the future.

Thanks for all your work!

ilg-ul:
> We're trying to get another OpenOCD release out today

I just did one release, binaries for all platforms are available (https://github.com/gnu-mcu-eclipse/openocd/releases/tag/v0.10.0-1-20170607-2132-dev); do you really need a new one?

timsifive:
This doesn't build for me. I get:
```
libtool: compile:  gcc -std=gnu99 -DHAVE_CONFIG_H -I. -I./src -I./src -I./src/helper -DPKGDATADIR=\"/opt/riscv/share/openocd\" -DBINDIR=\"/opt/riscv/bin\" -I./jimtcl -I./jimtcl -Wall -Wstrict-prototypes -Wformat-security -Wshadow -Wextra -Wno-unused-parameter -Wbad-function-cast -Wcast-align -Wredundant-decls -Werror -g -O2 -MT src/target/riscv/riscv-013.lo -MD -MP -MF src/target/riscv/.deps/riscv-013.Tpo -c src/target/riscv/riscv-013.c -o src/target/riscv/riscv-013.o
In file included from ./src/helper/system.h:28:0,
                 from ./config.h:333,
                 from src/target/riscv/riscv-013.c:11:
src/target/riscv/riscv-013.c: In function 'read_memory':
src/target/riscv/riscv-013.c:1316:34: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
                 assert (cur_addr >= address);
                                  ^
src/target/riscv/riscv-013.c: In function 'write_memory':
src/target/riscv/riscv-013.c:1511:34: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
                 assert (cur_addr > address);
                                  ^
```

I can work to fix those errors, but I'm curious if you know why you didn't encounter them.

palmer-dabbelt:
We usually build without -werror, as with all the compiler versions fixing all the warnings is a minefield and I don't want it to block users building things.

timsifive:
Looking over this diff, there's a lot of GNU_MCU_ECLIPSE_RISCV conditional code. Is that expected to make it to mainline OpenOCD at some point?

ilg-ul:
The GNU_MCU_ECLIPSE_RISCV conditional code was already fixed, in my repo I added `--enable-riscv` to configure, and the conditional code uses `#if BUILD_RISCV == 1`.

So I suggest we abandon this PR, and you merge directly from my repo.

timsifive:
What branch in your repo? I thought I was looking at ilg-ul/riscv
(502a73f9b8096fd25dee4237fb9a9f6b7d900b6f) but that didn't include the
--enable-riscv option. (Maybe I hosed your repo by clicking the resolve
conflicts button in the github UI?)

timsifive:
I'm not really concerned about the define. I'm concerned about there being
additional code just for your Eclipse integration that may not be suitable
for upstream. But the diff is huge and I'm having a hard time separating
what you did from what just came in from OpenOCD. Do you expect all your
code to be accepted by mainline OpenOCD?

ilg-ul:
the GNU MCU Eclipse OpenOCD repo, the `gnu-mcu-eclipse-dev` branch:

https://github.com/gnu-mcu-eclipse/openocd/commits/gnu-mcu-eclipse-dev

for this PR I created a temporary fork, where I merged changes from my repo.

I have no problems to keep my repo in sync with the OpenOCD master and with the RISC-V master, and I guess you should be able to sync directly to my repo too, it's not a big deal.

ilg-ul:
> I'm concerned about there being additional code just for your Eclipse integration that may not be suitable for upstream

which code? do you have an example?

> I'm having a hard time separating what you did from what just came in from OpenOCD

I also had a hard time to merge your changes over the OpenOCD master.

currently there is no GNU MCU Eclipse specific code in my branch, with two exceptions:

- all code coming from RISC-V is braced with `#if BUILD_RISCV/#endif`
- the code displaying the greeting message was extended with `--enable-branding` and there is a conditional code with `#if defined(BUILD_BRANDING)`

I suggest you do the following:

- checkout the OpenOCD/master in a separate local branch
- checkout my gnu-mcu-eclipse-dev branch in separate local branch
- compare the two local branches

if you find other changes than those braced with BUILD_RISCV & BUILD_BRANDING, please let me know to fix them, since there should be none.

ilg-ul:
> Do you expect all your code to be accepted by mainline OpenOCD?

without using `--enable-riscv` or `--enable-branding`, my branch should be identical to the OpenOCD/master.

generally adding code to OpenOCD is accepted.

what might not be OK is to alter functionality for other platforms when `--enable-riscv` is defined, and this happens in some cases.

you can identify these cases because the conditional code also has an `#else` case.

this is because while merging, I identified places where you changed functionality. if this is to correct obvious bugs, this should be submitted in separate patches to OpenOCD to be accepted.

anyway, my code has better chances to be integrated into mainstream OpenOCD than your current code, which has 0 chances.

as a general recommendation, when you change code in other projects and expect integration, I advise you to always use conditional code to keep both your changes and the original code in place; it'll also help you identify your changes. and it may sometimes help while solving conflicts. highly recommended.

this being said, I'll close this PR now.

timsifive:
So I did my own merge of OpenOCD mainline into riscv/riscv-openocd, and compared that against 
gnu-mcu-eclipse-dev. Below are my notes.

The obvious stuff that is in your branch but should not make it to OpenOCD are various documentation files, which point at your project. (There are lots of them, and they actually look like a great introduction to your project.)

The branding string seems like a reasonable feature, but has nothing to do with RISC-V.

I see changes in contrib/60-openocd.rules.

There are some build scripts etc, which don't seem relevant to RISC-V.

The diff is quite large, and there are probably a bunch of build fixes (I assume) in there that we want, but accepting the thing wholesale feels like too much.

timsifive:
>
> what might not be OK is to alter functionality for other platforms when
> --enable-riscv is defined, and this happens in some cases.
>
> you can identify these cases because the conditional code also has an
> #else case.
>
> this is because while merging, I identified places where you changed
> functionality. if this is to correct obvious bugs, this should be submitted
> in separate patches to OpenOCD to be accepted.
>
I noticed some of that. I'll have to go through those cases, because we
would like RISC-V to be built just like any other target. Having RISC-V
support compiled into an OpenOCD binary must not affect debugging of other
targets.

ilg-ul:
> are various documentation files, which point at your project

yes, those are specific to my project and are used during build. ignore the `gnu-mcu-eclipse` folder entirely.

> I see changes in contrib/60-openocd.rules.

yes, I simplified the rules, the `plugdev` group is not available on all linux distributions.

and I also added 3 configuration files for the SiFive boards.

> there are probably a bunch of build fixes (I assume) in there that we want

please feel free to cherry pick what you want.

PR #62, opened 2017-06-13
Merge mainline OpenOCD

timsifive:
The main thing we get is 64-bit support. 64-bit support doesn't fully work yet. This is bringing in the changes that allow us to implement it.

I've tested this with 32- and 64-bit spike, HiFive1, and U500, and everything looks normal.

The change that I actually made myself (aside from trivial conflicts during the merge) is https://github.com/riscv/riscv-openocd/commit/64af05291132811288d7af32c9fd27e254778d5c

The main change is to make riscv_addr_t be unsigned. The rest is mechanical fixing of types, print statements, and a few signed/unsigned compares.


ilg-ul:
what upstream commit did you use? 

if github shows commits in reverse chronological order, it seems you used one from Sep 29, 2016, which is quite old.

timsifive:
I merged 1025be363e2bf42f1613083223a2322cc3a9bd4c, which is from May 31.
github shows commits in forward chronological order.

ilg-ul:
in files where you modified something specific to risc-v, did you preserve the original content, preferably with conditional compilation?

timsifive:
In this merge I didn't modify anything in general code that's specific to RISC-V.
In the project in general I think some of these changes have been made, but I'm not attempting to fix that in this pull request.

ilg-ul:
ok, then I can't be of much help in this stage. when you're ready with all changes and your code is ready for prime time, like mine, please let me know, to compare the two versions.

PR #67, opened 2017-06-16
Various cosmetic improvements

timsifive:
Making mainly the debug output more readable, so I have a better chance of figuring out what's going on.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1155,6 +1155,11 @@ static int examine(struct target *target)
 			r->xlen[i] = 64;
 		}
 
+		/* Display this as early as possible to help people who are using
+		 * really slow simulators. */
+		LOG_DEBUG(" hart %d: XLEN=%d, program buffer at 0x%" PRIx64, i,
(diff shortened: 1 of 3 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1097,15 +1155,12 @@ static int examine(struct target *target)
 			r->xlen[i] = 64;
 		}
 
-		LOG_DEBUG("hart %d has XLEN=%d", i, r->xlen[i]);
```
mwachs5: Can this or something similar be left here? Enumerating the triggers takes a while in simulation, and this one is useful for an early-out if it's really broken.

timsifive: It got upgraded to an INFO at the end of this function. Is that soon enough?

mwachs5: I guess what I'm saying before the trigger enumeration, since that takes a long time in sim. It would be ncie to see that it at least identified the XLEN before going into that.

timsifive:
Megan, can you please review the latest changes? I'm unable to merge as it is.

timsifive:
@mwachs5

PR #69, opened 2017-06-21
Fix the multi-GDB mode bugs

palmer-dabbelt:
I just smoke tested this: I can run and halt both harts on my system, but that's all I tested.  I suspect we need to sprinkle riscv_hart_enabled() all over the place (at least in the hardware breakpoint code), but I think we're pretty close.

I haven't tested this at all in RTOS mode.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1143,6 +1143,9 @@ static int examine(struct target *target)
 	/* Find the address of the program buffer, which must be done without
 	 * knowing anything about the target. */
 	for (int i = 0; i < riscv_count_harts(target); ++i) {
+		if (!riscv_hart_enabled(target, i))
+			continue;
+
 		riscv_set_current_hartid(target, i);
 
 		/* Without knowing anything else we can at least mess with the
@@ -1213,6 +1216,9 @@ static int examine(struct target *target)
 
 	/* Then we check the number of triggers availiable to each hart. */
 	for (int i = 0; i < riscv_count_harts(target); ++i) {
+		if (!riscv_hart_enabled(target, i))
+			continue;
+
 		for (uint32_t t = 0; t < RISCV_MAX_TRIGGERS; ++t) {
 			riscv_set_current_hartid(target, i);
 
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1334,9 +1334,6 @@ static int read_memory(struct target *target, target_addr_t address,
 			size, address);
 
 	select_dmi(target);
-	/* There was a bug in the memory system and only accesses from hart 0 actually
-	 * worked correctly.  This should be obselete now. -palmer */
-	riscv_set_current_hartid(target, 0);
 
 	/* This program uses two temporary registers.  A word of data and the
 	 * associated address are stored at some location in memory.  The
@@ -1532,9 +1529,6 @@ static int write_memory(struct target *target, target_addr_t address,
 	LOG_DEBUG("writing %d words of %d bytes to 0x%08lx", count, size, (long)address);
 
 	select_dmi(target);
-	/* There was a bug in the memory system and only accesses from hart 0 actually
-	 * worked correctly.  This should be obselete now. -palmer */
-	riscv_set_current_hartid(target, 0);
 
 	/* This program uses two temporary registers.  A word of data and the
 	 * associated address are stored at some location in memory.  The
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1123,17 +1123,23 @@ static int examine(struct target *target)
 
 	/* Before doing anything else we must first enumerate the harts. */
 	RISCV_INFO(r);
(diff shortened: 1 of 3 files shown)


palmer-dabbelt:
@timsifive Do you mind taking a look at this?  We'd love to get this done by tomorrow night so we can deliver this in the next drop.

palmer-dabbelt:
Oh, also: I found the configuration I needed to use unexpected:

````
# JTAG adapter setup
adapter_khz     10000

interface ftdi
ftdi_device_desc "Olimex OpenOCD JTAG ARM-USB-TINY-H"
ftdi_vid_pid 0x15ba 0x002a

ftdi_layout_init 0x0808 0x0a1b
ftdi_layout_signal nSRST -oe 0x0200
#ftdi_layout_signal nTRST -data 0x0100 -oe 0x0100
ftdi_layout_signal LED -data 0x0800

set _CHIPNAME riscv
jtag newtap $_CHIPNAME cpu -irlen 5 -expected-id 0x20000001

target create $_TARGETNAME.0 riscv -chain-position $_TARGETNAME -coreid 0
$_TARGETNAME.0 configure -work-area-phys 0x80000000 -work-area-size 10000 -work-area-backup 1

target create $_TARGETNAME.1 riscv -chain-position $_TARGETNAME -coreid 1
$_TARGETNAME.1 configure -work-area-phys 0x80000000 -work-area-size 10000 -work-area-backup 1

# Un-comment these two flash lines if you have a SPI flash and want to write
# it.
#flash bank spi0 fespi 0x40000000 0 0 0 $_TARGETNAME.0 0x20004000
init
if {[ info exists pulse_srst]} {
  ftdi_set_signal nSRST 0
  ftdi_set_signal nSRST z
}

targets $_TARGETNAME.0
halt

targets $_TARGETNAME.1
halt

#flash protect 0 64 last off
echo "Ready for Remote Connections"
````

We should provide better error messages when users forget to halt.

**Comment by mwachs5** [comment]
Does this consider non 1-1 mappings between core ID and hartid? I guess from GDB's perspective it can always just use coreid, and read $mhartid to figure out what the actual hart id is.

**Comment by timsifive** [comment]
@mwachs5 the code in general should work with non 1-1 mappings, because it really only ever uses the core index. It's confusing though, because it calls that index the hart ID, which may not be true. On my mental TODO list is to clean this up to remove the confusion. (Part of that is to read the actual hart ID and display that to the user.)

**Comment by timsifive** [comment]
Build fails:
```
In file included from src/target/riscv/riscv.c:17:0:
src/target/riscv/riscv.c: In function 'riscv_hart_enabled':
src/target/riscv/riscv.c:1240:13: error: unused variable 'r' [-Werror=unused-variable]
  RISCV_INFO(r);
             ^
src/target/riscv/riscv.h:102:37: note: in definition of macro 'RISCV_INFO'
 #define RISCV_INFO(R) riscv_info_t *R = riscv_info(target);
                                     ^
```

File: src/target/riscv/riscv-013.c [hunk 1]
```diff
@@ -1123,17 +1123,19 @@ static int examine(struct target *target)
...
-				break;
-			r->hart_count = i + 1;
-		}
-	} else {
-		r->hart_count = 1;
+	int original_coreid = target->coreid;
+	for (int i = 0; i < RISCV_MAX_HARTS; ++i) {
+		/* Fake being a non-RTOS targeted to this core so we can see if
+		 * it exists. */
+		target->coreid = i;
```
timsifive: This feels fragile. Can you comment what code is being fooled here?

timsifive:
Hmm... but if I run this code against the multicore image I got from @mwachs5 then I get:
```
...
Warn : Bypassing JTAG setup events due to errors
Info : dtmcontrol_idle=5, dmi_busy_delay=1, ac_busy_delay=0
Info : dtmcontrol_idle=5, dmi_busy_delay=2, ac_busy_delay=0
openocd: src/target/riscv/riscv.c:1022: riscv_set_current_hartid: Assertion `riscv_hart_enabled(target, hartid)' failed.
```

This is with -rtos.

mwachs5:
@timsifive  when you said you added -rtos, then you shouldn't create two targets, right? Can you show the full openocd.cfg when you added -rtos?

palmer-dabbelt:
I just pushed a new patch set that should fix your CR comments.  I haven't tested this in RTOS mode, so I'm not surprised it failed.  Do you have time to look at the RTOS failure?

timsifive:
I prefer to continue writing a test that properly tests RTOS mode multicore, since that is what the customer is interested in. As it is our test coverage for multicore is almost nonexistent.
If you think they care about multi-GDB as well, I can change my priorities.

palmer-dabbelt:
OK, I fixed the RTOS problem -- it was just an assertion in examine.

palmer-dabbelt:
@timsifive I think everything has been fixed, is there anything I'm missing?

timsifive:
Now tests pass against the FPGA, but Hwbp2 fails against 32- and 64-bit spike. All tests pass on 32- and 64-bit spike in the riscv branch.

palmer-dabbelt:
@timsifive Do you have time to look at the failures?

PR #70, opened 2017-06-22
Ignore build directory

colinschmidt:
the riscv-tools build script make a subdirectory to build in, other submodules ignore this file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # stuff "git status" should ignore
 
 # build output
+build/
 .libs
 .deps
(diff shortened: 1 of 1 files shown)


timsifive:
This seems like a change that wouldn't be accepted upstream, because it's specific to riscv-tools.
I would prefer to modify riscv-tools so it creates the build directory somewhere else.
But it's also trivial to filter out from pull requests to upstream, so if this is important I'm OK with it.

colinschmidt:
Not important. Just a slight nuisance.

PR #72, opened 2017-06-30
Restore value of temporary register (s0) in examine OpenOCD procedure…

dmitryryzhov:
… in the case of the core can not execute 64 bit instruction.

Execution fails if the OpenOCD tries to connect to 32 bit core and halts it, when some program is already running and the register s0 is used.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1177,8 +1177,12 @@ static int examine(struct target *target)
 
 		/* This program uses a temporary register. If the core can not
 		 * execute 64 bit instruction, the original value of temporary
-		 * register will not be restored due to an exception. So we have to
-		 * restore it manually in that case. */
+		 * register (s0) will not be restored due to an exception.
+		 * So we have to save it and restore manually in that case.
+		 * If the core can execute 64 bit instruction, the saved value
+		 * is wrong, because it was read with 32 bit lw instruction,
+		 * but the value of s0 will be restored by the reverse swap
+		 * of s0 and dscratch registers. */
 		uint64_t s0 = riscv_get_register(target, GDB_REGNO_S0);
 
 		struct riscv_program program64;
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1175,6 +1175,12 @@ static int examine(struct target *target)
 		 * In order to make this work we first need to */
 		int offset = (progbuf_addr % 8 == 0) ? -4 : 0;
 
+		/* This program uses a temporary register. If the core can not
+		 * execute 64 bit instruction, the original value of temporary
+		 * register will not be restored due to an exception. So we have to
+		 * restore it manually in that case. */
+		uint64_t s0 = riscv_get_register(target, GDB_REGNO_S0);
```
mwachs5: does this solve the chicken-and-egg? Do we actually know how to 'riscv_get_register' before executing this function?

timsifive:
This code passes all riscv-tests/debug against spike (32 and 64 bit).

mwachs5:
Does your tests handle the case for a 32-bit program which is running and depends on its s0 value can be examined then just run successfully? 

What is actually used for program_lx when it hasn't been determined yet? I just want to understand why this works, not asserting that it *doesn't* work.

timsifive:
There are no tests for the bug this is fixing. I assume that Dmitri has tested that use case. I haven't looked closely enough to know how this works.

mwachs5:
I think it doesn't solve the underlying chicken-and-egg problem. Basically to fix this we need to read S0 in some 'safe' way before we know any other way to read it, which is either abstract command or by using the values in DMHARTINFO.

mwachs5:
Digging deeper, it solves the problem in the one specific case that:

1) this hart is actually 32-bit (because we 'guessed' that above, so riscv_program_lx will use 32-bit lw). In the case it was 64-bit the saved value of s0 is wrong, which we should add to the comment.
2) The program buffer is actually writable, because if it isn't, we would have lost s0 in the case above anyway.

Can you please clarify this in the comments? Functionally I am OK with the change.

dmitryryzhov:
Added a comment on the wrong saved value in the 64 bit case.
If the program buffer is not writable, then we really lose the value of s0 in both 32 and 64 bit cases at the moment of execution the 32 bit program. And with current procedure we can not even determine what kind of hart it is.

palmer-dabbelt:
Thanks!

PR #74, opened 2017-07-03
Fix 32-bit build errors.

### LINKED ISSUE #71: RISC-V target support doesn't build without --disable-werror
See here for the original issue report:

https://github.com/gnu-mcu-eclipse/openocd/issues/17

As mentioned there the RISC-V target support code does not build unless --disable-werror is specified at openocd configuration time even though openocd normally (and currently - i.e. openocd 0.10.0+dev development branch) compiles without having to suppress warnings as errors.

Maybe the RISC-V code could be cleaned up to eliminate these warnings?

Thanks.

timsifive:
I only compiled the source. Didn't have the tooling installed to link.
Hopefully that's good enough.
Fixes #71.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -94,7 +94,9 @@ size_t riscv_batch_add_dmi_read(struct riscv_batch *batch, unsigned address)
 	riscv_batch_add_nop(batch);
 
 	batch->read_keys[batch->read_keys_used] = batch->used_scans - 1;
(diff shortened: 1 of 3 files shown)


mwachs5:
Another common cause of build warnings (on macOS) is using DMI_DMCONTROL_HALTREQ with set_field. Because it is bit 31 it seems to not actually work with this macro.

timsifive:
If I can I'll set up a macos build on travis as well, but that doesn't seem relevant to this PR.

timsifive:
Agreed that more work will probably be needed.
Please file a separate issue for MacOS, along with all the errors you see there.

PR #78, opened 2017-07-06
Add 32-bit build

timsifive:
This should prevent Issue #74 from reoccurring.

### DIFF
--- a/src/target/mips32_pracc.h
+++ b/src/target/mips32_pracc.h
@@ -114,7 +114,7 @@ int mips32_cp0_read(struct mips_ejtag *ejtag_info,
 int mips32_cp0_write(struct mips_ejtag *ejtag_info,
 		uint32_t val, uint32_t cp0_reg, uint32_t cp0_sel);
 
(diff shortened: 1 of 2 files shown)


src/target/mips32_pracc.h
```diff
@@ -114,7 +114,7 @@ int mips32_cp0_read(struct mips_ejtag *ejtag_info,
 int mips32_cp0_write(struct mips_ejtag *ejtag_info,
 		uint32_t val, uint32_t cp0_reg, uint32_t cp0_sel);
 
-inline void pracc_swap16_array(struct mips_ejtag *ejtag_info, uint32_t *buf, int count)
+static inline void pracc_swap16_array(struct mips_ejtag *ejtag_info, uint32_t *buf, int count)
```
mwachs5: is this something that would need to be upstreamed? Why does mainline OpenOCD Just Work with this?

timsifive: Mainline OpenOCD does not Just Work with this. On some compilers this fix is required, but it hasn't made it to mainline yet. I got this fix from OpenOCD's gerrit, where it has been approved but not yet merged.

PR #79, opened 2017-07-08
Use abstract command to access registers if possible

timsifive:
This is the only way that the spec requires that GPRs are accessed. This gets us a lot closer to debugging cores which don't have a program buffer.

As a side benefit, it's a bit faster too (WriteGprs improved from 29s to 19s on spike, and from 17s to 15s on an FPGA target).

Some smaller changes:
* When an abstract command is executed and fails, immediately clear cmderr while returning the error.
* Also change debug code that decodes DMI values so it's easier to maintain.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -578,45 +578,6 @@ static int wait_for_idle(struct target *target, uint32_t *abstractcs)
 	}
 }
 
-static int register_read_direct(struct target *target, uint64_t *value, uint32_t number);
-
-static int register_write_direct(struct target *target, unsigned number,
-		uint64_t value)
-{
-	struct riscv_program program;
-
-	LOG_DEBUG("[%d] reg[0x%x] <- 0x%" PRIx64, riscv_current_hartid(target),
-			number, value);
-
-	riscv_program_init(&program, target);
-
-	riscv_addr_t input = riscv_program_alloc_d(&program);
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -573,16 +578,188 @@ static int wait_for_idle(struct target *target, uint32_t *abstractcs)
...
+		if (!r->abstract_read_csr_supported)
+			return ERROR_FAIL;
+		command = set_field(command, AC_ACCESS_REGISTER_REGNO,
+				number - GDB_REGNO_CSR0);
+	} else {
+		return ERROR_FAIL;
+	}
+
+	int result = execute_abstract_command(target, command);
+	if (result != ERROR_OK) {
```
mwachs5: should you check for the specific error code? For example, one might get error halt_resume here. Should you just check error_unsupported?

timsifive: Yes, that's better. I'll do that.

src/target/riscv/riscv-013.c
```diff
@@ -742,6 +925,13 @@ static int init_target(struct command_context *cmd_ctx,
 	info->dmi_busy_delay = 0;
 	info->ac_busy_delay = 0;
 
+	// Assume all these abstract commands are supported until we learn
+	// otherwise.
```
timsifive: I've added a TODO comment to do this right, but I don't want to spend the time on this right now.

PR #81, opened 2017-07-10
Use LL for 64-bit defines, as Windows is LLP64

palmer-dabbelt:
This should also fix bugs on ILP32 systems.  IIRC these are all autogenerated from somewhere, we should fix it there rather than just using sed.

### DIFF
--- a/src/target/riscv/debug_defines.h
+++ b/src/target/riscv/debug_defines.h
@@ -106,7 +106,7 @@
  */
 #define DTM_DMI_DATA_OFFSET                 2
 #define DTM_DMI_DATA_LENGTH                 32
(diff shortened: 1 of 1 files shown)


mwachs5:
Yes, they are autogenerated from riscv/riscv-debug-spec. Please excuse my ignorance, but why LL vs UL?

timsifive:
I'll fix the debug spec generating code once Palmer clarifies Megan's question.

palmer-dabbelt:
@mwachs5 LL is "long long" (which is 64-bit on Windows), while UL is "unsigned long" (which is 32-bit on Windows).  With UL DTM_DMI_DATA will still get truncated.

PR #84, opened 2017-07-14
Fix infinite loop in reset.

timsifive:


src/target/riscv/riscv-013.c
```diff
@@ -1874,7 +1882,16 @@ void riscv013_reset_current_hart(struct target *target)
 	control = set_field(control, DMI_DMCONTROL_NDMRESET, 0);
 	dmi_write(target, DMI_DMCONTROL, control);
```
mwachs5: This may take several seconds to come out of reset. I think there are  parameters you can specify in your OpenOCD CFG file that you should use here. Otherwise you may unnecessarily start bumping up the busy delay, which only ever goes up.

timsifive: Good point. Changing this to use a wall clock timeout.

mwachs5:
It would be nice to have this overridable at runtime, but I don't see a clean place other than a RISC-V specific command to do so

PR #86, opened 2017-07-27
Display register numbers in a more usable format.

timsifive:


src/target/riscv/riscv.c
```diff
@@ -1596,8 +1596,57 @@ int riscv_enumerate_triggers(struct target *target)
...
+		case GDB_REGNO_DSCRATCH:
+			return "dscratch";
+		case GDB_REGNO_MSTATUS:
+			return "mstatus";
+		case GDB_REGNO_PRIV:
+			return "priv";
+		default:
+			if (regno <= GDB_REGNO_XPR31) {
+				sprintf(buf, "x%d", regno - GDB_REGNO_XPR0);
+			} else if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095) {
```
mwachs5: what is the rationale of printing CSRs with %d instead of %x? The spec describes them as %x, does GDB use %d?

timsifive: GDB uses %d. It's also more compact. (eg. csr0x700 vs csr1792) But I'm happy to go with 0x if people prefer that. I was just getting tired of manually converting the gdb numbering to CSR numbers.

mwachs5:
yes either way this is an improvement

PR #90, opened 2017-08-10
Fix reads beyond requested memory range

p12nGH:
This change fixes bug that caused GDB to produce incorrect back traces.
Problem was with OpenOCD reading one additional word beyond the requested range. In this case GDB was reading program stack located at the very top of the RAM, so the last word was outside of valid memory range. The solution was to disable abstract command autoexec for the last read.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1354,8 +1354,9 @@ static int read_memory(struct target *target, target_addr_t address,
 	riscv_addr_t fin_addr = address + (count * size);
 	riscv_addr_t prev_addr = ((riscv_addr_t) address) - size;
 	bool first = true;
-	LOG_DEBUG("writing until final address 0x%" PRIx64, fin_addr);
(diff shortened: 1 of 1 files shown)


timsifive:
Did you run the debug tests in riscv-tests/debug with this change? `make clean && make all` in that directory should do it, assuming your modified openocd is in your PATH.

p12nGH:
No. Will try to do it now.

timsifive:
With this change, MemTest64 and MemTestBlock are failing for me on spike32, and MemTest64 fails against spike64.

palmer-dabbelt:
It made it through the tests on Travis <https://github.com/riscv/riscv-tools/pull/119>.  Are they false-passing?  That other PR you have open in riscv-tools seems to indicate that is the case. 

@p12nGH Did you get these to run the tests locally?

timsifive:
Yes, the Travis build is false-passing, and probably have been forever. I'm working on that.

PR #91, opened 2017-08-11
Be more clear in multi-core systems without -rtos

timsifive:
Don't print out there's a hart with XLEN of 0.


palmer-dabbelt (changes requested):
This looks good, but before we merge it I'd like to make sure riscv-tests is correctly running the tests and fix the regression from https://github.com/riscv/riscv-openocd/pull/90 .

timsifive:
Sure. I just wanted this to not get lost in my mess of local branches.

palmer-dabbelt:
Ya, no problem.

timsifive:
#90 merged a while ago. I just ran the sim tests against this code again, and it works fine. Can you approve the change?

palmer-dabbelt (approved):
Sorry about that, I just missed this going by again.

PR #92, opened 2017-08-12
Fe 402 fix

p12nGH:
My previous version of fix used reads from cached program buffer - this caused tests from riscv-tests/debug to fail. With this fix tests are passing and GDB shows the correct stack trace.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1374,7 +1374,7 @@ static int read_memory(struct target *target, target_addr_t address,
 		size_t rereads = reads;
 		for (riscv_addr_t i = start; i < count; ++i) {
 
-			if (i == count) {
+			if (i == count - 1) {
 				// don't do actual read in this batch,
 				// we will do it later after we disable autoexec
 				//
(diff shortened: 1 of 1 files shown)


timsifive:
This is a huge improvement. Right now MemTestBlock is still failing on the spike32 target, though:
```
tnewsome@compy-vm:~/SiFive/riscv-tools/riscv-tests/debug$ ./gdbserver.py targets/RISC-V/spike32.py Block
Running ExamineTarget > logs/20170813-111425-spike32-ExamineTarget.log ... pass in 2.32s
Running MemTestBlock > logs/20170813-111428-spike32-MemTestBlock.log ... fail in 5.32s
::::::::::::::::::::::::::::[ ran 2 tests in 8s ]:::::::::::::::::::::::::::::
1 tests returned fail
   MemTestBlock > logs/20170813-111428-spike32-MemTestBlock.log
1 tests returned pass
```

(`make all` in riscv-tests/debug tests on both 32- and 64-bit spike.)

timsifive:
The problem is caused by the addition of `- size` in:
```
+       while (count > 1 && (cur_addr = riscv_read_debug_buffer_x(target, d_addr)) < fin_addr - size) {
```
in change b5692585de669dc8562fd995fda5808625d6f7a8.

Removing that lets MemTestBlock pass, but now MemTest64 starts failing, so clearly it's needed sometimes.

PR #94, opened 2017-08-13
Fix a corner case in block memory read.

timsifive:
This makes all the MemTest* tests pass.

It looks like, only when run in a series of tests, DebugCompareSections fails on spike64. Still looking at that, but I think this is an improvement nonetheless.


timsifive:
Looks like the DebugCompareSections is to do with encountering abstractcs.busy, which happens when I run `make all` presumably because of the higher system load.

PR #95, opened 2017-08-14
Fix block memory reads on slow targets.

timsifive:
The interesting new code concerns ignore_prev_addr and
this_is_last_read.

Additionally, I tweaked some debug output, and optimized
riscv_batch_run() when the batch is empty.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -44,6 +44,11 @@ bool riscv_batch_full(struct riscv_batch *batch)
 
 void riscv_batch_run(struct riscv_batch *batch)
 {
+	if (batch->used_scans == 0) {
+		LOG_DEBUG("Ignoring empty batch.");
+		return;
+	}
+
 	LOG_DEBUG("running a batch of %ld scans", (long)batch->used_scans);
 	riscv_batch_add_nop(batch);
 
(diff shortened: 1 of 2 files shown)


timsifive:
This doesn't fix everything either. :-(

timsifive:
But it does make things better, so please review so we can merge it.

p12nGH-zz:
Do we have all the tests passing with this commit? What do you mean by it doesn't fix everything?

timsifive:
It passes all tests, when the machine running the tests is not heavily loaded. I'm still chasing out the details on when it fails. But in the meantime it's more reliable than it was.

p12nGH-zz:
What hardware do you use?

p12nGH-zz:
Nevermind, I realized we can only run tests on spike

timsifive:
You can run these tests against real hardware also, but I'm just working on spike right now. Real hardware doesn't become loaded in the same way.

timsifive:
Meanwhile, this change reduces the frequency of the problem, so I would still like it reviewed/merged.

mwachs5:
Has this been tested on the HW Targets?

p12nGH-zz:
@mwachs5 , no, I am  currently testing my binary build which doesn't have this commit. So far I didn't see any issues with real hardware.

timsifive:
riscv-tools build is failing, and this change would certainly improve things even if it won't fix anything.

Is there any reason nobody has approved it yet?

mwachs5:
I think riscv-tools has already merged a bunch of openocd and riscv-tests fixes. Is it still failing?

timsifive:
https://travis-ci.org/riscv/riscv-tools says it is.

timsifive:
What is this waiting for?

BTW, it appears that this fixes everything as long as spike still has the debug_module debug printf()s compiled in (which it does in the version that riscv-tools is pointing at, but not in the latest riscv-isa-sim).

palmer-dabbelt (approved):
@timsifive Can you bump this to riscv-tools to see if it fixes the problems?

PR #97, opened 2017-08-15
Add some keep_alive()s for use with slow targets.

timsifive:
The only real improvement is that now OpenOCD doesn't warn that keep_alive isn't called often enough anymore. Presumably there's some user improvement if you're using gdb interactively.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -44,6 +44,8 @@ bool riscv_batch_full(struct riscv_batch *batch)
 
 void riscv_batch_run(struct riscv_batch *batch)
 {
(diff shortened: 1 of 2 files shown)


mwachs5:
Is this orthogonal to the gdb set_remote_timeout command?

timsifive:
Increasing `remotetimeout` in gdb will work around this problem, but it's nice not to make the user do unnecessary things.

mwachs5 (approved):

timsifive:
Can I get a review on this?

timsifive:
Thank you.

Issue #96, opened 2017-08-15
abstractcs timeouts should not be hard-coded


mwachs5:
There is a timeout WALL_CLOCK_TIMEOUT for polling abstractcs.busy. This shouldn't be compiled in, as this timeout is going to be very different when interacting with a simulation target vs a HW target. Similarly, WALL_CLOCK_RESET_TIMEOUT should not be compiled in. We should create riscv target-specific commands for these items. 

I know it is a goal of this OpenOCD implementation that user should not have to specify a lot of knobs, but one could pretty easily give the user instructions on what to do if they *are* hitting these timeouts.

timsifive:
Can this be closed now that #100 merged?

mwachs5:
Yes thanks

PR #100, opened 2017-08-15
riscv: Add commands for setting timeouts

mwachs5:
Addresses #96.

Rather than using a #define for the timeouts, allow the user to specify them using OpenOCD's "Command Handler" interface. The defaults remain the same as they were with the #defines.

I tried making these target-specific variables, but then they can only be run with mode of COMMAND_EXEC, which can only be run after 'init'. These need to be set before attempting to interact with the target, so I changed it to COMMAND_ANY and moved them out of the target info structure. 

(There are some other whitespace changes because I used emacs 'tabify' to fight with the tab-based indentation, which tabified a few other things)

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -39,26 +39,26 @@
  *
  * There are a few functions to just instantly shift a register and get its
  * value:
- * 		dtmcontrol_scan
- * 		idcode_scan
- * 		dbus_scan
+ *		dtmcontrol_scan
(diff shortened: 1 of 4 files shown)


timsifive (approved):
This is a good improvement.

I do worry a bit about the name of the command. It's not obvious what the timeout really applies to, or what individual commands are. I'd be tempted to simply use
"riscv set_timeout"
"Set wall clock timeout for RISC-V operations."

PR #105, opened 2017-08-28
memory read is now completely stable even on intermittently slow targets

timsifive:
The downloaded program now post-increments, and there's no longer an attempt to read the current address from the target. This made it easier to fix the problem where at the start of the loop the current address was already read (in regular entry) or has not yet been read (when the first round through the loop encountered busy more than once, or busy was encountered at least once later on).

Fixed a few math errors regarding addresses/indexes.

Also:
* Removed unnecessary reset in increase_ac_busy_delay().
* Removed various unnecessary newlines in log messages.

This version passes a stress test (simultaneously run 4 spikes with random delays added in, covering 32- and 64-bit, as well as single and dual core), and it also passes against the hardware I have access to.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1273,7 +1273,6 @@ static int read_memory(struct target *target, target_addr_t address,
 	riscv_addr_t r_addr = riscv_program_alloc_x(&program);
 	riscv_program_fence(&program);
 	riscv_program_lx(&program, GDB_REGNO_S0, r_addr);
-	riscv_program_addi(&program, GDB_REGNO_S0, GDB_REGNO_S0, size);
 	switch (size) {
 		case 1:
 			riscv_program_lbr(&program, GDB_REGNO_S1, GDB_REGNO_S0, 0);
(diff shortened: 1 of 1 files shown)


palmer-dabbelt (changes requested):
I always find the read_memory function hard to read.  Does it still respect the constraint that it only reads from each address once?  IIRC that's why we had this off-by-one oddness here.

timsifive:
If the timing is right, and the loop doesn't encounter cmderr=busy, then every address is read exactly once. When cmderr=busy an entire block is read again, which means the addresses at the start of that block may be read more than once.

I will fix that, although I don't know of a good way to add a test for it. I'll also add a comment mentioning that behavior is important.

palmer-dabbelt:
It's probably more work to test for than just to keep sane manually.  The constraint is for memory-mapped register reads, which sometimes go through this path, so we could always test it with a special device in Spike.

timsifive:
Note that for reads of a single word, the read is never executed twice.

timsifive:
Fixed.

palmer-dabbelt (approved):

PR #107, opened 2017-09-11
Add timeout to another infinite loop.

timsifive:


palmer-dabbelt (dismissed):
Thanks

src/target/riscv/riscv-013.c
```diff
@@ -2041,9 +2041,17 @@ int riscv013_debug_buffer_register(struct target *target, riscv_addr_t addr)
...
 {
 	// Wait for busy to go away.
+	time_t start = time(NULL);
 	uint32_t abstractcs = dmi_read(target, DMI_ABSTRACTCS);
 	while (get_field(abstractcs, DMI_ABSTRACTCS_BUSY)) {
 		abstractcs = dmi_read(target, DMI_ABSTRACTCS);
+
+		if (time(NULL) - start > riscv_command_timeout_sec) {
+			LOG_ERROR("abstractcs.busy is not going low after %d seconds. "
+					"The target is either really slow, or broken. "
```
mwachs5: Can you add the same "suggestion" message on how to increase this timeout?

mwachs5 (approved):

Issue #103, opened 2017-08-25
clang warnings


ilg-ul:
it would be great to add an automated travis build on macOS, to spot clang warnings/errors.

timsifive:
There is a clean clang build now. Can we close this issue?

ilg-ul:
ok, I'll close it for now. 

in case I encounter further warnings, I'll let you know.

timsifive:
Thank you.

PR #113, opened 2017-09-18
Add clang build

timsifive:


timsifive:
Travis is very slow starting OSX builds, but this change adds clang builds on Linux which hopefully get us the same coverage.

timsifive:
This should fix #103.

ilg-ul:
> This should fix #103.

does the Travis masOS come with all tools required to build openocd? normally autotools are not part of the macOS distribution, not even part of the developer tools.

timsifive:
> does the Travis masOS come with all tools required to build openocd? normally autotools are not part of the macOS distribution, not even part of the developer tools.

I don't know. The Travis OSX build took so long to start that I never watched one happen. This change adds building with clang, which should get us most of the benefit.

palmer-dabbelt (approved):
LGTM

PR #114, opened 2017-09-21
Reset

timsifive:
Rewrite reset. Now it works on multicore systems, too.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1220,6 +1220,16 @@ static int assert_reset(struct target *target)
 				target->reset_halt ? 1 : 0);
 		control = set_field(control, DMI_DMCONTROL_HARTRESET, 1);
 		dmi_write(target, DMI_DMCONTROL, control);
+
+		// Read back to check if hartreset is supported.
+		uint32_t rb = dmi_read(target, DMI_DMCONTROL);
+		if (!get_field(rb, DMI_DMCONTROL_HARTRESET)) {
+			// Use ndmreset instead. That will reset the entire device, but
+			// that's probably what OpenOCD wants anyway.
+			control = set_field(control, DMI_DMCONTROL_HARTRESET, 0);
+			control = set_field(control, DMI_DMCONTROL_NDMRESET, 1);
+			dmi_write(target, DMI_DMCONTROL, control);
+		}
 	}
 
 	target->state = TARGET_RESET;
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1185,21 +1185,42 @@ static int examine(struct target *target)
 
 static int assert_reset(struct target *target)
 {
-	/*FIXME -- this only works for single-hart.*/
 	RISCV_INFO(r);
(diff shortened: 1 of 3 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1187,23 +1185,44 @@ static int examine(struct target *target)
...
+
 	} else {
-		LOG_DEBUG("TARGET RESET HALT NOT SET");
-		control = set_field(control, DMI_DMCONTROL_HALTREQ, 0);
+		// Reset just this hart.
+		uint32_t control = set_field(control_base, DMI_DMCONTROL_HARTSEL,
+				r->current_hartid);
+		control = set_field(control, DMI_DMCONTROL_HALTREQ,
+				target->reset_halt ? 1 : 0);
+		control = set_field(control, DMI_DMCONTROL_HARTRESET, 1);
```
mwachs5, line 1221: HARTRESET is optional. How do you handle the case when it doesn't work?

mwachs5 (changes requested):
I'm not sure how this works if you have something that doesn't use rtos, is only a single hart, and doesn't have HARTRESET functionality implemented.

src/target/riscv/riscv-013.c
```diff
@@ -1187,23 +1185,44 @@ static int examine(struct target *target)
...
-	control = set_field(control, DMI_DMCONTROL_NDMRESET, 1);
-	if (target->reset_halt) {
-		LOG_DEBUG("TARGET RESET HALT SET, ensuring halt is set during reset.");
-		control = set_field(control, DMI_DMCONTROL_HALTREQ, 1);
+
+	uint32_t control_base = set_field(0, DMI_DMCONTROL_DMACTIVE, 1);
+
+	if (target->rtos) {
+		// There's only one target, and OpenOCD thinks each hart is a thread.
+		// We must reset them all.
```
mwachs5, line 1196: Maybe should update this comment to say "we must reset them all, but we only halt the ones controlled by this target"

src/target/riscv/riscv-013.c
mwachs5: I think you should read back HARTRESET and if it's not actually set, you should fall back to writing NDMRESET (potentially with some debug information).

Also, I think that even if you DID have HARTRESET, the user may want to distinguish between "reset all the things" and "reset a single Hart". Is there an OpenOCD way of clarifying that intention?

src/target/riscv/riscv-013.c
timsifive: There is only one target in the RTOS case, and that target controls all the harts.

timsifive:
> I think you should read back HARTRESET and if it's not actually set, you should fall back to writing NDMRESET (potentially with some debug information).

Done.

> Also, I think that even if you DID have HARTRESET, the user may want to distinguish between "reset all the things" and "reset a single Hart". Is there an OpenOCD way of clarifying that intention?

I don't think there is. See http://openocd.org/doc/html/General-Commands.html#resetcommand

mwachs5 (approved):

PR #116, opened 2017-09-26
Fix trigger code to work with multi-gdb mode instead of RTOS mode

timsifive:


timsifive:
Can I get a review on this?

aswaterman:
Palmer's out sick, and I'm not familiar with the openocd code, but it looks OK to me save for one issue:

```assert``` can be compiled out with ```-DNDEBUG```, so it's good for making sure you don't have an internal error in your code, but not so good for checking for error conditions that might come from external sources (e.g., a glitch or a buggy chip).  I'm not sure which bucket this falls into, but might ```if (!x) return -1``` be more appropriate than ```assert(x)```?

aswaterman (approved):

timsifive:
The only asserts I see are `assert(first_hart >= 0);`, which check that internal bookkeeping is done properly. `riscv_hart_enabled()` does not talk to the target, and I don't think depends on it.

aswaterman:
k

PR #119, opened 2017-09-30
Update to master (Aug 10, 2017)

ilg-ul:
Keeping up to date with master would help distributions that keep track of both master and riscv.


ilg-ul:
as far as I can see, most changes are specific to the other targets.

for the common code, apart from the printf format improvements in target.c, there is a new log level added. I hope they will not impact the risc-v functionality.

timsifive:
Did you test this with `make all` in riscv-tests?

ilg-ul:
nope. aren't the Travis tests doing this automatically?

actually I do not have the environment for these tests, my environment is slightly different.

but if your question refers to how well these changes integrate with your changes, they do, in my previous release I already merged them to my openocd, together with yours, and the build on macOS, Debian and Debian32 was fine.

update: not only the build was ok, also the binary was functional for RISC-V; it was not ok for non-RISC-V, but the reason was the extra error processing you added.

ilg-ul:
if you don't feel comfortable, you don't necessarily have to accept this pull request, you can manually merge `master` in your build environment, and possibly perform any extra tests.

I don't mind how you do it, the result is identical.

the reason I need this update applied is that I planned a few more PRs to pass you my changes and fix functionality for non-RISC-V target (my most recent release is currently broken, and I need a fix as soon as possible); since I'm already in line with `master`, manually re-creating the patches for your version, that does not include it, would be more complicated than necessary.

timsifive:
> nope. aren't the Travis tests doing this automatically?

Only in riscv-tools, which does not often get the latest version of all its tools. It will catch problems, but not until quite late. :-( Ideally riscv-tools builds stick binaries somewhere, which we can then use in riscv-openocd to get gdb, gcc, etc.

> actually I do not have the environment for these tests, my environment is slightly different.

I'll run the tests then.

> if you don't feel comfortable, you don't necessarily have to accept this pull request, you can manually merge master in your build environment, and possibly perform any extra tests.

The change looks good. I just want to make sure all changes are tested to prevent regressions. (And ideally every pull request has already had the tests run. Travis would be ideal, like you say.)

ilg-ul:
thank you, Tim. let me know if I can be of any help.

timsifive (approved):
I ran the tests, and they pass.

palmer-dabbelt:
@p12nGH can you make sure this works on Eclipse?

ilg-ul:
@palmer-dabbelt, do you have anything specific in mind? 

I just released a new OpenOCD binary that is based on this latest commit and it seems ok.

palmer-dabbelt:
Gleb has been fixing all the bugs that crop up when used from Eclipse, and there's been a bunch of them.  I want to make sure we don't regress on anything.

ilg-ul:
you mean those: https://github.com/riscv/riscv-openocd/commits?author=p12nGH ?

as far as I can see, all are in the RISC-V specific code, so normally there should be no regressions.

palmer-dabbelt:
The idea is that we would test this pull request for regressions.

ilg-ul:
> we would test this pull request for regressions

sure, no problem. 

is there a way to automate these tests? for example, when I contributed to mbed, they had several automated services, like Travis, but, to save time, not all commits received a full treatment. to enable the full test suite, a special command was passed in the PR comments.

palmer-dabbelt:
We don't have any automated testing setup yet, sorry.  The tests involve FPGAs and clicking GUI buttons so we haven't automated them yet.

PR #122, opened 2017-10-02
Add configure option --enable-riscv

ilg-ul:
This PR helps isolate the RISC-V specific changes applied to common code, and provides a convenient way to disable all of them at once, for example to diagnose cases when the other targets are affected by RISC-V patches.

It includes a small addition to `configure.ac` to define `BUILD_RISCV`. For the moment, to do not brake your builds, I set the default to true (i.e. to include the RISC-V target), but generally it should be false.

Then (hopefully!) all RISC-V changes in common places were braced with `#if BUILD_RISCV == 1 / #endif`.

Some of them might need reflection.

The opportunity of the patches returning errors will be addressed in a separate PR.

This approach will also help when you'll try to upstream the RISC-V changes.

After this PR is applied, a diff to master should be done, and if there are other RISC-V specific changes not covered, they should be braced too.


ilg-ul:
I had a small glitch with an warning, but now it should be ok.

timsifive:
I'm not a fan of this change.

There don't seem to be existing --enable options for eg. ARM and MIPS, so I don't see why we need one for RISC-V.

The target-independent changes we made need to either be upstreamed or reverted. Putting them behind a --enable-riscv flag is not a good solution, because then RISC-V OpenOCD behaves differently for non-RISC-V targets, which is sure to end in tears.

ilg-ul:
so what is proposal?

mwachs5:
I understand the rationale for this change as part of upstreaming, but agree that this isn't ready to merge. At the very least,  we should separate the things that are really "RISC-V" vs things that are "FESPI" or "ISSI Flash". The latter don't really have anything to do with RISC-V. 

A number of things are just changes to the debugging prints, which again should be a different enable from RISC-V. Maybe "RISC-V Prints" or something.

timsifive:
> so what is proposal?

We need to go through every target-independent change we made, and either upstream it or revert it. Once that is done I don't see a need a for a --enable-riscv config option.

I've started this yesterday actually, upstreaming the simplest change I came across (http://openocd.zylin.com/#/c/4238/). I'll keep at this in the background.

ilg-ul:
I agree that if you upstream or revert the changes, this configure option is no longer necessary.

For me having a common code that works for all targets is of high priority, and I'm a bit concerned since I don't know how long it might take.

@TM1234? Any suggestions?

timsifive:
> For me having a common code that works for all targets is of high priority, and I'm a bit concerned since I don't know how long it might take.

What is the problem you're running into?

ilg-ul:
> What is the problem you're running into?

there are at least two show stoppers that I know of:

- the GDB error processing that breaks normal sessions for other targets
- viewing memory blocks crashes Eclipse; I need to diagnose this, it may be in openocd or in gdb

as annoyances, I need to do something to prevent showing the CSRs as regular registers.

timsifive:
> the GDB error processing that breaks normal sessions for other targets

For that specific failure, it makes sense to add a run-time option until we've figured out the final solution.

ilg-ul:
> a run-time option

can you be more specific what would do such an option?

timsifive:
Some option like --ignore-register-errors that reverts to the original behavior (don't propagate errors that occur while accessing registers).

ilg-ul:
> '--ignore-register-errors'

you mean to ask all existing openocd users to modify their configurations and add this extra option?

timsifive:
This is a temporary solution for people using the RISC-V fork. It will probably disappear once we figure out whether this code will be upstreamed/reverted.

ilg-ul:
> a temporary solution for people using the RISC-V fork

there is only one GNU MCU Eclipse OpenOCD binary, that should work for all targets.

right now it is marked as 'broken' and users are asked to use a previous release. (https://github.com/gnu-mcu-eclipse/openocd/releases)

timsifive:
If you don't want errors propagated in your binary, then the simplest is just to revert that code in your branch. It doesn't change often, so conflicts should be minimal.

ilg-ul:
> If you don't want errors propagated in your binary,

I don't think it is a matter of wanting or not, with that patch applied openocd fails for other targets.

in my opinion the solution is exactly the opposite, if you want errors propagated for some RISC-V devices, you need to explicitly set it, either via a command line option or a configuration file command.

otherwise I doubt your patches will be accepted upstream.

> revert that code in your branch

I will try it tomorrow and let you know the result.

ilg-ul:
just curious: if you need a different functionality for RISC-V in the common code, why don't you test at runtime the target type and do it without affecting other targets?

ilg-ul:
I applied a small change to my fork, to disable the code returning errors to gdb, and now my binary seems ok for both Cortex-M and RISC-V (I tested only with the HiFive1 board).

https://github.com/gnu-mcu-eclipse/openocd/commit/da43038f866cd7444405edd01780fd3316ad187c

it would be interesting to see if a similar change passes your tests.

TommyMurphyTM1234:
> @TM1234? Any suggestions?

I'm coming to this a bit late but my views for what they're worth:

1. Until such time as the RISC-V changes are upstreamed it makes a lot of sense to me to have a --enable-riscv configuration time option and associated #if BUILD_RISCV == 1 ... #endif wrappers because this means that the RISC-V specific changes (including those in non RISC-V target specific code) are clearly delineated and identifiable and the original unmodified OpenOCD base code is also available. The presence of these preprocessor wrappers helped me to identify the code in src/server/gdb_server.c that were made for RISC-V but which break Cortex-M (and maybe other CPU?) debugging: https://github.com/riscv/riscv-openocd/issues/115. 

1. Once the RISC-V changes are upstreamed the rationale for --enable-riscv diappears since OpenOCD only seems to make the inclusion or exclusion of specific JTAG and other debug probes configurable options and does not do the same for targets, flash drivers etc.

1. A problem that I have had with the master RISC-V OpenOCD code is that it's not always clear what master OpenOCD code (release, tag/branch etc.) it is based so diffing to find the changes made for RISC-V is not always practicable. Again Liviu's configuration option and preprocessor wrappers makes this easier but then Liviu also has to play catch up with the master RISC-V OpenOCD branch code base.

1. It's not clear to me what the status is of upstreaming the RISC-V changes. I presume that RISC-V development is ongoing (judging by the fact that the RISC-V OpenOCD branch is still getting fairly regular commits) and that the issue of upstreaming has not yet been tackled. I don't know how long that process might take but I am sure that some of the changes in the RISC-V branch will not be accepted as is.

1. On the general issue of adding different behaviour to OpenOCD my view is that such changes should be "opt in" rather than "opt out" otherwise they may break the flow for existing users - for example the suggestion of having a --ignore-register-errors does not make sense to me. It would make more sense to have a --check-for-register-errors or something like that to enable the changed functionality.

1. It would be good to have a clear explanation of the rationale and function of non RISC-V target specific changes in the RISC-V OpenOCD branch because in my experience the commit comments are often inadequate so the reasoning behind such changes is not clear.

1. Changes made to the RISC-V OpenOCD branch need to take cognizance of the fact that OpenOCD is used for more than just RISC-V or SiFive targets and needs to work with everything available to date as well as RISC-V going forward.

1. Liviu's structuring of the RISC-V OpenOCD code changes and his docker based scripted build process is much easier to use than simply pulling the code from the master RISC-V OpenOCD branch github. I think it'd be helpful if his approach was adopted in the branch github too for as long as the changes remain downstream.

Hope this helps.

ilg-ul:
> ... Until such time as the RISC-V changes are upstreamed ...

this is my concern too.

@timsifive, do you have an estimate for how long it'll take to upstream the current version?

except inocent changes in messages, I also think that you'll have a hard time to push the changes upstream, and the whole process might take longer than you expect. frankly, without run-time checks for the target, I seriously doubt the current changes will be accepted.

that's why Tommy's suggestion to use something like the `BUILD_RISCV` mechanism on a temporary basis might be useful until you manage to upstream everything.

timsifive:
> I applied a small change to my fork, to disable the code returning errors to gdb, and now my binary seems ok for both Cortex-M and RISC-V (I tested only with the HiFive1 board).

That sounds like a good temporary solution then.

> @timsifive, do you have an estimate for how long it'll take to upstream the current version?

There are 2 parts:
1. RISC-V-specific code. That likely won't be upstreamed until there is a ratified spec (or until things seem to be stable).
2. Target-independent code. I've started looking at that, and plan to upstream/revert our changes starting with the simplest. I might achieve one change per week or so, but there's no fixed timeline.

timsifive:
> that's why Tommy's suggestion to use something like the BUILD_RISCV mechanism on a temporary basis might be useful until you manage to upstream everything.

I don't see how BUILD_RISCV helps you. You want one binary that can debug both ARM and RISC-V, so you must use BUILD_RISCV, which would (presumably) turn on the error propagating behavior that breaks ARM debugging.

ilg-ul:
> I don't see how BUILD_RISCV helps you. 

exactly how Tommy explained it: if I encounter a problem that I suspect is related to the RISC-V patches, I can easily turn off all at once.

and if I need to browse through the RISC-V changes, I simply search for all occurences of BUILD_RISCV.

and so on...

ilg-ul:
> There are 2 parts: ... until there is a ratified spec ... there's no fixed timeline

so the risc-v openocd will continue to have, in one form or another, different functionality compared to the upstream one, for an undetermined period of time, and changes in the common code will continue to be applied, further widening the gap.

in other words, it would be safer to release two binaries, a standard one and a risc-v specific one. (in this case BUILD_RISCV is even more useful).

this is not a very good news for those who need both versions. Tommy? (@TM1234)

timsifive:
> exactly how Tommy explained it: if I encounter a problem that I suspect is related to the RISC-V patches, I can easily turn off all at once.

> and if I need to browse through the RISC-V changes, I simply search for all occurences of BUILD_RISCV.

That's what version control is for. If you want to "turn off" the RISC-V code:
```git checkout `git merge-base master riscv` ```

If you want to see the RISC-V differences:
```git diff riscv `git merge-base master riscv` `

This assumes you have the following remotes/branches defined:
```
tnewsome@compy-vm:~/SiFive/riscv-tools/riscv-openocd$ git remote -v
openocd	https://git.code.sf.net/p/openocd/code (fetch)
openocd	https://git.code.sf.net/p/openocd/code (push)
origin	https://github.com/riscv/riscv-openocd.git (fetch)
origin	https://github.com/riscv/riscv-openocd.git (push)
tnewsome@compy-vm:~/SiFive/riscv-tools/riscv-openocd$ git branch -vv
  master                    2168c47 [openocd/master] nrf5: Add nRF52832-QFAA support
  riscv                     abe7eba [origin/riscv] Merge pull request #118 from riscv/priv
```

TommyMurphyTM1234:
> That's what version control is for. If you want to "turn off" the RISC-V code:
> git checkout `git merge-base master riscv`
> If you want to see the RISC-V differences:
> ```git diff riscv git merge-base master riscv `

I want an openocd that works for RISC-V and all other supported CPUs - in particular Cortex-M at the moment. So I can't just switch off RISC-V functionality. I need to use the RISC-V functionality but hack it to make it not break Cortex-M debugging. I don't see why there is any confusion or contention here - if OpenOCD is to work for RISC-V then it needs not to break existing functionality.

TommyMurphyTM1234:
> RISC-V-specific code. That likely won't be upstreamed until there is a ratified spec (or until things seem to be stable).

I presume you mean a ratified or at least stable privileged ISA spec?

timsifive:
@TM1234 This discussion is about a change request for putting all RISC-V changes under a compile flag, which doesn't help with that problem. Eg. https://github.com/riscv/riscv-openocd/issues/115 is better place for that discussion. Everybody agrees with the end goal. The only contention is how to get there, and what to do in the meantime.

timsifive:
> I presume you mean a ratified or at least stable privileged ISA spec?

I meant a ratified/stable debug spec (which is unlikely to happen before the ISA spec).

PR #123, opened 2017-10-03
Add read buffer to bitbang, improving performance.

timsifive:
This reduces the time for one testcase where OpenOCD connects to a
simulator from 12.30s to 5.35s!

Running all our tests went from 13m13s to 3m55s.


palmer-dabbelt (dismissed):
This touches a bunch of upstream stuff, but it should be OK for now.

timsifive:
I remembered last night that I need to deal with the buffer filling up, so this is not quite ready to go yet.

timsifive:
Can I get another review of this code?

palmer-dabbelt (approved):

PR #133, opened 2017-10-26
No Russian

palmer-dabbelt:


timsifive:
Why do we need this change to non-RISC-V code?

palmer-dabbelt:
The Russian characters break something on @terpstra and @p12nGH's systems.

timsifive:
So this is just a stop-gap measure until that's resolved?

The second file still has a backwards R in it. I don't know if that was intentional or an oversight.

palmer-dabbelt:
I think I got them all now.  I don't know how long we'll need to carry the patch for.

mwachs5:
For the record, this isn't actually a problem with the OpenOCD build, it's just confusing another tool. Since riscv-openocd is included with riscv-tools, it gets sucked into a lot of projects.

timsifive:
This change seems like a bad idea, since we'll never be able to upstream it (unless perhaps this is a common problem that I just haven't heard about). It would be better to change the files before invoking the broken tool, or to have a special SiFive branch of OpenOCD that has this change.

p12nGH-zz (commented):
I would do it this way
1986ве1т -> 1986be1t
к1879хб1я -> k1879hb1ya
But I agree with Tim's comments regarding potential problems with upstreaming our port. This renaming is  a workaround, and I wonder how difficult would it be to fix the actual problem with some tool(not sure which one) not understanding Unicode characters.

terpstra:
Putting non-ascii characters in filenames is problematic for a lot of reasons. Some filesystems won't even support creating the files! Even linux does not have a defined character set for filenames, so anything high ascii is basically just undefined behaviour. I would be surprised if they don't change this upstream on their own sooner rather than later.

richardxia:
Can someone here show me how to reproduce the problem? I think I might be able to fix the issue without making changes to this repo if the problem you all are running to is because of `sbt`.

PR #134, opened 2017-10-27
Fix compile warning with gcc 6.3.0

timsifive:
Change-Id: I14ebf597f41429c0fc3ebac8da9c9f62c78fb1ae


ilg-ul:
could you be more specific on which version of gcc are you using?

timsifive:
Can I get a review for this trivial change?

mwachs5 (approved):

timsifive:
Thanks, @mwachs5

PR #135, opened 2017-10-31
Build OpenOCD the way we expect users to build it.

timsifive:
Change-Id: I3769137bc3109b44da76f2ca689d351bb93e7832


palmer-dabbelt (approved):
LGTM

PR #137, opened 2017-11-02
Add a RISC-V Compliance Test Command

mwachs5:
I don't expect to merge this PR as-is, but soliciting feedback.

This PR adds a RISC-V command "riscv test-compliance" (which is probably not that great of a name) to test a lot of aspects of the specification that aren't tested in normal GDB/OpenOCD operation. For example, it ensures that if an implementation says it has some number of Data / Program Buffer words, it actually has that many words and that all bits can be read and written. 

Certainly there is a lot of stuff missing from this test, we can always add to it. What do you think of adding something like this at all?


src/target/riscv/riscv-013.c
```diff
@@ -1787,7 +1789,7 @@ static void riscv013_halt_current_hart(struct target *target)
 
 	/* Issue the halt command, and then wait for the current hart to halt. */
 	uint32_t dmcontrol = dmi_read(target, DMI_DMCONTROL);
-	dmcontrol = set_field(dmcontrol, DMI_DMCONTROL_HALTREQ, 1);
+	dmcontrol |= DMI_DMCONTROL_HALTREQ;
```
palmer-dabbelt: Is there something wrong with set_field?  IIRC we use it all over the place, if it's broken the we should fix it.  As far as I can tell, these do the same thing so I'm fine with either way.

src/target/riscv/riscv-013.c
```diff
@@ -1803,7 +1805,7 @@ static void riscv013_halt_current_hart(struct target *target)
 		abort();
 	}
 
-	dmcontrol = set_field(dmcontrol, DMI_DMCONTROL_HALTREQ, 0);
+	dmcontrol &= ~DMI_DMCONTROL_HALTREQ;
```
palmer-dabbelt: Like above.

src/target/riscv/riscv-013.c
```diff
@@ -1923,6 +1925,7 @@ int riscv013_dmi_write_u64_bits(struct target *target)
 	return info->abits + DTM_DMI_DATA_LENGTH + DTM_DMI_OP_LENGTH;
 }
 
+
```
palmer-dabbelt: Spurious newline.

src/target/riscv/riscv-013.c
```diff
@@ -2073,3 +2076,418 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  do {
+    dmstatus = dmi_read(target, DMI_DMSTATUS);
+  } while ((dmstatus & DMI_DMSTATUS_ALLHALTED) == 0);
+  dmcontrol &= ~DMI_DMCONTROL_HALTREQ;
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  // Not clear that this read is required according to the spec.
+  dmi_read(target, DMI_DMSTATUS);
+  
+  // HARTINFO: Read-Only. This is per-hart, so need to adjust hartsel.
+  for (int hartsel = 0; hartsel < riscv_count_harts(target); hartsel++){
```
palmer-dabbelt: You should call hart_enabled here so this works on non-contiguous hart lists.

src/target/riscv/riscv-013.c
```diff
@@ -2073,3 +2076,418 @@ void riscv013_clear_abstract_error(struct target *target)
...
+    struct riscv_program program;
+    riscv_program_init(&program, target);
+    riscv_addr_t addr = riscv_program_alloc_x(&program);
+    riscv_program_lx(&program, GDB_REGNO_S0, addr);
+    // Also testing that WFI() is a NOP during debug mode.
+    riscv_program_insert(&program, wfi());
+    riscv_program_addi(&program, GDB_REGNO_S0, GDB_REGNO_S0, 1);
+    riscv_program_sx(&program, GDB_REGNO_S0, addr);
+    riscv_program_write_ram(&program, addr, 0);
+    if (riscv_xlen(target) > 32) {
```
palmer-dabbelt: IIRC there's a helper for this maybe riscv_program_write_x.

src/target/riscv/riscv-013.c
```diff
@@ -2073,3 +2076,418 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  }
+  
+  for (unsigned int i = 0; i < get_field(abstractcs, DMI_ABSTRACTCS_DATACOUNT); i ++){
+    testvar = (i + 1) * 0x11111111;
+    COMPLIANCE_TEST(dmi_read(target, DMI_DATA0 + i) == testvar, "DATA words must not be affected by NDMRESET");
+  }
+
+  // verify that DPC *is* affected by ndmreset. Since we don't know what it *should* be,
+  // just verify that at least it's not the bogus value anymore.
+  COMPLIANCE_TEST(bogus_dpc != 0xdeadbeef, "BOGUS DPC should have been set somehow (bug in compliance test)");
```
palmer-dabbelt: The low bit of SEPC is always 0.  It's not specified for DPC, but if that works the same way then this test will always pass.

palmer-dabbelt (dismissed):
I like the idea.  Upstream might push back, but I think it's better to try and keep this all in one place so we don't have to maintain a fork.

src/target/riscv/riscv-013.c
mwachs5: The issue I had when I wrote this code was something about overflow/sign extension, because I think haltreq is the MSB. This may have been fixed elsewhere when we cleaned up the rest of the code, so I can probably try reverting this change.

palmer-dabbelt: Eh, it's all the same.  As long as set_field works properly then it's not a big deal.

timsifive:
Ideally this would be external to OpenOCD, but I suspect that's a big pain.

Within OpenOCD, however, can you put the compliance tests in a separate file, with an eye on maybe making it a compile-time option in the future?

src/target/riscv/riscv-013.c
```diff
@@ -2073,3 +2076,418 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  int total_tests = 0;
+  int passed_tests = 0;
+
+  uint32_t dmcontrol_orig = dmi_read(target, DMI_DMCONTROL);
+  uint32_t dmcontrol;
+  uint32_t testvar;
+  dmcontrol = set_field(dmcontrol_orig, DMI_DMCONTROL_HARTSEL, RISCV_MAX_HARTS-1);
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  dmcontrol = dmi_read(target, DMI_DMCONTROL);
+  COMPLIANCE_TEST(get_field(dmcontrol, DMI_DMCONTROL_HARTSEL) == (RISCV_MAX_HARTS-1), "DMCONTROL.hartsel should hold MAX_HARTS - 1");
```
mwachs5: This needs to be updated based on spec now allowing these bit to not be fully RW

src/target/riscv/riscv-013.c
mwachs5: This is just trying to make sure you set the `bogus_dpc` variable, it's checking for a bug in the test, not in the HW.

mwachs5:
I would like to merge this. Can I get a review for merging?
@timsifive I know you want me to move it to a seperate file, not really sure how to do that given all the intertwined code. I mean what parts do you think should be moved out? Should we just move all the riscv commands to a different file?

src/target/riscv/riscv-013.c
```diff
@@ -263,7 +264,7 @@ static dm013_info_t *get_dm(struct target *target)
 static uint32_t hartsel_mask(const struct target *target)
 {
 	RISCV013_INFO(info);
-	return ((1L<<info->hartsellen)-1) << DMI_DMCONTROL_HARTSEL_OFFSET;
+	return ((1L<<info->hartsellen)-1) << DMI_DMCONTROL_HARTSELLO_OFFSET;
```
timsifive: Can you file an issue that we need to add support for hartselhi? This change is not complete. I assume you just made it to get things to build, but we shouldn't lose track of it.

src/target/riscv/riscv-013.c
```diff
@@ -277,7 +278,7 @@ static void decode_dmi(char *text, unsigned address, unsigned data)
 		{ DMI_DMCONTROL, DMI_DMCONTROL_RESUMEREQ, "resumereq" },
 		{ DMI_DMCONTROL, DMI_DMCONTROL_HARTRESET, "hartreset" },
 		{ DMI_DMCONTROL, DMI_DMCONTROL_HASEL, "hasel" },
-		{ DMI_DMCONTROL, ((1L<<10)-1) << DMI_DMCONTROL_HARTSEL_OFFSET, "hartsel" },
+		{ DMI_DMCONTROL, ((1L<<10)-1) << DMI_DMCONTROL_HARTSELLO_OFFSET, "hartsel" },
```
timsifive: Change the string to "hartsello"

src/target/riscv/riscv-013.c
```diff
@@ -464,7 +465,11 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 
 	int retval = jtag_execute_queue();
 	if (retval != ERROR_OK) {
-		LOG_ERROR("dmi_scan failed jtag scan");
+		static int once = 1;
```
timsifive: Hiding errors makes me nervous. If a target becomes inaccessible I'd prefer a constant stream of errors, just to make sure a user doesn't miss this very important message.

src/target/riscv/riscv-013.c
```diff
@@ -502,6 +507,7 @@ static int dmi_read(struct target *target, uint32_t *value, uint32_t address)
 			LOG_ERROR("failed read from 0x%x, status=%d", address, status);
 			return ERROR_FAIL;
 		}
+		usleep(100000);
```
timsifive: Why not try continuously?

Also note that https://github.com/riscv/riscv-openocd/pull/226 changes this code to use a wall clock timeout in any case.

src/target/riscv/riscv-013.c
```diff
@@ -1494,7 +1500,8 @@ static int init_target(struct command_context *cmd_ctx,
 	generic_info->authdata_write = &riscv013_authdata_write;
 	generic_info->dmi_read = &dmi_read;
 	generic_info->dmi_write = &dmi_write;
-	generic_info->version_specific = calloc(1, sizeof(riscv013_info_t));
+	generic_info->test_compliance = &riscv013_test_compliance;
+        generic_info->version_specific = calloc(1, sizeof(riscv013_info_t));
```
timsifive: Looks like there's an extra space at the beginning of the line.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
 	dmi_write(target, DMI_ABSTRACTCS, abstractcs & DMI_ABSTRACTCS_CMDERR);
 }
+
+#define COMPLIANCE_TEST(b, message) {		\
+    int pass = 0;				\
+    if (b) {					\
+      pass = 1;					\
+      passed_tests ++;				\
+    }									\
+    LOG_INFO("%s test %d (%s)\n", (pass) ? "PASSED":"FAILED",  total_tests, message); \
```
timsifive: Does the OpenOCD style checker like this? I would think it wants exactly one space around the colon, and exactly one space after the comma.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+    assert(pass);							\
+    total_tests ++;							\
+  }
+
+int riscv013_test_compliance(struct target *target) {
+  LOG_INFO("Testing Compliance against RISC-V Debug Spec v0.13");
```
timsifive: OpenOCD style guide mandates 4-space tabs as indents.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  dmi_read(target, &dmcontrol_orig, DMI_DMCONTROL);
+  uint32_t dmcontrol;
+  uint32_t testvar;
+  uint32_t testvar_read;
+  riscv_reg_t value;
+
+  dmcontrol = set_field(dmcontrol_orig, hartsel_mask(target), RISCV_MAX_HARTS-1);
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+  COMPLIANCE_TEST(get_field(dmcontrol, hartsel_mask(target)) == (RISCV_MAX_HARTS-1), "DMCONTROL.hartsel should hold all the harts allowed by HARTSELLEN.");
```
timsifive: The spec explicitly allows implementations that only have a few writable bits in hartsel.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  if (!riscv_rtos_enabled(target)) {
+    LOG_ERROR("Please run with -rtos riscv to run compliance test.");
+    return ERROR_FAIL;
+  }
+
+  int total_tests = 0;
+  int passed_tests = 0;
+
+  uint32_t dmcontrol_orig;
+  dmi_read(target, &dmcontrol_orig, DMI_DMCONTROL);
```
timsifive: I'd hardcode a dmcontrol value here. Otherwise you'll run into trouble if eg. some reset field is already set when you read this.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+  COMPLIANCE_TEST(get_field(dmcontrol, hartsel_mask(target)) == 0, "DMCONTROL.hartsel should hold Hart ID 0");
+
+  // hartreset
+  dmcontrol = set_field(dmcontrol_orig, DMI_DMCONTROL_HARTRESET, 1);
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+  testvar = get_field(dmcontrol, DMI_DMCONTROL_HARTRESET);
+  dmcontrol = set_field(dmcontrol_orig, DMI_DMCONTROL_HARTRESET, 1);
```
timsifive: Typo? Should this be set to 0?

I don't quite understand this testcase.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+  COMPLIANCE_TEST(((testvar == 0) || (get_field(dmcontrol, DMI_DMCONTROL_HARTRESET)) == 0), "DMCONTROL.hartreset can be 0 or RW.");
+
+  // hasel
+  dmcontrol = set_field(dmcontrol_orig, DMI_DMCONTROL_HASEL, 1);
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+  testvar = get_field(dmcontrol, DMI_DMCONTROL_HASEL);
+  dmcontrol = set_field(dmcontrol_orig, DMI_DMCONTROL_HASEL, 1);
```
timsifive: Ditto.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+  COMPLIANCE_TEST(((testvar == 0) || (get_field(dmcontrol, DMI_DMCONTROL_HASEL)) == 0), "DMCONTROL.hasel can be 0 or RW.");
+  //TODO: test that hamask registers exist if hasel does.
+
+  // haltreq
+  riscv_halt_all_harts(target);
+  // Writing haltreq should not cause any problems for a halted hart, but we
+  // should be able to read and write it.
+  dmi_read(target, &dmcontrol, DMI_DMCONTROL);
```
timsifive: riscv_halt_all_harts() already wrote HALTREQ for every hart, and confirmed the halt harted. I don't think you're testing anything extra with the code below.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  dmcontrol |= DMI_DMCONTROL_HALTREQ;
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+  COMPLIANCE_TEST(dmcontrol & DMI_DMCONTROL_HALTREQ, "DMCONTROL.haltreq should be R/W");
+  uint32_t dmstatus, dmstatus_read;
+  do {
```
timsifive: These infinite loops seem pretty user unfriendly, especially in a compliance test that's being run against dubious implementations.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  // Halt the hart again because the target isn't aware that we resumed it.
+  dmcontrol = set_field(dmcontrol, DMI_DMCONTROL_RESUMEREQ, 0);
+  dmcontrol |= DMI_DMCONTROL_HALTREQ;
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  do {
+    dmi_read(target, &dmstatus, DMI_DMSTATUS);
+  } while ((dmstatus & DMI_DMSTATUS_ALLHALTED) == 0);
+  dmcontrol &= ~DMI_DMCONTROL_HALTREQ;
+  dmi_write(target, DMI_DMCONTROL, dmcontrol);
+  // Not clear that this read is required according to the spec.
```
timsifive: I don't think any reads have side effects (besides a few explicitly set up around program buffer and bus access), so they're not required.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+
+    uint32_t nscratch = get_field(hartinfo, DMI_HARTINFO_NSCRATCH);
+    for (unsigned int d = 0; d < nscratch; d++) {
+
+      //TODO: DSCRATCH CSRs should be 64-bit on 64-bit systems.
+      riscv_reg_t testval;
+      for (testval = 0x0011223300112233; testval != 0xDEAD ; testval = testval == 0x0011223300112233 ? ~testval : 0xDEAD ) {
+        COMPLIANCE_TEST(register_write_direct(target, GDB_REGNO_S0, testval) == ERROR_OK, "Need to be able to write S0 in order to test DSCRATCH.");
+        struct riscv_program program32;
+	riscv_program_init(&program32, target);
```
timsifive: Several lines in this blocks aren't indented matching with the rest of the code.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+	riscv_program_csrw(&program32, GDB_REGNO_S0, GDB_REGNO_DSCRATCH + d);
+	riscv_program_csrr(&program32, GDB_REGNO_S1, GDB_REGNO_DSCRATCH + d);
+	riscv_program_fence(&program32);
```
timsifive: You should check that the hart actually implements a Program Buffer of the appropriate size to execute this program.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+      // TODO: datasize
+      // TODO: dataaddr
+    }
+
+  }
+
+  // HALTSUM -- TODO: More than 32 harts
+  // TODO: HALTSUM2, HALTSUM3
+  uint32_t expected_haltsum0 = 0;
+  for (int i = 0; i < riscv_count_harts(target); i +=32){
```
timsifive: Please make sure the OpenOCD style checker is happy:
```git diff `git merge-base riscv HEAD` | ./tools/scripts/checkpatch.pl -```

File: src/target/riscv/riscv-013.c [hunk 25]
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+      if (invert) {testvar = ~testvar;}
+      dmi_read(target, &testvar_read, DMI_PROGBUF0 + i);
+      COMPLIANCE_TEST(testvar_read == testvar, "All reported PROGBUF words must be R/W");
+    }
+  }
+
+  // TODO: Cause and clear all error types
+
+  // COMMAND
+  // TODO: Unclear from the spec whether all these bits need to truly be R/W.
```
timsifive: Seems like it should be a valid implementation to only support writing actually implemented commands, but I don't think the spec says that.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+		  "Illegal COMMAND should result in UNSUPPORTED");
+  dmi_write(target, DMI_ABSTRACTCS, DMI_ABSTRACTCS_CMDERR);
+
+  // Basic Abstract Commands
+  uint32_t command = 0;
+  uint32_t busy;
+  command = set_field(command, AC_ACCESS_REGISTER_SIZE, riscv_xlen(target) > 32 ? 3:2);
+  command = set_field(command, AC_ACCESS_REGISTER_TRANSFER, 1);
+  for (unsigned int i = 1 ; i < 32 ; i = i << 1) {
+    command = set_field(command, AC_ACCESS_REGISTER_REGNO, 0x1000 + GDB_REGNO_ZERO + i);
```
timsifive: You might be able to use access_register_command() here.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+      dmi_write(target, DMI_DATA0 + 1, i + 1);
+    }
+    dmi_write(target, DMI_COMMAND, command);
+    do {
+      dmi_read(target, &testvar_read, DMI_ABSTRACTCS);
+      busy = get_field(testvar_read, DMI_ABSTRACTCS_BUSY);
+    } while (busy);
+    dmi_read(target, &testvar_read, DMI_ABSTRACTCS);
+    COMPLIANCE_TEST(get_field(testvar_read, DMI_ABSTRACTCS_CMDERR) == 0, "GPR Writes should be supported.");
+    dmi_write(target, DMI_DATA0, 0xDEADBEEF);
```
timsifive: Can probably use write_abstract_arg()

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+    dmi_write(target, DMI_COMMAND, command);
+    do {
+      dmi_read(target, &testvar_read, DMI_ABSTRACTCS);
+      busy = get_field(testvar_read, DMI_ABSTRACTCS_BUSY);
+    } while (busy);
+    dmi_read(target, &testvar_read, DMI_ABSTRACTCS);
+    COMPLIANCE_TEST(get_field(testvar_read, DMI_ABSTRACTCS_CMDERR) == 0, "GPR Reads should be supported.");
+    dmi_read(target, &testvar_read, DMI_DATA0);
+    COMPLIANCE_TEST(testvar_read == i, "GPR Reads and writes should be supported.");
+    if (riscv_xlen(target) > 32) {
```
timsifive: Can probably use read_abstract_arg()

timsifive:
I'm working through this, but it's a lot of code...

mwachs5:
Sure, NP. Happy to take a partial review, would like to try to merge it sometime next week.

src/target/riscv/riscv-013.c
```diff
@@ -2926,3 +2933,483 @@ void riscv013_clear_abstract_error(struct target *target)
...
+  // See which bits are actually writable
+  dmi_write(target, DMI_ABSTRACTAUTO, 0xFFFFFFFF);
+  uint32_t abstractauto;
+  dmi_read(target, &abstractauto, DMI_ABSTRACTAUTO);
+  dmi_write(target, DMI_ABSTRACTAUTO, 0x0);
+  if (abstractauto > 0) {
+    testvar = 0;
+    // TODO: This mechanism only works when you have a reasonable sized progbuf, which is not
+    // a true compliance requirement.
+    uint32_t result = riscv_set_register(target, GDB_REGNO_S0, 0);
```
timsifive: The "right way" to set S0 when you're not making it visible to the user of the debugger, is to call register_write_direct(), and to read it with register_read_direct(). That makes no difference in your codebase, but it will once the optimizations merge. One of the optimizations is to cache S0 so we're not reading it over and over again, and I think with the functions you use here you may end up reading a cached S0, not the one that was actually modified.

register_set() and register_get() are for use by the user (probably gdb). Since the user can't execute arbitrary programs, they don't handle that case.

(This problem appears numerous times in this function, I just hadn't realized it until I'd read this far.)

timsifive (changes requested):

src/target/riscv/riscv-013.c
mwachs5: https://github.com/riscv/riscv-openocd/issues/240

src/target/riscv/riscv-013.c
mwachs5: I think at the time there was no check that haltreq bit was actually R/W. But then the spec changed so it's not actually required to be readable. So this test is invalid anyway, I'll fix it.

src/target/riscv/riscv-013.c
mwachs5: Can you add some commentary to those functions prototype in a follow-up PR to explain the differences?

src/target/riscv/riscv-013.c
mwachs5: The spec has been updated since I wrote this test to say that `COMMAND` is `W`

src/target/riscv/riscv-013.c
mwachs5: both of these changes were attempting to avoid the problem when you have a slow simulator, you get a massive log file full of not-too-useful error messages that becomes a problem in and of itself.

src/target/riscv/riscv-013.c
timsifive: #241

mwachs5:
@timsifive this is failing the lint checks for a file I neither wrote nor modified:

```
$ git diff -U20 HEAD~40 | ./tools/scripts/checkpatch.pl --no-signoff -
WARNING: braces {} are not necessary for single statement blocks
#17007: FILE: src/flash/nor/stm32f2x.c:1330:
+	if (stm32x_info->has_optcr2_pcrop) {
+		stm32x_info->option_bytes.optcr2_pcrop = OPTCR2_PCROP_RDP | (~1U << bank->num_sectors);
+	}
WARNING: braces {} are not necessary for single statement blocks
#18594: FILE: src/flash/nor/stm32lx.c:791:
+	if (retval == ERROR_OK && (device_id & 0xfff) == 0x429) {
+		flash_size_in_kb &= 0xff;
+	}
```

timsifive:
This is clearly a problem with the format checking, but the simplest solution I can think of is to simply fix the code by removing those braces. I can do that in a separate PR if you like, but I'm not sure that would fix the build error you're running into.

mwachs5:
@timsifive can you re-review?

src/target/riscv/riscv-013.c
```diff
@@ -3043,3 +3045,472 @@ void riscv013_clear_abstract_error(struct target *target)
...
+	assert(pass);		    \
+	total_tests++;		    \
+}
+
+int riscv013_test_compliance(struct target *target)
+{
+	LOG_INFO("Testing Compliance against RISC-V Debug Spec v0.13");
+
+	if (!riscv_rtos_enabled(target)) {
+		LOG_ERROR("Please run with -rtos riscv to run compliance test.");
```
timsifive, line 3081: That's an odd requirement, especially for a single-hart system. The long-term plan is to get rid of `-rtos riscv` altogether in favor of a target-independent solution.

Can you at least add a comment (or just add it to the message) why this requirement exists?

src/target/riscv/riscv-013.c
```diff
@@ -3043,3 +3045,472 @@ void riscv013_clear_abstract_error(struct target *target)
...
+	uint32_t testvar_read;
+	riscv_reg_t value;
+	RISCV013_INFO(info);
+
+	/* TODO: Support HARTSELLHI as well */
+	dmcontrol = dmcontrol_orig |  DMI_DMCONTROL_HARTSELLO;
+	dmi_write(target, DMI_DMCONTROL, dmcontrol);
+	dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+	COMPLIANCE_TEST(get_field(dmcontrol, DMI_DMCONTROL_HARTSELLO) == (uint32_t) ((1 << info->hartsellen) - 1),
+			"DMCONTROL.hartsello should hold all the harts allowed by its hartsellen.");
```
timsifive: info->hartsellen is configured by writing all ones to hartsel, and reading this back. So I'm not sure this actually tests anything.

src/target/riscv/riscv-013.c
```diff
@@ -3043,3 +3045,472 @@ void riscv013_clear_abstract_error(struct target *target)
...
+	/* This field is optional. Either we can read and write it to 1/0,
+	or it is tied to 0. */
+	dmcontrol = set_field(dmcontrol_orig, DMI_DMCONTROL_HARTRESET, 1);
+	dmi_write(target, DMI_DMCONTROL, dmcontrol);
+	dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+	testvar = get_field(dmcontrol, DMI_DMCONTROL_HARTRESET);
+	dmcontrol = set_field(dmcontrol_orig, DMI_DMCONTROL_HARTRESET, 0);
+	dmi_write(target, DMI_DMCONTROL, dmcontrol);
+	dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+	COMPLIANCE_TEST(((testvar == 0) || (get_field(dmcontrol, DMI_DMCONTROL_HARTRESET)) == 0),
```
timsifive: I think the only thing you care about is that hartreset is currently 0, since you just wrote 0.

testvar can contain either 0 or 1, depending on whether the feature is implemented. At that point there's no point in tracking the value at all. Whether the read is useful is another story. I guess it depends on whether you think that might affect the target in some interesting way.

I'd much prefer a coding style like this, over reusing variables over and over:
```
￼ 	dmi_write(target, DMI_DMCONTROL, set_field(dmcontrol_orig, DMI_DMCONTROL_HARTRESET, 1));
￼ 	dmi_read(target, &dmcontrol, DMI_DMCONTROL);
￼ 	int hartreset_readback = get_field(dmcontrol, DMI_DMCONTROL_HARTRESET);
￼ 	dmi_write(target, DMI_DMCONTROL, set_field(dmcontrol_orig, DMI_DMCONTROL_HARTRESET, 0));
￼ 	dmi_read(target, &dmcontrol, DMI_DMCONTROL);
```

src/target/riscv/riscv-013.c
```diff
@@ -3043,3 +3045,472 @@ void riscv013_clear_abstract_error(struct target *target)
...
+
+	/* hasel */
+	dmcontrol = set_field(dmcontrol_orig, DMI_DMCONTROL_HASEL, 1);
+	dmi_write(target, DMI_DMCONTROL, dmcontrol);
+	dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+	testvar = get_field(dmcontrol, DMI_DMCONTROL_HASEL);
+	dmcontrol = set_field(dmcontrol_orig, DMI_DMCONTROL_HASEL, 0);
+	dmi_write(target, DMI_DMCONTROL, dmcontrol);
+	dmi_read(target, &dmcontrol, DMI_DMCONTROL);
+	COMPLIANCE_TEST(((testvar == 0) || (get_field(dmcontrol, DMI_DMCONTROL_HASEL)) == 0),
```
timsifive: Same as above. testvar can be 0 or 1 so checking it doesn't tell you anything, and hasel must be 0 since that's what you wrote it most recently.

src/target/riscv/riscv-013.c
```diff
@@ -3043,3 +3045,472 @@ void riscv013_clear_abstract_error(struct target *target)
...
+	/* haltreq */
+	riscv_halt_all_harts(target);
+	/* This bit is not actually readable according to the spec, so nothing to check.*/
+
+	/* DMSTATUS */
+	uint32_t dmstatus, dmstatus_read;
+	dmi_read(target, &dmstatus, DMI_DMSTATUS);
+	dmi_write(target, DMI_DMSTATUS, ~dmstatus);
+	dmi_read(target, &dmstatus_read, DMI_DMSTATUS);
+	COMPLIANCE_TEST(dmstatus_read == dmstatus, "DMSTATUS is R/O");
```
timsifive: This pattern happens several times. Can you refactor it into a function that does this for any DMI register?

src/target/riscv/riscv-013.c
```diff
@@ -3043,3 +3045,472 @@ void riscv013_clear_abstract_error(struct target *target)
...
+		}
+	}
+
+	/* TODO: Cause and clear all error types */
+
+	/* COMMAND
+	According to the spec, this register is only W, so can't really check the read result.
+	But at any rate, this is not legal and should cause an error. */
+	dmi_write(target, DMI_COMMAND, 0xAAAAAAAA);
+	dmi_read(target, &testvar_read, DMI_COMMAND);
```
timsifive: Is this helpful?

src/target/riscv/riscv-013.c
```diff
@@ -3043,3 +3045,472 @@ void riscv013_clear_abstract_error(struct target *target)
...
+	See which bits are actually writable */
+	dmi_write(target, DMI_ABSTRACTAUTO, 0xFFFFFFFF);
+	uint32_t abstractauto;
+	uint32_t busy;
+	dmi_read(target, &abstractauto, DMI_ABSTRACTAUTO);
+	dmi_write(target, DMI_ABSTRACTAUTO, 0x0);
+	if (abstractauto > 0) {
+		/* This mechanism only works when you have a reasonable sized progbuf, which is not
+		a true compliance requirement. */
+		if (info->progbufsize >= 3) {
```
timsifive: Just roll this into the `abstractauto > 0` check above.

src/target/riscv/riscv-013.c
```diff
@@ -3043,3 +3045,472 @@ void riscv013_clear_abstract_error(struct target *target)
...
+			dmi_write(target, DMI_ABSTRACTAUTO, 0xFFFFFFFF);
+			dmi_read(target, &abstractauto, DMI_ABSTRACTAUTO);
+			uint32_t autoexec_data = get_field(abstractauto, DMI_ABSTRACTAUTO_AUTOEXECDATA);
+			uint32_t autoexec_progbuf = get_field(abstractauto, DMI_ABSTRACTAUTO_AUTOEXECPROGBUF);
+			for (unsigned int i = 0; i < 12; i++) {
+				dmi_read(target, &testvar_read, DMI_DATA0 + i);
+				do {
+					dmi_read(target, &testvar_read, DMI_ABSTRACTCS);
+					busy = get_field(testvar_read, DMI_ABSTRACTCS_BUSY);
+				} while (busy);
```
timsifive: This should probably have a timeout.

src/target/riscv/riscv-013.c
```diff
@@ -3043,3 +3045,472 @@ void riscv013_clear_abstract_error(struct target *target)
...
+	dmi_read(target, &abstractauto, DMI_ABSTRACTAUTO);
+
+	/* Pulse reset. */
+	target->reset_halt = true;
+	riscv_set_current_hartid(target, 0);
+	COMPLIANCE_TEST(ERROR_OK == assert_reset(target), "Must be able to assert NDMRESET");
+	COMPLIANCE_TEST(ERROR_OK == deassert_reset(target), "Must be able to deassert NDMRESET");
+
+	/* Verify that most stuff is not affected by ndmreset. */
+	dmi_read(target, &testvar_read, DMI_COMMAND);
```
timsifive: command is not readable.

src/target/riscv/riscv-013.c
```diff
@@ -3043,3 +3045,472 @@ void riscv013_clear_abstract_error(struct target *target)
...
+
+	COMPLIANCE_TEST(riscv_halt_reason(target, 0) == RISCV_HALT_INTERRUPT,
+			"After NDMRESET halt, DCSR should report cause of halt");
+
+	/* DMACTIVE -- deasserting DMACTIVE should reset all the above values. */
+
+	/* Toggle dmactive */
+	dmi_write(target, DMI_DMCONTROL, 0);
+	dmi_write(target, DMI_DMCONTROL, DMI_DMCONTROL_DMACTIVE);
+	dmi_read(target, &testvar_read, DMI_COMMAND);
```
timsifive: command isn't readable.

timsifive (changes requested):
The lack of error checking is going to cause the clang use-of-unitialized-values check to go crazy. I realize it's painful to write if ()... around every line. I'm not sure what the right answer is, but I wanted to mention it.

src/target/riscv/riscv-013.c
mwachs5: It may no longer be necessary, I'll try just removing the check

src/target/riscv/riscv-013.c
mwachs5: right, this predates that. I'll remove this check.

src/target/riscv/riscv-013.c
mwachs5: one of the points of this test is to write all bits that are potentially writable, even if checking them allows both values.

src/target/riscv/riscv-013.c
mwachs5: We are checking that it causes an error... is that what you were asking about?

src/target/riscv/riscv-013.c
mwachs5: right, that was clarified later, so I'll update the test.

src/target/riscv/riscv-013.c
mwachs5: Will fix as above

mwachs5:
@timsifive 

> "The lack of error checking is going to cause the clang use-of-unitialized-values check to go crazy. I realize it's painful to write if ()... around every line. I'm not sure what the right answer is, but I wanted to mention it."

Not quite sure what you mean... can you give an example of what I need to do to fix what you're talking about?

src/target/riscv/riscv-013.c
mwachs5: I think the reason I required that -rtos be enabled is because we assume we control all harts in this test. But perhaps those two concerns are orthogonal?

src/target/riscv/riscv-013.c
mwachs5: but sure I will simplify the check to somethign that makes sense.

src/target/riscv/riscv-013.c
mwachs5: ah I see, you mean why are we reading DMI_COMMAND into testvar_read, I'll remove that dmi_read

src/target/riscv/riscv-013.c
mwachs5: I'd rather not, since someone could fill in the guts of this with a better check of abstract auto that doesn't require progbufsize >=3... they're seperate concerns

src/target/riscv/riscv-013.c
timsifive: The only way to do a proper test is to have control of all harts. That doesn't really fit with the OpenOCD model, but we can more-or-less shoehorn it in. Given that OpenOCD is single-threaded, I don't think you need to do anything to ensure full control, except either inform the user that all harts are now in an unknown state, or try hard to restore their state. (I'd pick the former.)

src/target/riscv/riscv-013.c
timsifive: "one of the points of this test is to write all bits that are potentially writable, even if checking them allows both values."

That seems worth a comment somewhere.

src/target/riscv/riscv-013.c
timsifive: That makes sense.

timsifive:
How to use the clang static checker is described in http://openocd.zylin.com/gitweb?p=openocd.git;a=blob;f=HACKING;h=0d24957e141268a997274cdbb4ca0e4394e6968a;hb=refs/heads/master#l51

As an example, if dmi_read() fails it returns an error code without writing data to the pointer passed to it. If you don't check the error and then proceed to look at the data at that pointer, the static checker will complain (rightfully) that you're using a potentially uninitialized variable.

mwachs5:
Got it. Does my COMPLIANCE_MUST_PASS macros basically address that issue?

src/target/riscv/riscv-013.c
```diff
@@ -3058,6 +3058,21 @@ void riscv013_clear_abstract_error(struct target *target)
 	total_tests++;		    \
 }
 
+#define COMPLIANCE_MUST_PASS(b) COMPLIANCE_TEST(ERROR_OK == (b), "Regular calls must return ERROR_OK")
+
+#define COMPLIANCE_READ(target, addr, value) COMPLIANCE_MUST_PASS(dmi_read(target, addr, value))
+#define COMPLIANCE_WRITE(target, addr, value) COMPLIANCE_MUST_PASS(dmi_write(target, addr, value))
+
+#define COMPLIANCE_CHECK_RO(target, addr)                               \
```
timsifive, line 3066: Can this be a function?

src/target/riscv/riscv-013.c
```diff
@@ -3077,71 +3092,48 @@ int riscv013_test_compliance(struct target *target)
...
-	dmi_read(target, &dmcontrol, DMI_DMCONTROL);
-	COMPLIANCE_TEST(get_field(dmcontrol, DMI_DMCONTROL_HARTSELLO) == (uint32_t) ((1 << info->hartsellen) - 1),
-			"DMCONTROL.hartsello should hold all the harts allowed by its hartsellen.");
-
-	dmcontrol = set_field(dmcontrol_orig, DMI_DMCONTROL_HARTSELLO, 0);
-	dmi_write(target, DMI_DMCONTROL, dmcontrol);
-	dmi_read(target, &dmcontrol, DMI_DMCONTROL);
-	COMPLIANCE_TEST(get_field(dmcontrol, DMI_DMCONTROL_HARTSELLO) == 0,
-			"DMCONTROL.hartsello should hold Hart ID 0");
+	/* All the bits of HARTSEL is covered by the examine sequence. */
```
timsifive: is -> are

timsifive (dismissed):

timsifive:
Yes, I think checking errors everywhere probably fixes it.

src/target/riscv/riscv-013.c
mwachs5: well, not as currently written because the passed_test macro is a local variable (yes its not beautiful). So then I can't use any of the macros.

src/target/riscv/riscv-013.c
mwachs5: thanks fixed

src/target/riscv/riscv-013.c
timsifive: Alright. I'm not a fan of macros, but if you like it this way that's good enough.

timsifive (approved):

Issue #138, opened 2017-11-10
The new remote_bitbang code fails to compile on mingw-w64


ilg-ul:
```console
/Host/Work/openocd/openocd.git/src/jtag/drivers/remote_bitbang.c: In function 'remote_bitbang_fill_buf':
/Host/Work/openocd/openocd.git/src/jtag/drivers/remote_bitbang.c:63:2: error: implicit declaration of function 'fcntl' [-Werror=implicit-function-declaration]
  fcntl(remote_bitbang_fd, F_SETFL, O_NONBLOCK);
  ^~~~~
/Host/Work/openocd/openocd.git/src/jtag/drivers/remote_bitbang.c:63:27: error: 'F_SETFL' undeclared (first use in this function)
  fcntl(remote_bitbang_fd, F_SETFL, O_NONBLOCK);
                           ^~~~~~~
/Host/Work/openocd/openocd.git/src/jtag/drivers/remote_bitbang.c:63:27: note: each undeclared identifier is reported only once for each function it appears in
/Host/Work/openocd/openocd.git/src/jtag/drivers/remote_bitbang.c:63:36: error: 'O_NONBLOCK' undeclared (first use in this function)
  fcntl(remote_bitbang_fd, F_SETFL, O_NONBLOCK);
                                    ^~~~~~~~~~
/Host/Work/openocd/openocd.git/src/jtag/drivers/remote_bitbang.c: In function 'remote_bitbang_rread':
/Host/Work/openocd/openocd.git/src/jtag/drivers/remote_bitbang.c:157:27: error: 'F_SETFL' undeclared (first use in this function)
  fcntl(remote_bitbang_fd, F_SETFL, 0);
                           ^~~~~~~
```

For my new OpenOCD release I had to manually revert to the original code.

https://github.com/gnu-mcu-eclipse/openocd/blob/gnu-mcu-eclipse-dev/src/jtag/drivers/remote_bitbang.c

timsifive:
Ugh. mingw appears to not support POSIX. Looks like there's already some code in OpenOCD to abstract that away. See eg. socket_nonblock() in src/helper/replacements.h.

PR #144, opened 2017-12-13
Try to build on Windows.

timsifive:
Change-Id: I93843274dba56809fd43358a219c3fe2ba9e8a3e

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,14 +28,14 @@ matrix:
                       - gcc-multilib
 
         - os: linux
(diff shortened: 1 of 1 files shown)


ilg-ul:
I'm curious if you manage to compile the new remote_bitbang code with mingw [#138], since I had to revert it to pass the build.

timsifive:
It won't, which is why I'm doing this in the first place.

ilg-ul:
I don't know your development environment, but a local mingw might be faster, not only that you do not have to wait for Travis to schedule your job, but you can restart a make from the place it failed. at least this is how I run my builds.

PR #149, opened 2017-12-19
Send gdb an XML target description that contains only a list of registers we think exist on this target

timsifive:
Works for 0.11 and 0.13. (Some more code is now shared, which required some reworking of 0.13 examine().)
Report ABI names for GPRs and FPRs.
Custom CSRs can be exposed with the `riscv expose_csrs` command.

The main registers that still hang around and look annoying are all the performance counters. The spec provides no way to detect whether they're implemented, and there are a lot of them.


palmer-dabbelt (approved):
LGTM.  Thanks!

Issue #150, opened 2017-12-21
Syntax error in src/target/riscv/riscv.c


TommyMurphyTM1234:
https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv.c#L1203

`Line 1203: if isspace(c) {`

should be 

`Line 1203: if (isspace(c)) {`

As it is the following results when compiling:

```
libtool: compile:  i686-w64-mingw32-gcc -DHAVE_CONFIG_H -I. -I/Host/Work/openocd/openocd.git -D__USE_MINGW_ANSI_STDIO -I/Host/Work/openocd/openocd.git/src -I./src -I/Host/Work/openocd/openocd.git/src/helper -DPKGDATADIR=\"\" -DBINDIR=\"bin\" -I/Host/Work/openocd/openocd.git/jimtcl -I./jimtcl -Werror -m32 -pipe -Wall -Wstrict-prototypes -Wformat-security -Wshadow -Wextra -Wno-unused-parameter -Wbad-function-cast -Wcast-align -Wredundant-decls -Werror -g -O2 -MT src/target/riscv/riscv.lo -MD -MP -MF src/target/riscv/.deps/riscv.Tpo -c /Host/Work/openocd/openocd.git/src/target/riscv/riscv.c -o src/target/riscv/riscv.o
/Host/Work/openocd/openocd.git/src/target/riscv/riscv.c: In function 'riscv_set_expose_csrs':
/Host/Work/openocd/openocd.git/src/target/riscv/riscv.c:1203:7: error: expected '(' before 'isspace'
    if isspace(c) {
       ^~~~~~~
/Host/Work/openocd/openocd.git/src/target/riscv/riscv.c:1203:7: error: statement with no effect [-Werror=unused-value]
    if isspace(c) {
       ^~~~~~~~~~
/Host/Work/openocd/openocd.git/src/target/riscv/riscv.c:1200:12: error: unused variable 'high' [-Werror=unused-variable]
   unsigned high = 0;
            ^~~~
/Host/Work/openocd/openocd.git/src/target/riscv/riscv.c:1199:8: error: unused variable 'parse_low' [-Werror=unused-variable]
   bool parse_low = true;
        ^~~~~~~~~
/Host/Work/openocd/openocd.git/src/target/riscv/riscv.c:1198:12: error: unused variable 'low' [-Werror=unused-variable]
   unsigned low = 0;
            ^~~
cc1: all warnings being treated as errors
[...]

PR #151, opened 2017-12-21
Use parens after if.

timsifive:
I'm surprised this built with gcc before.

Fixes Issue #150.

Change-Id: I24d2957783c66ad53d5b532a4e930349a2059a97


palmer-dabbelt (approved):
Thtanks

PR #152, opened 2017-12-21
riscv: update defines for *havereset and dmerr fields

mwachs5:
Update defines generated from the spec to match the current 0.13-DRAFT version of the spec.


mwachs5:
same as #155

PR #154, opened 2017-12-21
[WIP] Test macOS build on Travis

mwachs5:
Is there any reason we can't test a macoS build as well, to avoid things like #153?

I'm not too familiar with this use of Travis, so this PR is going to be a WIP while I figure it out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,13 @@ matrix:
                   packages:
                       - binutils-mingw-w64-i686 gcc-mingw-w64-i686 g++-mingw-w64-i686
 
+        - os: macos
+          env:
+              - BUILD=macos
+              - EXECUTABLE=openocd
+          compiler: gcc
+
+
 script:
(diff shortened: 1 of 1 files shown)


mwachs5:
well, I am not sure this did anything interesting because it didn't complain about #147

timsifive:
When I tried a while ago, it seemed that MacOS builds on Travis took forever to/didn't start. I decided clang was close enough. Evidently there are still some differences.

mwachs5:
yeah but whatever they are aren't really exposed by this PR anyway.

timsifive:
Looking at the Travis build, it only built one variation (instead of the 6 we should now be up to including MacOS), and it was using some Ubuntu gcc despite having BUILD=macos

mwachs5:
yeah, I didn't want to waste Travis resources testing the ones we knew worked already. Anyway I cleaned up this PR but not sure it's worth merging. I can look more into my macos configuration that had the problem and see if there is a way to duplicate it more closely

PR #155, opened 2017-12-21
Update debug_defines to the one used with spike.

timsifive:
Change-Id: I627c6ee557d98239227324c33f9b89f6280cbf93


mwachs5 (approved):
same as #152 right? But I approve both :)

timsifive:
Same, yes. I didn't realize you had made the same change as well.

PR #156, opened 2017-12-22
fix fespi flash after registers were renamed.

timsifive:
Also made a number of functions static.

### DIFF
--- a/src/flash/nor/fespi.c
+++ b/src/flash/nor/fespi.c
@@ -598,7 +598,7 @@ struct algorithm_steps {
 	uint8_t **steps;
 };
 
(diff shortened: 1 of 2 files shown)


palmer-dabbelt:
I think @mwachs5 was in the process of testing this.

mwachs5:
Is this the fix you think you need to fix Flashing? Or is this just one piece of it?

mwachs5:
This worked for me to flash HiFive1 with a few demos.

mwachs5 (approved):
works for me to flash HiFive1

timsifive:
This should be the entire fix to fix HiFive1 flashing.

Issue #39, opened 2017-04-19
V11 dbus_read should work more like V13.


mwachs5:
V11 has the following code in dbus_read, which results in either the  DBUS_OP_READ being sent twice or you getting stale data back (because it is conditioned on address_in).

V13 does it better by changing the second read to a NOP. We should fix V11 code to work the same way.

```
	do {
		status = dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
		if (status == DBUS_STATUS_BUSY) {
			increase_dbus_busy_delay(target);
		}
	} while (((status == DBUS_STATUS_BUSY) || (address_in != address)) &&
			i++ < 256);
```

timsifive:
Is there an observable advantage to making this change? IIRC the 0.11 spec explicitly says that reads may not have side effects.

mwachs5:
Are you relying on the previous `address` being reported from the last scan (which could have been infinitely long ago) so you don't actually *have* to do the second scan to get the read data? The fact that you might get infinitely old data isn't super intuitive, but if that's your design choice and there aren't any cases where that's actually a problem, maybe a comment is necessary to clarify the intent.

PR #166, opened 2017-12-29
In dmi_read(), ignore stale data.

timsifive:
Fixes Issue #39. Reduces performance about 1%.

Change-Id: I3bfc71cb8e632ecce35c6adeb744177300d54b98

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -455,6 +455,9 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
 	dbus_status_t status;
 	uint16_t address_in;
 
+	// First, discard stale data
+	dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
+
 	unsigned i = 0;
 	do {
 		status = dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -459,7 +459,11 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
 	 * If there happens to be a busy state pending, we will clear it after we
 	 * discover it as part of the NOP scan. */
 	for (unsigned i = 0; i < 256; i++) {
-		dbus_scan(target, NULL, NULL, DBUS_OP_READ, address, 0);
+		status = dbus_scan(target, NULL, NULL, DBUS_OP_READ, address, 0);
+		if (status == DBUS_STATUS_BUSY) {
+			increase_dbus_busy_delay(target);
+			continue;
+		}
 		status = dbus_scan(target, &address_in, &value, DBUS_OP_NOP, address, 0);
 		if (status == DBUS_STATUS_BUSY)
 			increase_dbus_busy_delay(target);
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -455,7 +455,11 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
 	dbus_status_t status;
 	uint16_t address_in;
 
-	// First, discard stale data
+	// First scan sends the desired address/operation. This is 
+	// slightly inefficient as, if the previous read was to the same
+	// address, one could have simply used the data directly. However,
+	// that data may may have been captured significantly earlier,
+	// and could be considered stale at the start of this operation.
 	dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
 
 	unsigned i = 0;
@@ -463,7 +467,7 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
 		status = dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
 		if (status == DBUS_STATUS_BUSY)
 			increase_dbus_busy_delay(target);
-	} while (((status == DBUS_STATUS_BUSY) || (address_in != address)) &&
+	} while (((status == DBUS_STATUS_BUSY)) &&
 			i++ < 256);
 
 	if (status != DBUS_STATUS_SUCCESS)
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -455,6 +455,9 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
 	dbus_status_t status;
 	uint16_t address_in;
 
+	/* Assume dbus is idle and there aren't any busy/error states pending.
+	 * If there happens to be a busy state pending, we will clear it after we
+	 * discover it as part of the NOP scan. */
 	for (unsigned i = 0; i < 256; i++) {
 		dbus_scan(target, NULL, NULL, DBUS_OP_READ, address, 0);
 		status = dbus_scan(target, &address_in, &value, DBUS_OP_NOP, address, 0);
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -455,23 +455,21 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
 	dbus_status_t status;
 	uint16_t address_in;
 
-	// First scan sends the desired address/operation. This is 
-	// slightly inefficient as, if the previous read was to the same
-	// address, one could have simply used the data directly. However,
-	// that data may may have been captured significantly earlier,
-	// and could be considered stale at the start of this operation.
-	dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
-
-	unsigned i = 0;
-	do {
-		status = dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
+	for (unsigned i = 0; i < 256; i++) {
+		dbus_scan(target, NULL, NULL, DBUS_OP_READ, address, 0);
+		status = dbus_scan(target, &address_in, &value, DBUS_OP_NOP, address, 0);
 		if (status == DBUS_STATUS_BUSY)
 			increase_dbus_busy_delay(target);
-	} while (((status == DBUS_STATUS_BUSY)) &&
-			i++ < 256);
+		else if (address_in != address)
+			LOG_ERROR("Tried to read from 0x%x, but actually read from 0x%x",
+					address, address_in);
+		else
+			break;
+	}
 
 	if (status != DBUS_STATUS_SUCCESS)
-		LOG_ERROR("failed read from 0x%x; value=0x%" PRIx64 ", status=%d\n", address, value, status);
+		LOG_ERROR("failed read from 0x%x; value=0x%" PRIx64 ", status=%d",
+				address, value, status);
 
 	return value;
 }
@@ -486,7 +484,7 @@ static void dbus_write(struct target *target, uint16_t address, uint64_t value)
 			increase_dbus_busy_delay(target);
 	}
 	if (status != DBUS_STATUS_SUCCESS)
-		LOG_ERROR("failed to write 0x%" PRIx64 " to 0x%x; status=%d\n", value, address, status);
+		LOG_ERROR("failed to write 0x%" PRIx64 " to 0x%x; status=%d", value, address, status);
 }
 
 /*** scans "class" ***/


src/target/riscv/riscv-011.c
```diff
@@ -455,6 +455,9 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
 	dbus_status_t status;
 	uint16_t address_in;
 
+	// First, discard stale data
```
mwachs5: This comment is a bit misleading... you aren't JUST discarding stale data... you're also actually sending the address that you want to read.

src/target/riscv/riscv-011.c
```diff
@@ -455,12 +455,19 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
...
+	// First scan sends the desired address/operation. This is 
+	// slightly inefficient as, if the previous read was to the same
+	// address, one could have simply used the data directly. However,
+	// that data may may have been captured significantly earlier,
+	// and could be considered stale at the start of this operation.
+	dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
+
 	unsigned i = 0;
 	do {
 		status = dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
```
mwachs5: Also, I think if we go this route we need to change this to DBUS_OP_NOP. Otherwise, you won't be detecting the second BUSY that could be caused by *this* DBUS_OP_READ. This code (and the write code) assumes that the first operation always is accepted, which is only possible if it wasn't busy. In fact we should probably check that after the first dbus_scan.

mwachs5: I just looked again at the dbus_write code. There you are explicitly checking that the write was accepted in one loop, but we never check to make sure that the write was actually completed (BUSY = 0). In theory, if dbus_read and dbus_write both use DBUS_OP_NOP for a second loop and make sure BUSY = 0, there is no reason that you would ever *get* BUSY for the first loop, but otherwise it could be BUSY from the previous read or write.

mwachs5 (changes requested):
As it is, this code isn't right. In the original issue, I believe my main complaint was that it was nonintuitive. A clarifying comment on the original code would probably be sufficient to close the issue, but if you want to change the code, further adjustments to this are needed.

timsifive:
I think it's right now.

Thanks for continuing to check my work here.

src/target/riscv/riscv-011.c
```diff
@@ -455,16 +455,21 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
 	dbus_status_t status;
 	uint16_t address_in;
 
-	unsigned i = 0;
-	do {
-		status = dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
+	for (unsigned i = 0; i < 256; i++) {
+		dbus_scan(target, NULL, NULL, DBUS_OP_READ, address, 0);
```
mwachs5: I think you should add a status = dbus_scan on the first and check that it wasn't busy. There may be no case in which it would be busy, but it's at least a sanity that may help us find bugs.

timsifive: But if it returns, that's a stale result. It doesn't tell you anything about whether the scan that we're actually doing right now is going to succeed or not.

mwachs5: mmmm, this still doesn't quite make sense to me. What you really want to do is make sure that 1) the read command was seen and 2) that you got the result back. So you need to check for busy twice.

I suggest you first put a loop of `dbus_scan(DBUS_OP_READ)` where you check for busy. Then you would do  a second loop of  `dbus_scan(DBUS_OP_NOP)` until you didn't get busy (again), in which case you know you got the actual read data.

I'm checking to see if that is how v013 does this.

mwachs5: yeah, confirmed that v013 has 2 separate loops.

mwachs5: Though now that I think about it, in v11 NOP wasn't really a thing. So I think you could still get BUSY response even for a NOP.

mwachs5: Oh just saw your comment.

BUSY = 0 just means "there was no other operation in flight when you asked me to do this one". So that means two things: 1) I am free to start the operation that you just asked me to do, otherwise I would have ignored it, and 2) the data that you read was the result from the previous operation.

By doing a read with 2 loops, you are covering the two meanings: the first one makes sure that the read goes through, and the second one makes sure that the data you saw was completed.

In your original code you were basically doing the same thing except you were using the address match as the check for case 1) and just re-sending the request over and over. Which is why I was pretty much fine with it, just thought it was confusing without a comment.

timsifive: Won't the second scan get a busy response if it happens before the read has completed? I'm thinking this is what happens:
```
scan read
Update-DR, read begins
Capture-DR, captures either busy or result
scan NOP, scanning out either busy or result
```

mwachs5: Ah yeah I see what you mean. Let me go back to the v13 code to convince myself what I really meant with #39

mwachs5: Argh. I had a huge write-up with fast/slow examples but accidentally closed the window.

Long story short, you need to consider that Capture-DR happens BEFORE Update-DR. The busy bit is recorded during Capture-DR, and if it is set, nothing gets sent across the DBUS/DMI during Update-DR.

So your example above is more correct as:

```
Scan READ:
  Capture-DR: busy & unused result captured
  Update-DR: read begins (assuming busy =0)
Scan NOP:
   Capture-DR: busy & actual result captured (assuming busy = 0)
   Update-DR: NOP gets sent out
```

So yes, either of those scans can get busy responses, which is why you need to keep doing them in two separate loops. If the Scan READ reports a busy response, it won't actually send anything out over DBUS/DMI. If the Scan NOP reports a busy response, it won't actually have the right data.

Your old code worked (fine, except for the issue raised that it was kind of weird) like:
```
Scan READ:
  Capture-DR: busy/ result & address captured
  Update-DR: Useless read (or read that is used next time) sent into the DMI
```
or
```
Scan READ:
  Capture-DR: busy/ result & (wrong) address captured
  Update-DR: Actual read to correct address sent over DMI
Scan READ:
  Capture-DR: busy/ result & address captured
  Update-DR: Useless read (or read that is used next time) sent into the DMI
```

So yes, you would only have to do one READ as long as the address matched.

timsifive:
This code handles being busy upon entry (which should never be the case),
although not in the most efficient way. It will discover the busy status
after the NOP scan, and then clear it. It has the nice property that we
could some day optimize it to put both scans in a single USB transaction,
since the second one doesn't depend on the first one.

src/target/riscv/riscv-011.c
```diff
@@ -455,16 +455,24 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
...
 	uint16_t address_in;
 
-	unsigned i = 0;
-	do {
-		status = dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
+	/* Assume dbus is idle and there aren't any busy/error states pending.
+	 * If there happens to be a busy state pending, we will clear it after we
+	 * discover it as part of the NOP scan. */
+	for (unsigned i = 0; i < 256; i++) {
+		dbus_scan(target, NULL, NULL, DBUS_OP_READ, address, 0);
```
mwachs5: I'm sorry, this first scan can absolutely encounter busy. The next scan can *create* the busy condition (v11 doesn't specify that DBUS_OP_NOP isn't allowed to generate a BUSY result. In fact I think DBUS_OP_NOP just does a read as well). If the first scan encounters busy, the DTM doesn't send anything over DMI. Then the DBUS_OP_NOP needlessly is sent to see if you got the result. Then misleadingly you could interpret the result of a DBUS_OP_NOP as the result of the DBUS_OP_READ. Probably this happens to work, but it's weird.

Here is what I mean. 
Let's say you first read from FOO, and then read from BAR.

Here's the FOO read:
```
dbus_scan(DBUS_OP_READ, FOO)
   -> busy=0, as recorded in CAPTURE_DR.
   -> during UPDATE_DR, the READ is sent across DMI. This corresponds to the fact that busy=0.
...
   -> read completes on DMI.
dbus_scan(DBUS_OP_NOP, FOO)
   -> since the read completed, CAPTURE_DR gets busy = 0 and the read data, and address = FOO.
   -> Update DR *still sends the NOP operation into DMI* This can result in busy for the next operation!
```
Since busy was 0, address = FOO, this read was correctly interpreted as having finished. So you break out of the loop.

Now you want to read from BAR
*Let's say you do this really fast and the second FOO DMI operation has not yet completed.*

```
dbus_scan(DBUS_OP_READ, BAR)
   -> result *is* busy=1, as recorded in CAPTURE_DR, because the FOO NOP has not completed.
   -> during UPDATE_DR, the READ is NOT sent across DMI. This corresponds to the reported busy=1.
...
   -> FOO NOP finally completes on DMI.
dbus_scan(DBUS_OP_NOP, BAR)
   -> since the *NOP* completed, CAPTURE_DR gets busy = 0. But, busy is sticky so the reported busy=1 since you haven't cleared it yet. (It took me a while to realize that the stickyness of busy was the only reason this made any real sense).
   -> Update DR does nothing because you haven't cleared the busy yet.
```
You observe that busy = 1 so you finally clear it and go back through the loop again. This seems... convoluted. It costs nothing to check the first busy result and clear it when it first affects you. I don't understand, why not do it?

timsifive:
On Thu, Jan 4, 2018 at 10:51 AM, Megan Wachs <notifications@github.com>
wrote:

> *@mwachs5* commented on this pull request.
> ------------------------------
>
> In src/target/riscv/riscv-011.c
> <https://github.com/riscv/riscv-openocd/pull/166#discussion_r159727505>:
>
> > @@ -455,16 +455,24 @@ static uint64_t dbus_read(struct target *target, uint16_t address)
>  	dbus_status_t status;
>  	uint16_t address_in;
>
> -	unsigned i = 0;
> -	do {
> -		status = dbus_scan(target, &address_in, &value, DBUS_OP_READ, address, 0);
> +	/* Assume dbus is idle and there aren't any busy/error states pending.
> +	 * If there happens to be a busy state pending, we will clear it after we
> +	 * discover it as part of the NOP scan. */
> +	for (unsigned i = 0; i < 256; i++) {
> +		dbus_scan(target, NULL, NULL, DBUS_OP_READ, address, 0);
>
> I'm sorry, this first scan can absolutely encounter busy.
>
It can, but we're assuming it won't. None of this code is supposed to leave
the bus in a busy/error state.

> The next scan can *create* the busy condition (v11 doesn't specify that
> DBUS_OP_NOP isn't allowed to generate a BUSY result. In fact I think
> DBUS_OP_NOP just does a read as well). If the first scan encounters busy,
> the DTM doesn't send anything over DMI. Then the DBUS_OP_NOP needlessly is
> sent to see if you got the result. Then misleadingly you could interpret
> the result of a DBUS_OP_NOP as the result of the DBUS_OP_READ.
>
The busy state is sticky, so there is no confusion:
"The previous operation is still in progress. The data scanned into dbus in
this access will be ignored. This status is sticky and can be cleared by
writing dbusreset in dtmcontrol."

> Here is what I mean.
>
> Let's say you first read from FOO, and then read from BAR.
>
> Here's the FOO read:
>
> dbus_scan(DBUS_OP_READ, FOO)
>    -> busy=0, as recorded in CAPTURE_DR.
>    -> during UPDATE_DR, the READ is sent across DMI. This corresponds to the fact that busy=0.
> ...
>    -> read completes on DMI.
> dbus_scan(DBUS_OP_NOP, FOO)
>    -> since the read completed, CAPTURE_DR gets busy = 0 and the read data, and address = FOO.
>    -> Update DR *still sends the NOP operation into DMI* This can result in busy for the next operation!
>
> Since busy was 0, address = FOO, this read was correctly interpreted as
> having finished. So you break out of the loop.
>
> Now you want to read from BAR
> *Let's say you do this really fast and the second FOO DMI operation has
> not yet completed.*
>
> dbus_scan(DBUS_OP_READ, BAR)
>    -> result *is* busy=1, as recorded in CAPTURE_DR, because the FOO NOP has not completed.
>    -> during UPDATE_DR, the READ is NOT sent across DMI. This corresponds to the reported busy=1.
> ...
>    -> FOO NOP finally completes on DMI.
> dbus_scan(DBUS_OP_NOP, BAR)
>    -> since the *NOP* completed, CAPTURE_DR gets busy = 0. But, busy is sticky so the reported busy=1 since you haven't cleared it yet. (It took me a while to realize that the stickyness of busy was the only reason this made any real sense).
>    -> Update DR does nothing because you haven't cleared the busy yet.
>
> You observe that busy = 1 so you finally clear it and go back through the
> loop again. This seems... convoluted. It costs nothing to check the first
> busy result and clear it when it first affects you. I don't understand, why
> not do it?
>
The code works as-is, except that it's not optimal in a case that probably
never happens. Adding more code doesn't improve anything.
But I'll add it because this discussion has gone on for too long already,
about a change that has no observable effect.

Tim

PR #167, opened 2017-12-29
Add config files for SiFive RISC-V hardware.

timsifive:
Copied from https://github.com/gnu-mcu-eclipse/openocd

Change-Id: Ia0b3e192ca8b3bae6035623d605c9980e9bccd2c

### DIFF
--- a/tcl/board/sifive-coreplexip-e31-arty.cfg
+++ b/tcl/board/sifive-coreplexip-e31-arty.cfg
@@ -1,0 +1,31 @@
+# JTAG adapter setup
+adapter_khz     10000
+
+interface ftdi
+ftdi_device_desc "Olimex OpenOCD JTAG ARM-USB-TINY-H"
+ftdi_vid_pid 0x15ba 0x002a
+
+ftdi_layout_init 0x0808 0x0a1b
+ftdi_layout_signal nSRST -oe 0x0200
+#ftdi_layout_signal nTRST -data 0x0100 -oe 0x0100
+ftdi_layout_signal LED -data 0x0800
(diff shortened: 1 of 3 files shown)


palmer-dabbelt (approved):

ilg-ul:
you are using older files. 

the recent version of GNU MCU Eclipse OpenOCD use the now 'official' board names:

- sifive-hifive1.cfg
- sifive-e31arty.cfg
- sifive-e51arty.cfg

(https://github.com/gnu-mcu-eclipse/openocd/tree/gnu-mcu-eclipse-dev/tcl/board)

it would be good to use the same files.

also the usage for the arty files is slightly different; since the Arty board does not have an on-board JTAG, and requires an external one, including the Olimex ARM-USB-TINY-H in the board definition is not the best way.

the ARM-USB-TINY-H is already defined in OpenOCD as 'interface/ftdi/olimex-arm-usb-tiny-h.cfg', so the metadata associated with the Arty board defines the OpenOCD configuration as:

https://github.com/micro-os-plus/sifive-arty-boards-xpack/blob/36399e3ea63af9d28b46d0d01f594e31a3f0c479/boards-xcdl.json#L78

timsifive:
Can you submit a pull request to fix this? I just copied some files that Drew pointed me to.

ilg-ul:
#168

timsifive:
Thanks. With that merged, should I delete the 3 files added by this PR?

ilg-ul:
> should I delete the 3 files added by this PR?

In my opinion, yes.

PR #168, opened 2017-12-29
Add config files for the SiFive boards

ilg-ul:
- the HiFive1 board definition includes the FTDI interface
- the Arty boards require external interface definitions


timsifive (approved):
Looks good. Thanks.

PR #170, opened 2018-01-02
Parse 64-bit CRC addrs even on 32-bit hosts

timsifive:
Change-Id: I38720163eff292b2c24f25da4e25feb8245ff672


palmer-dabbelt:
Does this mean that things no longer blow up when I mix 32-bit and 64-bit targets?

palmer-dabbelt (approved):

timsifive:
Unlikely. This fixes running a CRC of a 64-bit target on a 32-bit host.

Issue #108, opened 2017-09-14
riscv013_get_register can't return an error


timsifive:
That's a problem if a user requests to read a non-existent register.

palmer-dabbelt:
ya, sorry, I was lazy when implementing some of these.  IIRC OpenOCD didn't pass back the correct error messages anyway, but we should refactor all the function pointers in that structure to be able to return error codes.

mwachs5:
This is related to #115 . If we revert the change in gdb_server.c, @palmer-dabbelt is correct that these errors would be meaningless anyway. What *should* happen when a user requests to read a non-existent register? Or should this be made a moot point by some other communication between OpenOCD/GDB (e.g. passing an XML file about which CSRs actually exist)?

timsifive:
> What should happen when a user requests to read a non-existent register?

The user should get an error.

Note that the XML exchange happens once, so in order to use that OpenOCD would have to attempt to access every CSR to check whether it exists. (Not the end of the world, but will be pretty slow on simulators.) And a platform might have CSRs that only are accessible part of the time.

ilg-ul:
why do you insist on considering the CSRs to belong to the same class as the general registers? they are actually system/peripheral registers. most single core architectures expose them as memory mapped registers; favouring multi-core, RISC-V could not do this and mapped them to a separate address space. but they are not general registers.

what other architectures show all system registers together with the general registers? Cortex-M definitely does **not** do this.

timsifive:
As I understand the code, CSRs are not part of the general register group. Only the 32 data registers and the PC are. See https://github.com/riscv/riscv-binutils-gdb/blob/riscv-binutils-2.29/gdb/riscv-tdep.c#L756

ilg-ul:
attached are some screenshots from the 'General Registers' view in Eclipse, when debugging the HiFive1 board.

As you can see, not only the float registers are there, but all 4096 CSs are also there.

I don't know if the mis-configuration is in GDB or OpenOCD, but, in my opinion, the CSRs **should not** be displayed as 'General Registers'.

![screen shot 2017-09-30 at 00 30 01](https://user-images.githubusercontent.com/362732/31036805-84f58ca8-a576-11e7-99bd-31beab9ca77e.png)
![screen shot 2017-09-30 at 00 28 39](https://user-images.githubusercontent.com/362732/31036772-63c517f6-a576-11e7-9c54-3d8b31ab1baa.png)
[...]

PR #181, opened 2018-01-09
Propagate register read errors

timsifive:
Let the caller know when reading a register failed.

Add code that temporarily changes mstatus if an FPR is read without any F bits set. (The same change is not necessary for writing, because in that case the hardware will automatically set the F bit.)

Fixes Issue #108

### DIFF
--- a/src/rtos/riscv_debug.c
+++ b/src/rtos/riscv_debug.c
@@ -281,7 +281,12 @@ static int riscv_get_thread_reg_list(struct rtos *rtos, int64_t thread_id, char 
 	*hex_reg_list[0] = '\0';
 	for (size_t i = 0; i < n_regs; ++i) {
 		if (riscv_has_register(rtos->target, thread_id, i)) {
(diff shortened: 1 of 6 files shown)


mwachs5:
"The same change is not necessary for writing, because in that case the hardware will automatically set the F bit."

What do you mean? Would that mean that the debugger is modifying the machine state and not restoring it? Couldn't that have an effect on the program being debugged?

timsifive:
If you write the FPR registers (in code), the mstatus.FS bits become set.

AFAIK the reason the FS bits exist is to save a kernel context switch time. It can ignore the FPRs if the FS bits are clear. In that use case, it feels like the right thing for the debugger is to allow the bits to be set, so the kernel can appropriately save context.

I could imagine a user wanting to change an FPR without affecting the FS bits, but that feels like more  of a corner case. A user can always change the FPR, and then reset mstatus to whatever value they want.

FWIW, the 0.11 code has had this behavior for a long time.

palmer-dabbelt (approved):
I agree with Tim: leaving the dirty bit seems like the sane thing to do here.

PR #188, opened 2018-01-19
Incorporate review feedback from OpenOCD team

timsifive:
Remove unused file handle, and rename the only remaining one to make
more sense.
Close file descriptor if initialization fails.
Review: http://openocd.zylin.com/#/c/4312/
Change-Id: I383567aaadb1aa59d86f814eba8bc65f24e91928

### DIFF
--- a/src/jtag/drivers/remote_bitbang.c
+++ b/src/jtag/drivers/remote_bitbang.c
@@ -288,7 +288,7 @@ static int remote_bitbang_init(void)
 	if (remote_bitbang_fd < 0)
 		return remote_bitbang_fd;
 
-	remote_bitbang_file = fdopen(remote_bitbang_fd, "w");
+	remote_bitbang_file = fdopen(remote_bitbang_fd, "w+");
 	if (remote_bitbang_file == NULL) {
 		LOG_ERROR("fdopen: failed to open write stream");
 		close(remote_bitbang_fd);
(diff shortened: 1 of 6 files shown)


src/jtag/drivers/remote_bitbang.c
```diff
@@ -289,17 +288,10 @@ static int remote_bitbang_init(void)
...
-	remote_bitbang_in = fdopen(remote_bitbang_fd, "r");
-	if (remote_bitbang_in == NULL) {
-		LOG_ERROR("fdopen: failed to open read stream");
-		close(remote_bitbang_fd);
-		return ERROR_FAIL;
-	}
-
-	remote_bitbang_out = fdopen(remote_bitbang_fd, "w");
-	if (remote_bitbang_out == NULL) {
+	remote_bitbang_file = fdopen(remote_bitbang_fd, "w");
```
palmer-dabbelt: I thought you needed "w+" to open for both read and write?

timsifive: You're right. It's been this way forever so clearly it works, but "w+" is more correct. I fixed it.

timsifive:
This just merged upstream, so no need to separately merge it here as well. I'll take care of it when we pull from upstream again. (Hopefully this week.)

PR #197, opened 2018-01-26
Track misa per-hart even in -rtos mode

### LINKED ISSUE #194: Can't set HBreaks on non-homogeneous cores
In a system with heterogenous cores (for example, one core that is M/U-mode only and other cores that are M/S/U), running with `-rtos riscv` assigns a single `$misa` value for the entire target. This is likely to cause a number of problems so may not be reasonable to expect it to work in all cases, but in particular this one should be easy to fix...

If I want to set an `hbreak` on all the cores, the `tdata1` logic tries to set the M/S/U bits according to the common $misa. On the core with no S mode support it fails and then erases the trigger, which is annoying, since the `hbreak` would work fine.

It's also annoying that GDB doesn't report any sort of failure in this case, it just doesn't actually hit the breakpoint until some other hart hits it.

```
Debug: 82338 130893 riscv-013.c:1055 register_read_direct(): [0] reg[0x7e2] = 0x288000000000104c
Debug: 82339 130893 riscv.c:1571 riscv_get_register_on_hart(): [0] tdata1: 288000000000104c
Debug: 82340 130893 riscv.c:374 maybe_add_trigger_t2(): tdata1=0x288000000000104c
Debug: 82341 130893 riscv.c:379 maybe_add_trigger_t2(): Trigger doesn't support what we need; After writing 0x288000000000105c to tdata1 it contains 0x288000000000104c
Debug: 82342 130893 riscv.c:1557 riscv_set_register_on_hart(): [0] tdata1 <- 0
Debug: 82343 130893 riscv-013.c:1876 riscv013_set_register(): writing 0x0 to register tdata1 on hart 0
```

timsifive:
This works around some side effects of the -rtos hack, namely that we
were unable to set hardware breakpoints on harts whose misa differed
from the first one. There may be other bugs like this one lurking
elsewhere. The only proper solution is for gdb to have a better user
interface when talking to a server that exposes multiple targets, but
that's a very big project.

This fixes #194.

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1572,11 +1572,11 @@ static int examine(struct target *target)
 	}
 	LOG_DEBUG("Discovered XLEN is %d", riscv_xlen(target));
 
-	if (read_csr(target, &r->misa, CSR_MISA) != ERROR_OK) {
+	if (read_csr(target, &r->misa[0], CSR_MISA) != ERROR_OK) {
 		const unsigned old_csr_misa = 0xf10;
 		LOG_WARNING("Failed to read misa at 0x%x; trying 0x%x.", CSR_MISA,
 				old_csr_misa);
-		if (read_csr(target, &r->misa, old_csr_misa) != ERROR_OK) {
+		if (read_csr(target, &r->misa[0], old_csr_misa) != ERROR_OK) {
 			/* Maybe this is an old core that still has $misa at the old
 			 * address. */
 			LOG_ERROR("Failed to read misa at 0x%x.", old_csr_misa);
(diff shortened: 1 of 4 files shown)


mwachs5:
This looks good, but do you  know why this error:
https://github.com/riscv/riscv-openocd/pull/197/files#diff-3f1d7a8875de5c486b524a60b3652904R337

Doesn't get reflected at all in GDB? Having that error message in GDB would have helped figure out what was going on much earlier (especially since this message is only LOG_DEBUG).

timsifive:
I don't know. Can you file a separate issue for that?

mwachs5 (commented):
This looks good, but I haven't tested it out yet.

mwachs5:
FYI this hasn't slipped off my radar, still trying to find a time to test it out.

mwachs5 (approved):
Tested this version, it works.

PR #198, opened 2018-01-27
Detect hartsellen, limiting which harts we probe

timsifive:
Tested with doctored spike with hartsellens of 0, 1, 3, and 10.

Change-Id: I97f57c7d03b076792d5ecd66545d9b9e853ed515


palmer-dabbelt (approved):
I'm assuming there was a change to the debug spec here?

timsifive:
Yes, the spec was changed.

PR #209, opened 2018-02-07
Added fence.i after each memory load

p12nGH-zz:


timsifive:
Are there case where we want to execute just a fence and not a fence.i? I'd be inclined to just add the fence_i to execute_fence(), and rename it to execute_fences().

Did you run `make all` in riscv-tests/debug to test for any regressions?

p12nGH-zz:
It is a good idea to have execute_fences that does both types of fences.

mwachs5 (dismissed):

src/target/riscv/program.c
```diff
@@ -121,14 +121,11 @@ int riscv_program_csrw(struct riscv_program *p, enum gdb_regno s, enum gdb_regno
...
-int riscv_program_fence_i(struct riscv_program *p)
+int riscv_program_fences(struct riscv_program *p)
 {
-	return riscv_program_insert(p, fence_i());
-}
-
-int riscv_program_fence(struct riscv_program *p)
-{
-	return riscv_program_insert(p, fence());
+	if (riscv_program_insert(p, fence()) != ERROR_OK) return ERROR_FAIL;
```
timsifive: I believe the OpenOCD style guide does not allow for putting `return` on the same line is `if`. You can check with tools/checkpatch.sh (although that will only check the last git commit, and exit early until you add a Signed-off line).

timsifive (dismissed):
I prefer to not modify riscv_program_fence_i(), but instead call
riscv_program_fence() and riscv_program_fence_i() from execute_fences(). This will allow us in the future to add some logic to execute_fences() to deal with a target that only has 1 word of program buffer, by executing first one and then the other.

src/target/riscv/riscv-013.c
```diff
@@ -1436,6 +1436,30 @@ static int execute_fence(struct target *target)
 	return result;
 }
 
+static int execute_fence_i(struct target *target)
```
timsifive, line 1439: Do you still need this?

I think you can just replace all calls to execute_fence() with execute_fences().

timsifive (approved):

palmer-dabbelt (approved):
I think leaving `execute_fence_i()` doesn't hurt.

PR #216, opened 2018-02-20
target/riscv: add some switch fallthrough comments

kaspar030:
This PR adds some comments that prevent gcc 7 errors like:

```
src/target/riscv/riscv-013.c: In function ‘read_abstract_arg’:
src/target/riscv/riscv-013.c:605:10: error: this statement may fall through [-Werror=implicit-fallthrough=]
    value |= ((uint64_t) dmi_read(target, DMI_DATA0 + offset + 1)) << 32;
          ^~
src/target/riscv/riscv-013.c:606:3: note: here
   case 32:
   ^~~~
```


timsifive (approved):
Thanks for cleaning this up!

PR #218, opened 2018-02-27
Add `riscv authdata_read` and `riscv authdata_write` commands to support arbitrary authentication through TCL scripts

timsifive:
Example usage (very simple challenge-response protocol):
```
init

set challenge [ocd_riscv authdata_read]
riscv authdata_write [expr $challenge + 1]

reset halt
```

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -141,7 +141,14 @@ typedef struct {
 	int abs_chain_position;
 	/* Indicates we already reset this DM, so don't need to do it again. */
 	bool was_reset;
+	/* Targets that are connected to this DM. */
+	struct list_head target_list;
 } dm013_info_t;
+
+typedef struct {
+	struct list_head list;
+	struct target *target;
+} target_list_t;
 
 typedef struct {
 	/* Number of address bits in the dbus register. */
@@ -198,6 +205,7 @@ typedef struct {
 	/* The width of the hartsel field. */
 	unsigned hartsellen;
 
+	/* DM that provides access to this target. */
 	dm013_info_t *dm;
 } riscv013_info_t;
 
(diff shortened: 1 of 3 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -525,6 +588,19 @@ static int dmi_write(struct target *target, uint16_t address, uint64_t value)
 	return ERROR_OK;
 }
 
+int dmstatus_read(struct target *target, uint32_t *dmstatus,
+		bool authenticated)
+{
+	*dmstatus = dmi_read(target, DMI_DMSTATUS);
+	if (authenticated && !get_field(*dmstatus, DMI_DMSTATUS_AUTHENTICATED)) {
+		LOG_ERROR("Debugger is not authenticated to target Debug Module. "
+				"(dmstatus=0x%x). Use `riscv authdata_read` and "
```
mwachs5, line 597: is it `riscv_authdata_read` or `riscv authdata_read` ? The comment mixes the two so seems likely that one is wrong

src/target/riscv/riscv-013.c
```diff
@@ -1111,57 +1187,31 @@ static int register_read_direct(struct target *target, uint64_t *value, uint32_t
...
+	while (1) {
+		uint32_t value;
+		if (dmstatus_read(target, &value, false) != ERROR_OK)
+			return ERROR_FAIL;
+		if (dmstatus)
+			*dmstatus = value;
+		if (!get_field(value, DMI_DMSTATUS_AUTHBUSY))
+			break;
+		if (time(NULL) - start > riscv_command_timeout_sec) {
+			LOG_ERROR("Timed out after %ds waiting for authbusy to go low (dmstatus=0x%x). "
```
mwachs5, line 1202: wouldn't this be a completely different timeout?

mwachs5:
What does `ocd_riscv authdata_read` mean? Is that different than `riscv authdata_read`?

src/target/riscv/riscv-013.c
timsifive: Fixed.

src/target/riscv/riscv-013.c
timsifive: Yes, this could be a separate timeout. But nobody wants to configure a zillion timeouts (or even one), so I prefer to reuse the same one if possible. Do you think the authentication timeout would be significantly different from the regular command timeout? In real hardware I imagine 10 seconds would work for either of them, but in simulation maybe not.

mwachs5: I mean, they are totally different mechanisms potentially. But sure, if one timeout is simpler I guess that's OK. they can always be configured dynamically before/after authentication.

timsifive:
> What does ocd_riscv authdata_read mean? Is that different than riscv authdata_read?

It is different. I think what happens with the ocd_ variant is that instead of printing out the result, it's captured by tcl. Just `riscv authdata_read` always "returns" 0 while `ocd_riscv authdata_read` actually returns the value that was read. This is apparently a long-standing OpenOCD hack (I'm not sure what problem it solves) that isn't likely to go anywhere.

mwachs5 (approved):
I'd like to test this against https://github.com/freechipsproject/rocket-chip/pull/1126, but don't really have an ETA for that, so if this passes all tests for systems with no authentication I'm fine with merging it

PR #224, opened 2018-03-07
Remove debug code.

timsifive:
I tried to upstream this, but it's turning out to be too painful to be
worth the effort. (See http://openocd.zylin.com/#/c/4451/)

Change-Id: Ie5c9d8ea84a4c8ab55ce1a6ba7df260d3963291d

### DIFF
--- a/src/jtag/core.c
+++ b/src/jtag/core.c
@@ -836,68 +836,7 @@ int default_interface_jtag_execute_queue(void)
 		return ERROR_FAIL;
 	}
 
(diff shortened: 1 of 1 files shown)


timsifive:
Looks like this may be upstreamed after all.
I'll leave this open in case it doesn't.

timsifive:
Upstream has stalled again, and I don't want to spend my time pushing a change I don't really care about. Can I get a review for this?

palmer-dabbelt (approved):

PR #226, opened 2018-03-16
Notice when a hart is reset while it's being debugged, and let the user know that it happened

### LINKED ISSUE #211: Debugging a hart not robust across non-debugger controlled reset
Consider the case where openOCD is debugging a hart and one of its actions cause the hart to be reset (e.g writing a memory location which leads to soft reset), or someone presses an external reset button. OpenOCD doesn't re-halt the hart or otherwise gracefully recover (it works fine if OpenOCD knowingly caused the reset with `ndreset`).

Most of the problems could be avoided if OpenOCD left `haltreq` set until it actually wanted to resume. This would cause a hart which was being debugged and got reset to halt again even if it was reset out of the debugger's control.

timsifive:
I think this resolves #211.


src/target/riscv/riscv-013.c
```diff
@@ -2557,9 +2563,17 @@ static bool riscv013_is_halted(struct target *target)
 	if (dmstatus_read(target, &dmstatus, true) != ERROR_OK)
 		return false;
 	if (get_field(dmstatus, DMI_DMSTATUS_ANYUNAVAIL))
-		LOG_ERROR("hart %d is unavailiable", riscv_current_hartid(target));
+		LOG_ERROR("hart %d is unavailable", riscv_current_hartid(target));
 	if (get_field(dmstatus, DMI_DMSTATUS_ANYNONEXISTENT))
 		LOG_ERROR("hart %d doesn't exist", riscv_current_hartid(target));
+	if (get_field(dmstatus, DMI_DMSTATUS_ANYHAVERESET)) {
+		int hartid = riscv_current_hartid(target);
+		LOG_INFO("hart %d unexpectedly reset!", hartid);
```
mwachs5: there are a lot of changes in the PR... is this basically the only functional change?

src/target/riscv/riscv-013.c
```diff
@@ -2567,6 +2566,14 @@ static bool riscv013_is_halted(struct target *target)
 		LOG_ERROR("hart %d is unavailable", riscv_current_hartid(target));
 	if (get_field(dmstatus, DMI_DMSTATUS_ANYNONEXISTENT))
 		LOG_ERROR("hart %d doesn't exist", riscv_current_hartid(target));
+	if (get_field(dmstatus, DMI_DMSTATUS_ANYHAVERESET)) {
+		int hartid = riscv_current_hartid(target);
+		LOG_INFO("hart %d unexpectedly reset!", hartid);
+		/* TODO: Can we make this more obvious to eg. a gdb user? */
+		dmi_write(target, DMI_DMCONTROL,
```
mwachs5: if this was the case, shouldn't you do something like replace breakpoints? Or halt? Or...?

mwachs5 (commented):
Not sure what the point of just telling the user is that reset happened... seems like you'd want to halt the hart or reset breakpoints or something.

src/target/riscv/riscv-013.c
timsifive: Halting would be reasonable if the hart was halted when it was reset. I'll add that. Whether breakpoints should be replaced is much more iffy. Upon reset the code could be completely different, and breakpoints completely meaningless. The DRAM controller might not be set up so memory might not even be accessible. Really the debugger (gdb/eclipse) should know a bit more and be able to deal with a reset, but unfortunately there isn't even an interface to tell them that the hart was reset.

src/target/riscv/riscv-013.c
timsifive: There are a few, all to do with looking for the havereset bits.
Also, we now check dmstatus in poll() while the target is halted, which previously didn't happen.

src/target/riscv/riscv-013.c
mwachs5: yeah, good point. I guess it makes sense if you were single-stepping through the code in GDB if you could at least get it to print a message through GDB then the user could re-add breakpoints as necessary. I guess if you wanted to get fancy you could read the BP registers and see if they were still set or not so that the GDB `info b` command would give the right information, but I'm not sure if OpenOCD->GDB can "cancel" breakpoints in this way.

src/target/riscv/riscv-013.c
```diff
@@ -2557,9 +2563,25 @@ static bool riscv013_is_halted(struct target *target)
...
-		LOG_ERROR("hart %d doesn't exist", riscv_current_hartid(target));
+		LOG_ERROR("Hart %d doesn't exist.", riscv_current_hartid(target));
+	if (get_field(dmstatus, DMI_DMSTATUS_ANYHAVERESET)) {
+		int hartid = riscv_current_hartid(target);
+		LOG_INFO("Hart %d unexpectedly reset!", hartid);
+		/* TODO: Can we make this more obvious to eg. a gdb user? */
+		uint32_t dmcontrol = DMI_DMCONTROL_DMACTIVE |
+			DMI_DMCONTROL_ACKHAVERESET;
+		dmcontrol = set_field(dmcontrol, hartsel_mask(target), hartid);
+		/* If we had been halted when we reset, request another halt. If we
```
mwachs5, line 2768: this is the fundamental point... do you HAVE to let the hart run away after reset before noticing and halting it again? If you leave haltreq set the whole time, then it's possible for the debug module to stop it right away, not just when the debugger notices that it got reset.

Does that make sense? I know it's sort of a fundamental change, but that was what I was really getting at in #211

timsifive: You can leave haltreq set the whole time. I agree with you and what you want with #211, I just haven't implemented it yet.

palmer-dabbelt (dismissed):

src/target/riscv/riscv.c
```diff
@@ -188,6 +188,8 @@ int riscv_reset_timeout_sec = DEFAULT_RESET_TIMEOUT_SEC;
 bool riscv_use_scratch_ram;
 uint64_t riscv_scratch_ram_address;
 
+bool riscv_havereset_not_supported;
```
mwachs5: what is the default of this?

timsifive: The default is false, which is why I put a negation in the variable name.

So the default behavior is for OpenOCD to assume havereset is implement.

timsifive:
So I think this is a real improvement with no negative side effects. Can I get a review?

mwachs5:
I'd like to try this out a bit on my side, which isn't going to happen until next week. In the meantime, can you comment on what sorts of targets you've tested against? Did you try this with the following cases for HW which does support `hasreset`:
* hitting reset button on a non-halted hart while debugger is polling 
* hitting reset button on a halted hart that is waiting for debugger commands
* using the debugger, do an action that causes reset (e.g. memory map a GPIO pin to the RESET input, write to it using debugger so that reset happens while executing the abstract command

What is the behavior in each case (or, are handling these cases out of scope of this PR)?

For a target that doesn't support `hasresetack`, is the idea that this PR wouldn't really change anything about the behavior in any of those cases?

timsifive:
This code barely helps with pushing the reset button, because that also resets the DTM. I've just been testing the hart itself spontaneously resetting.

I'd like to merge it because it improves things, but it doesn't actually help in many cases of #211.

src/target/riscv/riscv-013.c
```diff
@@ -1579,57 +1567,74 @@ static int deassert_reset(struct target *target)
...
-			if (time(NULL) - start > riscv_reset_timeout_sec) {
-				LOG_ERROR("Hart didn't halt coming out of reset in %ds; "
-						"dmstatus=0x%x; "
-						"Increase the timeout with riscv set_reset_timeout_sec.",
-						riscv_reset_timeout_sec, dmstatus);
-				return ERROR_FAIL;
-			}
-			target->state = TARGET_HALTED;
-		} while (get_field(dmstatus, DMI_DMSTATUS_ALLHALTED) == 0);
+	for (int i = 0; i < riscv_count_harts(target); ++i) {
```
mwachs5: I don't get what this loop is doing.

mwachs5: Isn't NDMRESET still asserted at this point? Why does it say "waiting for hart to halt out of reset?

timsifive: The line you selected loops over all of the harts, in case `-rtos` is being used.

In there there is a loop, starting on line 1594, that waits for the hart to halt. This is the same as in the                                                                           old code. What's new is proper handling of `-rtos`.

NDMRESET is no longer asserted here. It's deasserted in the write on line 1574.

src/target/riscv/riscv-013.c
```diff
@@ -1579,57 +1567,74 @@ static int deassert_reset(struct target *target)
...
 
-		control = set_field(control, DMI_DMCONTROL_HALTREQ, 0);
-		dmi_write(target, DMI_DMCONTROL, control);
+		if (target->reset_halt) {
+			LOG_DEBUG("Waiting for hart %d to halt out of reset.", index);
+			do {
+				if (dmstatus_read(target, &dmstatus, true) != ERROR_OK)
+					return ERROR_FAIL;
+				if (time(NULL) - start > riscv_reset_timeout_sec) {
+					LOG_ERROR("Hart %d didn't halt coming out of reset in %ds; "
```
mwachs5: So, I ran into this on the current master (I think this code is basically the same) when debugging something that takes a few seconds to come out of reset. The `dmstatus_read` following the reset deassertion can actually take the entire time for the `riscv_reset_timeout_sec`.  In other words, the `dmi_read` won't complete until you come out of reset. So then I get this error about "Hart didn't come out of reset in XYZ seconds", but who cares, because you aren't actually using the user-specified timeout value for that. Anyway in this case I do need to bump the timeout I guess, but it's just sort of misleading. For this particular `dmstatus_read` you should be applying the timeout the user gave you with `set_reset_timeout_sec`. And shouldn't you be checking `DMI_STATUS_ALLHALTED` before you conclude the hart still hasn't come out of reset and return ERROR_FAIL?

mwachs5: in other words, something about the order of these checks seems off.

mwachs5: I see now that your code above now uses the riscv_command_timeout_sec in the dmi_op loop instead of the hard-coded 256. I like this, someone who is asserting reset can just set riscv_command_timeout_sec = riscv_reseet_timeout_sec temporarily. So I'll just merge this PR into my current one to see if it solves my problem.

mwachs5: This is what I mean... trying this out: https://github.com/riscv/riscv-openocd/pull/137/commits/8ce4f787ca429cedac4da9c8931ac702d270ebb4

timsifive: I like the idea of using the reset timeout during reset, but I don't like mucking with the global variable like that. I'll take a stab at an implementation and update this PR.

src/target/riscv/riscv-013.c
```diff
@@ -549,11 +549,11 @@ static int dmi_op(struct target *target, uint32_t *data_in, int dmi_op,
 					status);
 			return ERROR_FAIL;
 		}
-		if (time(NULL) - start > riscv_command_timeout_sec) {
+		if (time(NULL) - start > timeout_sec) {
 			LOG_ERROR("dmi.op is still busy after %d seconds. The target is "
 					"either really slow or broken. You could increase the "
 					"timeout with riscv set_command_timeout_sec.",
```
mwachs5: is this message true in the case that this fails when you use a different comment?

mwachs5: Just a thought, could you return an ERROR_TIMEOUT and put the message in the caller?

src/target/riscv/riscv-013.c
```diff
@@ -1592,7 +1607,8 @@ static int deassert_reset(struct target *target)
 		if (target->reset_halt) {
 			LOG_DEBUG("Waiting for hart %d to halt out of reset.", index);
 			do {
-				if (dmstatus_read(target, &dmstatus, true) != ERROR_OK)
+				if (dmstatus_read_timeout(target, &dmstatus, true,
+							riscv_reset_timeout_sec) != ERROR_OK)
 					return ERROR_FAIL;
 				if (time(NULL) - start > riscv_reset_timeout_sec) {
```
mwachs5: You'll basically never get this error message in the case of a timeout, because you'll hit the timeout in the `dmstatus_read_timeout`, print an unhelpful message for the user, and never print the `increase riscv_reset_timeout_sec` message.

src/target/riscv/riscv-013.c
mwachs5: The general approach seems good, but I think there is still something off in the order of the checks and how you report the fix to the user.

src/helper/log.h
```diff
@@ -149,6 +149,7 @@ extern int debug_level;
  */
 #define ERROR_FAIL						(-4)
 #define ERROR_WAIT						(-5)
+#define ERROR_TIMEOUT					(-6)
```
mwachs5: strange spacing here?

timsifive: Probably because github shows 8-space tabs, but OpenOCD code uses 4-space tabs.

mwachs5 (approved):

PR #228, opened 2018-03-20
Add `riscv set_prefer_sba`

timsifive:
This allows a user to tell OpenOCD to prefer system bus access for
memory access, which can be useful for testing, or when there really is
a difference in behavior.

Change-Id: I8c2f15b89a2ccdae568c68ee743b75a74f9ad6bd

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2183,7 +2183,7 @@ static int read_memory(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, uint8_t *buffer)
 {
 	RISCV013_INFO(info);
-	if (info->progbufsize >= 2)
+	if (info->progbufsize >= 2 && !riscv_prefer_sba)
 		return read_memory_progbuf(target, address, size, count, buffer);
 
 	if ((get_field(info->sbcs, DMI_SBCS_SBACCESS8) && size == 1) ||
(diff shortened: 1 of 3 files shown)


src/target/riscv/riscv.h
```diff
@@ -127,6 +127,8 @@ extern int riscv_reset_timeout_sec;
 extern bool riscv_use_scratch_ram;
 extern uint64_t riscv_scratch_ram_address;
 
+extern bool riscv_prefer_sba;
```
mwachs5, line 130: so this is global for all targets? Probably fine since this isn't expected to be used much, though it seems like it would make sense as a per-target thing.

src/target/riscv/riscv-013.c
```diff
@@ -2550,8 +2553,9 @@ static int write_memory(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, const uint8_t *buffer)
 {
 	RISCV013_INFO(info);
-	if (info->progbufsize >= 2)
+	if (info->progbufsize >= 2 && !riscv_prefer_sba)
```
mwachs5, line 2556: What if I prefer SBA and have no SBA support? I guess this fails then. It might be nice to make it work even if you preferred SBA but don't support SBA, but again this isn't expected to be used much.

src/target/riscv/riscv.h
timsifive: I don't know how to do per-target settings. There may be an example of it in the code somewhere, but the things I can imagine aren't very pretty.

src/target/riscv/riscv-013.c
timsifive: Then a few lines down (line 2570) the memory write will happen using the program buffer.

mwachs5: ah gotcha, looks good.

timsifive:
So can I merge this?

palmer-dabbelt (approved):

mwachs5 (approved):

PR #230, opened 2018-03-23
Make m*deleg regs conditional on U/S/N

timsifive:
Change-Id: I544fc15625400d8ad64d4a65f0fc9d77f428ca84

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2328,8 +2328,7 @@ int riscv_init_registers(struct target *target)
 					/* "In systems with only M-mode, or with both M-mode and
 					 * U-mode but without U-mode trap support, the medeleg and
 					 * mideleg registers should not exist." */
-					r->exist = (riscv_supports_extension(target, 'S') ||
-							riscv_supports_extension(target, 'U')) &&
+					r->exist = riscv_supports_extension(target, 'S') ||
 						!riscv_supports_extension(target, 'N');
 					break;
 			}
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2329,7 +2329,7 @@ int riscv_init_registers(struct target *target)
 					 * U-mode but without U-mode trap support, the medeleg and
 					 * mideleg registers should not exist." */
 					r->exist = riscv_supports_extension(target, 'S') ||
-						!riscv_supports_extension(target, 'N');
+						riscv_supports_extension(target, 'N');
 					break;
 			}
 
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv.c
```diff
@@ -2323,6 +2323,15 @@ int riscv_init_registers(struct target *target)
 				case CSR_SATP:
 					r->exist = riscv_supports_extension(target, 'S');
 					break;
+				case CSR_MEDELEG:
```
mwachs5, line 2326: this looks good, do you generally also try to read the SEDELEG or SIDELEG registers? The same limitation about 'N' extension applies there.

mwachs5 (approved):

mwachs5 (changes requested):
actually wait I think the logic is wrong...

src/target/riscv/riscv.c
```diff
@@ -2323,6 +2323,15 @@ int riscv_init_registers(struct target *target)
...
+				case CSR_MIDELEG:
+					/* "In systems with only M-mode, or with both M-mode and
+					 * U-mode but without U-mode trap support, the medeleg and
+					 * mideleg registers should not exist." */
+					r->exist = (riscv_supports_extension(target, 'S') ||
```
mwachs5: this should be `r->exist = (supports(S) || supports (N))`. If you have `N`, you SHOULD have this register, which is not what the code currently does.

src/target/riscv/riscv.c
mwachs5: I guess they are not there at all, maybe they should be with the caveat they only appear if ` supports(S) && supports(N)`

src/target/riscv/riscv.c
timsifive: Right. I tried to derive this again. Can you check my logic, starting with what it says in the spec (I think):
                    r->exist = !((M && !U && !S) || (M && U && !N))
                             = !(M && !U && !S) && !(M && U && !N)
                             = (!M || U || S) && (!M || !U || N)
                             = (U || S) && (!U || N)

src/target/riscv/riscv.c
timsifive: s*deleg aren't in riscv-opcodes, which is where my list of registers comes from. Should they be?

mwachs5: I think since nothing implements `N` yet they're probably missing

src/target/riscv/riscv.c
mwachs5: I think the phrase "or with both M-Mode and U-Mode  but without U-mode trap support" has an implied "only"... as in, "only M-Mode and U-Mode, but without U-mode trap support".

Also there are other implied things like you can't have S without U, and you can't have N without U (presumably stated somewhere else in the spec).

timsifive:
Can I get a review for this?

mwachs5 (approved):
LGTM. I haven't tested it but the logic looks right now!

timsifive:
Thanks, Megan!

PR #232, opened 2018-04-02
Add gdb_report_register_access_error command

timsifive:
This replaces the earlier mechanism which would propagate errors only
for targets that decided they wanted to. It was suggested by Matthias
Welwarsky from the OpenOCD team.

Change-Id: Ibe8e97644abb47aff26d74b8280377d42615a4d3


mwachs5 (approved):
lgtm

PR #235, opened 2018-04-04
Tests for SBA feature

rmac-sifive:
Adds OpenOCD command `riscv test_sba_config_reg <illegal addr>`


src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2815,162 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
 	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, 0);
 }
 
+static int riscv013_test_sba_config_reg(struct target *target, uint32_t address)
```
timsifive: Why is this limited to 32-bit addresses?

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2815,162 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+    for(int i = 0; i < 100; i++){
+      uint32_t addr = 0x80000000 + (i << sbaccess);
+      write_memory_sba_simple(target,addr,i,sbcs_orig);
+    }
+
+    for(uint32_t i = 0; i < 100; i++){
+      uint32_t addr = 0x80000000 + (i << sbaccess);
+      uint32_t val = read_memory_sba_simple(target,addr,sbcs_orig);
+      if(i != val) {
+        LOG_ERROR("System Bus Access Test: Error reading non-autoincremented address %x, expected val = %d, read val = %d",addr,i,val);
```
timsifive: If you print this error, below it will still print out that the test passed.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2815,162 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+  // Test 1: Simple write/read test, no address autoincrement
+  sbcs = set_field(sbcs_orig,DMI_SBCS_SBAUTOINCREMENT,0);
+  dmi_write(target,DMI_SBCS,sbcs);
+
+  for(int sbaccess = 0; sbaccess < 0x5; sbaccess++) {
+    sbcs = set_field(sbcs_orig,DMI_SBCS_SBACCESS,sbaccess);
+    dmi_write(target,DMI_SBCS,sbcs);
+
+    for(int i = 0; i < 100; i++){
+      uint32_t addr = 0x80000000 + (i << sbaccess);
```
timsifive: Are you assuming that 0x80000000 is a valid address? I don't think that has to be true on all RISC-V cores.

src/target/riscv/riscv-013.c
timsifive: Don't call this variable "address." Something like "bad_address" is much better. Having both addr and address in the same function is pretty confusing.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2815,162 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+
+    sbcs = set_field(sbcs_orig,DMI_SBCS_SBREADONDATA,1);
+
+    dmi_write(target,DMI_SBADDRESS0,0x80000000);
+    for(uint32_t i = 0; i < 100; i++){
+      read_sbcs_nonbusy(target,&sbcs);
+      uint32_t val;
+      dmi_read(target,&val,DMI_SBDATA0);
+      if(i != val) {
+        LOG_ERROR("System Bus Access Test: Error reading autoincremented address, expected val = %d, read val = %d",i,val);
```
timsifive: Same comment about error/pass.

timsifive (changes requested):
The OpenOCD has a very strict, enforced, style guide. tools/checkpatch.sh will check a single commit for violations (which there are many in this PR). Or you can craft a diff and feed it to tools/scripts/checkpatch.pl.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2815,162 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+  dmi_read(target, &sbcs_orig, DMI_SBCS);
+
+  uint32_t sbcs = sbcs_orig;
+
+  // Test 1: Simple write/read test, no address autoincrement
+  sbcs = set_field(sbcs_orig,DMI_SBCS_SBAUTOINCREMENT,0);
+  dmi_write(target,DMI_SBCS,sbcs);
+
+  for(int sbaccess = 0; sbaccess < 0x5; sbaccess++) {
+    sbcs = set_field(sbcs_orig,DMI_SBCS_SBACCESS,sbaccess);
```
mwachs5: In this code you keep using sbcs_orig... don’t you want to keep adding to the fields? Like when you do autoincrement you also need to set sbaccess

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2815,162 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+    }
+  }
+  LOG_INFO("System Bus Access Test: Read/write test, no addr autoincrement PASSED");
+
+  // Test 2: Simple write/read test, with address autoincrement 
+  sbcs = set_field(sbcs_orig,DMI_SBCS_SBAUTOINCREMENT,1);
+  dmi_write(target,DMI_SBCS,sbcs);
+
+  for(int sbaccess = 0; sbaccess < 0x5; sbaccess++){
+    sbcs = set_field(sbcs_orig,DMI_SBCS_SBACCESS,sbaccess);
```
mwachs5: Yeah here you are clearing the autoincrement bit when you just set it

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2815,162 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+      sbcs = set_field(sbcs_orig,DMI_SBCS_SBERROR,0x1);
+      dmi_write(target,DMI_SBCS,sbcs);
+    } else {
+      LOG_ERROR("System Bus Access Test: SBCS Alignment error test FAILED");
+    }
+  }
+   
+  // Test 5: Set sbbusyerror
+  dmi_write(target,DMI_SBADDRESS0,0x80000000);
+  dmi_write(target,DMI_SBADDRESS0,0x80000000);
```
mwachs5: This isn’t guaranteed to give you a busy error. Maybe you’d have to batch up writes and even that may not be fast enough (JTAG is slow!)

mwachs5 (changes requested):
Good start!

src/target/riscv/riscv-013.c
rmac-sifive: Fixed in most recent version

src/target/riscv/riscv-013.c
rmac-sifive: Good catch, fixed this in most recent version

src/target/riscv/riscv-013.c
rmac-sifive: Yes, we're assuming that here. I'll add another parameter for a legal base address for these tests

src/target/riscv/riscv-013.c
rmac-sifive: Fixed

src/target/riscv/riscv-013.c
rmac-sifive: Fixed

src/target/riscv/riscv-013.c
rmac-sifive: Fixed

src/target/riscv/riscv-013.c
rmac-sifive: Yeah, I think this flag may be impossible to hit unless we add some sim-only logic to increase TileLink's read/write response delay

src/target/riscv/riscv-013.c
```diff
@@ -57,6 +57,10 @@ static void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a,
 static void riscv013_fill_dmi_read_u64(struct target *target, char *buf, int a);
 static int riscv013_dmi_write_u64_bits(struct target *target);
 static void riscv013_fill_dmi_nop_u64(struct target *target, char *buf);
+static int riscv013_test_sba_config_reg(struct target *target, target_addr_t legal_address,
+		target_addr_t illegal_address);
+uint32_t read_memory_sba_simple(struct target *target, uint32_t addr, uint32_t sbcs);
```
timsifive: This is also limited to 32-bit addresses.

src/target/riscv/riscv-013.c
```diff
@@ -57,6 +57,10 @@ static void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a,
...
+void write_memory_sba_simple(struct target *target, uint32_t addr, uint32_t data, uint32_t sbcs);
```
timsifive: This is also limited to 32-bit addresses.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
 	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, 0);
 }
 
+static int get_max_sbaccess(struct target *target)
+{
+	uint32_t sbcs;
+	dmi_read(target, &sbcs, DMI_SBCS);
+
+	uint32_t sbaccess128 = get_field(sbcs, DMI_SBCS_SBACCESS128);
+	uint32_t sbaccess64	 = get_field(sbcs, DMI_SBCS_SBACCESS64);
```
timsifive: I'm not a fan of the line-up-the-equals-signs whitespace, but if you're going to do that you should at least do it for every statement in a block.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
 	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, 0);
 }
 
+static int get_max_sbaccess(struct target *target)
+{
+	uint32_t sbcs;
+	dmi_read(target, &sbcs, DMI_SBCS);
```
timsifive: After examine(), the value of sbcs is already available to you, so you don't need to read it again here. See read_memory().

If you get rid of this then the function also cannot fail, meaning you don't have to worry about possibly returning an error.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+	else if (sbaccess64)
+		return 3;
+	else if (sbaccess32)
+		return 2;
+	else if (sbaccess16)
+		return 1;
+	else if (sbaccess8)
+		return 0;
+	else
+		return ERROR_FAIL;
```
timsifive: Please don't mix error results with legit values. Functions returning int should either return one of the ERROR_* codes, or something else.
You can return -1 if you like, but I want to avoid callers thinking 0 is ERROR_OK, which is not what it means here.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+
+	int max_sbaccess = get_max_sbaccess(target);
+
+	if (max_sbaccess == ERROR_FAIL) {
+		LOG_ERROR("System Bus Access not supported in this config.");
+		return ERROR_FAIL;
+	}
+
+	if (get_field(sbcs, DMI_SBCS_SBVERSION) != 1) {
+		LOG_ERROR("System Bus Access unsupported SBVERSION");
```
timsifive: Try to make error messages as helpful as possible. Here, include the value of get_field(sbcs, DMI_SBCS_SBVERSION) so the user doesn't have to wonder what the version actually was. For bonus points, include that the only supported version is 1.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+			if (i != val) {
+				LOG_ERROR("System Bus Access Test 1: Error reading non-autoincremented address %x, expected val = %d, read val = %d", addr, i, val);
+				test_passed = false;
+			}
+		}
+	}
+	if (test_passed)
+		LOG_INFO("System Bus Access Test 1: Read/write test, no addr autoincrement PASSED");
+
+	// Test 2: Simple write/read test, with address autoincrement
```
timsifive: In this test you don't seem to confirm that incrementing happens by the right amount. Eg. hardware that always increments by 4, regardless of sbaccess, will pass this test. It would be better to write auto-incrementing and read back without it, as well as vice versa.
Or you could read back sbaddress after accesses and confirm it was incremented correctly.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+	}
+
+	// Test 4: Write to illegal address
+	write_memory_sba_simple(target, illegal_address, 0xdeadbeef, sbcs_orig);
+
+	dmi_read(target, &rd_val, DMI_SBCS);
+	if (get_field(rd_val, DMI_SBCS_SBERROR) == 2) {
+		LOG_INFO("System Bus Access Test 4: Illegal address write test PASSED");
+		sbcs = set_field(sbcs_orig, DMI_SBCS_SBERROR, 1);
+		dmi_write(target, DMI_SBCS,sbcs);
```
timsifive: OpenOCD style guide mandates a space after that comma.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+		} else {
+			LOG_ERROR("System Bus Access Test 5: SBCS Alignment error test FAILED");
+		}
+	}
+
+	// Test 6: Set sbbusyerror
+	sbcs = set_field(sbcs_orig, DMI_SBCS_SBREADONADDR, 1);
+	dmi_write(target, DMI_SBCS, sbcs);
+
+	dmi_write(target, DMI_SBDATA0, 0xdeadbeef);
```
timsifive: I assume you've unrolled this loop to go faster to hope to set busy. Unrolling the loop will have no noticeable impact because dmi_write() does so much that the few extra instructions are in the noise. (Unrolling may actually hurt you due to i-cache misses.) If you want to do this as fast as possible, you should use the riscv_batch* functions to create a batch that does those scans, and then execute the whole thing at once. For a (confusing) example see write_memory_progbuf() and search for riscv_batch in that function.

If you want to not go that far, at least I'd inline dmi_write() and get rid of the read that's baked into it. That read is only necessary if the target might be too slow to handle consecutive writes.

Note that both of those options may end up with you setting busy in the DTM (dmi register). You might have to add extra run-test/idle cycles, just like the progbuf memory access functions do. (Calling increase_ac_busy_delay() should be good enough.)

Having said all that, you're trying to win a race against hardware. Unless it's really slow hardware, it will always be able to write to memory faster than you can scan over JTAG bit-by-bit.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+	dmi_write(target, DMI_SBDATA0, 0xdeadbeef);
+	dmi_write(target, DMI_SBDATA0, 0xdeadbeef);
+	dmi_write(target, DMI_SBDATA0, 0xdeadbeef);
+	dmi_write(target, DMI_SBDATA0, 0xdeadbeef);
+	dmi_write(target, DMI_SBDATA0, 0xdeadbeef);
+	dmi_write(target, DMI_SBDATA0, 0xdeadbeef);
+	dmi_write(target, DMI_SBDATA0, 0xdeadbeef);
+
+	dmi_write(target, DMI_SBADDRESS0, legal_address);
```
timsifive: Same as above.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+			LOG_ERROR("System Bus Access Test 6: SBCS sbbusyerror test FAILED, unable to clear to 0");
+	} else {
+		LOG_ERROR("System Bus Access Test 6: SBCS sbbusyerror test FAILED, unable to set");
+	}
+
+	return ERROR_OK;
+
+}
+
+void write_memory_sba_simple(struct target *target, uint32_t addr, uint32_t data, uint32_t sbcs)
```
timsifive: These 2 functions should be static. I also prefer to put the definition before the use, so you don't have to separately have the prototypes.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+{
+	uint32_t rd_sbcs;
+
+	read_sbcs_nonbusy(target, &rd_sbcs);
```
timsifive: You call this function in a loop that doesn't do much else. You could speed things up a bit by moving the read/write of sbcs outside of that loop.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+	dmi_write(target, DMI_SBADDRESS0, addr);
+	dmi_write(target, DMI_SBDATA0, data);
+}
+
+uint32_t read_memory_sba_simple(struct target *target, uint32_t addr, uint32_t sbcs)
+{
+	uint32_t rd_val;
+	uint32_t rd_sbcs;
+
+	read_sbcs_nonbusy(target, &rd_sbcs);
```
timsifive: You call this function in a loop that doesn't do much else. You could speed things up a bit by moving the read/write of sbcs outside of that loop.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2816,248 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+
+	uint32_t sbcs_readonaddr = set_field(sbcs, DMI_SBCS_SBREADONADDR, 1);
+	dmi_write(target, DMI_SBCS,sbcs_readonaddr);
```
timsifive: Needs space after comma.

src/target/riscv/riscv.c
```diff
@@ -1428,6 +1428,29 @@ COMMAND_HANDLER(riscv_dmi_write)
...
+{
+	if (CMD_ARGC != 1) {
+		LOG_ERROR("Command takes exactly 1 argument");
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	uint32_t legal_address;
```
timsifive: 64-bit addresses?

src/target/riscv/riscv.c
```diff
@@ -1495,6 +1518,15 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
 		.help = "Perform a 32-bit DMI write of value at address."
 	},
+	{
+		.name = "test_sba_config_reg",
+		.handler = riscv_test_sba_config_reg,
+		.mode = COMMAND_ANY,
+		.usage = "riscv test_sba_config_reg legal_address illegal_address",
+		.help = "Perform a series of tests on the SBCS register."
+			"Inputs are a legal address for read/write tests,"
+			"and an illegal address to for error flag/handling cases."
```
timsifive: Remove "to"

timsifive (changes requested):
Thanks for addressing the OpenOCD style guide changes. I know those are a pain, but it's a requirement to upstream some day.

src/target/riscv/riscv-013.c
rmac-sifive: Yeah, this error case is likely not reachable in actual hardware use cases so long as the JTAG interface exists in the system. I'm going to try to hit this error case in simulation, so I'll comment it out for now

src/target/riscv/riscv-013.c
rmac-sifive: Fixed in most recent update

src/target/riscv/riscv-013.c
rmac-sifive: Fixed in most recent update

src/target/riscv/riscv-013.c
rmac-sifive: Fixed in most recent update

src/target/riscv/riscv-013.c
rmac-sifive: Fixed in most recent update

src/target/riscv/riscv-013.c
```diff
@@ -2816,16 +2816,15 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
 	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, 0);
 }
 
-static int get_max_sbaccess(struct target *target)
+static uint32_t get_max_sbaccess(struct target *target)
```
timsifive: Why did you change this to uint32_t? Now the return -1 at the end is pretty dubious. I'd leave this as returning int.

src/target/riscv/riscv-013.c
```diff
@@ -2892,17 +2891,27 @@ static int riscv013_test_sba_config_reg(struct target *target,
...
 
 	// Test 2: Simple write/read test, with address autoincrement 
+	uint32_t curr_addr;
+	uint32_t prev_addr;
 	test_passed = true;
 	sbcs = set_field(sbcs_orig, DMI_SBCS_SBAUTOINCREMENT, 1);
 	dmi_write(target, DMI_SBCS, sbcs);
 
-	for (int sbaccess = 0; sbaccess <= max_sbaccess; sbaccess++) {
+	for (uint32_t sbaccess = 0; sbaccess <= (uint32_t)max_sbaccess; sbaccess++) {
```
timsifive: Why did you change this type to uint32_t? It looks like it just got you a lot of trouble.

src/target/riscv/riscv-013.c
```diff
@@ -2892,17 +2891,27 @@ static int riscv013_test_sba_config_reg(struct target *target,
 		LOG_INFO("System Bus Access Test 1: Read/write test, no addr autoincrement PASSED");
 
 	// Test 2: Simple write/read test, with address autoincrement 
+	uint32_t curr_addr;
```
timsifive: uint64_t (or target_addr_t) for this one and the next line.

src/target/riscv/riscv-013.c
```diff
@@ -2973,83 +2989,113 @@ static int riscv013_test_sba_config_reg(struct target *target,
...
-	dmi_write(target, DMI_SBADDRESS0, legal_address);
-	dmi_write(target, DMI_SBADDRESS0, legal_address);
-	dmi_write(target, DMI_SBADDRESS0, legal_address);
-	dmi_write(target, DMI_SBADDRESS0, legal_address);
-	dmi_write(target, DMI_SBADDRESS0, legal_address);
-	dmi_write(target, DMI_SBADDRESS0, legal_address);
-	dmi_write(target, DMI_SBADDRESS0, legal_address);
-	dmi_write(target, DMI_SBADDRESS0, legal_address);
-	dmi_write(target, DMI_SBADDRESS0, legal_address);
+		dmi_write(target, DMI_SBDATA0, 0xdeadbeef);
```
timsifive: Can you use a loop here? I bet it will still work just as well.

src/target/riscv/riscv.c
```diff
@@ -190,6 +190,8 @@ uint64_t riscv_scratch_ram_address;
 
 bool riscv_prefer_sba;
 
+bool riscv_run_sim_only_tests;
```
timsifive: Why is this global? You can just pass it into test_sba_config_reg() instead.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2818,324 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+	}
+
+	uint32_t num_sbdata_regs = get_num_sbdata_regs(target);
+
+	/* Test 1: Simple write/read test */
+	test_passed = true;
+	sbcs = set_field(sbcs_orig, DMI_SBCS_SBAUTOINCREMENT, 0);
+	dmi_write(target, DMI_SBCS, sbcs);
+
+	uint32_t test_patterns[4] = {0xdeadbeef, 0xfeedbabe, 0x12345678, 0x08675309};
```
mwachs5, line 3025: hah

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2818,324 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+	for (uint32_t sbaccess = 0; sbaccess <= (uint32_t)max_sbaccess; sbaccess++) {
+		sbcs = set_field(sbcs, DMI_SBCS_SBACCESS, sbaccess);
+		dmi_write(target, DMI_SBCS, sbcs);
+
+		uint32_t compare_mask = (sbaccess == 0) ? 0xff :
+														(sbaccess == 1) ? 0xffff : 0xffffffff;
```
mwachs5: whats going on with this whitespace?

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2818,324 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+						"expected val = %x, read val = %x", i, val);
+				test_passed = false;
+			}
+		}
+	}
+	if (test_passed)
+		LOG_INFO("System Bus Access Test 2: Address auto-increment test PASSED.");
+
+	/* Test 3: Read from illegal address */
+	uint32_t *illegal_addr_read = read_memory_sba_simple(target, illegal_address, 1, sbcs_orig);
```
mwachs5: it seems sort of weird to have the malloc and free in different functions.... couldn't you just have `read_memory_sba_simple` take in a `uint32_t* buf`, and for something small like this you wouldn't need to malloc/free anything?

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2818,324 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+			dmi_write(target, DMI_SBADDRESS3, masked_addr);
+	}
+
+	/* Write SBDATA registers starting with highest address, since write to
+	 * SBDATA0 triggers write */
+	for (int i = write_size-1; i >= 0; i--)
+		dmi_write(target, DMI_SBDATA0+i, write_data[i]);
+}
+
+uint32_t *read_memory_sba_simple(struct target *target, target_addr_t addr,
```
mwachs5: i think you should take the `read_data` in as a buffer here rather than mallocing it in this function. That seems likely to be overkill, as you could just have one buffer in the calling function that you use for everything

src/target/riscv/riscv-013.c
rmac-sifive: Not sure... looks completely different in my local directory. I'll try reformatting

rmac-sifive:
@timsifive, @mwachs5, any more feedback before I merge?

mwachs5:
Did you run the style checker @timsifive pointed out? I think he made a PR for it... https://github.com/riscv/riscv-openocd/pull/239

src/target/riscv/riscv-013.c
```diff
@@ -57,6 +57,12 @@ static void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a,
 static void riscv013_fill_dmi_read_u64(struct target *target, char *buf, int a);
 static int riscv013_dmi_write_u64_bits(struct target *target);
 static void riscv013_fill_dmi_nop_u64(struct target *target, char *buf);
+static int riscv013_test_sba_config_reg(struct target *target, target_addr_t legal_address,
```
mwachs5: I would move these down, as these are specific to the test command and seem weirdly mixed in with the functional stuff

src/target/riscv/riscv-013.c
```diff
@@ -1488,6 +1494,7 @@ static int init_target(struct command_context *cmd_ctx,
 	generic_info->fill_dmi_read_u64 = &riscv013_fill_dmi_read_u64;
 	generic_info->fill_dmi_nop_u64 = &riscv013_fill_dmi_nop_u64;
 	generic_info->dmi_write_u64_bits = &riscv013_dmi_write_u64_bits;
+	generic_info->test_sba_config_reg = &riscv013_test_sba_config_reg;
```
mwachs5: same here... this seems weirdly mixed in to the more "useful" commands

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2818,320 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
 	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, 0);
 }
 
+static int get_max_sbaccess(struct target *target)
```
mwachs5: Can you bracket this with some sort of comments that this is all related to the SBA test command?

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2818,320 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+	else
+		return -1;
+}
+
+static uint32_t get_num_sbdata_regs(struct target *target)
+{
+	RISCV013_INFO(info);
+
+	uint32_t sbaccess128 = get_field(info->sbcs, DMI_SBCS_SBACCESS128);
+	uint32_t sbaccess64	= get_field(info->sbcs, DMI_SBCS_SBACCESS64);
```
mwachs5: why this weird whitespace?

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2818,320 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+		dmi_write(target, DMI_SBCS, sbcs);
+
+		dmi_write(target, DMI_SBADDRESS0, legal_address);
+		read_sbcs_nonbusy(target, &sbcs);
+		curr_addr = legal_address;
+		for (uint32_t i = 0; i < num_words; i++) {
+			prev_addr = curr_addr;
+			read_sbcs_nonbusy(target, &sbcs);
+			curr_addr = sb_read_address(target);
+			if ((curr_addr - prev_addr != (uint32_t)(1 << sbaccess)) && i != 0) {
```
mwachs5: I'm a fan of parens around this sort of thing (the curr_add - prev_addr) and the two sides of the && but not sure what the OpenOCD style guide is

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2818,320 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+			}
+		}
+	}
+	if (test_passed)
+		LOG_INFO("System Bus Access Test 2: Address auto-increment test PASSED.");
+
+	/* Test 3: Read from illegal address */
+	read_memory_sba_simple(target, illegal_address, rd_buf, 1, sbcs_orig);
+
+	dmi_read(target, &rd_val, DMI_SBCS);
```
mwachs5, line 3116: if it's not already updated, we should use the new `debug_defines.h` as generated by the spec, so that these error codes match what the spec says

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2818,320 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+	if (get_field(rd_val, DMI_SBCS_SBERROR) == 2) {
+		LOG_INFO("System Bus Access Test 3: Illegal address read test PASSED.");
+		sbcs = set_field(sbcs_orig, DMI_SBCS_SBERROR, 1);
```
mwachs5: what does setting this to 1 do if the error code was 2? I don't get what this bit is doing.

rmac-sifive:
I ran the style checker scripted and it didn't have any issues, besides needing a sign-off: `ERROR: Missing Signed-off-by: line(s)`. @timsifive, any idea what to do here?

src/target/riscv/riscv-013.c
rmac-sifive: I think this is a GitHub bug. In my local file there's no weird whitespace

src/target/riscv/riscv-013.c
rmac-sifive: This is testing that we can write 1 to this register in order to clear it to 0. This is part of the register's spec

mwachs5: Ah, I thought you had to write exactly what you read in order to clear it

src/target/riscv/riscv-013.c
rmac-sifive: Moved in my latest PR

src/target/riscv/riscv-013.c
rmac-sifive: Moved in my latest PR

src/target/riscv/riscv-013.c
rmac-sifive: Added a comment in my latest PR

src/target/riscv/riscv-013.c
rmac-sifive: I added a paren around the != statement for clarity. The OpenOCD style guide doesn't have a preference

src/target/riscv/riscv-013.c
rmac-sifive: Bumped in my latest PR

src/target/riscv/riscv-013.c
mwachs5: So the spec says 
```
  When the debug module's system bus
            master causes a bus error, this field gets set. The bits in this
            field remain set until they are cleared by writing 1 to them.
            While this field is non-zero, no more system bus accesses can be
            initiated by the debug module.
```

This I think is intended to mean that *each* bit is cleared by writing 1 to it...

src/target/riscv/riscv-013.c
timsifive: Github shows this is a tab instead of a space. Probably in your editor things line up just so that it looks like a space, but it's actually a tab. The same is true for the next 2 lines.

src/target/riscv/riscv-013.c
```diff
@@ -2811,6 +2818,329 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+	RISCV013_INFO(info);
+
+	uint32_t sbaccess128 = get_field(info->sbcs, DMI_SBCS_SBACCESS128);
+	uint32_t sbaccess64 = get_field(info->sbcs, DMI_SBCS_SBACCESS64);
+
+	if (sbaccess128)
+		return 4;
+	else if (sbaccess64)
+		return 2;
+	else
```
timsifive: Can you check sbaccess32 and return 0? That seems trivial, and will help any future callers for the function who might not realize it's only useful if it's already confirmed that SBA is implemented.

src/target/riscv/riscv-013.c
```diff
@@ -3070,14 +3100,19 @@ void write_memory_sba_simple(struct target *target, target_addr_t addr, uint32_t
...
 }
 
-uint32_t read_memory_sba_simple(struct target *target, target_addr_t addr, uint32_t sbcs)
+uint32_t* read_memory_sba_simple(struct target *target, target_addr_t addr,
+		uint32_t read_size, uint32_t sbcs)
 {
 	RISCV013_INFO(info);
 
-	uint32_t rd_val;
+	uint32_t* rd_val = malloc(read_size*sizeof(uint32_t));
```
timsifive: This kind of construct is really prone to memory leaks. Instead, can you have the caller pass in the pointer? That also allows you to return an error if you want.

src/target/riscv/riscv.h
```diff
@@ -133,7 +133,7 @@ extern uint64_t riscv_scratch_ram_address;
 
 extern bool riscv_prefer_sba;
 
-extern bool riscv_run_sim_only_tests;
+extern bool riscv_run_sbbusyerror_test;
```
timsifive: You don't need this anymore, right?

timsifive (changes requested):

src/target/riscv/riscv-013.c
rmac-sifive: It's a space in my editor as well. I think it might be a bug in how GitHub counts whitespace between commits

mwachs5:
If you merge in the current master you'll get the automatic regression checks on your whitespace

mwachs5:
@timsifive can you re-review?

src/target/riscv/riscv-013.c
```diff
@@ -2935,6 +2942,332 @@ void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
...
+
+	int max_sbaccess = get_max_sbaccess(target);
+
+	if (max_sbaccess == -1) {
+		LOG_ERROR("System Bus Access not supported in this config.");
+		return ERROR_FAIL;
+	}
+
+	if (get_field(sbcs, DMI_SBCS_SBVERSION) != 1) {
+		LOG_ERROR("System Bus Access unsupported SBVERSION (0). Only version 1 is supported.");
```
timsifive: Why does it say (0) here? It would make sense to use (%d) and print out the actually detected version.

timsifive (dismissed):
This code also does very little error checking.

rmac-sifive:
Updated with Tim's feedback, and added a more explicit pass message to both the compliance and SBA tests

timsifive (approved):

mwachs5 (approved):

mwachs5:
sweet. Let's update the commit hash in this PR to point to the right thing and make sure all the tests pass. In the meantime, I'm workign on bubbling OpenOCD up through riscv-tools/rocket-chip. I'm also attempting to honor Tim's request to add some tests to riscv-tests of our new compliance commands, but won't block this PR on that.

mwachs5:
(and by "let's" I mean I will)

PR #236, opened 2018-04-06
Small speed improvements.

timsifive:
DebugBreakpoint went from 3.41s to 3.05s!

Change-Id: Icfc4ad5fb663b3607bf2027fda744b43be662fc5


mwachs5:
Trying this out

mwachs5:
Currently the debug tests are failing on riscv-tools regression. Is there a version of riscv-tools PR that bumps to this version of OpenOCD? It's really hard to approve if the tests are not shown to pass.

mwachs5:
How much speed improvement would this get you on the initial connection from GDB, where it queries all the registers?

timsifive:
The optimized version does about half the number of DMI scans in the ExamineTarget test (which just connects with GDB and reads MISA).

mwachs5:
"just connects with GDB" is actually a huge amount of transactions, as GDB wants you to read every single register/relevant CSR. So, if it cuts in half that sounds great.

timsifive:
So can I merge it?

timsifive:
What do I need to do to get you to approve this change?

mwachs5:
Sorry, lost track of this completely. What exactly is this doing, as in what does it mean for a register to be cached (is the idea that if you're halted registers don't change their values)? How do you keep track of the cache being dirty or not?

Presumably it's been tested against all the HW targets you can?

timsifive:
The idea is exactly that when halted registers don't change their value. OpenOCD already contains a register cache, which is what I'm using.

I've tested against the usual mystery FPGA image that implements 0.13. I don't know what would be different about hardware vs. spike in this case.

mwachs5:
See the following: 
https://github.com/riscv/riscv-tools/pull/203
and 
https://github.com/freechipsproject/rocket-chip/pull/1406

timsifive:
All those builds passed. Can you approve this?

mwachs5 (approved):

PR #238, opened 2018-04-12
Fix FESPI assert when guessing few algorithm steps

timsifive:
Instead of trying to predict exactly how many steps will be required
(doable but error-prone), just allocate more memory when we need it.

Tested against HiFive1, and Arty board image.

Change-Id: I3cd9798432e65176616c700ba122daf7a5ed6209

### DIFF
--- a/src/flash/nor/fespi.c
+++ b/src/flash/nor/fespi.c
@@ -607,11 +607,11 @@ struct algorithm_steps {
 	uint8_t **steps;
 };
 
-static struct algorithm_steps *as_new(unsigned size)
+static struct algorithm_steps *as_new(void)
 {
 	struct algorithm_steps *as = calloc(1, sizeof(struct algorithm_steps));
-	as->size = size;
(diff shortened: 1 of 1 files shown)


mwachs5 (commented):
What exactly was the bug here?

timsifive:
Lazy coding. Probably because I wanted to get you something that worked soon, and this felt safe. I should've tagged it with TODO when I wrote that code.

mwachs5 (approved):
If you've tested this I'm fine with it.

timsifive:
Sorry, I read your "what" as a "why."

When flashing, we build up a list of steps that need to happen, then we "compile" that list, and finally send it to the target to be run by our "algorithm." The code contained a heuristic of how many steps would be required based on the size of the write. It used size/4, but realistically it's probably more like 15+size/16 or something. If the writes are very small, the heuristic would end up allocating too few steps.

PR #239, opened 2018-04-14
Enforce OpenOCD style guide.

timsifive:
Change-Id: I579a9f54ed22a774bf52f6aa5bc13bcbd2e82cd8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,7 @@ matrix:
                       - binutils-mingw-w64-i686 gcc-mingw-w64-i686 g++-mingw-w64-i686
 
 script:
-    - set -o pipefail && git diff `git merge-base master HEAD` | ./tools/scripts/checkpatch.pl -
+    - mergebase=`git merge-base master HEAD`
+    - set -o pipefail && git diff $mergebase | ./tools/scripts/checkpatch.pl -
     - ./bootstrap && ./configure --enable-remote-bitbang --enable-jtag_vpi $CONFIGURE_ARGS && make
     - file src/$EXECUTABLE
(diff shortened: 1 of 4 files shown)


mwachs5 (dismissed):
If this is the check we need to run, I'm all for this. Is `master` the right branch, or should it be `riscv`?

mwachs5:
Will this fail on `WARNING`s or just `ERROR`s?

mwachs5 (dismissed):

timsifive:
This is still WIP, but I need the PR to get the builds to happen. Right now the builds should actually be failing, which they're not. I need to make that work first. :-)

.travis.yml
```diff
@@ -50,7 +50,6 @@ matrix:
                       - binutils-mingw-w64-i686 gcc-mingw-w64-i686 g++-mingw-w64-i686
 
 script:
-    - set -o pipefail
-    - git diff `git merge-base master HEAD` | ./tools/scripts/checkpatch.pl -
+    - set -o pipefail && git diff `git merge-base master HEAD` | ./tools/scripts/checkpatch.pl -
```
mwachs5: FWIW when I run this locally I use `riscv` instead of `master`. Not sure if that would help you fail automatically? Or maybe you should make this its own test step?

timsifive:
OK. Now it's ready to merge. @mwachs5, can you approve this?

This will fail a build on both errors and warnings.

mwachs5 (approved):
Can you squash all the ones related to Travis into one commit?

timsifive:
How about I just squash+merge this entire thing instead of regular merge?

mwachs5:
SGTM

PR #248, opened 2018-05-07
Don't error if hart select isn't implemented.

timsifive:
It's not implemented for 0.11 because we don't need it. Returning error
caused 0.11 targets to not be debuggable since change
848062d0d11679de25be573981df45e2c4880db8.

Change-Id: I8b04a1fcf3c3e8bf8340cbf39aaf475d2a213519


mwachs5 (commented):
Why not create a version for v11 that returns ERROR_OK?

mwachs5 (approved):

timsifive:
A better fix is probably to only check for and call select_current_hart() if there is more than one hart in the system.

PR #255, opened 2018-05-17
riscv: remove unexpected check during reset

mwachs5:
I'm not sure what this check is adding, and it causes problems for implementations that take some time to report that they are halted out of reset (e.g. by executing Debug ROM).

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1644,15 +1644,12 @@ static int deassert_reset(struct target *target)
 
 		char *operation;
 		uint32_t expected_field;
-		uint32_t unexpected_field;
 		if (target->reset_halt) {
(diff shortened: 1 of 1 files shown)


timsifive (approved):
I believe the intent of this code was to ensure that if we request a reset+halt, the hart never runs any code. That's what the spec says, but it doesn't actually say that that means the state must go straight to halted. It feels a little awkward to have the state say running in between, but it's not against the letter of the spec.

If this is causing trouble, it's fine to remove it.

mwachs5:
Yeah, the intent makes sense, but in practice the debug module + debug ROM (Spike included) assumes that a hart is running until it learns otherwise. It has to assume *something*, and we didn't really specify what the "default" is in the spec.

Issue #200, opened 2018-01-30
Should deinit_target() deassert dmactive?


TommyMurphyTM1234:
After reset the RISC-V debug module signal dmactive is low.
When a debug connection is established openocd asserts dmactive (snippet from riscv-013.c):

```
	/* Reset the Debug Module. */
	dmi_write(target, DMI_DMCONTROL, 0);
	dmi_write(target, DMI_DMCONTROL, DMI_DMCONTROL_DMACTIVE);
	uint32_t dmcontrol = dmi_read(target, DMI_DMCONTROL);
```

When the debug session terminates dmactive stays asserted/high and only gets deasserted on the next power on reset.
Should dmactive be deasserted - presumably in deinit_target() - or is there some specific reason that it 's left asserted after the first debug connection?
Is it perhaps for debug across an ndmreset driven reset of the CPU?

mwachs5:
It is not for debug across ndmreset driven reset.

 I think this is a reasonable request.

PR #256, opened 2018-05-17
Disable the DM when OpenOCD exits.

timsifive:
gdb/OpenOCD already remove all the breakpoints when exiting, so I think
this is all we need to do to restore the hart to the state it was in
before connecting.

Issue #200.

Change-Id: Ibcde7a3b68d221dd2718b8b51dfc7946350a8fcb

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1334,6 +1334,9 @@ int wait_for_authbusy(struct target *target, uint32_t *dmstatus)
 static void deinit_target(struct target *target)
 {
 	LOG_DEBUG("riscv_deinit_target()");
+
+	dmi_write(target, DMI_DMCONTROL, 0);
+
 	riscv_info_t *info = (riscv_info_t *) target->arch_info;
 	free(info->version_specific);
 	/* TODO: free register arch_info */


mwachs5:
Is this a specific command that one calls when wanting to quit OpenOCD quits, or does this get called any time OpenOCD quits? Is there a way to quit OpenOCD without doing this? Like, if I want to leave a hart halted and quit and restart OpenOCD, is that impossible?

timsifive:
I believe this always happens when OpenOCD exits (short of eg. `kill -9`). I don't think OpenOCD has a concept of different ways of exiting. Do you need this behavior to be controlled by a config option?

mwachs5:
Also, it's not correct that "this is all we need to do to restore the hart to the state it was in
before connecting." All this does is reset the debug module, which doesn't:

* resume a hart that was halted (if it should be)
* restore any registers that the debugger clobbered

mwachs5:
"Do you need this behavior to be controlled by a config option?"

Not sure, need to think about it... it has provided some utility in the past that one session of OpenOCD can halt a hart and a later session can notice it being halted. This is a little bit tied up with #211, since OpenOCD can crash  in that case but it might be nice to leave the hart halted, or something.

timsifive:
This is all getting a lot more complicated than I'd initially assumed. I'm tempted to punt on this until it causes problems for somebody, or there's nothing important to be done.

mwachs5:
Might a better solution be to just provide a way to clear dmactive from cfg script? I guess that's not that general.

timsifive:
That would be awkward to make work, since OpenOCD continuously tries to poll the target status, which is no longer possible when dmactive is cleared. A config option that controls what happens when OpenOCD shuts down makes more sense to me.

mwachs5:
> " A config option that controls what happens when OpenOCD shuts down makes more sense to me."

That makes sense to me too. The default could be to clear dmactive when OpenOCD quits, but we could turn it to "don't clear dmactive when OpenOCD quits" in the cfg script. I think that would cover my use cases. 

Questionable whether you would also want to resume harts, restore registers with some level of control, it seems like that should be more the user's responsibility. That sort of thing is more likely to fail if the system is in a bad place when OpenOCD quits anyway.

TommyMurphyTM1234:
Is there any plan for this change to be merged and if not why not?

timsifive:
There's no clear plan. This was original prompted by #200, which simply asks a question and doesn't describe desired behavior, or give a rationale. This change is just kind of lingering right now.

TommyMurphyTM1234:
Thanks.
Surely the rationale is obvious.
dmactive should be asserted when the debug module is active.
The DM is active when OpenOCD (or whatever) connects, resets/initializes it and starts a debug session.
The DM is no longer active when OpenOCD (or whatever) disconnects/tears down the connection (literally deinit_target() in the OpenOCD code).
Why can't it just do that?
Does it HAVE to do some or all of the other deinitialization of the target when disconnecting and tearing down the connection?

timsifive:
It's not obvious to me. What undesirable consequence is there to leaving dmactive high when the debugger disconnects?

I can't think of a real problem either way.

TommyMurphyTM1234:
We have a case where a customer wants the ability to enter/leave debug halt "out of band" using a gpio rather than the debug module. But this out of band mechanism also needs to know when the debug module is active (e.g. OpenOCD is connected). At the moment it can't know this as OpenOCD asserts dmactive on first connect and it stays asserted until reset even though once OpenOCD has disconnected the debug module is NOT active.

mwachs5:
In the discussion above I think we agreed on a config option (default is to deassert dmactive when you quit)? Makes sense to me, but I don't think it's the top of my priority list.

timsifive:
@mwachs5 I believe that harts will stay halted even if dmactive goes low. Is that not true? If it is true, does this need to be an option?

mwachs5:
Right, Harts would stay halted when dmactive goes low. Do you mean does THAT also need to be an option, or clearing dmactive at all?

mwachs5:
The point is, dmactive could be used to gate a lot of things, like perhaps debugged Harts only THINK they get powered down, and so on. So I agree with Tommy this is what OpenOCD should do probably by default, but maybe we want an option to not do it for some reason

TommyMurphyTM1234:
On reflection maybe deasserting dmactive in deinit_target() is not always the right thing to do?
There's an argument that it should be done when the debugger/gdb detaches even if openocd remains running waiting for further connections.
Not sure if this has been considered here?

timsifive:
Inactive

PR #257, opened 2018-05-18
Comment riscv_set_register, register_write_direct

### LINKED ISSUE #241: Comment on difference between riscv_set_register and register_write_direct

timsifive:
Fixes #241

Change-Id: Ia199f15106a0bda465d3918d052ddd4d03655031

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1763,6 +1763,10 @@ bool riscv_has_register(struct target *target, int hartid, int regid)
 	return 1;
 }
 
+/**
+ * This function is called when the debug user wants to change the value of a
+ * register. The new value may be cached, and may not be written until the hart
+ * is running again. */
 int riscv_set_register(struct target *target, enum gdb_regno r, riscv_reg_t v)
 {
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv.c
```diff
@@ -1763,6 +1763,10 @@ bool riscv_has_register(struct target *target, int hartid, int regid)
 	return 1;
 }
 
+/**
+ * This function is called when the debug user wants to change the value of a
+ * register. The new value may be cached, and may not be written until the hart
+ * is running again. */
```
mwachs5: "is resumed"?

mwachs5 (dismissed):
Pretty good, but is one preferable to the other in the general case? If you call register_write_direct does it invalidate the cache?

timsifive:
There is no general case.

Either you're passing on a user/higher level request, in which case you're using riscv_set_register(). Otherwise when the register writing is done for reasons internal to the RISC-V debug support you're using register_write_direct()

timsifive:
Could I get a review for this?

palmer-dabbelt (approved):

Issue #259, opened 2018-05-22
NULL deref in get_target_type


dlrobertson:
I hit a null deref on the latest master in `get_target_type` in `riscv_deinit_target`. The following is the minimal repro I was able to minimize it to.

```
$ cat sigsegv.cfg
interface remote_bitbang
remote_bitbang_host localhost
remote_bitbang_port 9824

init

jtag newtap riscv cpu -irlen 5 -expected-id 0x10e31913

target create riscv.cpu riscv -chain-position riscv.cpu
$ openocd -f sigsegv.cfg
# another terminal
$ echo "shutdown" | nc 127.0.0.1 4444
```

timsifive:
FWIW, I believe OpenOCD "wants" `init` to be called after taps/targets are created. It shouldn't crash, though.

PR #260, opened 2018-05-22
Fix posible null deref in get_target_type

dlrobertson:
A null deref occurs if `riscv_deinit_target` is called and the
target has not been initialized.

Fixes: #259 

Change-Id: Ic34057508ed6686eb48e9fe8220110c42ba2fc5e


timsifive (approved):
Looks good. Thanks for fixing this!

Issue #258, opened 2018-05-18
NFR: Delay trigger enumeration?


mwachs5:
New Feature Request:

Is it possible to bypass the Trigger enumeration at startup time? And only do the enumeration once a trigger is requested? For targets with large number of triggers, this can be a significant amount of runtime before GDB can even connect.

Also, what does enumerating the triggers up front get you, because at the time you want a trigger, you still need to check if M-Mode is using it, so you may not actually have any triggers that you can use at that time?

PR #261, opened 2018-05-22
Delay trigger enumeration until it's required.

timsifive:
This improves startup time, which is important when connecting to
simulators. One problem is that triggers that are set when the debugger
connects are not cleared until enumeration happens. Execution may halt
due to a trigger set by a previous debug session, which could confuse
the user. If this happens, triggers will be instantly enumerated, so it
will only happen once per session.

Addresses #258

Change-Id: I3396f713f16980a8b74745a1672fe8b8a2d4abae

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1453,9 +1453,6 @@ static int examine(struct target *target)
 		return ERROR_FAIL;
 	}
 
-	/* Then we check the number of triggers availiable to each hart. */
-	riscv_enumerate_triggers(target);
-
 	/* Resumes all the harts, so the debugger can later pause them. */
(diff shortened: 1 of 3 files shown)


mwachs5 (commented):
Awesome! Thanks.

Do you think there is any benefit to making this a config option (I don't really see one other than timing predictability when you insert your first trigger...)

timsifive:
The only point would be to deal with the rare case where triggers are already set when first connecting. I'm happy to leave it as-is until somebody decides that's a feature that really needs to be added.

mwachs5 (approved):
This works great for simulations, haven't tried on any HW or against riscv-tests. I'll open a PR against riscv-tools for it to test it out there.

aswaterman:
Nice

PR #266, opened 2018-06-12
target/riscv: add semihosting support

ilg-ul:
- use the semihosting_common code
- to enable it, use the same commands as for arm

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1530,17 +1530,17 @@ extern __COMMAND_HANDLER(handle_common_semihosting_resumable_exit_command);
 extern __COMMAND_HANDLER(handle_common_semihosting_cmdline);
 
 /*
- * To be noted that RISC-V targets use the same semihosting commands as 
+ * To be noted that RISC-V targets use the same semihosting commands as
  * ARM targets.
- * 
- * The main reason is compatibility with existing tools. For example the  
- * Eclipse OpenOCD/SEGGER J-Link/QEMU plug-ins have several widgets to  
- * configure semihosting, which generate commands like `arm semihosting 
- * enable`. 
- * A secondary reason is the fact that the protocol used is exactly the 
- * one specified by ARM. If RISC-V will ever define its own semihosting 
- * protocol, then a command like `riscv semihosting enable` will make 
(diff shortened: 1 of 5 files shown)


palmer-dabbelt (dismissed):
Thanks!

ilg-ul:
Thank you, Palmer, for your review.

src/target/riscv/riscv.c
```diff
@@ -1523,6 +1569,13 @@ const struct command_registration riscv_command_handlers[] = {
 		.usage = "",
 		.chain = riscv_exec_command_handlers
 	},
+	{
+		.name = "arm",
```
timsifive, line 1586: Having the RISC-V code add ARM commands is very confusing. Since these modify the behavior of the RISC-V support, they ought to be under RISC-V.

I believe you're trying to allude to the fact that this uses the ARM semihosting protocol. You could put that in the individual command's help.

timsifive (changes requested):

src/target/riscv/riscv.c
ilg-ul: No, the main reason is compatibility with existing tools. For example the Eclipse OpenOCD/SEGGER J-Link/QEMU plug-ins have some buttons to configure semihosting, which generate commands like `arm semihosting enable`. It is not reasonable to expect all tools to be updated. A secondary reason is the fact that the protocol used is indeed the one specified by ARM. If RISC-V will ever define its own semihosting protocol, then a command like `riscv semihosting enable` will make sense, but for now it does not.

timsifive: OK. Can you add a comment explaining that to the code?

ilg-ul: Yes, please check the text.

timsifive (approved):

ilg-ul:
Thank you, Tim.

ilg-ul:
Perhaps the manual must be also updated, to refer to the same ARM semihosting commands. What do you think? Actually I could not find a RISC-V section in the OpenOCD manual...

timsifive:
The manual should be updated for all the RISC-V commands. One of the things
I need to do before we can merge to mainline, but I'd happily accept some
help.

PR #273, opened 2018-06-19
Leave haltreq high while the target is halted.

timsifive:
This should help if a target unexpectedly resets while the debugger has
it halted. See #211.

I don't have a target to test this one. When I push reset on the Arty
board it also resets the Debug Module. (There's a small fix in this
change as well to prevent OpenOCD from increasing the busy delays when
that happens.)

Tested in `riscv-tests` and against all my hardware.

Change-Id: I894bbadbfd1005fa74f5fbc1303697c6f513be13

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -204,7 +204,7 @@ typedef struct {
 	 * target->state, because it'll reflect halted status even if we're
 	 * pretending to the client that the hart is running (ie. during
 	 * examine()). */
-	bool halted[RISCV_MAX_HARTS];
+	bool really_halted[RISCV_MAX_HARTS];
 
 	/* When a function returns some error due to a failure indicated by the
 	 * target in cmderr, the caller can look here to see what that error was.
@@ -2864,7 +2864,7 @@ static int riscv013_select_current_hart(struct target *target)
 
 	uint32_t dmcontrol = DMI_DMCONTROL_DMACTIVE;
 	dmcontrol = set_hartsel(dmcontrol, r->current_hartid);
-	if (info->halted[r->current_hartid])
+	if (info->really_halted[r->current_hartid])
 		dmcontrol |= DMI_DMCONTROL_HALTREQ;
 	int result = dmi_write(target, DMI_DMCONTROL, dmcontrol);
 	dm->current_hartid = r->current_hartid;
@@ -2902,7 +2902,7 @@ static int riscv013_halt_current_hart(struct target *target)
 		return ERROR_FAIL;
 	}
 
-	info->halted[r->current_hartid] = true;
+	info->really_halted[r->current_hartid] = true;
 
 	return ERROR_OK;
 }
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -195,6 +195,12 @@ typedef struct {
 	bool abstract_read_fpr_supported;
 	bool abstract_write_fpr_supported;
 
+	/* Track whether the hart is currently halted. This is more precise than
+	 * target->state, because it'll reflect halted status even if we're
+	 * pretending to the client that the hart is running (ie. during
+	 * examine()). */
+	bool halted[RISCV_MAX_HARTS];
```
mwachs5: should you call this something like "really_halted" to distinguish?

src/target/riscv/riscv-013.c
```diff
@@ -547,7 +563,19 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in, int dmi_op,
...
 				false);
-		if (status == DMI_STATUS_BUSY) {
+		if (status == 3 && address_in == (1U << info->abits) - 1 &&
+				*data_in == 0xffffffff) {
+			LOG_ERROR("TDO seems to be stuck high; target might be held in "
+					"reset, or there might be a connection failure");
+			return ERROR_FAIL;
+		} else if (dmi_op == DMI_OP_READ && address != 0 && status == 0 &&
+				address_in == 0 && *data_in == 0) {
+			/* On read ops we should at least get back the address as non-zero.
```
mwachs5: Is that actually required by the spec? Or just happens to be the way the implementations we have do it (there is a proposed PR in rocket-chip which would actually break this assumption for writes, which is why I ask).

mwachs5 (commented):
In general LGTM, but agree it's problematic you can't test it. 

Is it possible in OpenOCD to turn down the polling frequency (e.g. to ~1s) ? That might make this testable

src/target/riscv/riscv-013.c
timsifive: Yes, that's better.

src/target/riscv/riscv-013.c
timsifive: Good point. It's not required by the spec.

timsifive:
Can I get a review on this?

timsifive:
I tried Megan's suggestion on Jun 25, 2018:

You can turn off polling in OpenOCD, and then turn it on again. If I do that, things behave as expected. Not a great user experience, but I think it shows OpenOCD doing the right thing:
```
(gdb) target extended-remote localhost:3333
Remote debugging using localhost:3333
0x00000000 in ?? ()
(gdb) load
Loading section .text, size 0xa20 lma 0x80000000
Loading section .rodata, size 0x2b lma 0x80000a20
Loading section .data, size 0xc lma 0x80000a4c
Start address 0x80000670, load size 2647
Transfer rate: 14 KB/sec, 882 bytes/write.
(gdb) c
Continuing.

Breakpoint 1, main () at programs/debug.c:52
52        __malloc_freelist = 0;
(gdb) monitor poll off
# Push reset button
(gdb) monitor poll on
Hart 0 unexpectedly reset!
(gdb) p/x $pc
$10 = 0x8000017a
(gdb) flushregs 
Register cache flushed.
(gdb) p/x $pc
$11 = 0x40400000
(gdb) stepi
0x40400004 in ?? ()
(gdb) 
0x40400008 in ?? ()
```

mwachs5:
Is the user experience improved in the case where you leave polling on and the hart gets unexpectedly reset? (I'm just making sure we don't make the experience worse for the user).

ctt-wis:
Hello,
I am facing reset (while core in debug mode) issues on my system as well that could be fixed with these changes.
Any chance this pull request will be merged?
Thanks,
CTt

timsifive:
Inactive

PR #280, opened 2018-07-16
Document RISC-V commands.

timsifive:
Change-Id: I7e858368e9bbdb866d2569c686f700693e313eb5

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -8900,6 +8900,84 @@ Display all registers in @emph{group}.
  "timer" or any new group created with addreg command.
 @end deffn
 
+@section RISC-V Architecture
+
+@uref{http://riscv.org/, RISC-V} is a free and open ISA. OpenOCD supports JTAG
+debug of targets that implement version 0.11 and 0.13 of the RISC-V Debug
+Specification.
+
+@subsection RISC-V Terminology
+
+A @emph{hart} is a hardware thread. A hart may share resources (eg. FPU) with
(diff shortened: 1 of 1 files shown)


doc/openocd.texi
```diff
@@ -8900,6 +8900,84 @@ Display all registers in @emph{group}.
...
+standard ones. This must be executed before `init`.
+
+By default OpenOCD attempts to expose only CSRs that are mentioned in a spec,
+and then only if the corresponding extension appears to be implemented. This
+command can be used if OpenOCD gets this wrong, or a target implements custom
+CSRs.
+@end deffn
+
+@deffn Command {riscv set_command_timeout_sec} [seconds]
+Set the wall-clock timeout (in seconds) for individual commands. The default
```
mwachs5, line 8928: Should we say “abstract commands”? It’s unclear what kind of command you mean

mwachs5 (approved):

doc/openocd.texi
timsifive: It's a bit of a kitchen-sink timeout. In addition to timing out abstract commands it's also used for completing a DMI operation, authentication to complete, system bus accesses, and in 0.11 there are also a few use cases.

It doesn't really matter what we put here, as long as it's a bit hand-wavy. I don't think people care about what timeouts are for. They just increase them when they run into a problem due to it, and I think the code is pretty good about printing which timeout needs to be increased when one is hit.

PR #282, opened 2018-07-18
Mimic openrisc Makefile structure

timsifive:
That's better than inventing our own. Also this fixes a build issue in
the official OpenOCD regression build.

Change-Id: I042faa5b93b26e6f6b2d62bee62f21474ec74131


mwachs5 (approved):
Rubber stamp review

PR #283, opened 2018-07-21
Fix target not halting when GDB jumps to a hardware breakpoint

craigblackmore:
This issue affects riscv-0.11. It is caused by OpenOCD manually
stepping over hardware breakpoints to resume or step after a halt.
This is not necessary as GDB should remove and add breakpoints as
required.

At the moment OpenOCD still steps over hardware watchpoints manually
as GDB needs to know which address triggered the watchpoint and
OpenOCD does not currently provide this information.

Tested on the freedom-e310-arty using the GDB regression suite.
There is one regression which is a corner case caused by a GDB bug.
If a breakpoint is set in GDB and then the executable file is
discarded, GDB reverts to its default information about address
sizes e.g. on the freedom-e310-arty, 0x20400000 becomes
0xffff20400000. As a result, GDB is unable to step over breakpoints
set before the executable was discarded.

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1852,9 +1852,8 @@ static int handle_halt(struct target *target, bool announce)
 			 * which watchpoint to step over. Setting need_strict_step to true
 			 * flags OpenOCD to disable triggers -> step -> enable triggers on
 			 * the next step or resume*/
-			if (! riscv_breakpoint_hit(target, 0, info->dpc)) {
+			if (!riscv_breakpoint_hit(target, 0, info->dpc))
 				info->need_strict_step = true;
-			}
 			break;
 		case DCSR_CAUSE_DEBUGINT:
 			target->debug_reason = DBG_REASON_DBGRQ;
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1852,8 +1852,9 @@ static int handle_halt(struct target *target, bool announce)
 			 * which watchpoint to step over. Setting need_strict_step to true
 			 * flags OpenOCD to disable triggers -> step -> enable triggers on
 			 * the next step or resume*/
-			if (!riscv_breakpoint_hit(target, 0, info->dpc))
+			if (! riscv_breakpoint_hit(target, 0, info->dpc)) {
 				info->need_strict_step = true;
+			}
 			break;
 		case DCSR_CAUSE_DEBUGINT:
 			target->debug_reason = DBG_REASON_DBGRQ;
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1848,9 +1848,13 @@ static int handle_halt(struct target *target, bool announce)
 			break;
 		case DCSR_CAUSE_HWBP:
 			target->debug_reason = DBG_REASON_WPTANDBKPT;
-			/* If we halted because of a data trigger, gdb doesn't know to do
-			 * the disable-breakpoints-step-enable-breakpoints dance. */
-			info->need_strict_step = true;
+			/* If we halted because of a watchpoint trigger, gdb doesn't know
+			 * which watchpoint to step over. Setting need_strict_step to true
+			 * flags OpenOCD to disable triggers -> step -> enable triggers on
+			 * the next step or resume*/
+			if (! riscv_breakpoint_hit(target, 0, info->dpc)) {
+				info->need_strict_step = true;
+			}
 			break;
 		case DCSR_CAUSE_DEBUGINT:
 			target->debug_reason = DBG_REASON_DBGRQ;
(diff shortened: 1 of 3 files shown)


timsifive:
It's not exactly clear to me what problem you're solving. I think it is:
1. Set a hardware execute breakpoint at <address>
2. Change PC to <address> from gdb (eg. using `p $pc=<address>`)
3. Tell gdb to resume execution.

In that case OpenOCD might not hit the trigger if it thinks it needs to do a strict step.

Is that the problem you're fixing?

craigblackmore:
The problem is that the target does not halt when the GDB jump command is used to jump to a hardware breakpoint e.g.

(gdb) hbreak *0x20004000
(gdb) jump *0x20004000

The target should halt at 0x20004000, but OpenOCD steps over the breakpoint and resumes.

I have just pushed a cleaner solution to this problem.

timsifive:
Thanks for clarifying. I assume you do `c` after those two commands. Is that really a bug? If you use a software breakpoint instead of a hardware breakpoint in that example, then you also won't halt there.

craigblackmore:
This is without doing ```c``` after those two commands. The bug is that ```jump *<address>``` is supposed to resume execution at ```<address>```, but halt immediately if there is a breakpoint at ```<address>``` (https://sourceware.org/gdb/onlinedocs/gdb/Jumping.html).

timsifive:
Playing with this, the problem only reproduces if the hart is currently halted because hitting a trigger. If it's at a software breakpoint then there's no issue.

The same problem also exists in riscv-013.c (which doesn't have strict_step()). I wonder if there's a fix that works for both implementations

timsifive:
I don't see how OpenOCD can differentiate between:
```
p/x $pc=<address>
c
```
and
```
jump <address>
```

In both cases, gdb sends the following packets:
```
Debug: 5977 1751 gdb_server.c:3142 gdb_input_inner(): received packet: 'P20=ce020110'
Debug: 6053 1764 gdb_server.c:400 gdb_put_packet_inner(): sending packet '$OK#9a'
Debug: 6059 1764 gdb_server.c:3142 gdb_input_inner(): received packet: 'Z1,100102ce,2'
Debug: 6314 1817 gdb_server.c:400 gdb_put_packet_inner(): sending packet '$OK#9a'
Debug: 6326 1818 gdb_server.c:3142 gdb_input_inner(): received packet: 'Z0,100102d0,2'
Debug: 6470 1839 gdb_server.c:400 gdb_put_packet_inner(): sending packet '$OK#9a'
Debug: 6476 1839 gdb_server.c:3142 gdb_input_inner(): received packet: 'vCont;c'
```

Is it always expected that we hit a trigger if it's set on the current address? Then how do you run off of triggers?

craigblackmore:
> The same problem also exists in riscv-013.c (which doesn't have strict_step()). I wonder if there's a fix that works for both implementations

I don't see this problem on 0.13 as riscv_openocd_resume only steps over watchpoints, whereas strict_step was stepping over both breakpoints and watchpoints.

craigblackmore:
Both
```
p/x $pc=<address>
c
```
and
```
jump <address>
```
should hit the trigger at ```<address>```.

Once a breakpoint is hit, GDB sends a packet to remove it. The next time you ```continue```, GDB sends the packets to do a single step -> add the breakpoint back in -> resume. For example:
```
(gdb) j main
Sending packet: $g#67...Packet received: 00000000a6054040b03f0080280c008000000000040000806861732000000000000000001d00000001000000910640400d0000000a00000000000020630000009106404065786974ffffffff0000008000000000e002008000000080605840400000008064000000408340400000000067616d200a50726f0000000000000000
Continuing at 0x40400092.
Sending packet: $P20=92004040#82...Packet received: OK
Sending packet: $Z1,40400092,2#a8...Packet received: OK
Sending packet: $vCont;c#a8...Packet received: T05
Sending packet: $p20#d2...Packet received: 92004040
Sending packet: $qXfer:threads:read::0,fff#03...Packet received: l<?xml version="1.0"?>\n<threads>\n</threads>\n
Sending packet: $z1,40400092,2#c8...Packet received: OK

Breakpoint 1, main () at jump.c:13
13        int i = 99;
(gdb) c
Continuing.
Sending packet: $vCont;s:0;c:0#2a...Packet received: T05
Sending packet: $p20#d2...Packet received: 96004040
Sending packet: $Z1,40400092,2#a8...Packet received: OK
Sending packet: $vCont;c#a8...
```

timsifive (approved):
I think I finally have this straight in my head. Thanks for bearing with me.

For software and hardware execute breakpoints, gdb takes care of disabling/reenabling breakpoints, so we shouldn't do that.

For data hardware breakpoints, gdb doesn't do that. (Probably because we don't tell gdb exactly which trigger we hit, which we can't know in 0.11, and no target implements in 0.13 AFAIK.) In that case we need to disable the watchpoints and step around them ourselves, which is what strict_step() in riscv-011.c does. There's no need for strict_step() to do that for execute breakpoints, though, and that's what the latest version of your patch removes.

I'd appreciate if you can add some comments that would have made your life easier when figuring this out, but the code LGTM.

Separately I'm looking into why this same bug happens with 0.13 against spike. It looks like it may be a bug in the spike implementation.

Thank you.

PR #285, opened 2018-08-03
Update s25fl256 flash device id, cypress now.

kenzhang82:
The device ID for s25fl256 seems to be changed from 0x00190201 to 0x00196001 .Tested using Freedom Arty with this device attached. Also change the brand of flash to cypress.


timsifive:
Shouldn't we still keep the old entry to support older boards?

kenzhang82:
Oops, sorry. Could have done that. Could you put that entry back and approve?

kenzhang82:
Yes, I just put it back. Can you approve and merge?

timsifive (approved):

PR #288, opened 2018-08-20
From upstream

timsifive:


timsifive:
I merged another change between the latest from_upstream creation and merging, which means git can't find the "proper" merge-base, which makes things confusing when comparing riscv vs upstream master.

This PR should fix that.

cgsfv (approved):

PR #292, opened 2018-08-23
Handle the qC packet

craigblackmore:
GDB sends the qC packet to query the current OpenOCD thread. The packet gets sent after ```target remote :port``` and if GDB receives a thread id, it switches to that thread, otherwise it defaults to thread 1.

The following examples show how this can be tested - they are from a test I have written for the GDB test suite.

Without qC packet handling, GDB defaults to thread 1:
```
(gdb) target remote :3333
Remote debugging using :3333
warning: Target-supplied registers are not supported by the current architecture
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0x00007fff404005c6 in ?? ()
(gdb) info threads
  Id   Target Id                     Frame 
* 1    Thread 1 (Name: Hart 0, RV32) 0x00007fff404005c6 in ?? ()
  2    Thread 2 (Name: Hart 1, RV32) 0x00007fff404005c6 in ?? ()
(gdb) PASS: gdb.riscv/connect-active-thread.exp: Start on thread 1
thread 2
[Switching to thread 2 (Thread 2)]
#0  0x00007fff404005c6 in ?? ()
(gdb) PASS: gdb.riscv/connect-active-thread.exp: Switch to thread 2
info threads
  Id   Target Id                     Frame 
  1    Thread 1 (Name: Hart 0, RV32) 0x00007fff404005c6 in ?? ()
* 2    Thread 2 (Name: Hart 1, RV32) 0x00007fff404005c6 in ?? ()
(gdb) PASS: gdb.riscv/connect-active-thread.exp: info threads: current thread is 2
disconnect
Ending remote debugging.
(gdb) PASS: gdb.riscv/connect-active-thread.exp: disconnect
target remote :3333
Remote debugging using :3333
warning: Target-supplied registers are not supported by the current architecture
warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0x00007fff404005c6 in ?? ()
(gdb) info threads
  Id   Target Id                     Frame 
* 1    Thread 1 (Name: Hart 0, RV32) 0x00007fff404005c6 in ?? ()
  2    Thread 2 (Name: Hart 1, RV32) 0x00007fff404005c6 in ?? ()
(gdb) FAIL: gdb.riscv/connect-active-thread.exp: Current thread is 2 after reconnecting to remote target
```

With qC packet handling, GDB switches to the correct thread:
```
(gdb) target remote :3333
Remote debugging using :3333
[...]

PR #294, opened 2018-08-27
Match on qC, but not qCRC.

timsifive:
Now we pass MulticoreRtosSwitchActiveHartTest again.
It was broken by #292.

Change-Id: I61f0ae41efda09fd48f732b2122fed2400a43d29


palmer-dabbelt (approved):
Thanks!

PR #296, opened 2018-08-28
Fix gdb_signal_reply() allocating too small buffer

timsifive:
In my test-case (64-bit OpenOCD, 64-bit target), OpenOCD ended up
sending gdb '$T05rwatch:1212340a00;thread:0000000000000002#89'
That's missing the final semi-colon after the thread id.

This fix increases the buffer size, and also removes the 0 padding on
the thread id.

This bug showed up when running MulticoreRtosSwitchActiveHartTest
against dual-hart, 64-bit spike.

Change-Id: I8c7d88e2d37b00cf3099f226a1a32671219802d5


ilg-ul:
This looks easy to accept upstream.

mwachs5 (approved):

Issue #297, opened 2018-08-29
This repo depends on unreliable submodule


mwachs5:
This repo depends on a submodule for jimtcl, located at http://repo.or.cz/r/jimtcl.git. This repo is frequently unavailable when cloning submodules. Is there any reason not to clone this repo to something else (e.g. riscv/jimtcl.git) and point to that version instead?

aswaterman:
agreed

timsifive:
Fine by me. Upstream does occasionally change which version we point to, but I've never noticed any problems from using a mismatched version.

mwachs5:
From the JimTCL page, there is an official mirror we could use instead: 
See http://repo.or.cz/w/jimtcl.git or the github mirror at https://github.com/msteveb/jimtcl

mwachs5:
Fixed by #298

PR #298, opened 2018-08-29
Don't use unreliably hosted submodules (git2cl, jimtcl, libjaylink)

mwachs5:
The repo.or.cz version is often unavailable.

Presumably we want to do something similar for the other submodules, but I didn't find as obvious official mirrors for those.

Addresses #297


aswaterman (dismissed):

timsifive (dismissed):

timsifive:
@aswaterman Can you add a README.riscv or something with shell commands required to update these repos?

aswaterman:
We don't need any new commands, fortunately.  Simply running the existing bootstrap program does the right thing.

aswaterman (dismissed):

aswaterman (approved):

Issue #295, opened 2018-08-27
gdb assumes watchpoints fire after the write happened


timsifive:
The RISC-V spec recommends write triggers fire before the write actually occurred. This causes gdb to not fire on watchpoints, because gdb attempts to only halt when the memory value actually changed. To do that it reads the memory the watchpoint was set at, and if it's the same as when the watchpoint was set then it automatically resumes. (See `case WP_VALUE_NOT_CHANGED` in bpstat_check_watchpoint() in gdb/breakpoint.c.)

Solutions that come to mind:
1. Have OpenOCD single step if it detects this. That makes gdb happy, but now users lose the ability to see what the memory value was before the write happened.
2. Change this gdb behavior when debugging RISC-V targets.

I'm not a fan of any of these.

@craigblackmore FYI.

craigblackmore:
GDB should be able to handle watchpoints that fire before the write happens - ```set_gdbarch_have_nonsteppable_watchpoint``` can be used to tell it to step over them first, but this has not been necessary in the C examples (e.g. below) I have tried (and currently OpenOCD manually steps over them anyway).

```
int main()
{
  int data;
  int t0;
  t0 = 1;
  data = t0;

  return 0;
}
```

```
(gdb) load 
Loading section .init, size 0x7c lma 0x40400000
Loading section .text, size 0x5d0 lma 0x4040007c
Loading section .rodata, size 0x54 lma 0x4040064c
Loading section .eh_frame, size 0x2c lma 0x404006a0
Loading section .data, size 0x434 lma 0x404006cc
Start address 0x40400000, load size 2816
Transfer rate: 3 KB/sec, 563 bytes/write.
(gdb) continue
Continuing.
Note: automatically using hardware breakpoints for read-only addresses.

Breakpoint 1, main () at /home/craig/work/projects/sifive/testing-public/gd
b/gdb/testsuite/gdb.riscv/trigger.c:6
6         t0 = 1;
(gdb) watch data
Hardware watchpoint 2: data
(gdb) continue
Continuing.

Hardware watchpoint 2: data

Old value = 0
New value = 1
```

In ```TriggerStoreAddressInstant ```, the ```bpstat_check_watchpoint()```function is not actually getting called, so I think the problem occurs earlier in ```bpstat_stop_status``` where it seems that GDB isn't even checking the watchpoint.

timsifive:
See also insightful responses at https://sourceware.org/ml/gdb/2018-08/msg00047.html

[...]

PR #300, opened 2018-09-06
Don't report exact watchpoint to gdb.

timsifive:
We should have a fix for #295 first.

Change-Id: Ic72a7a3fa866fbb5aaed22689adfebf9ce913b50


timsifive:
@craigblackmore Your change was causing problems for tests that assume gdb halts on watchpoints, which seems like a more basic feature than the lesser detail of reporting exactly which watchpoint hit.

aswaterman (approved):

mwachs5 (approved):

mwachs5:
Are there any tests in riscv-tests that need to be disabled accordingly?

timsifive:
No. The test that hit this problem was TriggerStoreAddressInstant, which was made to tolerate the new (now switched off again) behavior in 40dbc5118c9ac4beb4fc0a28cf4ad4cb56536111.

PR #301, opened 2018-09-12
Add wall clock timeout to mpsse_flush()

timsifive:
I think that libusb_handle_events_timeout_completed is supposed to make
progress or time out, but sometimes we hit a case where it makes no
progress, and mpsse_flush() loops forever. This wall clock timeout kicks
it out of that loop. OpenOCD appears to die afterwards, but that's still
an improvement.

Change-Id: Id9220557625834fb5b7dccf65251651a11a887f0

### DIFF
--- a/src/jtag/drivers/mpsse.c
+++ b/src/jtag/drivers/mpsse.c
@@ -22,6 +22,7 @@
 
 #include "mpsse.h"
 #include "helper/log.h"
+#include "helper/time_support.h"
 #include <libusb.h>
 
 /* Compatibility define for older libusb-1.0 */
--- a/src/jtag/drivers/mpsse.c
+++ b/src/jtag/drivers/mpsse.c
@@ -892,6 +892,7 @@ int mpsse_flush(struct mpsse_ctx *ctx)
 	}
 
 	/* Polling loop, more or less taken from libftdi */
+	int64_t start = timeval_ms();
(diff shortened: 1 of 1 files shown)


mwachs5 (commented):
Is this a bug introduced upstream? Do you know why this suddenly started to matter?

timsifive:
This function hasn't been touched since 2016. I don't know what changed that caused us to start seeing the problem now.

mwachs5:
I get this compile error:
```
../src/jtag/drivers/mpsse.c: In function ‘mpsse_flush’:
../src/jtag/drivers/mpsse.c:895:18: error: implicit declaration of function ‘timeval_ms’ [-Werror=implicit-function-declaration]
  int64_t start = timeval_ms()
```

mwachs5:
Do you need to:
```
#include <helper/time_support.h>
```

timsifive:
Evidently. I don't know why I didn't run into that problem.

mwachs5:
This seems to at least not break things, though I haven't hit the case in my testing of this PR.

mwachs5:
Is there any upstream discussion of this issue?

timsifive:
Not AFAIK.

mwachs5 (approved):
Works no worse than before and hopefully solves hanging problems.

PR #304, opened 2018-10-05
Revert "Don't report exact watchpoint to gdb. (#300)"

timsifive:
This reverts commit 933cb875a8fb6243df52208c725b1c6ebc9662e3.

https://github.com/riscv/riscv-openocd/issues/295 was fixed in gdb.


cgsfv (approved):
Tested it with the update you made to the ProgramHwWatchpoint test in riscv-tests/debug, and it works as expected.

PR #306, opened 2018-10-12
Fix segfault in riscv_deinit_target().

timsifive:
This would happen when OpenOCD is unable to connect to the JTAG device.

Change-Id: I1785fd5f5a20db9b4b574bdddfe3eab9bdc0b0bc

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -280,12 +280,17 @@ static void riscv_deinit_target(struct target *target)
 		free(info);
 	}
 	/* Free the shared structure use for most registers. */
-	free(target->reg_cache->reg_list[0].arch_info);
-	/* Free the ones we allocated separately. */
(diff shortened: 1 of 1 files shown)


paul-walmsley-sifive:
FSDK-146

cgsfv (approved):
I think it looks way better with all the checking of pointers before calling free.
Does it make any sense to add a test of this to riscv-tests repo?

timsifive:
A test case would be nice to have, but I was going to ignore it because this bug is unlikely to regress (this code is not often touched), the test infrastructure doesn't pay attention to OpenOCD's return code, and it's not a bug that is likely to cause much trouble for people. These are all individually weak reasons, but combine into my feeling. If you think it's important we can at least open an issue for a test case in riscv-tests.

cgsfv:
I totally trust your experience with the openocd code base and any need for testing - just felt like I had to ask.

PR #307, opened 2018-10-12
dmi_scan() allocate bytes depending on abits value

### LINKED ISSUE #303: in dmi_scan function 'in' and 'out' buff array size should be 9 bytes
in riscv-013.c dmi_scan function.
"in" and "out" buffer size are 8bytes (64bits). Buffer size should be 66bits.
it is corrupting scan_field structure.
Please change it to 9 byte array.

uint8_t in[8] = {0};
	uint8_t out[8];
	struct scan_field field = {
		.num_bits = info->abits + DTM_DMI_OP_LENGTH + DTM_DMI_DATA_LENGTH,
		.out_value = out,
		.in_value = in
	};

timsifive:
Fixes #303.

Change-Id: Iac45959cf342180c60cd0b5462f864ad81beddd2

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -459,15 +459,16 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 {
 	riscv013_info_t *info = get_info(target);
 	unsigned num_bits = info->abits + DTM_DMI_OP_LENGTH + DTM_DMI_DATA_LENGTH;
-	uint8_t in[(num_bits + 7) / 8];
-	uint8_t out[(num_bits + 7) / 8];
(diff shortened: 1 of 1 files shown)


cgsfv (dismissed):
Looks correct.
"(num_bits + 7) / 8" should only be calculated once.
I expect a test in riscv-tests would be able to show the error in the old code.

cgsfv (approved):
Perfect.

timsifive:
I've the code change you asked for. I'm on the fence about testing. It would require adding a --abits argument to spike, and then using that in riscv-tests just to test this one code path. If it turns out there's another bug regarding abits handling then I'm all for it, but I'm not convinced about the value.

PR #308, opened 2018-10-18
Fixing Eclipse block memory read

cgsfv:
OpenOCD was failing to read a block if one address within the block would cause an access exception in target. This did not work for Eclipse, as it is reading memory in blocks, meaning that values for addresses within the block that would not cause an access exception could not be read. This fix reverts to single element reads if the block read fails, thereby always returning correct values.

### DIFF
--- a/src/server/gdb_server.c
+++ b/src/server/gdb_server.c
@@ -390,7 +390,7 @@ static int gdb_put_packet_inner(struct connection *connection,
 			break;
 		}
 
-		LOG_WARNING("Discard unexpected char %c", reply);
+		LOG_DEBUG("Discard unexpected char %c", reply);
 	}
 #endif
 
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2293,7 +2293,7 @@ static int read_memory_progbuf(struct target *target, target_addr_t address,
 		target_addr_t address_i = address;
 		uint32_t size_i = size;
 		uint32_t count_i = 1;
-		uint8_t* buffer_i = buffer;
+		uint8_t *buffer_i = buffer;
 
 		for (uint32_t i = 0; i < count; i++, address_i += size_i, buffer_i += size_i) {
 			result = read_memory_progbuf_inner(target, address_i, size_i, count_i, buffer_i);
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2036,52 +2036,12 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
  * Read the requested memory, taking care to execute every read exactly once,
  * even if cmderr=busy is encountered.
  */
-static int read_memory_progbuf(struct target *target, target_addr_t address,
+static int read_memory_progbuf_inner(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, uint8_t *buffer)
 {
 	RISCV013_INFO(info);
 
 	int result = ERROR_OK;
-
-	LOG_DEBUG("reading %d words of %d bytes from 0x%" TARGET_PRIxADDR, count,
-			size, address);
-
-	select_dmi(target);
-
-	/* s0 holds the next address to write to
-	 * s1 holds the next data value to write
-	 */
-	uint64_t s0, s1;
-	if (register_read(target, &s0, GDB_REGNO_S0) != ERROR_OK)
-		return ERROR_FAIL;
-	if (register_read(target, &s1, GDB_REGNO_S1) != ERROR_OK)
-		return ERROR_FAIL;
-
-	if (execute_fence(target) != ERROR_OK)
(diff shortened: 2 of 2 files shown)


timsifive:
Does `make all` still pass in riscv-tests/debug?

src/server/gdb_server.c
```diff
@@ -390,7 +390,7 @@ static int gdb_put_packet_inner(struct connection *connection,
 			break;
 		}
 
-		LOG_WARNING("Discard unexpected char %c", reply);
+		LOG_DEBUG("Discard unexpected char %c", reply);
```
timsifive, line 393: Why did you make this change? Since it's in the target-independent part of the code, we'll have to give some justification when I merge this to the official OpenOCD branch.

src/target/riscv/riscv-013.c
```diff
@@ -2268,13 +2228,88 @@ static int read_memory_progbuf(struct target *target, target_addr_t address,
...
+	}
+	riscv_program_addi(&program, GDB_REGNO_S0, GDB_REGNO_S0, size);
+
+	if (riscv_program_ebreak(&program) != ERROR_OK)
+		return ERROR_FAIL;
+	riscv_program_write(&program);
+
+	result = read_memory_progbuf_inner(target, address, size, count, buffer);
+
+	/* The full read did not succeed, so we will try to read each word individually. */
```
timsifive: Can you move these 3 lines inside the if()? As it is it makes it sound like the read never succeeds.

src/target/riscv/riscv-013.c
```diff
@@ -2268,13 +2228,88 @@ static int read_memory_progbuf(struct target *target, target_addr_t address,
...
+
+		for (uint32_t i = 0; i < count; i++, address_i += size_i, buffer_i += size_i) {
+			result = read_memory_progbuf_inner(target, address_i, size_i, count_i, buffer_i);
+
+			/* The read of a single word failed, so we will just return 0 for that instead */
+			if (result != ERROR_OK) {
+				LOG_DEBUG("error reading single word of %d bytes from 0x%" TARGET_PRIxADDR,
+						size_i, address_i);
+
+				uint64_t value_i = 0;
```
timsifive, line 2308: This doesn't need to be it's own variable.

src/target/riscv/riscv-013.c
```diff
@@ -2268,13 +2228,88 @@ static int read_memory_progbuf(struct target *target, target_addr_t address,
...
+		return ERROR_FAIL;
+	riscv_program_write(&program);
+
+	result = read_memory_progbuf_inner(target, address, size, count, buffer);
+
+	/* The full read did not succeed, so we will try to read each word individually. */
+	/* This will not be fast, but reading outside actual memory is a special case anyway. */
+	/* It will make the toolchain happier, especially Eclipse Memory View as it reads ahead. */
+	if (result != ERROR_OK) {
+		target_addr_t address_i = address;
```
timsifive, line 2295: Why do all these variables have an _i suffix?

cgsfv:
All tests that passed before this fix, also passes after this fix.
I will be adding a specific test for this fix.

timsifive (changes requested):
I'm not a fan of hiding memory read failures from the users of OpenOCD. However, per the meeting, let's ignore that problem for now in the interest of getting everybody using at least the same OpenOCD branch again.

That just leaves a few small questions/comments.

src/server/gdb_server.c
cgsfv: This message is noise for the user and is of debug character and not warning, as the user has no possibilities of doing anything about it.

src/target/riscv/riscv-013.c
cgsfv: You mean the 3 comment lines, right?
If so, I agree that is better.

src/target/riscv/riscv-013.c
cgsfv: I think it makes the code easier to read.

src/target/riscv/riscv-013.c
cgsfv: It was just my way to signal corresponding variables that is used in the loop with the loop variable named i.

timsifive: Alright. It's not my habit, but I can see the use in that. The compiler will catch it if you do end up reusing the same variable outside the loop.

src/server/gdb_server.c
timsifive: I will try to push this upstream, but the counter argument is probably that this message indicates a bug either in OpenOCD or the debugger that's connected to it.

timsifive (approved):

src/server/gdb_server.c
ilg-ul: FYI, in my latest release of GME OpenOCD I'm not carrying in any of the changes you make in the target-independent code until they are upstreamed.

PR #309, opened 2018-10-18
Add m4 macro dir

jcpunk:
This silences the suggestion to add the m4 macrodir automatically created by autotools.


timsifive:
That looks like a good change, but really has nothing to do with RISC-V. Instead of sending them here, can you contribute them to OpenOCD mainline? They will then filter down to this branch. Instructions on contributing to OpenOCD are at http://openocd.org/doc-release/doxygen/patchguide.html.

PR #311, opened 2018-10-19
Old fixes from June

cgsfv:


src/rtos/riscv_debug.c
```diff
@@ -124,7 +124,12 @@ static int riscv_gdb_thread_packet(struct connection *connection, const char *pa
 			return ERROR_OK;
 		}
 
-		if (strcmp(packet, "qC") == 0) {
+		if (strncmp(packet, "qTStatus", 8) == 0) {
```
timsifive: Why strncmp() instead of strcmp()? qTStatus doesn't appear to take any arguments (https://sourceware.org/gdb/current/onlinedocs/gdb/Tracepoint-Packets.html)

src/rtos/riscv_debug.c
```diff
@@ -124,7 +124,12 @@ static int riscv_gdb_thread_packet(struct connection *connection, const char *pa
...
+			gdb_put_packet(connection, "T0", strlen("T0"));
```
timsifive: The other code in gdb_server.c all uses 2 instead of strlen("T0"), so I'd mimic that. (Really there ought to be a gdb_put_string() function, but it's probably not the time to make that improvement.)

src/server/gdb_server.c
```diff
@@ -781,6 +781,10 @@ static void gdb_signal_reply(struct target *target, struct connection *connectio
 		gdb_connection->ctrl_c = 0;
 	}
 
+	char sig_reply_str[46];
+	memset(sig_reply_str, '\0', 46);
+	strncpy(sig_reply_str, sig_reply, 46);
+	LOG_DEBUG("signal reply: %s", sig_reply_str);
```
timsifive: Is this important, or just some temp code added when debugging?

src/target/riscv/batch.c
```diff
@@ -51,7 +51,7 @@ int riscv_batch_run(struct riscv_batch *batch)
 
 	keep_alive();
 
-	LOG_DEBUG("running a batch of %ld scans", (long)batch->used_scans);
+	LOG_DEBUG_IO("running a batch of %ld scans", (long)batch->used_scans);
```
timsifive: I'd be happier getting rid of these debug messages in the batch functions. It's easy enough to add them in if there turns out to be an issue with this code, but it's been very stable.

src/target/riscv/batch.c
```diff
@@ -135,15 +135,15 @@ void riscv_batch_add_nop(struct riscv_batch *batch)
...
+	LOG_DEBUG_IO("  added NOP with in_value=0x%p", field->in_value);
 }
 
 void dump_field(const struct scan_field *field)
 {
 	static const char * const op_string[] = {"-", "r", "w", "?"};
 	static const char * const status_string[] = {"+", "?", "F", "b"};
 
-	if (debug_level < LOG_LVL_DEBUG)
+	if (debug_level < LOG_LVL_DEBUG_IO)
```
timsifive: The dump_field() ones, however, I'd like to keep at LOG_LVL_DEBUG. This is the stuff I look at when something goes wrong and people send me debug logs. We're just about at the point where people have started to learn to run openocd with `-d` to get a debug log, but with this change those statements won't be included and they'll have to be retrained to use `-d4` instead.

src/target/riscv/riscv-013.c
```diff
@@ -1771,13 +1771,38 @@ static void write_to_buf(uint8_t *buffer, uint64_t value, unsigned size)
...
+		struct riscv_program program;
+		riscv_program_init(&program, target);
+		riscv_program_fence_i(&program);
+		riscv_program_fence(&program);
+		int result = riscv_program_exec(&program, target);
+		if (result != ERROR_OK)
+			LOG_DEBUG("Unable to execute pre-fence");
+	}
+
+	for (int i = 0; i < riscv_count_harts(target); ++i) {
```
timsifive, line 1788: What happens in this loop if one of the harts isn't halted?

timsifive: In `-rtos riscv` all other harts are probably halted, but in that case it seems like a lot of unnecessary work to execute fences on each hart.

This function is called at the beginning of read_memory_progbuf(), and the end of write_memory_progbuf(). What exactly is the problem we're trying to fix? The only one I can think of off the top of my head is to write through one hart, and then resume another which depends on the data written by the first. That could be fixed with way less overhead by adding a call to execute_fence() from riscv013_step_or_resume_current_hart().

src/target/riscv/riscv-013.c
```diff
@@ -3305,19 +3330,7 @@ int riscv013_dmi_write_u64_bits(struct target *target)
 
 static int maybe_execute_fence_i(struct target *target)
 {
-	RISCV013_INFO(info);
-	RISCV_INFO(r);
-	if (info->progbufsize + r->impebreak >= 2) {
```
timsifive, line 3310: By removing this check, you're always reporting an error on debug systems where you can't execute a fence, eg. because they don't have a Program Buffer. They already get a warning about this at https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv-013.c#L1443.

src/target/riscv/riscv.c
```diff
@@ -1200,8 +1204,12 @@ int riscv_openocd_halt(struct target *target)
...
 	register_cache_invalidate(target->reg_cache);
 	if (riscv_rtos_enabled(target)) {
-		target->rtos->current_threadid = r->rtos_hartid + 1;
-		target->rtos->current_thread = r->rtos_hartid + 1;
+		if (r->rtos_hartid != -1) {
+			LOG_DEBUG("halt requested on RTOS hartid %d", r->rtos_hartid);
+			target->rtos->current_threadid = r->rtos_hartid + 1;
+			target->rtos->current_thread = r->rtos_hartid + 1;
+		} else
+			LOG_DEBUG("halt requested, but no known RTOS hartid");
```
timsifive, line 1212: Isn't this a WARNING at least?

src/target/riscv/riscv.c
```diff
@@ -1946,14 +1956,15 @@ int riscv_set_current_hartid(struct target *target, int hartid)
 	if (!target_was_examined(target))
 		return ERROR_OK;
 
+#if 0
```
timsifive: Seems a bit heavy-handed, but given that hopefully we'll switch away from `-rtos riscv` it might not be worth the effort to fix properly.

If we are removing this code, let's remove it instead of using `#if 0`.

src/target/riscv/riscv.c
```diff
@@ -2035,7 +2046,15 @@ int riscv_get_register_on_hart(struct target *target, riscv_reg_t *value,
 		int hartid, enum gdb_regno regid)
 {
 	RISCV_INFO(r);
+
+	if (hartid != riscv_current_hartid(target))
+		riscv_invalidate_register_cache(target);
+
 	int result = r->get_register(target, value, hartid, regid);
+
+	if (hartid != riscv_current_hartid(target))
```
timsifive, line 2045: I'm curious how the current hartid can not match the hartid of the last hart we read a register from.

src/target/startup.tcl
```diff
@@ -203,7 +203,6 @@ proc init_target_events {} {
 	foreach t $targets {
 		set_default_target_event $t gdb-flash-erase-start "reset init"
 		set_default_target_event $t gdb-flash-write-end "reset halt"
-		set_default_target_event $t gdb-attach "halt"
```
timsifive: I don't know what exactly happens in this file, but changing some generic startup that applies to all targets is impossible to test for regressions. We shouldn't do that. Even though I agree this "halt" should be a nop.

timsifive (commented):
Looks alright, but I have a bunch of comments/questions.

src/rtos/riscv_debug.c
palmer-dabbelt: I must have just been pattern matching from whatever else I did.

src/rtos/riscv_debug.c
palmer-dabbelt: OK.

src/server/gdb_server.c
palmer-dabbelt: it's probably just debug code.

src/target/riscv/batch.c
palmer-dabbelt: OK.

src/target/riscv/batch.c
palmer-dabbelt: OK.

src/target/riscv/riscv-013.c
palmer-dabbelt: This is necessary to prevent false-sharing errors on non-coherent systems.

src/target/riscv/riscv-013.c
palmer-dabbelt: IIRC I removed this warning because it was firing on systems I knew there was a progbuf and I just didn't want to deal with it at the time, but I might be wrong about that.

src/target/riscv/riscv.c
palmer-dabbelt: I don't think so: we're just reverting to the heuristics here, which IIRC trigger quite frequently.

src/target/riscv/riscv.c
palmer-dabbelt: I think this might be masking another issue, but I'm not sure.  If it's not then removing it seems fine.

src/target/riscv/riscv.c
palmer-dabbelt: My commit says "The 0.13 code now caches registers interally, so when reading registers on a different hart we need to invalidate the cache.", though I'm not sure how that's relevant.

src/target/startup.tcl
palmer-dabbelt: Makes sense.  I don't remember exactly why I did this, but it was probably some request to not implicitly halt under all conditions.

palmer-dabbelt:
@cgsfv I left some comments, if you have any more questions feel free to ask.

src/target/riscv/riscv.c
timsifive: Alright.

src/rtos/riscv_debug.c
timsifive: Use strcmp() instead of strncmp().

src/rtos/riscv_debug.c
timsifive: Use 2 instead of strlen("T0")

src/server/gdb_server.c
timsifive: Remove these 4 lines.

src/target/riscv/batch.c
timsifive: Remove the LOG_DEBUG messages that were changed to LOG_DEBUG_IO (except in dump_field() below).

src/target/riscv/batch.c
timsifive: Revert this change, so use LOG_LVL_DEBUG again.

src/target/riscv/riscv-013.c
timsifive: Let's put in these changes now, and file a new issue to track making it an option that should only be used on rare non-coherent systems.

src/target/riscv/riscv-013.c
timsifive: Let's leave this check, but inside of it call execute_fence() to get the new ultra-paranoid behavior.

src/target/riscv/riscv.c
timsifive: Remove the entire #if 0 block.

src/target/riscv/riscv.c
timsifive: Let's leave it for now. It should be a harmless check.

src/target/startup.tcl
timsifive: Keep the original version of this file.

src/target/riscv/riscv-013.c
```diff
@@ -3308,14 +3333,7 @@ static int maybe_execute_fence_i(struct target *target)
 	RISCV013_INFO(info);
 	RISCV_INFO(r);
 	if (info->progbufsize + r->impebreak >= 2) {
```
timsifive: This needs to >= 3 now that execute_fence() executes both a fence and a fence.i.

timsifive (changes requested):
This looks good, except for one tiny change.

timsifive (approved):

PR #313, opened 2018-10-22
Clean up fespi flashing code

timsifive:
These changes are all in response to the review I got when attempting to send this code upstream: http://openocd.zylin.com/#/c/4656/

There's potentially one more change, which is to move a lot more smarts into the target algorithm code. I'll at least take a look to see what's involved in that. In any case I think this will be much easier to review this batch and then the final change than all at once.


cgsfv (approved):
Looks fine.

PR #320, opened 2018-10-30
Support for two-wire cJTAG OSCAN1 signaling thru FTDI devices with appropriate pinout 

GregSavin:
cJTAG OSCAN1, in lieu of 4-wire JTAG, is starting to be a configuration option for some SiFive hardware.  An FTDI-based adapter that can be configured to drive the bidirectional pin TMSC is assumed for this topology.  Specifically, the Olimex ARM-USB-TINY-H with the ARM-JTAG-SWD adapter, connected to a SiFive cJTAG-enabled target board is the only known concrete topology, currently.  But in theory, other FTDI based devices that can drive a two-wire bidirectional signaling pattern could be made to work in this scheme in the future.

These code changes are offered as a way to drive that topology.  It's translating IR/DR and JTAG traversal commands to the two-wire clocking and signaling.


src/jtag/drivers/oscan1_ftdi.c
```diff
@@ -0,0 +1,1538 @@
+/**************************************************************************
+*   Copyright (C) 2012 by Andreas Fritiofson                              *
```
timsifive: I'm not sure what the copyright should be, but this doesn't look right. In the RISC-V specific code I've not bothered with a copyright line at all, nor with the GPL info. Upstream has accepted that without comment.

src/jtag/drivers/oscan1_ftdi.c
```diff
@@ -0,0 +1,1538 @@
...
+ * Hereafter this is called the "MPSSE Spec".
+ *
+ * The datasheet for the ftdichip.com's FT2232D part is here:
+ * http://www.ftdichip.com/Documents/DataSheets/DS_FT2232D.pdf
+ *
+ * Also note the issue with code 0x4b (clock data to TMS) noted in
+ * http://developer.intra2net.com/mailarchive/html/libftdi/2009/msg00292.html
+ * which can affect longer JTAG state paths.
+ *
+ * This driver is based on ftdi.c, but is a variant that assumes a two-wire
```
timsifive: I would remove the comment that's identical to the one in ftdi.c, and add a line to this paragraph something like:
See ftdi.c for more comments about these JTAG adapters.

timsifive (changes requested):
oscan1_ftdi.c duplicates way too much of ftdi.c.
ftdi.c is 1289 lines
oscan1_ftdi.c is 1538
The diff between them inserts 415 lines and deletes 166 lines. And at least half of those simply change ftdi to oscan1 without any functional change at all.

The simplest solution is probably to add to ftdi.c what you need added there (checking for FTDI_OSCAN1, probably) and reuse as much of the ftdi.c as you currently are. The other option is to make 2 new files: one for the functions shared between ftdi and oscan1, and one for the oscan1-specific stuff. It might be worth asking openocd-devel@lists.sourceforge.net which solution is preferred before doing a bunch of work.

Both of those solutions also have the benefit that I only need to review code that is actually new, instead of reading 1000 lines of code that haven't actually changed (and are probably solid).

GregSavin:
Thanks Tim.  Yeah I was conflicted about modifying ftdi.c to add support a for such a niche topology, thinking there might be more resistance to upstreaming later, but modifying ftdi.c would be a much smaller diff for sure.  I'll see about refactoring my changes by adding them to ftdi.c (not a whole lot of work), and then we'll see if that looks better.

GregSavin:
My most recent commit on this branch has incorporated the previously differentiating content of oscan1_ftdi.c into ftdi.c (establishing OSCAN1 as a non-default mode that the ftdi driver can be placed in), and removed oscan1_ftdi.c.

src/jtag/drivers/ftdi.c
```diff
@@ -95,6 +108,29 @@ static uint8_t ftdi_jtag_mode = JTAG_MODE;
 
 static bool swd_mode;
 
+#if BUILD_FTDI_OSCAN1 == 1
+/*
+  cJTAG OSCAN1, in lieu of 4-wire JTAG, is starting to be a configuration option
+  for some SiFive hardware. An FTDI-based adapter that can be configured to drive
+  the bidirectional pin TMSC is assumed for this topology. Specifically, the Olimex
+  ARM-USB-TINY-H with the ARM-JTAG-SWD adapter, connected to a SiFive cJTAG-enabled
+  target board such a topology, currently. But in theory, other FTDI based devices
```
timsifive: missing a verb?

src/jtag/drivers/ftdi.c
```diff
@@ -711,6 +796,230 @@ static int ftdi_quit(void)
 	return ERROR_OK;
 }
 
+#if BUILD_FTDI_OSCAN1 == 1
+static void oscan1_mpsse_clock_data(struct mpsse_ctx *ctx, const uint8_t *out, unsigned out_offset, uint8_t *in,
+		     unsigned in_offset, unsigned length, uint8_t mode)
+{
+	uint8_t zero = 0;
```
timsifive: Can these two be const?

timsifive: Might want to make them static, as well.

src/jtag/drivers/ftdi.c
```diff
@@ -83,6 +84,18 @@
 /* FTDI access library includes */
 #include "mpsse.h"
 
+#define DEBUG_IO(expr...) LOG_DEBUG(expr)
```
timsifive: DEBUG_JTAG_IO might be more suitable than LOG_DEBUG. Its output is only displayed with -d4, and only compiled in if the OpenOCD build is configured with --enable-verbose-jtag-io.

src/jtag/drivers/ftdi.c
```diff
@@ -711,6 +796,230 @@ static int ftdi_quit(void)
...
+	    
+  
+	  }
+}
+
+
+static void oscan1_mpsse_clock_tms_cs(struct mpsse_ctx *ctx, const uint8_t *out, unsigned out_offset, uint8_t *in,
+		       unsigned in_offset, unsigned length, bool tdi, uint8_t mode)
+{
+	uint8_t zero = 0;
```
timsifive: Ditto.

src/jtag/drivers/ftdi.c
```diff
@@ -711,6 +796,230 @@ static int ftdi_quit(void)
...
+		LOG_ERROR("Can't run cJTAG online/activate escape sequences: TCK signal is not defined");
+
+	if (!tdi)
+		LOG_ERROR("Can't run cJTAG online/activate escape sequences: TDI signal is not defined");
+
+	if (!tms)
+		LOG_ERROR("Can't run cJTAG online/activate escape sequences: TMS signal is not defined");
+
+	if (!tdo)
+		LOG_ERROR("Can't run cJTAG online/activate escape sequences: TDO signal is not defined");
```
timsifive, line 1019: Do you need to return here if any of these signals is not defined?

src/jtag/drivers/ftdi.c
```diff
@@ -95,6 +108,29 @@ static uint8_t ftdi_jtag_mode = JTAG_MODE;
 
 static bool swd_mode;
 
+#if BUILD_FTDI_OSCAN1 == 1
+/*
+  cJTAG OSCAN1, in lieu of 4-wire JTAG, is starting to be a configuration option
+  for some SiFive hardware. An FTDI-based adapter that can be configured to drive
```
timsifive: If possible, can you emphasize that this is not some SiFive-specific invention, but there may be other hardware out there that this code is useful for?

src/jtag/drivers/ftdi.c
```diff
@@ -711,6 +796,230 @@ static int ftdi_quit(void)
...
+
+	if (!tms)
+		LOG_ERROR("Can't run cJTAG online/activate escape sequences: TMS signal is not defined");
+
+	if (!tdo)
+		LOG_ERROR("Can't run cJTAG online/activate escape sequences: TDO signal is not defined");
+	
+
+	/* TCK=0, TMS=1, TDI=0 (drive TMSC to 0 baseline) */
+	oscan1_set_tck_tms_tdi(tck,'0',tms,'1',tdi,'0');
```
timsifive: I love the comments here, but I think the compiled code would be much smaller if you had something like:
sequence[] = {{1,1,0},
{1,1,1},
...};

and then a loop to iterate over the sequence and call oscan1_set_tck_tms_tdi for each triple.

But I don't feel strongly, and I'm not convinced upstream would care too much either.

src/jtag/drivers/ftdi.c
```diff
@@ -951,6 +1260,29 @@ COMMAND_HANDLER(ftdi_handle_tdo_sample_edge_command)
...
 
+#if BUILD_FTDI_OSCAN1 == 1
+COMMAND_HANDLER(ftdi_handle_oscan1_mode_command)
+{
+	if (CMD_ARGC > 1) {
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	if (CMD_ARGC == 1) {
+		if (!strcmp(CMD_ARGV[0], "on")) {
```
timsifive: There's a COMMAND_PARSE_ON_OFF macro that will do this for you.

timsifive (changes requested):
Overall this looks good. I have a bunch of feedback, but this feels really close to what we want.

This code violates the OpenOCD coding standards in all kinds of ways. The Travis build is supposed to flag that, but somehow doesn't. I'll fix that. In your own checkout can you run:
```git diff -U20 HEAD~40 | ./tools/scripts/checkpatch.pl --no-signoff -```
and fix whatever it complains about?

GregSavin:
Thanks @timsifive , the latest commits to this branch are intended to resolve your most recent set of comments (first a commit to address the non-formatting things and then a separate commit that addressed formatting problems flagged by checkpatch).

timsifive (approved):

GregSavin:
Thanks @timsifive for reviewing/approving.  At this point, do I need to take further action for this to be merged?  Or is that something that somebody other than myself would do?

timsifive:
My workflow is:
1. Create PR.
2. Get review approval.
3. Click "Squash and merge"

You're welcome to do the same.

In addition, I try to "Squash and merge" any changes that I've approved after giving the 24 hours for other people to comment on.

PR #321, opened 2018-11-01
Install patchutils for the build.

timsifive:
This contains filterdiff, which we need to check that our changes
conform to OpenOCD style.

Change-Id: Id522f4e62fee3efad4e0e00933abfeada9635624

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,7 +64,7 @@ script:
     # so this should work fine most of the time, and be a lot better than not
     # checking at all.
     - git diff -U20 HEAD~40 |
-        filterdiff -x "src/jtag/drivers/libjaylink/*" -x "tools/git2cl/*" |
+        filterdiff -x "b/src/jtag/drivers/libjaylink/*" -x "b/tools/git2cl/*" |
         ./tools/scripts/checkpatch.pl --no-signoff -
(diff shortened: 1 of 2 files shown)


GregSavin (approved):
I'm not familiar with travis, but I get the intent is to trigger a style check (checkpatch) on commit, excluding checking code in submodules that aren't part of OpenOCD proper.  That looks right to me, as do the resulting style adjustments in riscv-013.c

.travis.yml
```diff
@@ -64,7 +64,7 @@ script:
     # so this should work fine most of the time, and be a lot better than not
     # checking at all.
     - git diff -U20 HEAD~40 |
-        filterdiff -x "src/jtag/drivers/libjaylink/*" -x "tools/git2cl/*" |
+        filterdiff -x "b/src/jtag/drivers/libjaylink/*" -x "b/tools/git2cl/*" |
```
cgsfv, line 67: Did libjaylink and git2cl move or does the "b/..." mean something special?

timsifive: They didn't move. This line probably never worked.
In git diff output it shows the original version with a path of a/..., and the new version with a path of b/.... filterdiff filters on the new path, so we need to mention b/... in the paths we give it.

PR #322, opened 2018-11-02
FIX(src/target/riscv/riscv.c): riscv_add_breakpoint: invalid 32bit transactions size

aka-sps:
RVC: invalid 32bit transactions for 32bit instruction with 16bit word alignment.


timsifive (approved):
This looks great. Thank you!

Do you have an example program where this problem happened? If I have time I'll turn it into a test case.

aka-sps:
You can use any RVC program with a mix of 16-bit and 32-bit instructions, where 32-bit instructions cross the 32-bit border.

PR #325, opened 2018-11-06
Doxygen style, too.

timsifive:
Change-Id: I85e60e8577c4177ac7094ae41ee84357b292a89c

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -754,7 +754,7 @@ static int write_abstract_arg(struct target *target, unsigned index,
 }
 
 /**
- * @size in bits
+ * @par size in bits
  */
 static uint32_t access_register_command(struct target *target, uint32_t number,
 		unsigned size, uint32_t flags)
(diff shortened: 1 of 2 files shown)


GregSavin (approved):
Do Doxygen style issues get flagged by the checkpatch script?

timsifive:
No. @aka-sps brought them up.

aka-sps:
There are a lot of invalid Doxygen tags in the openocd srs (most of them are out of 'riscv' target).
In 'riscv' there are, for example, `@size` instead `@par size`.

Doxygen itself is the best checker.

aka-sps:
Probably we should not bother about the style because the document doc/manual/style.txt itself violates official standards.
For example, typedef naming: http://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html

timsifive:
We have to conform to the OpenOCD style as enforced by the script, otherwise our work will not be accepted upstream. OpenOCD is not a GNU project.

As for Doxygen, I'm happy to make our code clean with it, but I don't want to clean up code that we didn't write for RISC-V support. OpenOCD has its own system for accepting contributions to fix that.

aka-sps:
ANSI/ISO standards reserve '_t'-ended identifiers too.
There were commits with their own definitions of time_t.

PR #327, opened 2018-11-09
examine() should leave halted harts halted

### LINKED ISSUE #326: riscv target `examine` changes HART state
HART in `halt` state becomes `running` after `examine`.

Uncontrolled and unexpected launch.

timsifive:
Previously all harts would be resumed at the end of examine().
Fixes #326.

Change-Id: Id82b361e98f151911f8679538ee4b3c754efd6a5


GregSavin (approved):
Can see that the code no longer resumes previously halted harts.

cgsfv (approved):
Looks correct.

PR #344, opened 2018-12-13
Deal with DMI busy in block reads/writes

timsifive:
This was trivial for writes, but more painful for reads. In doing so (I think) I simplified the block read code a little, although it's still a bit messy.

I finally went through the effort to test this properly. To do that I added a `riscv reset_delays` command. Combined with https://github.com/riscv/riscv-isa-sim/pull/265 and some riscv-tests changes which I'll push once these 2 have merged, I now get coverage of all the corner cases in the memory progbuf functions by running `make all` in riscv-tests/debug.

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -206,8 +206,6 @@ typedef struct {
 
 	bool need_strict_step;
 	bool never_halted;
-
-	int reset_delays_wait;
 } riscv011_info_t;
 
 typedef struct {
@@ -360,10 +358,11 @@ static void add_dbus_scan(const struct target *target, struct scan_field *field,
 		uint16_t address, uint64_t data)
 {
 	riscv011_info_t *info = get_info(target);
+	RISCV_INFO(r);
 
-	if (info->reset_delays_wait >= 0) {
-		info->reset_delays_wait--;
-		if (info->reset_delays_wait < 0) {
+	if (r->reset_delays_wait >= 0) {
+		r->reset_delays_wait--;
+		if (r->reset_delays_wait < 0) {
 			info->dbus_busy_delay = 0;
 			info->interrupt_high_delay = 0;
 		}
@@ -1385,13 +1384,6 @@ static int halt(struct target *target)
 	return ERROR_OK;
 }
 
-static int reset_delays(struct target *target, int wait)
-{
-	riscv011_info_t *info = get_info(target);
-	info->reset_delays_wait = wait;
-	return ERROR_OK;
-}
-
 static int init_target(struct command_context *cmd_ctx,
 		struct target *target)
 {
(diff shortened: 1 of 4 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -503,14 +513,20 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 	return buf_get_u32(in, DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH);
 }
 
-static int dmi_op_timeout(struct target *target, uint32_t *data_in, int dmi_op,
-		uint32_t address, uint32_t data_out, int timeout_sec)
+/* If dmi_busy_encountered is non-NULL, this function will use it to tell the
+ * caller whether DMI was every busy during this call. */
```
mwachs5: every -> ever

timsifive: Fixed.

timsifive:
Can I get a review/approval for this?

GregSavin:
I'll take a look today.  It reaches deeper RISCV debug architecture-level details than I've reviewed before, which is good (for me).

src/target/riscv/riscv-011.c
```diff
@@ -206,6 +206,8 @@ typedef struct {
 
 	bool need_strict_step;
 	bool never_halted;
+
+	int reset_delays_wait;
```
GregSavin: I wonder if a code comment generally describing the purpose of reset_delays_wait might be worthwhile here, and in riscv013_info_t.  Although I do see the command help text for the reset_delays Tcl command, which explains the same thing at a different level.

src/target/riscv/riscv-013.c
```diff
@@ -2095,31 +2104,43 @@ static int read_memory_progbuf_inner(struct target *target, target_addr_t addres
...
-		goto error;
+	if (execute_abstract_command(target, command) != ERROR_OK)
+		return ERROR_FAIL;
 
 	/* First read has just triggered. Result is in s1. */
 
-	dmi_write(target, DMI_ABSTRACTAUTO,
-			1 << DMI_ABSTRACTAUTO_AUTOEXECDATA_OFFSET);
+	if (count == 1) {
+		uint64_t value;
```
GregSavin, line 2147: Is this "if" block a new optimization special case?

GregSavin (dismissed):

src/target/riscv/riscv-013.c
timsifive: Yes, but it's also essential after the rewrite because I didn't add all the code below that's necessary to properly work when there's just 1 word being read.

src/target/riscv/riscv-011.c
timsifive: I'll add comments.

GregSavin (approved):

PR #345, opened 2019-01-09
Add 'riscv set_ir' command to set IR value for JTAG registers.

darius-bluespec:
This allows using different TAP addresses, for example, if using
BSCANE2 primitives on a Xilinx FPGA.

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -9145,6 +9145,17 @@ When on, prefer to use System Bus Access to access memory.  When off, prefer to
 use the Program Buffer to access memory.
 @end deffn
 
+@deffn Command {riscv set_ir} (@option{idcode}|@option{dtmcs}|@option{dmi}) [value]
+Set the IR value for the specified JTAG register.  This is useful, for
+example, when using the existing JTAG interface on a Xilinx FPGA by
+way of BSCANE2 primitives that only permit a limited selection of IR
+values.
+
+When utilizing version 0.11 of the RISC-V Debug Specification,
+@option{dtmcs} and @option{dmi} set the IR values for the DTMCONTROL
+and DBUS registers, respectively.
+@end deffn
+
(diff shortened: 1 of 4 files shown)


timsifive:
I'm not familiar with Xilinx' BSCANE2. Is the problem simply that it uses the same IR values that RISC-V specifies? If so, this change looks pretty reasonable. Please do fix the Travis build, which is failing because of the OpenOCD style guide checks.

darius-bluespec:
The Xilinx BSCANE2 primitive allows connecting user logic to the FPGA scan chain, and therefore the FPGA JTAG pins.  The issue is the primitive implements the JTAG state machine and only allows the user to provide DR for certain IR values (up to four, and the values are different for different parts).

I updated the PR to address the style issues.  There still is one failure, but it doesn't seem like it's my fault (something about build dependencies).

timsifive (changes requested):
Looks good overall. Have you looked at the 0.11 support? Quickly checking it appears to me that this change should take effect in the same way, but I may be wrong.

Can you document this command to the "RISC-V Debug Configuration Commands" in doc/openocd.texi, along with a sentence or two on why somebody might want to use it?

darius-bluespec:
I added documentation for the command and rebased.

I have not tested debug spec 0.11 support, but I did review the code and believe it will work similarly.  In the 0.13 code there was one location that hardcoded the IR value, which I fixed.  I didn't see anything similar for 0.11 support.  I put a note in the documentation about the difference in register names for 0.11.

timsifive (approved):
Thanks! I'll merge this tomorrow assuming nobody else comments.

jrrk:
The lowrisc project uses this feature to provide access to the debug transport module on low end boards such as Nexys4DDR, see https://www.lowrisc.org/docs/jtag/ for details. Needless to say it does not use this version of openocd as it was put together in Q3 of last year.

PR #346, opened 2019-01-22
Move version check until after dmactive=1.

timsifive:
This should allow OpenOCD to work with targets where version is not readable when dmactive=0, which is allowed by the spec.

I don't have access to hardware that does that to confirm this actually works.


timsifive:
This fixed a problem for at least one person, and should have no downsides. I had thought that this might break auto-detection of 0.11 targets, but the version number in the DTM also changed and is checked before we even get here.

aswaterman (approved):
I can't test this right now, but this definitely seems like the right thing to do.  The implication is that future versions of the spec had better not be backwards-incompatible with respect to the mechanism used to set dmactive=1, but that sort of restriction seems to me to be a feature, not a bug.

GregSavin (approved):
Makes sense to me.  (I saw some recent email discussion related to this issue, so somewhat familiar with the history now).

PR #347, opened 2019-01-25
`-rtos hwthread` support

timsifive:
`-rtos hwthread` is a target-independent alternative to `-rtos riscv`. It's not in mainline OpenOCD yet. I took the code from http://openocd.zylin.com/#/c/3999/, and made some changes to get all our tests to pass. Specifically there's now a mechanism to read and write arbitrary registers. Another independent change is to treat triggers on SMP targets the same as hardware breakpoints: e.g. make sure they're set on each target that's part of the SMP group.

All our existing tests pass with either of the -rtos options now. To use this new feature, openocd.cfg needs to explicitly list each hart. So instead of:
```
target create $_TARGETNAME riscv -chain-position $_TARGETNAME -rtos riscv
```

Use:
```
target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -rtos hwthread
target smp $_TARGETNAME_0 $_TARGETNAME_1
```

It would be nice to deprecate `-rtos riscv` because it's more of a hack than `-rtos hwthread` and getting rid of it would make the riscv code a lot simpler.

### DIFF
--- a/src/rtos/hwthread.c
+++ b/src/rtos/hwthread.c
@@ -1,0 +1,394 @@
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ *   This program is distributed in the hope that it will be useful,       *
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
+ *   GNU General Public License for more details.                          *
+ *                                                                         *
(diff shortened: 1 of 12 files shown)


src/rtos/hwthread.c
```diff
@@ -0,0 +1,382 @@
...
+	char tmp_str[HW_THREAD_NAME_STR_SIZE];
+	threadid_t tid = threadid_from_target(curr);
+
+	memset(tmp_str, 0, HW_THREAD_NAME_STR_SIZE);
+
+	/* thread-id is the core-id of this core inside the SMP group plus 1 */
+	rtos->thread_details[thread_num].threadid = tid;
+	/* create the thread name */
+	rtos->thread_details[thread_num].exists = true;
+	rtos->thread_details[thread_num].thread_name_str = strdup(target_name(curr));
```
GregSavin, line 78: Was going to ask about whether the strdup() calls are eventually balanced with free() calls, but now I think I see where that already-existent de-allocation code lives.

GregSavin (approved):

PR #350, opened 2019-02-06
Rtos riscv

timsifive:
My `-rtos hwthread` changes had made `-rtos riscv` really slow, because now it was frequently reading all registers. This change fixes that.

Unfortunately I ended up making more target-independent changes, but I think this is the most straightforward solution, making explicit a distinction that should have existed already.

### DIFF
--- a/src/rtos/rtos.c
+++ b/src/rtos/rtos.c
@@ -472,8 +472,8 @@ int rtos_get_gdb_reg(struct connection *connection, int reg_num)
 		struct rtos_reg *reg_list;
 		int num_regs;
 
-		LOG_DEBUG("RTOS: getting register %d for thread 0x%" PRIx64
-				  ", target->rtos->current_thread=0x%" PRIx64 "\r\n",
+		LOG_DEBUG("getting register %d for thread 0x%" PRIx64
+				  ", target->rtos->current_thread=0x%" PRIx64,
 										reg_num,
 										current_threadid,
 										target->rtos->current_thread);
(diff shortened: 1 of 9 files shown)


ilg-ul:
> making explicit a distinction that should have existed already

can you elaborate?

src/rtos/rtos.c
```diff
@@ -468,10 +468,7 @@ int rtos_get_gdb_reg(struct connection *connection, int reg_num)
 {
 	struct target *target = get_target_from_connection(connection);
 	int64_t current_threadid = target->rtos->current_threadid;
-	if ((target->rtos != NULL) && (current_threadid != -1) &&
```
GregSavin: Just curious why the current_threadid checks are no longer needed in the "if" statement

GregSavin (dismissed):

timsifive:
> > making explicit a distinction that should have existed already

> can you elaborate?

Differentiating between calls to target_get_gdb_reg_list() that require the register values to be read, and calls to the same function that do not need the values to be read (in which case reading them is a big waste of time).

ilg-ul:
you mean differentiating between getting the list of register names and getting the register values?

this makes sense, probably upstream would also benefit.

src/rtos/rtos.c
timsifive: That was something I did that turned out not to be necessary. I've reverted this change.

GregSavin (approved):

PR #351, opened 2019-02-11
From upstream

timsifive:
Just merged the latest OpenOCD master, and fixed the build. Then I fixed a small bug that looks ancient but apparently didn't manifest previously.


GregSavin (approved):
Sounds like it was a clean merge.

PR #353, opened 2019-02-26
Set up halt groups for SMP groups.

timsifive:
If the hardware supports it, when one hart in an SMP group halts all the
other harts in that same SMP group will automatically, quickly, halt as
well.

Change-Id: Ida81f1309c180674e8c9d8060e3d2a4bbb910a6f

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -314,7 +314,8 @@ static void decode_dmi(char *text, unsigned address, unsigned data)
 		{ DMI_DMSTATUS, DMI_DMSTATUS_ANYHALTED, "anyhalted" },
 		{ DMI_DMSTATUS, DMI_DMSTATUS_AUTHENTICATED, "authenticated" },
 		{ DMI_DMSTATUS, DMI_DMSTATUS_AUTHBUSY, "authbusy" },
-		{ DMI_DMSTATUS, DMI_DMSTATUS_DEVTREEVALID, "devtreevalid" },
+		{ DMI_DMSTATUS, DMI_DMSTATUS_HASRESETHALTREQ, "hasresethaltreq" },
+		{ DMI_DMSTATUS, DMI_DMSTATUS_CONFSTRPTRVALID, "confstrptrvalid" },
 		{ DMI_DMSTATUS, DMI_DMSTATUS_VERSION, "version" },
 
 		{ DMI_ABSTRACTCS, DMI_ABSTRACTCS_PROGBUFSIZE, "progbufsize" },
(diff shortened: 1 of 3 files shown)


GregSavin (commented):
Where can I read a debug specification revision that covers halt groups?  If possible, I'd like to use that as a reference when reviewing these changes.

timsifive:
It's in the latest draft, which you can find at https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-draft.pdf

src/target/riscv/riscv-013.c
```diff
@@ -314,7 +314,8 @@ static void decode_dmi(char *text, unsigned address, unsigned data)
 		{ DMI_DMSTATUS, DMI_DMSTATUS_ANYHALTED, "anyhalted" },
 		{ DMI_DMSTATUS, DMI_DMSTATUS_AUTHENTICATED, "authenticated" },
 		{ DMI_DMSTATUS, DMI_DMSTATUS_AUTHBUSY, "authbusy" },
-		{ DMI_DMSTATUS, DMI_DMSTATUS_DEVTREEVALID, "devtreevalid" },
```
GregSavin, line 317: Did devtreevalid disappear?  If so, is that change directly related to the halt groups?

timsifive: devtreevalid was renamed to confstrptrvalid in the spec. This change is not directly related to halt groups.

This file is auto-generated as part of building the spec, so I just wholesale replaced it with the latest one.

timsifive: Not "this file" but debug_defines.h.

src/target/riscv/riscv-013.c
```diff
@@ -1362,6 +1363,18 @@ static void deinit_target(struct target *target)
 	info->version_specific = NULL;
 }
 
+static int set_haltgroup(struct target *target, bool *supported)
+{
+	uint32_t write = set_field(DMI_DMCS2_HGWRITE, DMI_DMCS2_HALTGROUP, target->smp);
+	if (dmi_write(target, DMI_DMCS2, write) != ERROR_OK)
```
GregSavin, line 1369: I don't see a read/modify/write, just a write, but I assume this does not clobber any non-haltgroup fields of DMCS2 (or if it does potentially clobber then there's no bad effect from that)?

timsifive: So far writing 0 to the other fields of dmcs2 is what we want. If more fields are added in the future, we might have to update this code to use those features.

GregSavin (approved):

src/target/riscv/riscv-013.c
GregSavin: OK, yeah, looks good to me.

PR #354, opened 2019-03-01
Fixed erroneous space that made CentOS 6 build fail

cgsfv:


timsifive (approved):
Thanks for tracking this down!

PR #361, opened 2019-03-22
riscv: Do the sign-extending from bit 39 of DPC CSR

lbmeng:
The DPC CSR correctly retains only the low 40 bits of the PC when
entering debug mode. When the CSR is read, the result should be
sign-extended to 64 bits but is not. This is a published erratum
of the ROCK processor. Let's workaround this issue in the debugger.

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1818,6 +1818,13 @@ static riscv_error_t handle_halt_routine(struct target *target)
 	info->dpc = reg_cache_get(target, CSR_DPC);
 	info->dcsr = reg_cache_get(target, CSR_DCSR);
 
+	/*
+	 * ROCK erratum: The DPC CSR correctly retains only the low 40 bits of
+	 * the PC. Do the sign-extending from bit 39 of DPC.
+	 */
+	if (info->dpc & (1UL << 39))
+		info->dpc |= 0xffffff0000000000UL;
+
 	cache_invalidate(target);
 
 	return RE_OK;
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1321,6 +1321,15 @@ static int register_read_direct(struct target *target, uint64_t *value, uint32_t
 			return ERROR_FAIL;
 	}
 
+	/*
+	 * ROCK erratum: The DPC CSR correctly retains only the low 40 bits of
+	 * the PC. Do the sign-extending from bit 39 of DPC.
+	 */
+	if (number == GDB_REGNO_PC || number == GDB_REGNO_DPC) {
+		if (*value & (1UL << 39))
+			*value |= 0xffffff0000000000UL;
+	}
+
 	if (result == ERROR_OK) {
 		LOG_DEBUG("{%d} reg[0x%x] = 0x%" PRIx64, riscv_current_hartid(target),
 				number, *value);


aswaterman (commented):
Unfortunately this will break systems with address spaces larger than 39 bits. Is there a way to detect the bug and conditionally enable it? I’m thinking: check to see that Sv39 is supported, then check to see whether dpc is incapable of having bit 40 set. If both conditions are true, then enable this hack.

lbmeng:
> Unfortunately this will break systems with address spaces larger than 39 bits. Is there a way to detect the bug and conditionally enable it? I’m thinking: check to see that Sv39 is supported, then check to see whether dpc is incapable of having bit 40 set. If both conditions are true, then enable this hack.

Thanks for the review. Do you have an idea of how to determine SV39 is supported or not?

For checking whether dpc is incapable of having bit 40 set, I think we can test this by writing dpc with bit 40 set, then reading back to check the bit 40 is zero.

aswaterman:
To check for Sv39, see if satp bit 63 is writable.

timsifive:
Is there a way to check for this specific hardware, that we know has the bug? E.g. in JTAG IDCODE, or mvendorid/marchid/mimpid?

mwachs5:
> Is there a way to check for this specific hardware, that we know has the bug? E.g. in JTAG IDCODE, or mvendorid/marchid/mimpid?

@aswaterman's algorithmic approach is better.

timsifive:
The benefit of keying off some version/build field is that it scales better. There will be more silicon bugs that need workarounds. Having to test for a dozen different bugs when first connecting will start to be cumbersome, especially on very slow targets like HDL simulations. (And there's always the chance that your test thinks it's one bug, but in reality it's a different bug which will now be harder to figure out because OpenOCD is doing something unexpected.) Unless you think this mistake is so easy to make that we expect other cores to also have the same problem.

mwachs5:
Well, the issue existed in a shared library of code that could have been used by any number of implementations. 

Anyway the answer to your actual question is "no, there isn't a specific IDCODE/CSR combination to answer this".

lbmeng:
> To check for Sv39, see if satp bit 63 is writable.

For RV32, how to write bit 63 of satp, given the GPR is 32-bit?

For RV64, from spec 1.10 chapter 4.5 below,
> Implementations that support Sv48 should also support Sv39.

So SV39 is always supported. Besides, writing bit 63 to test this has a side effect of turning on MMU, so we should prepare PTEs in the debugger I believe, no?

I am not sure how to do this.

aswaterman:
On Thu, Mar 28, 2019 at 4:31 AM Bin Meng <notifications@github.com> wrote:

> To check for Sv39, see if satp bit 63 is writable.
>
> For RV32, how to write bit 63 of satp, given the GPR is 32-bit?
>

This issue does not apply to RV32, because the virtual-address size equals
XLEN.  For RV32, you should skip this hack.

For RV64, from spec 1.10 chapter 4.5 below,
>
> Implementations that support Sv48 should also support Sv39.
>
> So SV39 is always supported. Besides, writing bit 63 to test this has a
> side effect of turning on MMU, so we should prepare PTEs in the debugger I
> believe, no?
>
The system might have no MMU, in which case satp is either unimplemented
(i.e. trap), or have satp but hardwire bit 63 to 0.

In D-mode, address translation is always disabled, regardless of satp
setting, so you do not need to set up page tables.

I am not sure how to do this.

lbmeng:
thanks, I will prepare a proper fix.

timsifive:
Inactive

PR #364, opened 2019-03-28
Support simultaneous resume using hasel

timsifive:
This involved a lot of refactoring, breaking up resume into prep/go/finish. I took the opportunity to pull some common code out of riscv-0{11,13} into riscv.c, and to get rid of a some of the unused layers of abstraction.

This code works great with spike, but doesn't seem effective with the Arty board. I don't know where that problem lies, but I suspect if there is an OpenOCD bug it'll be a relatively small change to address it, so it's worth reviewing now.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -41,11 +41,11 @@ static int riscv013_get_register(struct target *target,
 static int riscv013_set_register(struct target *target, int hartid, int regid, uint64_t value);
 static int riscv013_select_current_hart(struct target *target);
 static int riscv013_halt_current_hart(struct target *target);
-static int riscv013_resume_current_hart(struct target *target);
+static int riscv013_resume_go(struct target *target);
 static int riscv013_step_current_hart(struct target *target);
 static int riscv013_on_halt(struct target *target);
 static int riscv013_on_step(struct target *target);
-static int riscv013_on_resume(struct target *target);
+static int riscv013_resume_prep(struct target *target);
 static bool riscv013_is_halted(struct target *target);
 static enum riscv_halt_reason riscv013_halt_reason(struct target *target);
 static int riscv013_write_debug_buffer(struct target *target, unsigned index,
@@ -147,6 +147,9 @@ typedef enum {
 typedef struct {
 	struct list_head list;
 	int abs_chain_position;
+
+	/* The number of harts connected to this DM. */
+	int hart_count;
 	/* Indicates we already reset this DM, so don't need to do it again. */
 	bool was_reset;
 	/* Targets that are connected to this DM. */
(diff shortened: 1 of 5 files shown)


timsifive:
The Arty board issue appears to be something to do with multiple harts and memory access on an Arty board. Changing the test case to only use registers, it's clear that simultaneous resume is working just fine.

src/target/riscv/riscv.c
```diff
@@ -846,48 +848,160 @@ static int riscv_deassert_reset(struct target *target)
...
+/**
+ * Resume all the harts that have been prepped, as close to instantaneous as
+ * possible.
+ */
+static int resume_go(struct target *target, int current,
+		target_addr_t address, int handle_breakpoints, int debug_execution)
+{
+	riscv_info_t *r = riscv_info(target);
+	int result;
+	if (r->is_halted == NULL) {
```
GregSavin, line 941: the "if (r->is_halted == NULL)", when would that expression be true?  If the target implements an older version of the RISC-V debug spec?

src/target/riscv/riscv-013.c
```diff
@@ -2945,17 +2970,74 @@ static int riscv013_halt_current_hart(struct target *target)
...
+	}
+
+	/* Don't use hasel if we only need to talk to one hart. */
+	if (total_selected <= 1) {
+		*use_hasel = false;
+		return ERROR_OK;
+	}
+
+	for (unsigned i = 0; i < hawindow_count; i++) {
+		if (dmi_write(target, DMI_HAWINDOWSEL, 0) != ERROR_OK)
```
GregSavin: The code seems to be writing hawindowsel to zero on each iteration.  Is that correct, or is it supposed to be writing it to the value of the variable "i" on each iteration?

src/target/riscv/riscv.c
timsifive: That's right. It's a pretty confusing way to check for that, but it's what's used throughout this file. Some day I'll clean this up, hopefully.

src/target/riscv/riscv-013.c
timsifive: Yes! Good catch.

GregSavin (approved):
Does the simulataneous resume code path only get driven if a core is designated as part of an SMP group (with more than one member)?

timsifive:
Yes. Probably also gets driven if you have an SMP group with one member, but in that case it's just a waste of time and shouldn't affect anything else.

That means it doesn't get used with `-rtos riscv` because that doesn't use SMP groups. It does get used with `-rtos hwthread` which does rely on SMP groups, and of course it also gets used if people simply configure SMP groups (which we have never done AFAIK). Since I want people to move from `-rtos riscv` to `-rtos hwthread` in any case, this fits nicely.

PR #366, opened 2019-04-01
RISC-V: Make compliance tests more verbose

imphil:
Currently the RISC-V compliance test suite doesn't output the test is
currently runs before it succeeds. It also uses the same message for
many tests. This makes it very hard to find out which test fails.

This commit makes things slightly easier by printing the test that's
being executed before it actually runs, and by adding the source code
line where the test is located, making it easier to look up the test in
the source code.

New output looks like this:

Info : Executing test 149 (riscv-013.c:3800): Regular calls must return ERROR_OK
Info :   PASSED

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3528,14 +3528,22 @@ void riscv013_clear_abstract_error(struct target *target)
 	dmi_write(target, DMI_ABSTRACTCS, abstractcs & DMI_ABSTRACTCS_CMDERR);
 }
 
+#ifdef _WIN32
+#define FILE_SEP '\\'
+#else
+#define FILE_SEP '/'
+#endif
 #define COMPLIANCE_TEST(b, message) \
-{                                   \
+{ \
+	const char *last_sep = strrchr(__FILE__, FILE_SEP); \
+	const char *fname = (last_sep == NULL ? __FILE__ : last_sep + 1); \
+	LOG_INFO("Executing test %d (%s:%d): %s", total_tests, fname, __LINE__, message); \
 	int pass = 0;		    \
 	if (b) {		    \
 		pass = 1;	    \
 		passed_tests++;     \
 	}			    \
-	LOG_INFO("%s test %d (%s)\n", (pass) ? "PASSED" : "FAILED",  total_tests, message); \
+	LOG_INFO("  %s", (pass) ? "PASSED" : "FAILED"); \
 	assert(pass);		    \
 	total_tests++;		    \
 }


src/target/riscv/riscv-013.c
```diff
@@ -3530,12 +3530,13 @@ void riscv013_clear_abstract_error(struct target *target)
 
 #define COMPLIANCE_TEST(b, message) \
 {                                   \
+	LOG_INFO("Executing test %d (riscv-013.c:%d): %s", total_tests, __LINE__, message); \
```
gurjeet: Why should we not replace `riscv-013.c` with `%s` format that uses `__FILE__` as a parameter?

```suggestion
	LOG_INFO("Executing test %d (%s:%d): %s", total_tests, __FILE__, __LINE__, message); \
```

gurjeet (changes requested):

src/target/riscv/riscv-013.c
imphil: __FILE__ gives a (more or less) full path, which is rather long and yet does not add useful information. I therefore chose to abbreviate it to only the file name.

gurjeet: Fair point. How about using the `strrchr()`, like so:

```suggestion
#ifdef _WIN32
#define FILE_SEP '\\'
#else
#define FILE_SEP '/'
#endif
#define COMPLIANCE_TEST(b, message) \
{\
	char *last_sep = strrchr(__FILE__, FILE_SEP);\
	char *fname = (last_sep == NULL ? __FILE__ : last_sep + 1);\
	LOG_INFO("Executing test %d (%s:%d): %s", total_tests, fname, __LINE__, message); \
```

imphil: I don't feel that this (rather ugly) complexity is necessary for this use case; ultimately, I'd love to see the test have a meaningful name so that we don't need to rely on the line number.

gurjeet: > (rather ugly) complexity

Beauty is in the eyes of the beholder :)

If the need to print the filename and line number is obviated, I'd agree with you. But if filename and line number are to be used, I'd rather rely on compiler to do the right thing automatically; that'd take care of the cases if this file is ever renamed, the macro is moved to a header and included by other `.c` files, etc.

imphil: updated

imphil:
Ping: what's the status on this one?

imphil:
@timsifive can you review this PR or whom should I ask for it?

timsifive (approved):
Fine by me.

PR #368, opened 2019-04-03
Merge upstream master into our branch.

timsifive:
Tested against spike, 2xE31 on Arty, HiFive1, and HiFive1-flash


GregSavin (approved):
Sounds like it was a clean merge.

PR #369, opened 2019-04-09
Propagate some errors in execute_abstract_command().

timsifive:
Change-Id: Ia3ec457dd5d65378a6c922802713e36d6828bcea


GregSavin (approved):
So these were theoretically possible error conditions that would have gotten silenced and lost (or not handled ideally) before.  These changes look good to me.

PR #370, opened 2019-04-11
Support for driving RISCV DM via Arty's own JTAG chain using BSCAN tunnel

GregSavin:
Arty bitstreams can be built to include a hardware tunnel between the Arty's own boundary scan chain and a RISCV DM TAP.  The code changes in this pull request are intended to enable using that topology from OpenOCD.

The Arty's own TAP is the one to be used in the newtap initialization command, and in the command to configure the RISCV target, the RISCV DM's IR width should be declared with the argument named "bscan-tunnel-ir-width".  For example:

set _CHIPNAME riscv
jtag newtap $_CHIPNAME cpu -irlen 6
target create $_TARGETNAME.0 riscv -chain-position $_TARGETNAME
$_TARGETNAME.0 configure -work-area-phys 0x80000000 -work-area-size 10000 -work-area-backup 1 -bscan-tunnel-ir-width 5

If no "bscan-tunnel-ir-width" argument is given, or if the value is 0, then OpenOCD will assume normal direct non-tunneled access to the DM.

The protocol to get through the tunnel is to, on the boundary scan chain, select the USER4 instruction into IR, and use a DR scan to tunnel the desired DM scan through to the DM itself.  A tunneled scan is formatted as:  1 bit specifying IR or DR (IR=0, DR=1), 7 bits specifying the tunneled scan width, scan width + 1 bits of tunneled scan input (use 0 for the high bit), 3 bits of 0 as epilogue because the tunnel requires these.  Then the numbits+1 of scan output from TDO needs to be right shifted by one bit because there is a 1-clock skew between the two "domains" (the boundary scan chain and the DM TAP).  That is the intent of the code in this branch.


configure.ac
```diff
@@ -111,6 +111,7 @@ m4_define([ADAPTER_OPT], [m4_translit(ADAPTER_ARG($1), [_], [-])])
 m4_define([USB1_ADAPTERS],
 	[[[ftdi], [MPSSE mode of FTDI based devices], [FTDI]],
 	[[ftdi_oscan1], [cJTAG OSCAN1 tunneled thru MPSSE], [FTDI_OSCAN1]],
+	[[riscv_arty_bscan], [Access to RISCV on Arty board via BSCAN], [RISCV_ARTY_BSCAN]],
```
timsifive: Why does this need to be a build option?

RISCV should probably include the hyphen in the English text.

src/target/riscv/riscv-011.c
```diff
@@ -281,6 +281,11 @@ static uint32_t dtmcontrol_scan(struct target *target, uint32_t out)
 	uint8_t in_value[4];
 	uint8_t out_value[4];
 
+#if BUILD_RISCV_ARTY_BSCAN == 1
```
timsifive: I assume you don't have a target to test 0.11 with on an Arty. That's fine, but probably worth a comment here if somebody ever runs into that.

src/target/target.h
```diff
@@ -210,6 +210,9 @@ struct target {
 
 	/* The semihosting information, extracted from the target. */
 	struct semihosting *semihosting;
+#if BUILD_RISCV_ARTY_BSCAN == 1
+	int bscan_tunnel_ir_width;  /* if zero, then tunneling is not present/active */
```
timsifive: Since this is only implemented for RISC-V, this option should be in riscv_info_t, or possibly dm013_info_t if we don't expect to ever need this for 0.11.

timsifive (changes requested):
Overall looks fine, but I have some structural comments below.

configure.ac
GregSavin: I wasn't sure about whether to make it a build option, and came down on the side of being conservative, but if I move that target flag to dm013_info_t and it's isolated to riscv-013, maybe it makes sense to always compile that logic.  I'll try going that way unless I find a reason not to.

src/target/riscv/riscv-011.c
GregSavin: Or should I just remove the code from riscv-011.c if it's not likely to be relevant?

src/target/target.h
GregSavin: I'll try the 013 only option, thanks.

GregSavin: Tim, the presence of this feature isn't something that can be dynamically queried, at least currently.  It's something that is declared in the startup script, as an attribute of the "target configure" command.  I guess that was the reason that putting it in struct target seemed natural, because the parsing of "target configure" happens in target.c.   If you think it is better to move the data to dm013_info_t, would declaring the data in "target configure" command still be the right thing, or should I declare it in some other way within the config file?  Some sort of risc-v specific initialization command?

GregSavin: Or maybe if the IR width is declared as 6 within the "jtag newtap", is that enough to assume that a tunnel should be used?  The Arty TAP is 6 bits wide, and is declared that way in the newtap.  If a 6-bit wide rather than 5-bit wide IR is enough for riscv*.c logic to assume a BSCAN tunnel, maybe then that would be a possibility that removes the need for any code/data members for this in target.c, and removes the need for a parameter in "target configure"

timsifive: Since this feature is RISC-V-specific, it makes a lot of sense to configure it with a riscv command. See e.g. https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv.c#L1830.

I think all those commands just do global configuration, but I think that will be OK. It would break if somebody ever has a JTAG chain consisting of an Arty board and a regular RISC-V TAP, but that sounds quite unlikely.

Relying on the IR width is more fragile.

GregSavin: On the most recent commit, I reworked it to take the global configuration approach.  I did try adding the "use bscan" flag to dm013_info_t, rather than global state in riscv.c, but a problem with that is that the dm013_info_t struct hasn't been allocated yet at the first point at which we need to know whether or not to use BSCAN tunnel.  The code compiled, but at runtime there was a NULL pointer dereference segfault; an active struct target instance already exists at that point of OpenOCD initialization but the target->arch_info hasn't yet been assigned a non-NULL value.  There's probably a different way around that, but right now I have the "use bscan" as global within riscv.c.

src/target/riscv/riscv-013.c
```diff
@@ -486,8 +493,47 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
...
+	/* I wanted to place this code in a different function, but the way JTAG command
+	   queueing works in the jtag handling functions, the scan fields either have to be
+	   heap allocated, global/static, or else they need to stay on the stack until
+	   the jtag_execute_queue() call.  Heap or static fields in this case doesn't seem
+	   the best fit.  Declaring stack based field values in a subsidiary function call wouldn't
+	   work. */
+	if (bscan_tunnel_ir_width != 0) {
+		jtag_add_ir_scan(target->tap, &select_user4, TAP_IDLE);
+
+		uint8_t tunneled_dr_width[4] = {num_bits};
```
timsifive: This doesn't need to be 4 bytes large, since you only scan 7 bits below.

src/target/riscv/riscv.c
```diff
@@ -201,12 +239,103 @@ range_t *expose_custom;
...
+void select_dmi_via_bscan(struct target *target)
+{
+	jtag_add_ir_scan(target->tap, &select_user4, TAP_IDLE);
+	jtag_add_dr_scan(target->tap, bscan_tunneled_select_dmi_num_fields, bscan_tunneled_select_dmi, TAP_IDLE);
+}
+
+
+uint32_t dtmcontrol_scan_via_bscan(struct target *target, uint32_t out)
+{
+	/* jtag_add_ir_scan(target->tap, &select_dtmcontrol, TAP_IDLE);	*/
```
timsifive: Can you remove this commented out code?

src/target/riscv/riscv.c
```diff
@@ -1933,6 +2085,13 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.usage = "riscv set_ir_idcode [idcode|dtmcs|dmi] value",
 		.help = "Set IR value for specified JTAG register."
 	},
+	{
+		.name = "use_bscan_tunnel",
```
timsifive, line 2087: Can you also document this command in doc/openocd.texi?

timsifive (changes requested):
Looks good. My only main comment is the documentation one.

src/target/riscv/riscv.c
GregSavin: removed

src/target/riscv/riscv.c
GregSavin: Yes, I've added, and viewed the resulting PDF as cross-check.

timsifive (dismissed):

src/target/riscv/riscv-013.c
```diff
@@ -502,32 +504,29 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
...
-				.out_value = bscan_zero,
-				.in_value = NULL,
-			}
-		};
+		/* I wanted to use struct initialization syntax, but that would involve either
+		   declaring the variable within this scope (which would go out of scope at runtime
+		   before the JTAG queue gets executed, which is an error waiting to happen), or
+		   initializing outside of the check for whether a BSCAN tunnel was active (which
+		   would be a waste of CPU time when BSCAN tunnel is not being used. So I declared the
+		   struct at the function's top-level, so it's lifetime exceeds the point at which
```
timsifive: its lifetime

timsifive (dismissed):

timsifive (approved):

towoe:
Hi,
thanks for the great work. I am currently trying to integrate a BSCANE2 into a design and wondering where this protocol is coming from? By protocol I mean the 1+7+(x+1)+3 bits formation.
Thanks a lot!
Tobias

GregSavin:
Hi Tobias, the protocol was designed (and one implementation made so far) by some SiFive engineers.  There are some more details here: https://github.com/riscv/riscv-openocd/pull/373#issuecomment-491510875

Regards,
Greg S.

towoe:
Hi Greg, thanks for the reference!

PR #372, opened 2019-05-07
Simultaneous halt

timsifive:
Use hart array selection to halt all harts near simultaneously when gdb requests a halt. Only guaranteed to work with `-rtos hwthread`.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2963,6 +2963,7 @@ static int select_prepped_harts(struct target *target, bool *use_hasel)
 		riscv013_info_t *info = get_info(t);
 		unsigned index = info->index;
 		LOG_DEBUG("index=%d, coreid=%d, prepped=%d", index, t->coreid, r->prepped);
+		r->selected = r->prepped;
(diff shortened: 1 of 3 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1650,11 +1651,13 @@ static int init_target(struct command_context *cmd_ctx,
 	generic_info->set_register = &riscv013_set_register;
 	generic_info->select_current_hart = &riscv013_select_current_hart;
 	generic_info->is_halted = &riscv013_is_halted;
-	generic_info->halt_current_hart = &riscv013_halt_current_hart;
+	generic_info->halt_go = &riscv013_halt_go;
```
GregSavin: Line 1654 appears to be duplicated on line 1660, I'm assuming only one of those lines is necessary?

GregSavin (dismissed):

src/target/riscv/riscv-013.c
timsifive: Good catch. Fixed.

GregSavin (approved):

PR #375, opened 2019-05-14
Don't write sbcs while sbbusy is set.

timsifive:
Probably not hurting anything, but the spec says we shouldn't.
Also propagate more errors, and fully decode sbcs in debug output.


GregSavin (approved):
Should all occurrences of writes to DMI_SBCS be preceded by a call to read_sbcs_nonbusy?  I think I see a couple other occurrences in riscv-013.c (in the vicinity of line 2073 and line 2099).

timsifive:
> Should all occurrences of writes to DMI_SBCS be preceded by a call to read_sbcs_nonbusy? I think I see a couple other occurrences in riscv-013.c (in the vicinity of line 2073 and line 2099).

Yes, unless we're being optimistic and properly handle the failure later to improve performance. But the lines you point out are in the *_memory_bus_v0 functions, which implement SBA support that was in the spec but was never ratified. It was contributed by somebody else, and I have no way of testing it. I think there are very few implementations, too, so I try not to touch that code.

I looked through all writes to DMI_SBCS, and I'm comfortable with the way the code is.

PR #380, opened 2019-06-13
Set mstatus.FS to access FPU CSRs.

timsifive:
This improves behavior when executing function calls from mainline gdb.

Change-Id: Ia37507a16cd76fc03d26457e84fd68402969c534

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1311,6 +1311,14 @@ static int register_read(struct target *target, uint64_t *value, uint32_t number
 	return ERROR_OK;
 }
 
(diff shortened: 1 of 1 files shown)


GregSavin (approved):
Prior to this change, would an attempted access to a FP related CSR sometimes or always generate an exception?

timsifive:
Prior to this change, accessing an FP-related CSR would generate an exception if mstatus.FS was 0, and program buffer was used for the access, or an abstract command was used and the specific implementation would return failure in that case (the spec is ambiguous).

In mainline gdb this would cause the user to get an error when trying a function call.
https://github.com/riscv/riscv-binutils-gdb would quietly ignore the register access error.

PR #381, opened 2019-06-18
Improve block read and checksum speed

timsifive:
Block read goes from ~20KB/s to ~40KB/s.
Checksum speed goes from ~40KB/s (assuming the read improvement) to ~300KB/s.

There's also an improvement to cache program buffer writes, which speeds up almost everything by a few percent.


src/target/riscv/riscv-013.c
```diff
@@ -3214,7 +3225,15 @@ static enum riscv_halt_reason riscv013_halt_reason(struct target *target)
...
 int riscv013_write_debug_buffer(struct target *target, unsigned index, riscv_insn_t data)
 {
-	return dmi_write(target, DMI_PROGBUF0 + index, data);
+	dm013_info_t *dm = get_dm(target);
+	if (dm->progbuf_cache[index] != data) {
+		if (dmi_write(target, DMI_PROGBUF0 + index, data) != ERROR_OK)
+			return ERROR_FAIL;
+		dm->progbuf_cache[index] = data;
+	} else {
+		LOG_DEBUG("cache hit for 0x%x @%d", data, index);
```
GregSavin, line 3234: So during a load sequence, there is enough commonality between successive progbuf contents that caching is a noticeable performance difference?  Makes sense.

src/target/riscv/riscv.c
```diff
@@ -1525,8 +1532,86 @@ static int riscv_checksum_memory(struct target *target,
 		target_addr_t address, uint32_t count,
 		uint32_t *checksum)
 {
-	*checksum = 0xFFFFFFFF;
-	return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
```
GregSavin, line 1529: So in the old version of this function that returns resource not available, OpenOCD would perform a slower checksum operation without using a target program?

GregSavin (approved):

src/target/riscv/riscv-013.c
rtwfroody: I wouldn't call it noticeable, but it is measurable.

Flashing already uses an on-chip algorithm to speed things up, so there's a bunch of overhead in reading/writing registers and halting/resuming to make that go for every 4K(?) block. The second word of the program buffer is very often set to ebreak, and I think that's where most of the cache hits come in. I didn't really look, though.

src/target/riscv/riscv.c
rtwfroody: That's right. If this function returns any kind of error, then OpenOCD falls back to reading the memory from the target, and computing the CRC itself.

PR #383, opened 2019-06-20
Reduce abstract command execution by one scan.

timsifive:
Speeds up Arty flashing another 7%.

```
wrote 2228224 bytes from file /media/sf_tnewsome/SiFive/arty_images/arty.E21TraceFPGAEvaluationConfig.mcs in 96.997032s (22.434 KiB/s)
verified 2192012 bytes in 6.642059s (322.285 KiB/s)
11.86user 12.75system 1:44.13elapsed 23%CPU (0avgtext+0avgdata 18684maxresident)k
```

Change-Id: If609ce3de1726332f420d131e9fa6e04a5d974a1

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -588,7 +588,7 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
  * caller whether DMI was ever busy during this call. */
 static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 		bool *dmi_busy_encountered, int dmi_op, uint32_t address,
-		uint32_t data_out, int timeout_sec, bool exec)
+		uint32_t data_out, int timeout_sec, bool exec, bool ensure_success)
 {
 	select_dmi(target);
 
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -815,7 +818,7 @@ static int execute_abstract_command(struct target *target, uint32_t command)
 		}
 	}
 
-	if (dmi_write_exec(target, DMI_COMMAND, command) != ERROR_OK)
+	if (dmi_write_exec(target, DMI_COMMAND, command, false) != ERROR_OK)
```
GregSavin, line 821: So this particular call to dmi_write_exec() doesn't need to make a subsequent scan to ensure success, because the call to wait_for_idle() is about to do another DMI scan which will detect if an error happened?

GregSavin (approved):

src/target/riscv/riscv-013.c
timsifive: That's right. There are probably other instances of this in the code, but this is one that gets accessed quite a lot.

PR #384, opened 2019-06-27
Redo fespi flash algorithm

timsifive:
Just put all the flashing logic into the algorithm, instead of using an intermediate format. This should reduce total data written while flashing by about 9%, and also makes the code much simpler.

Speeds up Arty flashing another 9%.

wrote 2228224 bytes from file media/sf_tnewsome/SiFive/arty_images/arty.E21TraceFPGAEvaluationConfig.mcs in 86.784538s (25.074 KiB/s)
verified 2192012 bytes in 6.693336s (319.816 KiB/s)
8.66user 13.03system 1:33.91elapsed 23%CPU (0avgtext+0avgdata 12272maxresident)k


contrib/loaders/flash/fespi/riscv_fespi.c
```diff
@@ -0,0 +1,255 @@
...
+
+static void fespi_enable_hw_mode(volatile uint32_t *ctrl_base)
+{
+	uint32_t fctrl = fespi_read_reg(ctrl_base, FESPI_REG_FCTRL);
+	fespi_write_reg(ctrl_base, FESPI_REG_FCTRL, fctrl | FESPI_FCTRL_EN);
+}
+
+static int fespi_txwm_wait(volatile uint32_t *ctrl_base)
+{
+	unsigned timeout = 1000;
```
GregSavin, line 159: Just curious about the timeout value.  Is this derived from the specification of this type of flash device, or is it more of a pragmatic value that has been observed over time to be sufficient in practice?

GregSavin (approved):
The loader target code looks really clean.

contrib/loaders/flash/fespi/riscv_fespi.c
timsifive: I think I carried over this number from the host-side code, where it reflects 1000ms. Here I don't have access to a clock, so we're just counting loops, which I'm sure are nowhere near 1ms per iteration. Empirically this works fine on HiFive1 and Arty.

PR #386, opened 2019-07-05
Access memory through the scope of current privilege level

niwis:
Hi,

currently, all memory reads and writes using the program buffer are performed on physical addresses. This makes it almost impossible to debug code running in s- or u-mode with the MMU enabled.

I've considered many approaches and to me this seems to be the most reliable one.

Setting `mstatus.mpp = dcsr.priv` ensures that the scope is the same as in our current environment. i.e. if we are entering debug mode from m-mode, `mstatus.mprv` won't have any effect and the MMU remains deactivated. If, however, `mstatus.mprv` was already set before entering debug mode, virtualization might have been active even though we came from m-mode. Thus, don't interfere with `mstatus.mpp`.

Next, `dcsr.mprven` is set and cleared right before and after the actual access inside the program buffer. This is to ensure that the enabled MMU doesn't interfere with the functionality of debug modules that rely on memory mapped I/O with the hart.

I'm not 100% certain whether the changes are placed properly or if there should be separate functions for this, but I'm open for suggestions!

With best regards,
Nils


timsifive:
This seems like it would work for a lot of use cases. The main one where it would fall down is if there's no page table entry for a given virtual address. Fixing that gets really hairy though, so it makes sense to have something like this for now.

I have a few comments so far:
1. Please fix the style issues that caused the build to fail. OpenOCD has a strict style guide, and I'm trying to keep our fork of it clean to facilitate upstreaming.
2. You're adding an extra command to the read/write program buffer programs. At least SiFive cores keep progbuf size to a minimum (2 instructions) so this will cause all memory access to fail. Can you add logic to only add these crc*i instructions when they'll actually be doing something?
3. Reading/writing memory (especially a few words) already has a lot of overhead, and this adds to it. It's worth thinking about only changing mstatus when halting/just before resuming.

niwis:
Thanks for the feedback!

1. I've fixed the style issues
2. The csr-instructions to set and clear `mprven` are now only added to the program buffer if
   - `progbufsize` is at least `4`
   - and `mprv` is set
3. I've reduced the average number of reads/writes. If `progbufsize >= 4`, the change now adds 2 register reads per memory access. If we furthermore do not come from m-mode, it adds another 2 register writes per access. Reading/writing when halting/resuming definitely sounds reasonable. However I couldn't find a short solution to prevent interferences with accesses to `mstatus` from other places during debug mode

Is there anything to add? And is it okay to leave out targets with a smaller program buffer for now or how would you suggest to proceed?

src/target/riscv/riscv-013.c
```diff
@@ -2460,7 +2460,7 @@ static int read_memory_progbuf_inner(struct target *target, target_addr_t addres
  * Read the requested memory, silently handling memory access errors.
  */
 static int read_memory_progbuf(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, uint8_t *buffer)
+		uint32_t size, uint32_t count, uint8_t *buffer, unsigned progbufsize)
```
timsifive: There's no need to pass progbufsize in here. You can access it through target using the RISCV013_INFO macro. See e.g. the start of read_memory().

src/target/riscv/riscv-013.c
```diff
@@ -2471,6 +2471,33 @@ static int read_memory_progbuf(struct target *target, target_addr_t address,
...
+	uint64_t mstatus = 0;
+	uint64_t mstatus_old = 0;
+	if (progbufsize >= 4) {
+		/* Read DCSR */
+		uint64_t dcsr;
+		if (register_read(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
+			return ERROR_FAIL;
+
+		/* Read and save MSTATUS */
+		if (register_read(target, &mstatus, GDB_REGNO_MSTATUS) != ERROR_OK)
```
timsifive: Is it necessary to read this if the DCSR_PRV check below fails?

src/target/riscv/riscv-013.c
```diff
@@ -2471,6 +2471,33 @@ static int read_memory_progbuf(struct target *target, target_addr_t address,
...
+		/* If we come from m-mode with mprv set, we want to keep mpp */
+		if (get_field(dcsr, DCSR_PRV) < 3) {
+			/* MPP = PRIV */
+			mstatus = set_field(mstatus, MSTATUS_MPP, get_field(dcsr, DCSR_PRV));
+
+			/* MPRV = 1 */
+			mstatus = set_field(mstatus, MSTATUS_MPRV, 1);
+
+			/* Write MSTATUS */
+			if (register_write_direct(target, GDB_REGNO_MSTATUS, mstatus) != ERROR_OK)
```
timsifive: We should only do this write if it will actually change the value of mstatus.

timsifive (changes requested):
Thanks for making these changes. This is looking much better. Having pondered the change some more, I have more requests, however.

I'm not comfortable with the user-visible behavior silently changing (whether addresses are interpreted as virtual/physical) depending on progbufsize which is something users have no control over, and most won't even be aware of.

We should decide which behavior we want by default, and then warn the user once per session (using LOG_INFO) if what they requested isn't possible. I'm also a bit concerned that there might be people who are unknowingly relying on the existing behavior. Can you add a configuration option that lets people pick what they want? Search riscv.c for "prefer_sba" to get an example on how to do that.

src/target/riscv/riscv-013.c
niwis: Done

src/target/riscv/riscv-013.c
niwis: Done

src/target/riscv/riscv-013.c
niwis: I guess so. If `dcsr.prv = 3` (m-mode), `mstatus.mprv` still could have been set from outside debug mode (e.g. to enable the MMU in m-mode). Thus we need it later for the final check whether the `csrr*i` instructions should be written into the program buffer.

niwis:
I've included your suggestions, in particular a configuration option to enable this feature. By default it is disabled now to prevent any confusion, but to me enabling it looks generally more intuitive. Perhaps changing the default behavior would be something for a release where people expect changes in the ui?

Do you have further comments or suggestions?

src/target/riscv/riscv-013.c
```diff
@@ -1580,6 +1580,12 @@ static int examine(struct target *target)
 				r->impebreak);
 	}
 
+	if (info->progbufsize < 4 && riscv_enable_virtual) {
+		LOG_WARNING("set_enable_virtual is not available on this target. It "
```
timsifive: Since this is now a configurable option, this really should be a fatal error. Users might not see/ignore warnings, and that would lead to debugging failing in all kinds of exciting ways.

Can you also include in the error message the config command the user can run to disable it?

src/target/riscv/riscv-013.c
```diff
@@ -2739,6 +2786,34 @@ static int write_memory_progbuf(struct target *target, target_addr_t address,
 
 	select_dmi(target);
 
+	uint64_t mstatus = 0;
+	uint64_t mstatus_old = 0;
+	if (riscv_enable_virtual && info->progbufsize >= 4) {
```
timsifive: Can you share this code between read and write?

timsifive (changes requested):
Can you resolve the conflicts as well? (Sorry about that.)

niwis:
Done. Anything else?

timsifive (approved):

timsifive:
Thanks for making this all nice and clean.

PR #387, opened 2019-07-08
RV32E support

timsifive:
It's a bit hacked up currently because gdb really doesn't like it if GPRs 16--31 don't exist. So we fake that they do exist, read as 0, and return an error if any value besides 0 is written.


GregSavin (approved):
So it sounds like GDB doesn't (yet) have realistic expectations for RV32E targets.  At least these compensating OpenOCD changes are clean and simple.

PR #388, opened 2019-07-11
Make resume order configurable.

timsifive:
This is a customer requirement. Using this option is discouraged.

Change-Id: I520ec19cc23d7837cb8576f69dadf2b922fa2628


GregSavin (approved):
Interesting feature.  You'd hope most people would fix their software instead, but I guess that (fixing race condition) might not always be an option at any given moment.

PR #389, opened 2019-07-12
Write all ones to clear cmderr.

timsifive:
Change-Id: Ia76e749ed9f5a5f3509f253eeb69d1208bcfc929


GregSavin (approved):
It looks like the previous code was compliant with the RISC-V debug specification (as is the new code which is slightly simpler).  Curious if there were any DM implementations that somehow weren't playing nice with the old code?

timsifive:
The spec actually changed (https://github.com/riscv/riscv-debug-spec/pull/472). AFAIK this code worked great with all existing implementations, but it might not with future implementations.

PR #390, opened 2019-07-12
Optimize reading a single byte/short/word.

timsifive:
gdb has developed a nasty habit of very often reading 30-some
half-words. This change speeds that up significantly.

Change-Id: Iab1b7575bec5c57051c6e630ae292dddf8fe6350


GregSavin (approved):
Looks good to me.

PR #393, opened 2019-08-02
Don't fake step for hwthread rtos.

timsifive:
Fake step is a hack introduced to make things work with real RTOSs that
have a concept of a current thread. The hwthread rtos always has access
to all threads, so doesn't need it.

This fixes a bug when running my MulticoreRegTest against HiFive
Unleashed where OpenOCD would return the registers of the wrong thread
after gdb stepped a hart.

Change-Id: I64f538a133fb078c05a0c6b8121388b0b9d7f1b8


timsifive:
@SiFiveGregS, can you review this?

GregSavin (approved):
So, sounds like the fake step workaround only comes into play in cases where target software, rather than target hardware, manages thread context switching; makes sense.

PR #214, opened 2018-02-16
System Bus access spec0.13

fhugodot:
hi,
i've updated the riscv-013.c to use System Bus Access using the draft spec 0.13
Would it be possible to have a look and to merge this ?
Do you want me to re-do it with the latest version?

Thank you for your comments

PR #394, opened 2019-08-06
Adds support for RISCV Access Memory Abstract Commands

dave-estes-syzexion:
The Access Memory Abstract Command is one of the three optional
methods for reading and writing memory on a complient RISCV
debug module. The previous two options were already implemented
in OpenOCD.

Implementation Notes:
- aamvirtual is hard-coded to false until the design for accessing
  virtual addresses is finalized.
- aamsizes corresponding to 8b, 16b, 32b, and 64b are supported.
  128b support is postponed until it is required, as it will mean
  changes to the read/write_abstract_arg() interface to pass 128b
  values.
- aampostincrement is not used and hard-coded to false.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1954,6 +1954,30 @@ static int deassert_reset(struct target *target)
 /**
  * @par size in bytes
  */
+static void read_from_buf(uint64_t *value, const uint8_t *buffer, unsigned size)
+{
+	switch (size) {
+		case 1:
+			*value = buffer[0];
+			break;
+		case 2:
+			*value = buffer[0]
+				| ((uint64_t) buffer[1] << 8);
+			break;
+		case 4:
+			*value = buffer[0]
+				| ((uint64_t) buffer[1] << 8)
+				| ((uint64_t) buffer[2] << 16)
+				| ((uint64_t) buffer[3] << 24);
+			break;
+		default:
+			assert(false);
+	}
+}
+
+/**
+ * @par size in bytes
+ */
 static void write_to_buf(uint8_t *buffer, uint64_t value, unsigned size)
 {
 	switch (size) {
@@ -2388,7 +2412,7 @@ static int read_memory_abstract(struct target *target, target_addr_t address,
 
 		/* Copy arg0 to buffer (rounded width up to nearest 32) */
 		riscv_reg_t value = read_abstract_arg(target, 0, width32);
-		memcpy(p, &value, size);
+		write_to_buf(p, value, size);
 
 		updateaddr = false;
 		p += size;
@@ -2428,7 +2452,7 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
 	for (uint32_t c = 0; c < count; c++) {
 		/* Move data to arg0 */
 		riscv_reg_t value = 0;
-		memcpy(&value, p, size);
+		read_from_buf(&value, p, size);
 		result = write_abstract_arg(target, 0, value, riscv_xlen(target));
 		if (result != ERROR_OK) {
 			LOG_ERROR("Failed to write arg0 during write_memory_abstract().");
@@ -2877,6 +2901,7 @@ static int write_memory_bus_v0(struct target *target, target_addr_t address,
 
 	/* B.8 Writing Memory, single write check if we write in one go */
 	if (count == 1) { /* count is in bytes here */
+		/* TODO: Test with read_from_buf(&value, t_buffer, size) */
 		/* check the size */
 		switch (size) {
 			case 1:
@@ -2921,6 +2946,7 @@ static int write_memory_bus_v0(struct target *target, target_addr_t address,
 		t_addr = address + offset;
 		t_buffer = buffer + offset;
 
+		/* TODO: Test with read_from_buf(&value, t_buffer, size) */
 		switch (size) {
 			case 1:
 				value = t_buffer[0];
@@ -3103,6 +3129,7 @@ static int write_memory_progbuf(struct target *target, target_addr_t address,
 			unsigned offset = size*i;
 			const uint8_t *t_buffer = buffer + offset;
 
+			/* TODO: Test with read_from_buf(&value, t_buffer, size)*/
 			uint32_t value;
 			switch (size) {
 				case 1:
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -844,7 +844,7 @@ static riscv_reg_t read_abstract_arg(struct target *target, unsigned index,
 	unsigned offset = index * size_bits / 32;
 	switch (size_bits) {
 		default:
-			LOG_ERROR("Unsupported size: %d", size_bits);
+			LOG_ERROR("Unsupported size: %d bits", size_bits);
 			return ~0;
 		case 64:
 			dmi_read(target, &v, DMI_DATA0 + offset + 1);
@@ -863,7 +863,7 @@ static int write_abstract_arg(struct target *target, unsigned index,
 	unsigned offset = index * size_bits / 32;
 	switch (size_bits) {
 		default:
-			LOG_ERROR("Unsupported size: %d", size_bits);
+			LOG_ERROR("Unsupported size: %d bits", size_bits);
 			return ERROR_FAIL;
 		case 64:
 			dmi_write(target, DMI_DATA0 + offset + 1, value >> 32);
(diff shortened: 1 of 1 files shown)


dave-estes-syzexion:
Hey @timsifive (et. al), this is that patch that I mentioned earlier in the week. Please let me know if it needs even the smallest of changes. Also, do you have a better idea for the logic in read_memory() and write_memory() where I fallback on access memory abstract command as a fall through if the other two methods fail.

src/target/riscv/riscv-013.c
```diff
@@ -2298,6 +2340,127 @@ static int batch_run(const struct target *target, struct riscv_batch *batch)
...
+	unsigned width = size << 3;
+	if (width > 64) {
+		// TODO: Add 128b support if it's ever used. Involves modifying
+		//       read/write_abstract_arg() to work on two 64b values.
+		LOG_ERROR("Unsupported size: %d bits", size);
+		return ERROR_FAIL;
+	}
+
+	// Create the command (physical address, postincrement, read)
+	uint32_t command = access_memory_command(target, false, width, true, false);
```
dave-estes-syzexion, line 2389: Both `read_memory_abstract()` and `write_memory_abstract()` build their access memory command using `aampostincrement = true`. My reading of the 0.14 debug spec indicates that access memory abstract commands are optional, but that none of the fields are optional. This is in contrast to the access register abstract command where `aarpostincrement` is specifically stated as being optional.

Can anyone confirm my reading, or should I create some probe code which tries to determine if the support for `aampostincrement` for the access memory abstract command?

src/target/riscv/riscv-013.c
```diff
@@ -986,6 +986,48 @@ static int register_write_abstract(struct target *target, uint32_t number,
...
+ * the width (bits).
+ */
+uint32_t abstract_memory_size(unsigned width)
+{
+	switch (width) {
+		case 8:
+			return set_field(0, AC_ACCESS_MEMORY_AAMSIZE, 0);
+		case 16:
+			return set_field(0, AC_ACCESS_MEMORY_AAMSIZE, 1);
+			break;
```
timsifive: Don't need break right after return (several times in this function).

src/target/riscv/riscv-013.c
```diff
@@ -986,6 +986,48 @@ static int register_write_abstract(struct target *target, uint32_t number,
...
+		default:
+			LOG_ERROR("Unsupported memory width: %d", width);
+			return 0;
+	}
+}
+
+/*
+ * Creates a memory access abstract command.
+ */
+static uint32_t access_memory_command(struct target *target, unsigned virtual,
```
timsifive: virtual, postincrement, and write should have bool type.

src/target/riscv/riscv-013.c
```diff
@@ -986,6 +986,48 @@ static int register_write_abstract(struct target *target, uint32_t number,
 	return ERROR_OK;
 }
 
+/*
+ * Sets the AAMSIZE field of a memory access abstract command based on
+ * the width (bits).
+ */
+uint32_t abstract_memory_size(unsigned width)
```
timsifive: Function should be static.

src/target/riscv/riscv-013.c
timsifive: aampostincrement is optional, as are all other arguments. The intro to "Abstract Commands" says: "Commands may be supported with some options set, but not with other options set."

Probe code would be great, but I'm happy to accept that we only support abstract memory access when aampostincrement is supported for now. We can add support for targets that don't support it when we encounter them.

src/target/riscv/riscv-013.c
```diff
@@ -2698,6 +2861,8 @@ static int read_memory(struct target *target, target_addr_t address,
 	if (info->progbufsize >= 2)
 		return read_memory_progbuf(target, address, size, count, buffer);
 
+	return read_memory_abstract(target, address, size, count, buffer);
+
 	LOG_ERROR("Don't know how to read memory on this target.");
```
timsifive: Remove these 2 lines.

src/target/riscv/riscv-013.c
```diff
@@ -3085,6 +3250,8 @@ static int write_memory(struct target *target, target_addr_t address,
 	if (info->progbufsize >= 2)
 		return write_memory_progbuf(target, address, size, count, buffer);
 
+	return write_memory_abstract(target, address, size, count, buffer);
+
 	LOG_ERROR("Don't know how to write memory on this target.");
```
timsifive: Remove these two lines.

timsifive (changes requested):
Thanks for contributing this code. Overall this looks pretty good, but I do have some comments:

1. OpenOCD has a pretty strict style guide, which is automatically run at build time. Please take a look at the failed builds and fix what it complains about. (Search for "error" in the log.) Looks like it's just the fact that you used // comments which OpenOCD doesn't like.
2. You could speed things up a bit by using abstractauto, but then you'd end up writing more code. Your choice.

src/target/riscv/riscv-013.c
dave-estes-syzexion: Fixed.

src/target/riscv/riscv-013.c
dave-estes-syzexion: Fixed.

src/target/riscv/riscv-013.c
dave-estes-syzexion: Fixed.

src/target/riscv/riscv-013.c
dave-estes-syzexion: Thanks. I'll leave it to that future target then, but I'm certainly happy to help write the probe code when that target arrives. I just don't want to write and push untested code.

src/target/riscv/riscv-013.c
dave-estes-syzexion: Done.

src/target/riscv/riscv-013.c
dave-estes-syzexion: Done.

dave-estes-syzexion (commented):
Thanks for your review, Tim. Sorry about those failures. I should have clicked through and dug through the logs. I've addressed all of the comments and will push a new revision shortly. I did leave the probe code up to a future date when we have a platform to test it on. I also punted on using abstractauto. I'm happy to address both if you change your mind. Thanks again.

src/target/riscv/riscv-013.c
```diff
@@ -2413,41 +2410,41 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
-	// Execute the writes
-	const uint8_t* p = buffer;
+	/* Execute the writes */
+	const uint8_t *p = buffer;
 	bool updateaddr = true;
 	for (uint32_t c = 0; c < count; c++) {
-		// Move data to arg0
-		riscv_reg_t value = *(riscv_reg_t*)p;
+		/* Move data to arg0 */
+		riscv_reg_t value = *p;
```
timsifive: Does that work? p is uint8_t*, so dereferencing it will just get you a byte and not a full value, right?

dave-estes-syzexion: Oops, yeah, sorry. Was chasing a clang error about ignoring alignment, and forgot that I intended this to grab a full word of data even if fewer bits are written. Will just use a memcpy instead.

src/target/riscv/riscv-013.c
```diff
@@ -2427,7 +2427,8 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
 	bool updateaddr = true;
 	for (uint32_t c = 0; c < count; c++) {
 		/* Move data to arg0 */
-		riscv_reg_t value = *p;
+		riscv_reg_t value = 0;
+		memcpy(&value, p, size);
```
timsifive: That probably doesn't work on big-endian hosts. Can you do something like https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv-013.c#L2764? (For bonus points, share the code.)

dave-estes-syzexion: Thanks for the feedback. Much cleaner and more portable now. I did a little digging and found `write_to_buf()`, which is already being used when reading from memory. So I used that for `read_memory_abstract()` replacing that memcpy. Then I wrote `read_from_buf()` and used it in `write_memory_abstract()` for its memcpy.

I didn't, however, replace that other replicated code. I just dropped a TODO in the three spots that can use it. I can't really test it on our platform. ...though I *just* realized that I might be able to use Spike. I'm happy to give that a shot if like.

timsifive (approved):
This looks good, thanks.

If you are interested in running tests against spike that would be great. The tests are in riscv-tests/debug. You need to have a recent gdb (I use mainline of regular gdb), a compiler, and spike itself. Then running `make` from riscv-tests/debug should run some tests. `make all` will be more exhaustive.

dave-estes-syzexion:
Thanks @timsifive for pointing me to these tests. I went ahead and made the code-cleanup changes to `write_memory_progbuf()` and `write_memory_bus_v0()` to match what I already have in `write_memory_abstract()`. Unfortunately, spike seems to use `write_memory_bus_v1()` (note v1 vs. v0 above) exclusively. I tried to get the main `write_memory()` to use something other than `write_memory_bus_v1()`, but the tests always failed...even without my code-cleanup. May have to leave in the code-cleanup TODOs.

It wasn't a wash though, these debug tests are going to be great for our targets. Thanks again.

timsifive:
Memory bus v0 is obsolete, and I don't have any targets that implement it. Support was added through #214. I am keeping the code around because it may be useful to someone.

PR #399, opened 2019-09-03
target/riscv-013: Fix assert failed in read_memory_progbuf_inner in 32-bit platform

diodep:
This function use a wrong way to calculate receive_addr variable. If you compile it on 32-bit system, it will mistakenly convert 'i-2' statement into 32-bit variable, Which makes line 2238 assert failed.


src/target/riscv/riscv-013.c
```diff
@@ -2619,7 +2619,7 @@ static int read_memory_progbuf_inner(struct target *target, target_addr_t addres
 		/* Now read whatever we got out of the batch. */
 		dmi_status_t status = DMI_STATUS_SUCCESS;
 		for (size_t i = 0; i < reads; i++) {
-			riscv_addr_t receive_addr = read_addr + (i-2) * size;
+			riscv_addr_t receive_addr = read_addr + ((int64_t)i-2) * size;
```
timsifive, line 2622: I think a better fix is to change the type of `i` to riscv_addr_t. Does that work?

rgwan:
@timsifive I don't think it works. Because 'riscv_addr_t' is an alias of 'uint64_t' 

'i-2' is a signed value, which means it could be either positive and negative. If you force conversation it to uint64_t, the compiler will not do sign extend. To keep it functioning normally. I think it have to be a signed 64-bit variable.

timsifive:
I'm a bit worried about explicitly using int64_t, which still feels like it's playing games with implicit conversion. Does this work:
```
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2619,7 +2619,8 @@ static int read_memory_progbuf_inner(struct target *target, target_addr_t addres
                /* Now read whatever we got out of the batch. */
                dmi_status_t status = DMI_STATUS_SUCCESS;
                for (size_t i = 0; i < reads; i++) {
-                       riscv_addr_t receive_addr = read_addr + (i-2) * size;
+                       riscv_addr_t receive_addr = read_addr + i * size;
+                       receive_addr -= 2 * size;
                        assert(receive_addr < address + size * count);
                        if (receive_addr < address)
                                continue;
```

orzel:
debugging on longan nano is currently broken:
          Error: flash driver 'gd32vf103' not found

I guess the related commit in this pull request would fix it .. ?

fabiovila:
> debugging on longan nano is currently broken:
> Error: flash driver 'gd32vf103' not found
> 
> I guess the related commit in this pull request would fix it .. ?

> Error: flash driver 'gd32vf103' not found

Some problem here. 
Did you find any solution for that?

rgwan:
> 
> 
> > debugging on longan nano is currently broken:
> > Error: flash driver 'gd32vf103' not found
> > I guess the related commit in this pull request would fix it .. ?
> 
> > Error: flash driver 'gd32vf103' not found
> 
> Some problem here.
> Did you find any solution for that?

Refer to here:
https://github.com/riscv-mcu/riscv-openocd/

timsifive:
Closing due to inactivity.

PR #402, opened 2019-09-09
Fix flashing HiFive Unleashed

timsifive:
The main change is to fix the erase command for issi is25wp256d.

There are a bunch of smaller improvements as well:
* Make flash algorithm work for RV32E.
* Make flash algorithm propagate all errors.
* Add a few features that make it easier to debug the flash algorithm.
* Align flash algorithm stack to XLEN.
* Make flash timeout depend on how much data is written.
* Improve debug messages.
* Make algorithm running on just one hart work with `-rtos riscv`.


GregSavin (approved):
Looks good to me.  Interesting approach with the optional extra context in the algorithm error return value.

PR #405, opened 2019-09-11
Perform SBA writes with batch transactions for improved performance.

darius-bluespec:
Using batch transactions avoids an unnecessary dmi read after every
dmi write, resulting in a significant performance improvement.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2982,24 +2982,39 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
 	target_addr_t next_address = address;
 	target_addr_t end_address = address + count * size;
 
+	int result;
+
 	sb_write_address(target, next_address);
 	while (next_address < end_address) {
+		LOG_DEBUG("transferring burst starting at address 0x%" TARGET_PRIxADDR,
+				next_address);
+
+		struct riscv_batch *batch = riscv_batch_alloc(
+				target,
+				32,
+				info->dmi_busy_delay + info->bus_master_write_delay);
+
 		for (uint32_t i = (next_address - address) / size; i < count; i++) {
 			const uint8_t *p = buffer + i * size;
+
+			if (riscv_batch_available_scans(batch) < (size + 3) / 4)
+				break;
+
 			if (size > 12)
-				dmi_write(target, DMI_SBDATA3,
+				riscv_batch_add_dmi_write(batch, DMI_SBDATA3,
 						((uint32_t) p[12]) |
 						(((uint32_t) p[13]) << 8) |
 						(((uint32_t) p[14]) << 16) |
 						(((uint32_t) p[15]) << 24));
+
 			if (size > 8)
-				dmi_write(target, DMI_SBDATA2,
+				riscv_batch_add_dmi_write(batch, DMI_SBDATA2,
 						((uint32_t) p[8]) |
 						(((uint32_t) p[9]) << 8) |
 						(((uint32_t) p[10]) << 16) |
 						(((uint32_t) p[11]) << 24));
 			if (size > 4)
-				dmi_write(target, DMI_SBDATA1,
+				riscv_batch_add_dmi_write(batch, DMI_SBDATA1,
 						((uint32_t) p[4]) |
 						(((uint32_t) p[5]) << 8) |
 						(((uint32_t) p[6]) << 16) |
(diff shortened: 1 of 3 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -2982,24 +2982,35 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
 	target_addr_t next_address = address;
 	target_addr_t end_address = address + count * size;
 
+	int result;
+
 	sb_write_address(target, next_address);
 	while (next_address < end_address) {
+		LOG_DEBUG("transferring burst starting at address 0x%016" PRIx64,
```
timsifive: next_address is a target_addr_t. use: "0x%" TARGET_PRIxADDR

src/target/riscv/riscv-013.c
```diff
@@ -3011,34 +3022,39 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
 			log_memory_access(address + i * size, value, size, false);
 
-			if (info->bus_master_write_delay) {
-				jtag_add_runtest(info->bus_master_write_delay, TAP_IDLE);
-				if (jtag_execute_queue() != ERROR_OK) {
-					LOG_ERROR("Failed to scan idle sequence");
-					return ERROR_FAIL;
-				}
-			}
+			if (riscv_batch_full(batch))
```
timsifive: I think you are getting away with only this check because, depending on size, you add 1, 2, or 4 scans to the batch. And the batch size is 32, which is divisible by all of those. That's pretty fragile, though. Can you make this a bit more robust by checking at the start of the loop there is enough space for all the writes? Simplest is probably just to make sure there is space for 4 scans, which handles the worst case. The slight inefficiency in that you allocate a few more scans than necessary isn't a big deal.

src/target/riscv/riscv-013.c
```diff
@@ -3011,34 +3022,39 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+			return result;
+
 		if (read_sbcs_nonbusy(target, &sbcs) != ERROR_OK)
 			return ERROR_FAIL;
 
+		next_address = sb_read_address(target);
+
+		if (next_address < address) {
+			/* This should never happen, probably buggy hardware. */
+			LOG_DEBUG("unexpected system bus address 0x%016" PRIx64, next_address);
```
timsifive: Same formatting comment as above.

src/target/riscv/riscv-013.c
```diff
@@ -3011,34 +3022,39 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
 
+		result = batch_run(target, batch);
+		riscv_batch_free(batch);
+		if (result != ERROR_OK)
+			return result;
+
 		if (read_sbcs_nonbusy(target, &sbcs) != ERROR_OK)
 			return ERROR_FAIL;
 
+		next_address = sb_read_address(target);
```
timsifive: It seems slow to read this unnecessarily. Maybe the overall code is still faster, but this seems like a good next candidate for optimization.

timsifive (changes requested):
Looks OK overall. What testing did you do?

src/target/riscv/riscv-013.c
darius-bluespec: I updated the branch to address this.

src/target/riscv/riscv-013.c
darius-bluespec: Indeed.  I copied the idiom from program buffer writes, but did not realize that code only supports a size up to 4 bytes.  I have updated the branch to more carefully check space in the batch.

src/target/riscv/riscv-013.c
darius-bluespec: Likewise, updated in branch.

src/target/riscv/riscv-013.c
darius-bluespec: This read and the following check was necessary to catch a system that would potentially reset the debug module in response to a memory write (to a MMIO GPIO device).  Without this check, OpenOCD would hang indefinitely when the debug module was unexpectedly reset.  Hopefully with the specification clarification (riscv/riscv-debug-spec#488) I can convince others that the platform with this behavior should be changed.

In any case, if you feel it is important, I can eliminate this now from the pull request.

darius-bluespec:
Changes have been tested on two RV64 systems (one with system bus, one without), loading and running ~200 small (~10 kiB) images and one large (~8MiB) image.  All images loaded and ran successfully.

src/target/riscv/riscv-013.c
timsifive: Why exactly did it end up looping infinitely?
Can you move this check outside of this loop, so it happens once per call to this function instead of once for every 32 words? And you can probably tighten up the check too. sbaddress should contain exactly what you expect. Anything else indicates some kind of failure.

darius-bluespec: If the debug module was spontaneously reset while the memory write was on going, the next_address read back would be zero.  Since zero is less than end_address, it would never exit that loop, but would also never write any bytes (probably due to overflowing the unsigned loop index i).

The check cannot be moved outside the while loop, since it the point of it was to check if it was necessary to terminate the loop early.

I have a pushed an updated branch that I think will address these issues.  next_address is only read back when necessary (when the target is busy), and is only checked when read.  Otherwise, next_address is updated by the loop without reading from the hardware.  This is tested working as before.

timsifive:
This introduces a regression in my MemTestBlock test. You can reproduce this from the https://github.com/riscv/riscv-tests repo by following the instructions in https://github.com/riscv/riscv-tests/blob/readme/debug/README.md.

If these instructions aren't clear, please let me know. I'm trying to get this to the point where I can just send people that way and I don't have to run any tests myself.

darius-bluespec:
I updated the branch with a fix for that test.  Changes tested as before, plus the tests from riscv-tests/debug, which all pass except for one timeout and eleven not-applicable.

Previously, this patch was checking for and handling sbbusyerror, but not checking for dmi busy.  The new patch checks for and handles both.

src/target/riscv/riscv-013.c
```diff
@@ -3011,34 +3026,44 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+		if (result != ERROR_OK)
+			return result;
+
+		bool dmi_busy_encountered;
+		if (dmi_op(target, &sbcs, &dmi_busy_encountered, DMI_OP_READ,
+			   DMI_SBCS, 0, false, false) != ERROR_OK)
 			return ERROR_FAIL;
 
-		if (get_field(sbcs, DMI_SBCS_SBBUSYERROR)) {
+		while (get_field(sbcs, DMI_SBCS_SBBUSY))
```
timsifive: Can you add a timeout to this loop? Buggy hardware might make it infinite.

timsifive (changes requested):

src/target/riscv/riscv-013.c
darius-bluespec: I updated the branch with a timeout.  Tested working as before.
FWIW, I copied the busy loop idiom from `write_memory_progbuf()`.  Both `write_memory_progbuf()` and `read_memory_progbuf_inner()` have similar loops without timeouts, which should probably be added at some point.

timsifive (approved):
Looks good. Thank you for making this improvement!

In the future, instead of updating your existing commits and force-pushing, can you just add more commits? When I merge it I squash-merge so you're not cluttering up the history that way. But it's harder for me to see what's changed from time to time when you don't create new commits each time.

darius-bluespec:
Sure, I'll keep that in mind in the future.

PR #406, opened 2019-09-12
Add TCK padding in the reset/online/activate sequence of cJTAG OSCAN1.

GregSavin:


timsifive (approved):
I don't know enough about cJTAG to tell whether this is reasonable. I assume you've tested this on something and that it works.

GregSavin:
This change has now been tested on multiple targets, and is reported to fix a timing issue that was being seen on one target (and the other target still works fine too).  So I'm going to merge this now.

sifivekevin:
ASTI just confirmed that this change allows them to connect to their target.

PR #409, opened 2019-09-23
The stack is relative to the program.

timsifive:
Just like everything else. Fixes flashing on targets that don't have RAM
at 0x80000000.

Change-Id: Ibf423c95ba333660e760d5110f5ce2375c1b762c


GregSavin (approved):
Looks correct to me.  

(Seems strange that the assembler didn't expand the "la" to that sequence, but that question isn't of immediate importance)

PR #410, opened 2019-09-24
Use more idiomatic assembly for flash code.

timsifive:
Also change the address that we use to link the PIC code, so that if
there is a PIC issue in the future it will show up on hardware I test
against.

Also remove fespi.S, which hasn't been used in a long time.

Change-Id: I667d930b48107a3522d619167c7afc335431b4b6


contrib/loaders/flash/fespi/riscv.lds
```diff
@@ -2,7 +2,7 @@ OUTPUT_ARCH( "riscv" )
 
 SECTIONS
 {
-  . = 0x80000000;
+  . = 0x12340000;
```
GregSavin, line 5: I like that idea.

contrib/loaders/flash/fespi/riscv_wrapper.S
```diff
@@ -11,10 +11,7 @@
 		.section .text.entry
 		.global _start
 _start:
-		/* Spell out `la sp, stack_end` because otherwise the assembler won't
-		generate position-independent code. */
-1:	    auipc sp,     %pcrel_hi(stack_end) # load stack_end(hi)
-	    addi  sp, sp, %pcrel_lo(1b)  # load stack_end(lo)
+		lla		sp, stack_end
```
GregSavin, line 14: Interesting, wasn't aware of that pseudo-instruction.

GregSavin (approved):

contrib/loaders/flash/fespi/riscv_wrapper.S
timsifive: Me either. :-)

PR #411, opened 2019-09-28
Get latest code from upstream

timsifive:
Testing:
1. Full tests against spike.
2. HiFive Unleashed (flash and non-flash) (plenty of failures, but no regressions)
3. HiFive1 (flash and non-flash) (plenty of failures, but no regressions)
4. Flashing an Arty 35T

This change removed the ocd_ prefix TCL commands, but they're not widely used (if at all) so I think that's OK. (To get the result of a command for use in a TCL variable, you used to have to prefix that command with ocd_. Now it just works.)


GregSavin (dismissed):
Something about Git I was wondering... if a branch like this with lots of commits gets squashed and merged, does the squashing effectively cast away the easy ability to subsequently do a "git bisect" over those commits, in-place, directly on the riscv branch?   If so, is it worth doing a non-squashing merge for these upstream sync-ups.  Just in case?

timsifive:
@SiFiveGregS, you're exactly right about git. And I will be doing a non-squash merge for this sync-up, for exactly that reason.

GregSavin (approved):

PR #417, opened 2019-10-11
Combine SMP group registers into one list for gdb

timsifive:
This makes behavior when you've configured an SMP group of heterogeneous
targets a bit less weird. (You still shouldn't be doing that, since gdb
and who knows what else assumes that the targets in an SMP group are
homogeneous.)

Specifically, if you have a HiFive Unleashed board (where the first core
is fairly basic and the other 4 or more full-featured) this lets you
connect to all 5, and still have access to the FPU etc. on the higher
numbered cores.

Change-Id: I2e01f63f8753f78c29d7f414ea603e02bf0390e0

### DIFF
--- a/src/server/gdb_server.c
+++ b/src/server/gdb_server.c
@@ -2196,6 +2196,78 @@ static int get_reg_features_list(struct target *target, char const **feature_lis
 	return ERROR_OK;
 }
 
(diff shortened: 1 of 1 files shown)


GregSavin (approved):
Looks good to me, as far as a pragmatic solution to the situation described.

PR #420, opened 2019-11-05
For BSCAN tunneling, USER4 only needs to be scanned into IR once

GregSavin:
Within a debug session, no value other than USER4 ends up getting scanned into the IR of the boundary scan chain.  So as an optimization, we can load IR once, and then refrain from loading it again.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2257,6 +2257,13 @@ COMMAND_HANDLER(riscv_use_bscan_tunnel)
 
 	bscan_tunnel_type = tunnel_type;
 	bscan_tunnel_ir_width = irwidth;
+	/* Set things up to generate another one-time loading of USER4 into IR the first time
+	   a RISCV scan happens from this point,  in case somebody transitioned out of bscan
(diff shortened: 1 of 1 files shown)


timsifive (approved):
Looks OK to me. I assume you've tested as much as is possible right now.

How does this play if there are multiple JTAG TAPs? I assume then this TAP would get BYPASS scanned into IR. Does that affect whether user4 is selected?

GregSavin:
Your question made me realize the current BSCAN tunneling solution is tailored to a particular topology, that being an outer JTAG chain with a single TAP (the Xilinx TAP) and then a RISC-V chain with a single nested TAP (the DM).   If the outer chain has more than a Xilinx TAP, and if the Xilinx TAP were put into bypass, then USER4 would no longer be selected and would need to be re-selected.  I don't know if it's common for other topologies to exist on the Arty, so I'm having second thoughts about making this change.  Maybe it's safer to leave this unmerged, for now.  It seemed like a low-risk adjustment, but now I'm not so sure, and I don't even know what the human-discernable difference in performance is.  So I think I'll defer merging for now, might come back to this idea later if need be (and if there's more confidence that alternate topologies wouldn't break things).

PR #421, opened 2019-11-05
fespi: Properly support large flash devices

timsifive:
My previous fix simply forced us to always use the flash commands that took a 3-byte address, meaning we couldn't address anything beyond 16MB. The new code always uses the right command, and infers the number of address bytes from the size of the flash device.


GregSavin (approved):

xfdzjz:
Hi, 
  I have a question about riscv openocd,when I was trying to generate a openocd, I am not sure is it the openocd that can be used by Windows system under Eclipse, should I add '--host=i686-w64-mingw32' after ./configure? cause I just use normal PC Windows system, I not quite sure my system is ming32 on Debian. Really hope you can give me some tips, thank you

xfdzjz:
One more question is that after sudo make install, I have the openocd file, should I add .exe after openocd?  Truly thanks for your help

timsifive:
@xfdzjz If you have a RISC-V problem with OpenOCD, please file a new issue under https://github.com/riscv/riscv-openocd/issues. If you have a problem creating an OpenOCD executable, that's best addressed to the main OpenOCD project: http://openocd.org/discussion/

PR #422, opened 2019-11-08
BSCAN batch fix

GregSavin:
The batch submission of DMI scans was not taking appropriate action in the case of BSCAN mode, but instead sending non-tunnel formatted JTAG scans instead, which was causing undefined behavior.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -8,6 +8,7 @@
 
 #define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
 #define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
+#define DIM(x)		(sizeof(x)/sizeof(*x))
 
 static void dump_field(int idle, const struct scan_field *field);
 
@@ -42,6 +43,9 @@ bool riscv_batch_full(struct riscv_batch *batch)
 
 int riscv_batch_run(struct riscv_batch *batch)
 {
+	uint8_t tunneled_dr_width;  
+	struct scan_field tunneled_dr[4];
+	
 	if (batch->used_scans == 0) {
 		LOG_DEBUG("Ignoring empty batch.");
 		return ERROR_OK;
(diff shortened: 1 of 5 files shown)


src/target/riscv/batch.c
```diff
@@ -8,6 +8,7 @@
 
 #define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
 #define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
+#define DIM(x)		(sizeof(x)/sizeof(*x))
```
timsifive: I recently discovered that OpenOCD already defines ARRAY_SIZE that does exactly this in src/helper/types.h, which is already included just about everywhere.

timsifive (changes requested):

src/target/riscv/batch.c
GregSavin: Ah OK, ARRAY_SIZE it is (probably not a bad idea to familiarize myself with other utility macros and functions)

src/target/riscv/batch.c
```diff
@@ -52,7 +55,45 @@ int riscv_batch_run(struct riscv_batch *batch)
 	riscv_batch_add_nop(batch);
 
 	for (size_t i = 0; i < batch->used_scans; ++i) {
-		jtag_add_dr_scan(batch->target->tap, 1, batch->fields + i, TAP_IDLE);
+		if (bscan_tunnel_ir_width != 0) {
```
timsifive: This block looks nearly identical to what we have in dmi_scan(). Can you refactor it to share that code?
(Sorry if I'm repeating myself here. I thought I had commented this earlier but I can't see it anywhere.)

timsifive (changes requested):

src/target/riscv/batch.c
GregSavin: Latest commit on this branch factors out the duplicated code, and also factors out the dependent data used by the code (it groups the dependent data into an opaque struct whose instances I figured can live on the stack because jtag_execute_queue is called within the lifetime of the corresponding stack-residing struct in the places it is instantiated/used... the data just needs to be live at least until the associated call to jtag_execute_queue finishes).

GregSavin: Actually I think I need to revisit this (in basic testing it seemed to work but I can't explain why), in the batch case I think there needs to be an array of riscv_bscan_tunneled_scan_context_t instances allocated within the batch structure, not just a single instance within riscv_batch_run.  I'll take another look and will probably post another commit.

GregSavin: OK I think I got the refactoring correct, it seems to work in focused smoke testing (attach w/ GDB, write some bytes to the load area in flash to clobber some values from any successful previous loads, load a program to same flash area, run compare-sections to verify the load result), but I haven't done a riscv-tests run yet.

timsifive (approved):
This looks good to me. If you're concerned about structures on the stack and how long they live, one run using valgrind on OpenOCD will definitely spit out an error if you got that wrong.

GregSavin:
Thanks Tim.  In one test run under valgrind, I'm not seeing any warnings that look to be focused on the code in this particular patch from this PR, although independently of this current PR I might take a look at that "conditional jump depends on uninitialized value" thing.

$ valgrind openocd -c "adapter_khz 10000" -f ./arty-onboard-ftdi.cfg -f ./sifive-e31arty-onboard-ftdi.cfg -c "riscv set_command_timeout_sec 10"
==30195== Memcheck, a memory error detector
==30195== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==30195== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==30195== Command: openocd -c adapter_khz\ 10000 -f ./arty-onboard-ftdi.cfg -f ./sifive-e31arty-onboard-ftdi.cfg -c riscv\ set_command_timeout_sec\ 10
==30195== 
--30195-- WARNING: Serious error when reading debug info
--30195-- When reading debug info from /lib/x86_64-linux-gnu/libusb-1.0.so.0.1.0:
--30195-- Ignoring non-Dwarf2/3/4 block in .debug_info
--30195-- WARNING: Serious error when reading debug info
--30195-- When reading debug info from /lib/x86_64-linux-gnu/libusb-1.0.so.0.1.0:
--30195-- Last block truncated in .debug_info; ignoring
--30195-- WARNING: Serious error when reading debug info
--30195-- When reading debug info from /lib/x86_64-linux-gnu/libusb-1.0.so.0.1.0:
--30195-- parse_CU_Header: is neither DWARF2 nor DWARF3 nor DWARF4
Open On-Chip Debugger 0.10.0+dev-00824-gf93ede5 (2019-11-07-09:57)
Licensed under GNU GPL v2
For bug reports, read
adapter speed: 10000 kHz

none separate

Info : auto-selecting first available session transport "jtag". To override use 'transport select <transport>'.
Info : Nested Tap based Bscan Tunnel Selected
Info : ftdi: if you experience problems at higher adapter clocks, try the command "ftdi_tdo_sample_edge falling"
Info : clock speed 10000 kHz
Info : JTAG tap: riscv.cpu tap/device found: 0x13631093 (mfg: 0x049 (Xilinx), part: 0x3631, ver: 0x1)
==30195== Conditional jump or move depends on uninitialised value(s)
==30195==    at 0x474A5E: buf_set_buf (binarybuffer.c:159)
==30195==    by 0x422E7B: bit_copy (binarybuffer.h:213)
==30195==    by 0x42415D: buffer_write (mpsse.c:468)
==30195==    by 0x4244C0: mpsse_clock_data (mpsse.c:524)
==30195==    by 0x419A14: clock_data (ftdi.c:286)
==30195==    by 0x41A866: ftdi_execute_scan (ftdi.c:568)
==30195==    by 0x41ADC9: ftdi_execute_command (ftdi.c:684)
==30195==    by 0x41AE6E: ftdi_execute_queue (ftdi.c:710)
==30195==    by 0x409302: default_interface_jtag_execute_queue (core.c:839)
==30195==    by 0x4171ED: interface_jtag_execute_queue (driver.c:369)
==30195==    by 0x40931C: jtag_execute_queue_noclear (core.c:845)
==30195==    by 0x409358: jtag_execute_queue (core.c:863)
==30195== 
Info : datacount=1 progbufsize=16
Info : Disabling abstract command reads from CSRs.
Info : Examined RISC-V core; found 1 harts
Info :  hart 0: XLEN=32, misa=0x40901125
Info : Listening on port 3333 for gdb connections
Info : Found flash device 'sp s25fl128s' (ID 0x00182001)
Ready for Remote Connections
Info : Listening on port 6666 for tcl connections
Info : Listening on port 4444 for telnet connections
Info : accepting 'gdb' connection on tcp/3333
Info : Disabling abstract command writes to CSRs.
Info : dropped 'gdb' connection
^Cshutdown command invoked
==30195== 
==30195== Process terminating with default action of signal 2 (SIGINT)
==30195==    at 0x5594767: kill (syscall-template.S:84)
==30195==    by 0x46E74C: exit_on_signal (server.c:715)
==30195==    by 0x4055B6: openocd_main (openocd.c:371)
==30195==    by 0x404EA1: main (main.c:39)
==30195== 
==30195== HEAP SUMMARY:
==30195==     in use at exit: 8,336 bytes in 4 blocks
==30195==   total heap usage: 79,952 allocs, 79,948 frees, 6,339,885,229 bytes allocated
==30195== 
==30195== LEAK SUMMARY:
==30195==    definitely lost: 0 bytes in 0 blocks
==30195==    indirectly lost: 0 bytes in 0 blocks
==30195==      possibly lost: 0 bytes in 0 blocks
==30195==    still reachable: 8,336 bytes in 4 blocks
==30195==         suppressed: 0 bytes in 0 blocks
==30195== Rerun with --leak-check=full to see details of leaked memory
==30195== 
==30195== For counts of detected and suppressed errors, rerun with: -v
==30195== Use --track-origins=yes to see where uninitialised values come from
==30195== ERROR SUMMARY: 16902 errors from 1 contexts (suppressed: 0 from 0)

timsifive:
That conditional jump has been in there since I first ran valgrind. It's in the target-independent code so I haven't taken the time to look at it.

PR #424, opened 2019-11-19
Fix: Take into account progbuf writability.

JanMatCodasip:
When allocating scratch memory within RISC-V target
(scratch_reserve()), take into account whether progbuf
is writable or not, as determined by examine_progbuf().


timsifive (approved):
Looks good. Thank you!

PR #426, opened 2019-11-19
Fixed write_memory_progbuf() on RV64.

JanMatCodasip:
Abstract write size (aarsize) shall always match the real
size of the register. This is because abstract write of smaller size
than the register need not be supported per the spec (pg. 13 of RISC-V
External Debug Support ver. 0.13.2).

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3204,7 +3204,7 @@ static int write_memory_progbuf(struct target *target, target_addr_t address,
 				/* Write and execute command that moves value into S1 and
 				 * executes program buffer. */
 				uint32_t command = access_register_command(target,
-						GDB_REGNO_S1, size > 4 ? 64 : 32,
+						GDB_REGNO_S1, riscv_xlen(target),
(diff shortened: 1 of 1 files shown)


timsifive:
This change looks like it does the opposite of your comment. It reads to me like the original code sets aarsize to the size of the register, and your change uses XLEN instead. What am I missing?

JanMatCodasip:
The original code selected the width of the abstract register access to S1 per the size of the memory transfer, which is incorrect. For instance, when writing 32-bit words to memory on RV64, the original code would only perform 32-bit abstract write to the 64-bit-wide register S1 - and that need not be supported by the target. (See also: https://github.com/riscv/riscv-debug-spec/issues/499)

The little fix ensures that the width of the abstract register write (aarsize) always matches the register width in the target.

I looked throughout the RISC-V target code and this is the only place that needs this fix - all other instances of abstract writes already have the correct size parameter.

timsifive:
That makes sense. Thanks for explaining it again.

timsifive (approved):

timsifive:
Still passes riscv-tests/debug with this change.

PR #429, opened 2019-12-03
Increase maximum number of harts

bluewww:
OpenOCD can't deal with systems that have more than 32 harts.

### DIFF
--- a/src/target/riscv/riscv.h
+++ b/src/target/riscv/riscv.h
@@ -9,7 +9,7 @@ struct riscv_program;
 #include "jtag/jtag.h"
 
 /* The register cache is statically allocated. */
-#define RISCV_MAX_HARTS 32
+#define RISCV_MAX_HARTS 1024
 #define RISCV_MAX_REGISTERS 5000
 #define RISCV_MAX_TRIGGERS 32
 #define RISCV_MAX_HWBPS 16


timsifive:
How much more memory does OpenOCD use when you do this? To keep supporting `-rtos riscv` (which is deprecated and will be removed eventually) I think memory usage is O(n^2) on the number of harts. Once we get rid of that we should be able to remove the limit altogether.

bluewww:
It increases memory usage from `2.4MiB` to `45.1MiB` on my machine.

timsifive:
How many harts are actually configured in that measurement?

bluewww:
One. Aren't we interested how this patch affects the regular case (low hart count/low hart id number since that seems to be the same)?

timsifive:
So if 1 hart uses 43MiB more after this change, then (assuming somebody was using OpenOCD for) a 32-hart system using might incur 1.3GiB of extra memory usage. That's kind of high.

Do you need the full 1024 harts, or did you just pick a number that should be good for a long time? I'd be a lot more comfortable with half or a quarter of this memory increase.

Over the long term `-rtos riscv` is going to go away, we can get rid of RISCV_MAX_HARTS, and memory usage won't blow up the way it does currently when increasing that number.

bluewww:
Well in our case it's a combination of both: We have setups with more than 32 cores and the maximum hartid is just slightly below 1000 ish.
Are there any low hanging fruits to alleviate the memory issues?

timsifive:
Turns out there's some very low hanging fruit: https://github.com/riscv/riscv-openocd/pull/431
Once that merges, your change will be fine.

timsifive (approved):

PR #431, opened 2019-12-04
Remove unused data structure.

timsifive:
Saves 1.4MiB of RAM too, with just 1 hart configured.

Change-Id: I68d8c003a67c280b62ff6c9285ac6f54865f99f2


GregSavin (approved):
Is this a case where no other code was referencing saved_registers[][]?   So basically unreferenced vestigial data left orphaned after a refactor or something?

timsifive:
Exactly.

PR #432, opened 2019-12-16
Fix of DMI batch scans over 64-bits

JanMatCodasip:
This fixes buffer overrun/data corruption during DMI scan batches
on targets where the DMI scans would be longer than 64 bits.
Needed for targets with large value of dtmcs.abits > 30 (which is 
unusual, but within the spec).

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -9,6 +9,9 @@
 #define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
 #define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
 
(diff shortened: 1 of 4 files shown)


timsifive (commented):
This looks right. Can you make sure to run `make all` in riscv-tests/debug? I'm always ware of changes to the memory reading code. There are a bunch of corner cases there, but the test suite hits most of them.

JanMatCodasip:
Thank you for looking at the code. I will run the test suite Friday/Monday and will reach back with the results.

JanMatCodasip:
It will take me a bit more time to get to the test suite than I expected. Please expect a reply from me in the week of Jan/6.

JanMatCodasip:
@timsifive - I have revived this dormant but still valid fix. It has been rebased on top of the current code in `riscv`.

Running `riscv-tests/debug` produces no failures (after applying fixes as discussed in https://github.com/riscv/riscv-tests/issues/287). Results are attached: [riscv-tests-results.txt](https://github.com/riscv/riscv-openocd/files/5034508/riscv-tests-results.txt).

Please let me know if there is more to do prior it can be merged.

timsifive (approved):
Looks good. Thanks for taking the time to do this. I'll probably merge it tomorrow.

PR #433, opened 2019-12-17
target/riscv : Fix resume PC value after semihosting file IO operation

ewft:
After performing semihosting file io operation the target is resumed at its current pc ( an ebreak ).
Which re-trigger a semihosting operation.

Adding 4 ( semihosting code use norvc ) to the PC  to resume on the next instruction.

Signed-off-by: Benjamin Vinot <benjamin.vinot@emmicroelectronic.com>

### DIFF
--- a/src/target/riscv/riscv_semihosting.c
+++ b/src/target/riscv/riscv_semihosting.c
@@ -190,5 +190,11 @@ static int riscv_semihosting_post_result(struct target *target)
 
 	LOG_DEBUG("0x%" PRIx64, semihosting->result);
 	riscv_set_register(target, GDB_REGNO_A0, semihosting->result);
+
+	riscv_reg_t dpc;
+	riscv_get_register(target, &dpc, GDB_REGNO_DPC);
+	riscv_set_register(target, GDB_REGNO_DPC, dpc + 4);
+
+
 	return 0;
 }


abidh (commented):
I faced the same issue and this fixes the problem for me. Although I noticed that in 4f9e2d7171f3cad8d1a99dff3eee5ec5e6d8ea2b, the GDB_REGNO_DPC was replaced by GDB_REGNO_PC in this file. I wonder if this patch should be updated accordingly.

timsifive:
I have a test that uses semihosting file I/O to write a file, and that test has been passing just fine. What exactly is the scenario where this problem occurs?

abidh:
This issue causes the following message on openocd console.
semihosting: unsupported call 0
I have only tried "write" and noticed that return value is wrong due to this problem. That does not seem to have any effect on printf but cout blocks after printing first string. (I am using llvm libc++).

TommyMurphyTM1234:
> This issue causes the following message on openocd console.
> semihosting: unsupported call 0
> I have only tried "write" and noticed that return value is wrong due to this problem. That does not seem to have any effect on printf but cout blocks after printing first string. (I am using llvm libc++).

Don't you, your program or your std lib need to provide the necessary support for semihosting?

https://github.com/riscv/riscv-openocd/blob/riscv/src/target/semihosting_common.c#L1326

I don't think that this is an issue with the RISC-V specific or openocd target generic semihosting support per se - rather something missing in your "board support package"?

abidh:
I guess I need to explain what is happening a bit more. We hit a semihosting breakpoint. Do the usual stuff and then continue from the same place and hit the same breakpoint again. I think this happens here. See target_resume with handle_breakpoint is false.
https://github.com/riscv/riscv-openocd/blob/95a8cd9b5d07501ac2243c61f331b092b1ea9894/src/server/gdb_server.c#L3185

The register GDB_REGNO_A0 stores the op at the time of first call. But when we hit the breakpoint 2nd time, its value has changed. It now stores the result of previous call. So semihosting complains that this is unsupported call. But in this case, it does resume from dpc+4 
https://github.com/riscv/riscv-openocd/blob/95a8cd9b5d07501ac2243c61f331b092b1ea9894/src/target/riscv/riscv_semihosting.c#L157

so thing seem be working ok although the return value is wrong now and can cause problem as I discovered with write and libc++.

TommyMurphyTM1234:
I don't know how all this tallies with what @timsifive said/asked here?

https://github.com/riscv/riscv-openocd/pull/433#issuecomment-634171325

abidh:
@TM1234  semihosting is working for me too. I see printf in my application correctly printing on GDB's console. But the problem is there and it can show up e.g. when I try cout. 

@timsifive can you confirm that you don't see the following warning in the openocd console when you use semihosting?
semihosting: unsupported call

timsifive:
I see no warnings in my semihosting test, which is very basic. (It's this program: https://github.com/riscv/riscv-tests/blob/master/debug/programs/semihosting.c)

Do you have an example program where the problem occurs? I'd love to expand the testcase so it hits the problem reported here, and then merge this (or a different) fix.

timsifive:
Closing due to inactivity. If you can provide a test case that produces a problem, I'm happy to look at it more.

PR #434, opened 2019-12-18
Deal with a case where DMI read returns busy.

timsifive:
This was exposed by the latest change, which modifies the number of
scans enough that this problem is now exposed by the testsuite.

This particular failure could happen in more cases, but I don't have the
time to address everything right now. In the real world they'll be very
rare. It only shows up in the test because it intentionally resets the
learned delays while a memory access is happening..

Change-Id: Ia865eb76ed630c3d836131a065ec2dad0657c9b4

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -606,6 +606,12 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 					false);
 			if (status == DMI_STATUS_BUSY) {
 				increase_dmi_busy_delay(target);
+				if (dmi_op == DMI_OP_READ) {
+					/* The scan where we should have gotten the read value
+					 * returned busy.  It's anybody's guess whether the read
+					 * actually happened or not. */
+					return ERROR_BUSY;
+				}
 			} else if (status == DMI_STATUS_SUCCESS) {
 				break;
 			} else {
@@ -2060,25 +2066,29 @@ static int read_memory_bus_word(struct target *target, target_addr_t address,
 {
 	uint32_t value;
 	if (size > 12) {
-		if (dmi_read(target, &value, DMI_SBDATA3) != ERROR_OK)
-			return ERROR_FAIL;
+		int result = dmi_read(target, &value, DMI_SBDATA3);
+		if (result != ERROR_OK)
+			return result;
 		write_to_buf(buffer + 12, value, 4);
 		log_memory_access(address + 12, value, 4, true);
 	}
 	if (size > 8) {
-		if (dmi_read(target, &value, DMI_SBDATA2) != ERROR_OK)
-			return ERROR_FAIL;
+		int result = dmi_read(target, &value, DMI_SBDATA2);
+		if (result != ERROR_OK)
+			return result;
 		write_to_buf(buffer + 8, value, 4);
 		log_memory_access(address + 8, value, 4, true);
 	}
 	if (size > 4) {
-		if (dmi_read(target, &value, DMI_SBDATA1) != ERROR_OK)
-			return ERROR_FAIL;
+		int result = dmi_read(target, &value, DMI_SBDATA1);
+		if (result != ERROR_OK)
+			return result;
 		write_to_buf(buffer + 4, value, 4);
 		log_memory_access(address + 4, value, 4, true);
 	}
-	if (dmi_read(target, &value, DMI_SBDATA0) != ERROR_OK)
-		return ERROR_FAIL;
+	int result = dmi_read(target, &value, DMI_SBDATA0);
+	if (result != ERROR_OK)
+		return result;
 	write_to_buf(buffer, value, MIN(size, 4));
 	log_memory_access(address, value, MIN(size, 4), true);
 	return ERROR_OK;
@@ -2296,10 +2306,19 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
 			}
 		}
 
-		for (uint32_t i = (next_address - address) / size; i < count - 1; i++) {
-			if (read_memory_bus_word(target, address + i * size, size,
-					buffer + i * size) != ERROR_OK)
-				return ERROR_FAIL;
+		bool busy = false;
+		for (uint32_t i = (next_address - address) / size;
+				i < (end_address - address) / size - 1;
+				i++) {
+			int result = read_memory_bus_word(target, address + i * size, size,
+					buffer + i * size);
+			if (result == ERROR_BUSY) {
+				busy = true;
+				break;
+			} else if (result != ERROR_OK) {
+				return result;
+			}
+			next_address += size;
 		}
 
 		uint32_t sbcs_read = 0;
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -606,6 +606,12 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 					false);
 			if (status == DMI_STATUS_BUSY) {
 				increase_dmi_busy_delay(target);
+				if (dmi_op == DMI_OP_READ) {
```
GregSavin, line 609: So, there used to be transparent retry at this level when a busy result came back, and now the retry responsibility has been moved upward to the caller(s)?

JanMatCodasip, line 609: Hello, 

per my understanding of the Debug Spec v0.13, DMI "busy" responses in case of DMI reads would not require any special handling apart from what is already in the code. The original code appears completely sufficient.

Per Debug Spec 0.13, page 66:

> dmi.op == 3: An operation was attempted while a DMI request is still in progress. The data scanned into dmi in this access will be ignored. (...)

Let's assume the following sequence of events:

* New DMI read transaction is issued by dmi_scan() call at line 579
* We obtain DMI_SUCCESS response (dmi.op == 0), therefore we know the transaction has been started correctly and is now in progress (line 585)
* We try to obtain the read result by issuing DMI NOP on line 605
* Let's assume the DMI read transaction is not yet finished (still runs in the background), therefore we obtain BUSY response (dmi.op == 3) to our NOP. The NOP is ignored and the read transaction continues.
* _Why to break the loop now?_ My understanding is that we should keep issuing additional DMI NOPs, and the first response that has dmi.op==0 (success) will also carry the data - result of the original "read" operation.

Therefore, I do not see the need for the code change. The existing code should, as I see it, cover all BUSY cases.

Please let me know if I missed/misunderstood anything. Thank you.

timsifive: busy is a sticky state, and can only be cleared by writing dmireset. No other operations (including NOPs) will be executed as long as the status is busy.

It is this way so that a debugger can issue a stream of commands without checking for the results, and only check for results at the end. Some hardware (e.g. Olimex adapter) can get much better performance that way. It's unfortunate because in this particular case it means the debugger ends up reading from memory twice, which might have side effects.

timsifive: Hmm... there is possibly some fudging in the spec language that means NOP might not make busy sticky.

timsifive: One thing that's wrong with @JanMatCodasip's summary is the second point. The status we get from the scan where we initiated the read is the result of the previous operation (nop in this case). It's not until the next scan that we learn whether the read took place or not. In my test case it did not, because the read was issued too soon after the previous nop. Does that make sense?

GregSavin: These changes look good to me (especially having read the latest comments).  Sounds like the problem was exposed through stress testing (Spike?), and also the fix verified by the same tests.

timsifive:
I'm looking into a different solution to this problem, but I do think the problem is real. If you issue a read, follow too quickly by a NOP, the target may set busy in response, and busy is sticky.

src/target/riscv/riscv-013.c
JanMatCodasip: Thank you for your replies.

> busy is a sticky state, and can only be cleared by writing dmireset. No other operations (including NOPs) will be executed as long as the status is busy.

I understand the sticky nature of the "busy" state, along with the reasons for it, but forgot to mention it in my fifth point in the sequence. To be precise, I should have said:

* ... we should keep issuing additional DMI NOPs, and the first response that has dmi.op==0 (success) will also carry the data - result of the original "read" operation. For responses with dmi.op==3 (busy), we reset the sticky busy state by writing to dtmcs.dmireset bit and then continue issuing NOPs. The original OpenOCD code already already handles the sticky state correctly by resetting it each time it is encountered - in `increase_dmi_busy_delay()`, called at line 608.

> One thing that's wrong with @JanMatCodasip's summary is the second point. The status we get from the scan where we initiated the read is the result of the previous operation (nop in this case). It's not until the next scan that we learn whether the read took place or not. 

My understanding of the DTM spec is that: 

* i) If we issue a "read" transaction (dmi.op==1) and receive success in that very same dmi scan (received dmi.op==0), that indeed means that the _previous_ DMI transaction succeeded, as you have said. 
* ii) For our discussion it should not matter what the previous DMI transaction was - it needn't have been just NOP but also read or write. 
* iii) The important thing is that this previous transaction succeeded == it completed and the DTM module is ready for a new transaction.
* iv) So receiving dmi.op==0 (success response) in the same DMI scan where we initiated the read **also means** that the new "read" operation **must have been started and is being processed**, because we know the DTM was idle and ready for transactions at the point when we tried to start the read.

In other words:

If we issue a read, followed too quickly by a NOP, I believe the current code in OpenOCD should already handle it correctly. If there are issues in this realm, then it appears more as a bug in the target than missing logic in OpenOCD code. 

What OpenOCD already does I believe is sufficient and correct - It will detect that the read has been started (by receiving dmi.op==0 in the initial DMI transaction which starts the read, indicating that DTM was idle when the read request came). OpenOCD will then keep issuing NOPs until the result of the running read operation becomes available (and will properly reset the sticky busy state via dmireset each time it encouters it).

Or, please let me know if there is still a piece of the spec that I missed or misunderstood. Thanks.

timsifive:
On Mon, Jan 6, 2020 at 12:10 AM Jan Matyas <notifications@github.com> wrote:

> My understanding of the DTM spec is that:
>
>    -
>       1. If we issue a "read" transaction (dmi.op==1) and receive success
>       in that very same dmi scan (received dmi.op==0), that indeed means that the
>       *previous* DMI transaction succeeded, as you said.
>    -
>       1. For our discussion it should not matter what the previous DMI
>       transaction was - it needn't have been just NOP but also read or write. The
>       important thing is that this previous transaction succeeded == it completed
>       and the DTM module is ready for a new transaction.
>    -
>       1. So receiving dmi.op==0 (success response) in the same DMI scan
>       where we initiated the read also means that the "read" operation *must
>       have been started and is being processed*, because we know the DTM
>       was idle and ready for transactions at the point when we started the read.
>
> I think that is right.
The problem is that once busy has been set, dmireset is the only way to
clear it, and dmireset might interfere with the read that was issued.
Depending on the implementation of the Debug Module Interface, the read we
issued might not even have made it to the DM by the time OpenOCD issues the
reset. The spec is frustratingly unclear about dmireset: "Writing 1 to this
bit clears the sticky error state and allows the DTM to retry or complete
the previous transaction."
That's even contradictory. If setting dmireset allows a retry, that assumes
the in-progress transaction is cancelled. If it allows it to complete, then
the in-progress transaction is still pending. Are there other ways to read
that?

timsifive:
But dmihardreset explicitly says that it cancels any outstanding transactions. So it's not unreasonable to read the spec as saying that dmireset only clears the busy bit but has no other side effects.

I'll check with some implementations, and if that's what people have done, then I think we can clarify the spec so it behaves that way, and then this OpenOCD code is unnecessary.

JanMatCodasip:
Thank you for replying - I completely agree.

**As for the interpretation of the spec -**

Our understanding was that `dmireset` only clears the sticky error state, without any other side effects. There were two main reasons for this interpretation:

* 1) The difference between `dmireset` vs. `dmihardreset` - as you pointed out above
* 2) The fact that this interpretation allows for "guaranteed" execution of transactions -- it becomes clear to OpenOCD (or anther tool) which DMI transactions have been carried out and which ones need to be retried due to the busy state - without any ambiguity. I believe that the guaranteed/reliable transport is a desired property of the DTM layer of RISC-V debug.

However, I do agree that the current wording of the spec leaves certain room for other interpretations.

**Specification of the dmireset bit -**

> The spec is frustratingly unclear about dmireset: "Writing 1 to this bit clears the sticky error state and allows the DTM to retry or complete the previous transaction." That's even contradictory. 

I fully agree and also read the sentence as contradictory. IMO, it is neccessary to specify what the behavior of `dmireset` is w.r.t. running transactions.



Thank you for looking into this matter with me. I appreciate the discussion.

timsifive:
Inactive

PR #436, opened 2019-12-30
Upcast mask value to work with 64-bit physical

timsifive:
Change-Id: I00f0d2a3c79a431e1aa49c7478fa6c17e2fa5256


timsifive:
@Hsiangkai, can you take a look at this?

GregSavin (approved):
The widening cast looks correct to me to cover physical addresses > 32 bits (though I'm not really familiar with RISC-V paging yet, other than having read through the RISC-V spec document describing supervisor mode etc)

PR #437, opened 2020-01-03
Handle DMI busy in sba write.

timsifive:
If we encounter DMI busy on the NOP after a read, we'll never get the
value out because DMI busy is sticky. The read must be retried, but we
don't know whether it was ever issued. Since the read has side effects
(incrementing of the address) this retry must be handled at a higher
layer. So now dmi_op_timeout can be told to retry or not, and if retry
is disabled it'll return an error when busy.

Also actually properly do the retry in dmi_op_timeout(). Previously the
code would not reissue the command and end up returning a garbage value.

This is a cleaner alternative to #434

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -654,11 +654,6 @@ static int dmi_read(struct target *target, uint32_t *value, uint32_t address)
 	return dmi_op(target, value, DMI_OP_READ, address, 0, false, true, true);
 }
 
-static int dmi_read_exec(struct target *target, uint32_t *value, uint32_t address)
-{
-	return dmi_op(target, value, DMI_OP_READ, address, 0, true, true, true);
-}
-
 static int dmi_write(struct target *target, uint32_t address, uint32_t value)
 {
 	return dmi_op(target, NULL, DMI_OP_WRITE, address, value, false, true, true);
@@ -2058,14 +2053,15 @@ static void log_memory_access(target_addr_t address, uint64_t value,
 /* Read the relevant sbdata regs depending on size, and put the results into
  * buffer. */
 static int read_memory_bus_word(struct target *target, target_addr_t address,
-		uint32_t size, uint8_t *buffer)
+		uint32_t size, uint8_t *buffer, bool retry)
 {
 	uint32_t value;
 	int result;
 	static int sbdata[4] = { DMI_SBDATA0, DMI_SBDATA1, DMI_SBDATA2, DMI_SBDATA3 };
 	assert(size <= 16);
 	for (int i = (size-1) / 4; i >= 0; i--) {
-		result = dmi_op(target, &value, DMI_OP_READ, sbdata[i], 0, false, true, false);
+		result = dmi_op(target, &value, DMI_OP_READ, sbdata[i], 0, false, true,
+				retry);
 		if (result != ERROR_OK)
 			return result;
 		write_to_buf(buffer + i * 4, value, MIN(size, 4));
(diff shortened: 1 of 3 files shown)


GregSavin (dismissed):
So, calling dmi_op with retry=true is generally appropriate if and only if repeating the particular DMI operation being invoked is known not to have stateful side effects on the target?

timsifive:
> So, calling dmi_op with retry=true is generally appropriate if and only if repeating the particular DMI operation being invoked is known not to have stateful side effects on the target?

Yes, that's exactly right. Should I add a comment somewhere to make that more clear?

GregSavin:
It might be worth including that in the comments (for new maintainers etc)

src/target/riscv/riscv-013.c
```diff
@@ -3266,9 +3274,14 @@ static int write_memory_progbuf(struct target *target, target_addr_t address,
 		 * to be incremented if necessary. */
 
 		uint32_t abstractcs;
-		bool dmi_busy_encountered;
-		if (dmi_op(target, &abstractcs, &dmi_busy_encountered, DMI_OP_READ,
-					DMI_ABSTRACTCS, 0, false, true) != ERROR_OK)
+		result = dmi_op(target, &abstractcs, DMI_OP_READ, DMI_ABSTRACTCS,
+				0, false, true, false);
```
GregSavin: Curious why retry is specified as false in this particular call, but then if the result comes back as busy, then a second read operation is made immediately, but with retry=true.  (Why not specify retry=true in the first call?  Is the first call variant more efficient, and usually expected not to have a busy result?)

JanMatCodasip:
> If we encounter DMI busy on the NOP after a read, we'll never get the
value out because DMI busy is sticky. The read must be retried, but we
don't know whether it was ever issued.

I am sorry but I believe this point is incorrect. If we issue a "DMI read" and get dmi.op==0 as a response in that very same DMI scan, we know that the new DMI read must have been started, because dmi.op==0 says all previous transactions had been finished by the time our new read request came. In short, we can be sure that our DMI read got really started.

I have elaborated more on this under #434 

So in the end, there should not be any need to expose the "BUSY" state and retry mechanisms to the upper layers above `dmi_op()/dmi_op_timeout()`. All the DTM retry logic can stay in `dmi_op_timeout()` and need not "pollute" the upper layers.

If there is a case when we issue a "DMI read", get back dmi.op==0 in the same DMI scan but the target for some reason does not carry out the DMI read, that is, in my opinion, a bug of the target and violation of the spec.

src/target/riscv/riscv-013.c
```diff
@@ -582,65 +584,73 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
...
-						false);
-				if (status == DMI_STATUS_BUSY) {
-					increase_dmi_busy_delay(target);
-					if (!retry)
-						return ERROR_TARGET_BUSY;
-					break;
-				} else if (status == DMI_STATUS_SUCCESS) {
-					return ERROR_OK;
+			if (status == DMI_STATUS_FAILED) {
+				if (data_in) {
```
GregSavin, line 625: Is the body of this "if" statement reachable?  I see the assignment to status in line 616, and if the value DMI_STATUS_FAILED gets assigned to status at line 616, then the "else" clause at line 624 will be executed, including the return statement on line 627, in which case the status variable seems guaranteed to be unequal to DMI_STATUS_FAILED at line 632.

timsifive: You're right. Looks like some code got duplicated there. Thanks!

GregSavin (approved):

PR #439, opened 2020-01-08
Don't issue extra FENCE+FENCE.i for the current hart.

JanMatCodasip:
The original OpenOCD code issued FENCE & FENCE.i twice for the current
hart (which is harmless, but takes time).

Avoiding this extra FENCE is a slight performance improvement. Per my rough
measurements, this improves performance of certain debugger actions
(single-stepping) by approx. 20% in single-hart systems.


timsifive (approved):
Looks good!

PR #444, opened 2020-01-30
Update the current thread when gdb requests a step.

timsifive:
Evidently this is what gdb expects.

Change-Id: I634cdbcbcfab149c1b916e3744ff4915a8f8669b


GregSavin (approved):
Makes sense, looks good to me.

PR #447, opened 2020-02-12
sysctl.h is deprecated

ubfx:
In file included from src/helper/options.c:38:
/usr/include/x86_64-linux-gnu/sys/sysctl.h:21:2: error: #warning "The <sys/sysctl.h> header is deprecated and will be removed." [-Werror=cpp]
   21 | #warning "The <sys/sysctl.h> header is deprecated and will be removed."
      |  ^~~~~~~

### DIFF
--- a/src/helper/options.c
+++ b/src/helper/options.c
@@ -34,9 +34,6 @@
 #if IS_DARWIN
 #include <libproc.h>
 #endif
-#ifdef HAVE_SYS_SYSCTL_H
-#include <sys/sysctl.h>
-#endif
 #if IS_WIN32 && !IS_CYGWIN
 #include <windows.h>
 #endif


timsifive (approved):
Given that it builds, that's fine by me. Not sure how upstream is going to handle this (has already handled it?) but this is a small enough change that I can deal with it.

TommyMurphyTM1234:
Any use?

https://repo.or.cz/openocd.git/blob/HEAD:/src/helper/options.c#l37

ubfx:
I mean, the impact of leaving it in is pretty big: You can't build riscv-openocd from source on newer Linux/gcc.

TommyMurphyTM1234:
I posted a link to the upstream fix.

ubfx:
Sorry, I didn't realize how this works. So do I do a new pull request now with the upstream fix or will this be fixed automatically?

timsifive:
Once in a while I merge all upstream into this branch. I just created #450 which takes just the one change that Tommy identified and brings it into our branch. That should solve the problem and prevent any future conflicts.

timsifive:
Fixed in #450.

PR #448, opened 2020-02-12
Add support for vector register access

timsifive:
Vector registers are exposed to gdb as a union of arrays of bytes, shorts, words, longs, and quad-words (gdb's largest native type).
The code is relatively readable but slow. The way registers are cached needs to be improved to get a bunch of performance back, without sacrificing code readability. That's beyond the scope of this change.

There are a few minor cleanups sprinkled in, too.


GregSavin (commented):
As far as getting in a good position to review this, what would be the best place to look for the most current/relevant specification information on RISC-V vector registers?

aswaterman:
https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc#3-vector-extension-programmers-model

src/target/riscv/riscv-013.c
```diff
@@ -1048,6 +1055,54 @@ static int examine_progbuf(struct target *target)
...
+		} else if (is_vector_reg(regno) && (*mstatus & MSTATUS_VS) == 0) {
+			if (register_write_direct(target, GDB_REGNO_MSTATUS,
+						set_field(*mstatus, MSTATUS_VS, 1)) != ERROR_OK)
+				return ERROR_FAIL;
+		}
+	}
+	return ERROR_OK;
+}
+
+static int cleanup_after_register_access(struct target *target,
```
GregSavin, line 1096: So the prep/cleanup pair is to avoid illegal instruction exceptions during vector accesses, while also avoiding returning to user code with mstatus in a different state from its original value?

timsifive: Exactly right.

src/target/riscv/riscv-013.c
```diff
@@ -1757,6 +1836,140 @@ static unsigned riscv013_data_bits(struct target *target)
...
+	for (unsigned i = 0; i < debug_vl; i++) {
+		if (riscv_program_exec(&program, target) != ERROR_OK)
+			return ERROR_FAIL;
+		uint64_t v;
+		if (register_read_direct(target, &v, GDB_REGNO_S0) != ERROR_OK)
+			return ERROR_FAIL;
+		buf_set_u64(value, xlen * i, xlen, v);
+	}
+
+	if (cleanup_after_vector_access(target, vtype, vl) != ERROR_OK)
```
GregSavin, line 1930: How likely is it that an earlier ERROR_FAIL in this function would cause the cleanup steps to be skipped?   Or are those potential failure points earlier in this function "catastrophic" enough that if they ever happen, then skipping these cleanup steps isn't going to cause additional significant adverse consequences compared to what's already gone wrong to have caused an earlier failure point?

timsifive: The only failure I can imagine that we may be able to recover from is somehow V registers not being accessible at all. Maybe we were misguided thinking they exist, or something. I've updated the code to deal with that.
The rest of the failures are a real stretch to think things are working enough that we could even restore anything.

src/target/riscv/riscv-013.c
```diff
@@ -1757,6 +1836,153 @@ static unsigned riscv013_data_bits(struct target *target)
...
+		return ERROR_FAIL;
+
+	if (cleanup_after_register_access(target, mstatus, regno) != ERROR_OK)
+		return ERROR_FAIL;
+	if (register_write_direct(target, GDB_REGNO_S0, s0) != ERROR_OK)
+		return ERROR_FAIL;
+
+	return result;
+}
+
```
GregSavin, line 1985: So both the reading and writing involve essentially repeatedly doing an element-wise rotate of the vector register, but for the write we're rotating in element values from the caller buffer, and for the read we're rotating in the corresponding original elements back in after we've copied the element values out to the caller buffer.  This looks good to me.

GregSavin (approved):

Issue #443, opened 2020-01-27
warning sysctl.h stop compilation


damiencarrier:
Hello,
I have a warning and cc1 says: all warnings being treated as errors
I don't find how to disable it in the makefile.
What can i do?
Thank you :)

![compilation openocd](https://user-images.githubusercontent.com/57895929/73163083-80c55980-40ef-11ea-93cc-a1d77b11feb4.png)

TommyMurphyTM1234:
Remove -Werror?
I don't get that warning/compilation failure when compiling the RISC-V openocd fork.
Have you modified the code to cause this or something?

rtwfroody:
Specifically, when running configure you can add `--disable-werror` to the command line to ignore warnings.

towoe:
The problem is glibc 2.30 deprecating `sys/sysctl.h`.
@damiencarrier probably uses a distribution which already uses the new glibc version.

Upstream fix http://openocd.zylin.com/#/c/5317/

damiencarrier:
Hello,
Thank you, i re-download openocd and find the -Werror to remove it in the makefile.

towoe:
Yes, removing `-Werrror` will help as it disables the handling of warnings as errors. But this should not be the final solution.
Not sure about the process of importing changes from upstream, but it would make sense to incorporate the upstream fix.

PR #450, opened 2020-02-13
helper: skip including sys/sysctl.h on Linux

timsifive:
Starting from glibc 2.30, the header file sys/sysctl.h gets
deprecated on Linux, after the commit 744e82963716 ("Linux:
Deprecate <sys/sysctl.h> and sysctl")
	https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=744e82963716

The associated NEWS reports
	The Linux-specific <sys/sysctl.h> header and the sysctl
	function have been deprecated and will be removed from a
	future version of glibc.

Latest automake 1.16.1 still does not handle this case.
Current OpenOCD build fails with warning and requires configure
with "--disable-werror" to build.

Prevent including sys/sysctl.h on Linux build.

Change-Id: I5310976573352a96e5aef123352f73475f0c35fe
Signed-off-by: Antonio Borneo <borneo.antonio@gmail.com>
Reviewed-on: http://openocd.zylin.com/5317
Tested-by: jenkins
Reviewed-by: Moritz Fischer <moritz.fischer.private@gmail.com>
Reviewed-by: Paul Fertser <fercerpav@gmail.com>

Cherry picked this from mainline, as an alternative to #447.

### DIFF
--- a/src/helper/options.c
+++ b/src/helper/options.c
@@ -34,8 +34,11 @@
 #if IS_DARWIN
 #include <libproc.h>
 #endif
+/* sys/sysctl.h is deprecated on Linux from glibc 2.30 */
+#ifndef __linux__
 #ifdef HAVE_SYS_SYSCTL_H
 #include <sys/sysctl.h>
+#endif
 #endif
 #if IS_WIN32 && !IS_CYGWIN
 #include <windows.h>


towoe (approved):
Not sure if I can approve this commit, but following #443 this PR LGTM.

GregSavin (approved):

PR #451, opened 2020-02-14
Give control over dcsr.ebreak[msu] bits.

timsifive:
This allows a user to debug code that uses software breakpoints itself.

Change-Id: If40cb626354e11703017cdf8c5919a31e83ebc3f

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -9511,6 +9511,21 @@ Enable or disable use of a BSCAN tunnel to reach DM.  Supply the width of
 the DM transport TAP's instruction register to enable.  Supply a value of 0 to disable.
 @end deffn
 
+@deffn Command {riscv set_ebreakm} on|off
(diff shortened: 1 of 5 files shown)


timsifive:
@SiFiveGregS, can you confirm this does what you want as part of reviewing?

GregSavin (approved):
Looks good to me, and I verified that "riscv set_ebreakm off" results in the machine mode exception being taken, rather than a trap into OpenOCD.  Then re-running after "riscv set_ebreakm on" generates a trap into OpenOCD again.  Thanks!    

Actually I had to test by "cherrypick" merging this branch into the commit of OpenOCD that was used in the latest Freedom Tools release (commit bf1e201) because the current HEAD of the riscv branch seems to be showing "Unable to allocate work area" problems for me when I use the load command.  I may end up trying to bisect to see which commit between bf1e201 and current HEAD introduced the load problem I'm seeing.

PR #453, opened 2020-02-29
Fix address translation when high bits are set.

### LINKED ISSUE #452: Incorrect physical address translation with `set_enable_virt2phys`
I'm running into problems with the features from #425 when trying to debug Linux (after `setup_vm()`), the following translation happens:

```
Debug: 3246 5880 riscv.c:1487 riscv_address_translate(): 0xffffffe0004a7d16 -> 0xff000000406a7d16
```

... which results in gdb still reading out all zeroes despite the translation succeeded.  Manually zeroing out the top byte of the translated address will make the read succeed with intended data.

Is there something wrong with the translation logic here?
https://github.com/riscv/riscv-openocd/blob/1449af5bdb99052c2f3bf92747f8438544cd8f3f/src/target/riscv/riscv.c#L1482-L1489

timsifive:
Fixes #452.
Also check that the high bits match the MSB of the virtual address.

Change-Id: Ib1d3d04db9ad9327ef71ea3736d5cf5d3b65b9c4

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -276,6 +276,8 @@ static enum {
 } resume_order;
 
 virt2phys_info_t sv32 = {
+	.name = "Sv32",
+	.va_bits = 32,
 	.level = 2,
 	.pte_shift = 2,
 	.vpn_shift = {12, 22},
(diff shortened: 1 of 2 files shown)


GregSavin (approved):
Makes sense to me.  Curious how this issue was exposed, was it through a failing test case in riscv-tests?

TommyMurphyTM1234:
> Makes sense to me. Curious how this issue was exposed, was it through a failing test case in riscv-tests?

Isn't it explained here how it was exposed?

https://github.com/riscv/riscv-openocd/issues/452

GregSavin:
Yes indeed it is explained in #452, thanks

timsifive:
BTW, I do have a test-case for this in riscv-tests now, but I need to polish it a little before making a PR for that.

PR #455, opened 2020-03-12
Update to 1.11 privileged spec.

timsifive:
Change-Id: I25029f7e83819464e71528fb4225b4761787793f


src/target/riscv/encoding.h
```diff
@@ -33,15 +32,18 @@
 #define SSTATUS_UPIE        0x00000010
 #define SSTATUS_SPIE        0x00000020
 #define SSTATUS_SPP         0x00000100
+#define SSTATUS_VS          0x00000600
```
GregSavin, line 35: So the vector context status position within MSTATUS is based on the latest draft of the vector spec, and not the 1.11 privileged spec?  Makes sense that the vector is more up to date on that detail, but just wanted to double check.

GregSavin (approved):

src/target/riscv/encoding.h
aswaterman: Yes, and this matches Spike.

PR #456, opened 2020-03-18
Add support for HiFive1 RevB board

sctincman:
Adds new PID (0x1051) used on board to libjaylink, and add config
and flash entry for RevB board and FE310-G002 respectively.


timsifive (approved):
Looks good. Thank you!

PR #463, opened 2020-03-25
Fix accesses for true-64bit destinations in RPC

KireinaHoro:
`mem2array` and `array2mem` are used to read and write memory from
a TCL socket, and the implementation did not support true-64bit
locations (i.e. addr >= 1 << 32) due to uint32_t.  This commit
fixes this.


timsifive:
Looks like this was already fixed upstream, and we got it.

PR #466, opened 2020-04-10
Don't propagate failure to read satp in riscv_mmu()

timsifive:
If we return failure, then the caller will think something's wrong. But
it could very well be that the hardware doesn't have SATP, in which case
we should just report that the MMU is disabled.

This fixes a bug where flashing wasn't using the target algorithm
because allocating a work area failed.

Change-Id: I16e8e660036d3f8584c0b17e842c4ec8961a8410


timsifive:
@SiFiveGregS can you review this? Carsten would like it merged.

GregSavin (approved):
Looks good to me.   I'm thinking I might've run into the associated bug once (months ago maybe), but haven't seen it lately (makes sense if it was target dependent scenario of SATP not present)

PR #467, opened 2020-04-20
Cache accesses through riscv_[sg]et_register.

timsifive:
This helps a lot with the address translation code, which checks satp
over and over again. Now satp is only read once per halt. It should also
help in a few other cases (but I don't have a good test setup to really
measure the impact).

Ran all my spike tests, and also tested against HiFive Unleashed.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3001,6 +3001,54 @@ bool riscv_has_register(struct target *target, int hartid, int regid)
 	return 1;
 }
 
+/** 
+ * If write is true:
+ *   return true iff we are guaranteed that the register will contain exactly
+ *       the value we just wrote when it's read.
+ * If write is false:
+ *   return true iff we are guaranteed that the register will read the same
+ *       value in the future as the value we just read.
+ */
+static bool gdb_regno_cacheable(enum gdb_regno regno, bool write)
+{
+	/* GPRs, FPRs, vector registers are just normal data stores. */
+	if (regno <= GDB_REGNO_XPR31 ||
+			(regno >= GDB_REGNO_FPR0 && regno <= GDB_REGNO_FPR31) ||
+			(regno >= GDB_REGNO_V0 && regno <= GDB_REGNO_V31) ||
+			regno == GDB_REGNO_PC)
+		return true;
+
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv.c
```diff
@@ -3001,6 +3001,54 @@ bool riscv_has_register(struct target *target, int hartid, int regid)
...
+static bool gdb_regno_cacheable(enum gdb_regno regno, bool write)
+{
+	/* GPRs, FPRs, vector registers are just normal data stores. */
+	if (regno <= GDB_REGNO_XPR31 ||
+			(regno >= GDB_REGNO_FPR0 && regno <= GDB_REGNO_FPR31) ||
+			(regno >= GDB_REGNO_V0 && regno <= GDB_REGNO_V31) ||
+			regno == GDB_REGNO_PC)
+		return true;
+
+	/* Most CSRs won't change value on us, but we can't assume it about rbitrary
```
GregSavin, line 3021: typo?  ("arbitrary"?)

src/target/riscv/riscv.c
```diff
@@ -3001,6 +3001,54 @@ bool riscv_has_register(struct target *target, int hartid, int regid)
...
+		case GDB_REGNO_MEPC:
+		case GDB_REGNO_MCAUSE:
+		case GDB_REGNO_SATP:
+			/*
+			 * WARL registers might not contain the value we just wrote, but
+			 * these ones won't spontaneously change their value either. *
+			 */
+			return !write;
+
+		case GDB_REGNO_TSELECT:	/* I think this should be above, but then it doesn't work. */
```
GregSavin, line 3044: So there are cases where tselect will spontaneously change value?   (Just curious, clearly there's no problem in not caching it).

GregSavin (approved):

src/target/riscv/riscv.c
timsifive: I'll fix this some other time, in the interest of merging this ASAP for Kevin.

src/target/riscv/riscv.c
timsifive: No. I don't quite know what happened, but not caching it made triggers work again.

PR #472, opened 2020-05-05
Add awareness of halt group cause.

timsifive:
Change-Id: I7f7b967ccaa3d1ff05a7e7d0c2a7ba4fa7d68ac0


src/target/riscv/riscv-013.c
```diff
@@ -3878,6 +3879,8 @@ static enum riscv_halt_reason riscv013_halt_reason(struct target *target)
 	case CSR_DCSR_CAUSE_DEBUGINT:
 	case CSR_DCSR_CAUSE_HALT:
 		return RISCV_HALT_INTERRUPT;
+	case CSR_DCSR_CAUSE_GROUP:
+		return RISCV_HALT_GROUP;
 	}
 
 	LOG_ERROR("Unknown DCSR cause field: %x", (int)get_field(dcsr, CSR_DCSR_CAUSE));
```
GregSavin, line 3886: So I guess if this case was encountered previously, the code was reflecting RISCV_HALT_UNKNOWN

GregSavin (approved):

src/target/riscv/riscv-013.c
timsifive: Yes, and you'd also get an error message to the console.

PR #473, opened 2020-05-05
Don't cache PC, but do cache DPC.

timsifive:
This fixes a bug where we read PC and marked it cached without actually
updating the cached value. The DPC value was correctly marked as valid
and updated.

Change-Id: Id6d3e94a96b981688b06f7f4a998019f2c02f6f5


GregSavin (approved):
Makes sense, DPC is a stable/consistent snapshot when in RISC-V debug mode; PC at a particular moment in debug mode is implementation specific from an ISA perspective anyway?

timsifive:
Yeah. It's a little weird but the code path to read the PC doesn't stuff the PC in the cache, and it seemed nicer not to have to add that. There isn't really a performance hit to translating PC to DPC and then hitting the cache, so I went with this solution.

PR #476, opened 2020-05-13
Improvements for the HiFive1 revB

alistair23:


githubquestions:
@alistair23 Could you add `ATTRS{idVendor}=="1366", ATTRS{idProduct}=="1061", MODE="660", GROUP="plugdev", TAG+="uaccess"` to `contrib/60-openocd.rules`? 

Full disclosure: not affiliated with this project, but documented [similar resolution](https://github.com/riscv/riscv-openocd/issues/471) when trying to work with Tock on HiFive1 revB.

alistair23:
I have sent a patch to upstream OpenOCD with that change, which will propagate down to systems. I didn't make the change in this fork as I don't think it will usually be installed on systems.

I have added it in case it can help. I have sent the same changes upstream as well, so you can use upstream OpenOCD.

timsifive (approved):
This looks reasonable enough. I assume you've used these changes with an actual board.

alistair23:
Yes, I have used it with my Hifive1 revB board.

timsifive:
Note that upstream and this branch have large differences in RISC-V support (and some target-independent changes). I occasionally spend time bringing the two branches closer together, but getting changes accepted upstream has been slow in the past.

In general I prefer everybody who uses RISC-V to use this branch, and any RISC-V-related changes to be submitted in this branch. At least until we get the two to be relatively close.

alistair23:
Yes, I did notice that the RISC-V fork has unfortunately diverged a lot. Generally upstream OpenOCD supports RISC-V well enough, although it seems there are some more advanced features on this fork.

I think it will usually be the other way around. Upstream OpenOCD is packaged by every major distro and is commonly included with other SDKs. I think most people will end up using upstream, unless they go hunting for this fork or get it from some SDK.

Preferring that people change the current OpenOCD to a fork when they do RISC-V work seems a little ambitious. Lots of people are probably happily using upstream unaware there is a fork. In which case most work will happen on upstream, and it will be up to someone to rebase/merge them to this fork.

I sent a similar patch set (but using some of the newer config options) upstream, so that will work for the HiFive1 revB in the future. Hopefully some of the other RISC-V features can be upstremed as well.

PR #477, opened 2020-05-15
Speed up SBA block reads roughly 2x.

timsifive:
Change-Id: I4e4f5530d4abae7470fd00308361e727904367d2


GregSavin (approved):
So it looks like this is faster because the previous code was using ensure_success=true on the scan for each word (forcing 2 DMI scans for every word), and the new code can get away with ensure_success=false and get the same data back with one scan per word in the best case?

timsifive:
That's exactly right.

PR #478, opened 2020-05-16
Fix semihosting for multicore targets

timsifive:
Now it works in multiple gdb scenarios, as well as with `-rtos hwthread`. It does not work with `-rtos riscv`, which is deprecated in any case.

Moved the automatic resume out of riscv_semihosting, because when multiple harts are halted we sometimes want to resume them all (when they're all part of a halt group halted because of a semihosting breakpoint) and sometimes now (when one hart hit a semihosting breakpoint and another hit a regular breakpoint).

### DIFF
--- a/src/target/riscv/riscv_semihosting.c
+++ b/src/target/riscv/riscv_semihosting.c
@@ -60,35 +60,35 @@ void riscv_semihosting_init(struct target *target)
 /**
  * Check for and process a semihosting request using the ARM protocol). This
  * is meant to be called when the target is stopped due to a debug mode entry.
- * If the value 0 is returned then there was nothing to process. A non-zero
- * return value signifies that a request was processed and the target resumed,
- * or an error was encountered, in which case the caller must return
- * immediately.
  *
  * @param target Pointer to the target to process.
  * @param retval Pointer to a location where the return code will be stored
  * @return non-zero value if a request was processed or an error encountered
  */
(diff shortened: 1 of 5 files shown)


src/target/riscv/riscv_semihosting.c
```diff
@@ -60,35 +60,35 @@ void riscv_semihosting_init(struct target *target)
...
 
 	/* Read the current instruction, including the bracketing */
-	*retval = target_read_memory(target, dpc - 4, 2, 6, tmp);
+	*retval = target_read_memory(target, pc - 4, 2, 6, tmp);
 	if (*retval != ERROR_OK)
-		return 0;
+		return SEMI_ERROR;
 
 	/*
 	 * The instructions that trigger a semihosting call,
```
GregSavin, line 94: Interesting, so those instructions surrounding the ebreak are particular flavors of no-op, being used solely as a signal to the debugger that this ebreak isn't a normal breakpoint.  Is this set of conventions documented in a RISC-V spec or elsewhere?

GregSavin (approved):

src/target/riscv/riscv_semihosting.c
TommyMurphyTM1234: It's discussed here and maybe also on the RISC-V mailing lists...

https://forums.sifive.com/t/openocd-semihosting/748

PR #479, opened 2020-05-20
Don't use MMU in M mode - https://github.com/riscv/riscv-openocd/issu…

TommyMurphyTM1234:
Fix for https://github.com/riscv/riscv-openocd/issues/474

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1383,15 +1383,26 @@ static int riscv_mmu(struct target *target, int *enabled)
 	if (riscv_rtos_enabled(target))
 		riscv_set_current_hartid(target, target->rtos->current_thread - 1);
 
-	riscv_reg_t value;
-	if (riscv_get_register(target, &value, GDB_REGNO_SATP) != ERROR_OK) {
+	/* Don't use MMU in explicit or effective M (machine) mode */
+	riscv_reg_t priv;
+	riscv_reg_t mstatus;
+	if ((riscv_get_register(target, &priv, GDB_REGNO_PRIV) == ERROR_OK) &&
+	    (riscv_get_register(target, &mstatus, GDB_REGNO_MSTATUS) == ERROR_OK) &&
+	    ((get_field(mstatus, MSTATUS_MPRV) ? get_field(mstatus, MSTATUS_MPP) : priv) == PRV_M)) {
+		LOG_DEBUG("SATP/MMU ignored in M mode.");
+		*enabled = 0;
+		return ERROR_OK;
+	}
+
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv.c
```diff
@@ -1383,15 +1383,26 @@ static int riscv_mmu(struct target *target, int *enabled)
 	if (riscv_rtos_enabled(target))
 		riscv_set_current_hartid(target, target->rtos->current_thread - 1);
 
-	riscv_reg_t value;
-	if (riscv_get_register(target, &value, GDB_REGNO_SATP) != ERROR_OK) {
+	/* Don't use MMU in explicit or effective M (machine) mode */
+	riscv_reg_t priv;
+	riscv_reg_t mstatus;
+	if ((riscv_get_register(target, &priv, GDB_REGNO_PRIV) == ERROR_OK) &&
+	    (riscv_get_register(target, &mstatus, GDB_REGNO_MSTATUS) == ERROR_OK) &&
```
timsifive: priv (part of DCSR) and mstatus are both required to be implemented, so if we can't read them I think we should return error here.

src/target/riscv/riscv.c
```diff
@@ -1383,15 +1383,26 @@ static int riscv_mmu(struct target *target, int *enabled)
...
+	    ((get_field(mstatus, MSTATUS_MPRV) ? get_field(mstatus, MSTATUS_MPP) : priv) == PRV_M)) {
+		LOG_DEBUG("SATP/MMU ignored in M mode.");
```
timsifive: Can you add the value of mstatus to this error message? That way I'll at least have something to go on when somebody reports they got this error message for no good reason.

timsifive (changes requested):
Thanks for addressing this, Tommy.

What testing did you do?

timsifive:
As to your performance impact questions. There are two more register reads, but both dcsr and mstatus reads should be cached. You should be able to confirm this easily by looking at the debug log, because riscv_mmu() is called every time OpenOCD starts any memory access.

TommyMurphyTM1234:
Thanks for the comments and feedback @timsifive.
I will address the issues that you mention above and resubmit it.
I have only tested in M mode so not when the target is in any lower privilege mode - explicitly or via mstatus.mprv/mpp.

timsifive (approved):
Thanks, Tommy! I'll merge this next week.

timsifive:
I just confirmed that all riscv-tests/debug still pass.

TommyMurphyTM1234:
Thanks a lot @timsifive.

PR #481, opened 2020-05-27
Cherry-picking commit 25efc1506 from upstream

JanMatCodasip:
Added are two new events: TARGET_EVENT_STEP_START and
TARGET_EVENT_STEP_END. These are needed for cache invalidation
prior to step/resume on cores that do not implement the Debug
Program Buffer.

Original commit in the upstream:
http://openocd.zylin.com/#/c/5551/

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -4849,6 +4849,10 @@ when reset disables PLLs needed to use a fast clock.
 @* After all targets have resumed
 @item @b{resumed}
 @* Target has resumed
+@item @b{step-start}
+@* Before a target is single-stepped
+@item @b{step-end}
+@* After single-step has completed
 @item @b{trace-config}
 @* After target hardware trace configuration was changed
 @end itemize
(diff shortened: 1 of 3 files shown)


JanMatCodasip:
I would like to add this commit from upstream also to `riscv-openocd` as it is needed for cache invalidation on cores that do not have RISC-V program buffer (e.g. SweRV cores).

Alternatively, the whole latest upstream could be merged to `riscv-openocd`, as has already been done in the past, I believe. (In that case this would be redundant and could be closed.)

Thank you, Jan

timsifive:
When do you need this? Merging upstream into our branch is working its way up on my todo list, and will probably happen in June.

JanMatCodasip:
If the upstream merge takes place in June, then I believe it is fine to do it that way. Should the merge slip to a later date, I would prefer to have this change in `riscv-openocd` before that to allow smooth use of this OpenOCD fork with SweRV cores.

In general - If another set of eyes would help when it comes to the merge, please feel free to let me know.

JanMatCodasip:
Resolved as part of https://github.com/riscv/riscv-openocd/pull/494. 

Closing this one.

PR #483, opened 2020-06-11
riscv: Avoid shadowing read_csr/write_csr macros

kraj:
The name conflict is picked by compiler and it fails to compile for rv64
Fixes
src/target/riscv/riscv-011.c:1014:44: error: too many arguments provided to function-like macro invocation
static int read_csr(struct target *target, uint64_t *value, uint32_t csr)
                                           ^
Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1011,7 +1011,7 @@ static int wait_for_state(struct target *target, enum target_state state)
 	}
 }
 
-static int read_csr(struct target *target, uint64_t *value, uint32_t csr)
+static int read_remote_csr(struct target *target, uint64_t *value, uint32_t csr)
 {
 	riscv011_info_t *info = get_info(target);
 	cache_set32(target, 0, csrr(S0, csr));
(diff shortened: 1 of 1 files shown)


TommyMurphyTM1234:
Just curious (and not questioning the issue raised) you're compiling openocd to run on rv64 to debug what target?

kraj:
I am just cross compiling openocd on x86 targeting rv64

TommyMurphyTM1234:
I understand, but why do you want to run openocd *on* RISC-V?
What sort of target will you be debugging when running openocd on a RISC-V host?

kraj:
> I understand, but why do you want to run openocd _on_ RISC-V?
> What sort of target will you be debugging when running openocd on a RISC-V host?

No I don't tend to use it on RV64 host yet but that should not be a consideration for this patch
the code is wrong and it should be fixed regardless

TommyMurphyTM1234:
Ok, fair enough.
I was just curious about how you intended to run openocd on RISC-V.
Thanks.

timsifive (changes requested):
This looks like a decent solution to the problem, but can you use read_remote_csr and write_remote_csr instead of the names you chose?

kraj:
> This looks like a decent solution to the problem, but can you use read_remote_csr and write_remote_csr instead of the names you chose?

done

timsifive (approved):
Thank you! I'll merge this tomorrow, unless somebody objects.

PR #486, opened 2020-06-18
Step/resume off manual hardware triggers

timsifive:
Manual hardware triggers are those set by the user by writing directly to tselect/tdata*, instead of using e.g. hbreak from gdb. We want to support them because the hardware can support many kinds of triggers that can't be communicated from gdb to OpenOCD.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1213,6 +1213,97 @@ int riscv_resume_prep_all_harts(struct target *target)
 	return ERROR_OK;
 }
 
+/* state must be riscv_reg_t state[RISCV_MAX_HWBPS] = {0}; */
+static int disable_triggers(struct target *target, riscv_reg_t *state)
+{
+	RISCV_INFO(r);
+
+	LOG_DEBUG("deal with triggers");
+
+	if (riscv_enumerate_triggers(target) != ERROR_OK)
+		return ERROR_FAIL;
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv.c
```diff
@@ -3563,6 +3637,17 @@ static int register_set(struct reg *reg, uint8_t *buf)
 	memcpy(reg->value, buf, DIV_ROUND_UP(reg->size, 8));
 	reg->valid = gdb_regno_cacheable(reg->number, true);
 
+	if (reg->number == GDB_REGNO_TDATA1 ||
+			reg->number == GDB_REGNO_TDATA2) {
+		r->manual_hwbp_set = true;
```
GregSavin, line 3642: So, once set to true, manual_hwbp_set stays true for the duration of the current OpenOCD session, which makes sense.  If no direct accesses to trigger registers are made in the entire session, then the original logic is used for disabling/reenabling triggers (presumably that has less overhead?).

GregSavin (approved):
Looks good to me.

src/target/riscv/riscv.c
timsifive: That's right. In theory we could be smarter by tracking what tselect was, and what value was written to tdata, but that's a lot more complicated and I'm not sure how helpful it really is.

PR #489, opened 2020-06-23
Accept dmstatus.version==3 (0.14)

### LINKED ISSUE #485: the version check for the version in  dtmcs and in dmstatus
Now the openOCD get version check info from dtmcs, but i think the version in dmstatus is what actully matrers and imply the actual feature（in spec） the DM supporting.
So what do we need to do to fully use both of the versions(in  dtmcs and in dmstatus) ?

timsifive:
Fixes #485.

Change-Id: I60b3d68827ca726558bc28035c0b74c5cf0d9754


GregSavin (approved):
Reminds me to read 0.14 at some point

PR #494, opened 2020-06-30
Get changes from upstream

timsifive:
There seem to be no regressions in tests against spike, HiFive1, and HiFive Unleashed.
Unfortunately the tests against the 2 hardware platforms are not clean right now, but that doesn't seem to be affected by this change.


GregSavin (approved):
Looks like most of the riscv specific changes were essentially lint adjustments, code formatting, etc.

PR #495, opened 2020-07-01
doc: fix missing parameter in dmi_read

yunowo:
The description says `dmi_read` command needs an `address` parameter, but it's missing in the command syntax.


timsifive (approved):
Thank you for catching and fixing this.

PR #497, opened 2020-07-08
Update jep106.inc using perl script and JEP106BB.pdf

jonathankimmitt:
The JEDEC manufacturers ID list in jep106.inc has been automatically updated based on JEDEC publication JEP106BB.pdf


timsifive:
Is this still required? Do we need it in the RISC-V fork or is it more something to go into upstream?

jonathankimmitt:
I think new vendor lists are released every six months so this patch is obsolete now, and as you say, is not RISCV specific so should be upstreamed.

Get Outlook for iOS<https://aka.ms/o0ukef>

PR #498, opened 2020-07-11
Enable remote_bitbang by default

Ma124:
RBB is the protocol supported by [spike](https://github.com/riscv/riscv-isa-sim#readme) so it would make sense to include it by default.

> To get remote bitbang support in OpenOCD, you need to run configure with `--enable-remote-bitbang` before building.
> I'd accept a change https://github.com/riscv/riscv-openocd that makes that the default.

_Originally posted by @timsifive in https://github.com/riscv/riscv-isa-sim/issues/102#issuecomment-656910235_


timsifive (approved):
Thank you.

I'll merge this tomorrow, unless somebody pipes up before then.

PR #505, opened 2020-08-10
Update to version 1.0 of the vector spec.

timsifive:
Accessing registers on targets that implement 0.9 or earlier will no
longer work. If you need that we can talk about making it a config
option.

Change-Id: I953b639cf9a92ee9b0422e035da57c1d07504237


GregSavin (approved):
gotta like one-line diffs

Issue #445, opened 2020-02-05
What is general way to access CPU internal RAMs?


agrobman:
If a CPU complex includes internal RAMs,  which may be configurable in size and base addresses, how does openOCD supposed to access these?

There are two ways to access memories - System Bus commands and abstract memory commands,
How does openOCD select which one to use?

Related topic : access size. Some memories ( internal or external) may be limited to specific access size,
for ex only accessed with 4 bytes (word),  how does openOCD "know" that? 
Is there way for openOCD "discover" these limitations?

If a CPU doesn't provide "program buffer",  how will openOCD set SW breakpoint if the CPU has instruction cache?

 Is there defined common way for debugger to invalidate caches, without running instructions or have specific knowledge about cache control registers?

How does openOCD discovers what is implemented in processor DM module?
Is the discovery done while CPU is in reset/halted/running?

Is there any configuration file, which can be read by the openOCD at startup to provide processor implementation details, including privet CSRs internal memories, features implemented, etc ?

timsifive:
> If a CPU complex includes internal RAMs, which may be configurable in size and base addresses, how does openOCD supposed to access these?

The same way all memory is accessed.

> There are two ways to access memories - System Bus commands and abstract memory commands,
> How does openOCD select which one to use?

Defaults to program buffer, uses System Bus if that doesn't work or if prefer_sba is set, and if all that fails then uses abstract.

> Related topic : access size. Some memories ( internal or external) may be limited to specific access size,
> for ex only accessed with 4 bytes (word), how does openOCD "know" that?
> Is there way for openOCD "discover" these limitations?

Not AFAIK.

> If a CPU doesn't provide "program buffer", how will openOCD set SW breakpoint if the CPU has instruction cache?

It will write the memory and not do anything else. The Debug Spec currently provides no other mechanism. If you need that, bring it up on https://lists.riscv.org/g/tech-debug/topics

> Is there defined common way for debugger to invalidate caches, without running instructions or have specific knowledge about cache control registers?

No.

> How does openOCD discovers what is implemented in processor DM module?

See examine() in src/target/riscv/riscv-013.c.

> Is the discovery done while CPU is in reset/halted/running?

OpenOCD always briefly halts every hart when first connecting.

> Is there any configuration file, which can be read by the openOCD at startup to provide processor implementation details, including privet CSRs internal memories, features implemented, etc ?

[...]

PR #508, opened 2020-08-11
Selection of memory access methods, aampostincrement detection

sobuch:
New configuration command introduced: "riscv set_mem_access". It allows to specify which RISC-V memory access methods (progbuf, sysbus and/or abstract access) should be tried and in which order of priority.

Command "riscv set_prefer_sba" is left and works in backward compatible way, but is marked as deprecated.

Support for the optional aampostincrement bit is detected the first time abstract memory access is used. This enables the use of OpenOCD with HW without support for aampostincrement.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -215,6 +215,8 @@ typedef struct {
 	bool abstract_read_fpr_supported;
 	bool abstract_write_fpr_supported;
 
+	yes_no_maybe_t has_aampostincrement;
+
 	/* When a function returns some error due to a failure indicated by the
 	 * target in cmderr, the caller can look here to see what that error was.
 	 * (Compare with errno.) */
@@ -1842,27 +1844,40 @@ static int riscv013_hart_count(struct target *target)
 	return dm->hart_count;
 }
 
+/* Try to find out the widest memory access size depending on the selected memory access methods. */
 static unsigned riscv013_data_bits(struct target *target)
 {
 	RISCV013_INFO(info);
-	/* TODO: Once there is a spec for discovering abstract commands, we can
-	 * take those into account as well.  For now we assume abstract commands
-	 * support XLEN-wide accesses. */
-	if (has_sufficient_progbuf(target, 3) && !riscv_prefer_sba)
-		return riscv_xlen(target);
+	RISCV_INFO(r);
 
-	if (get_field(info->sbcs, DM_SBCS_SBACCESS128))
-		return 128;
-	if (get_field(info->sbcs, DM_SBCS_SBACCESS64))
-		return 64;
-	if (get_field(info->sbcs, DM_SBCS_SBACCESS32))
-		return 32;
-	if (get_field(info->sbcs, DM_SBCS_SBACCESS16))
-		return 16;
-	if (get_field(info->sbcs, DM_SBCS_SBACCESS8))
-		return 8;
+	for (unsigned i = 0; i < RISCV_NUM_MEM_ACCESS_METHODS; i++) {
+		int method = r->mem_access_methods[i];
 
-	return riscv_xlen(target);
+		if (method == RISCV_MEM_ACCESS_PROGBUF) {
+			if (has_sufficient_progbuf(target, 3))
+				return riscv_xlen(target);
+		} else if (method == RISCV_MEM_ACCESS_SYSBUS) {
+			if (get_field(info->sbcs, DM_SBCS_SBACCESS128))
+				return 128;
+			if (get_field(info->sbcs, DM_SBCS_SBACCESS64))
+				return 64;
+			if (get_field(info->sbcs, DM_SBCS_SBACCESS32))
+				return 32;
+			if (get_field(info->sbcs, DM_SBCS_SBACCESS16))
+				return 16;
+			if (get_field(info->sbcs, DM_SBCS_SBACCESS8))
+				return 8;
+		} else if (method == RISCV_MEM_ACCESS_ABSTRACT) {
+			/* TODO: Once there is a spec for discovering abstract commands, we can
+			 * take those into account as well.  For now we assume abstract commands
+			 * support XLEN-wide accesses. */
+			return riscv_xlen(target);
+		} else if (method == RISCV_MEM_ACCESS_UNSPECIFIED)
+			/* No further mem access method to try. */
+			break;
+	}
+	LOG_ERROR("Unable to determine supported data bits on this target. Assuming 32 bits.");
+	return 32;
 }
 
 static int prep_for_vector_access(struct target *target, uint64_t *vtype,
@@ -2069,6 +2084,8 @@ static int init_target(struct command_context *cmd_ctx,
 	info->abstract_write_csr_supported = true;
 	info->abstract_read_fpr_supported = true;
 	info->abstract_write_fpr_supported = true;
+
+	info->has_aampostincrement = YNM_MAYBE;
 
 	return ERROR_OK;
 }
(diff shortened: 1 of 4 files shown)


sobuch:
Unsupported postincrement was tested on SweRV EH1 1.7, example of unsupported but "buggy" postincrement (address is not incremented but abstract command doesn't fail) is SweRV EH1 1.6. Help with testing with fully supported postincrement would be appreciated.

sobuch:
What is the expected behavior when multiple memory access methods are available and first one fails? Currently after first failure requested read/write fails as well. However, here https://github.com/riscv/riscv-openocd/issues/445#issuecomment-582595208 seems to be described a different behavior - that after a failure another method is tried, if possible. We implemented the second version, as it seems to be an improvement. Would this change of behavior be acceptable?

JanMatCodasip:
To further add on & rephrase what @sobuch already mentioned, this pull request deals with two items:

* Selection of memory access methods - merely an extension of current command `riscv set_prefer_sba`. It is motivated by SweRV cores, for which it is needed to have the option to use abstract memory access with priority over system bus (and that is not possible in current OpenOCD).

* Detection of whether `aampostincrement` bit is supported by the target - already discussed in https://github.com/riscv/riscv-debug-spec/issues/531

Any comments are appreciated. Most importantly, the answer to @sobuch's question above - how to treat multiple memory access methods?

* If a memory access method is enabled and supported (e.g. access via program buffer & the progbuf is large enough), but the access fails, what is the preferred course of action?
  * Should we try the next memory access method on the list? (this is what this branch does, and what we believe was mentioned in #445)
  * Or should we give up and report an error? (this is what the current OpenOCD code does)

Thank you.

JanMatCodasip:
To further add on & rephrase what @sobuch already mentioned, this pull request deals with two items:

* Selection of memory access methods - merely an extension of current command `riscv set_prefer_sba`. It is motivated by SweRV cores, for which it is needed to have the option to use abstract memory access with priority over system bus (and that is not possible in current OpenOCD).

* Detection of whether `aampostincrement` bit is supported by the target - already discussed in https://github.com/riscv/riscv-debug-spec/issues/531

Any comments are appreciated. Most importantly, the answer to @sobuch's question above - how to treat multiple memory access methods?

* If a memory access method is enabled and supported (e.g. access via program buffer & the progbuf is large enough), but the access fails, what is the preferred course of action?
  * Should we try the next memory access method on the list? (this is what this branch does, and what is mentioned in #445)
  * Or should we give up and report an error? (this is what the current OpenOCD code does)

Thank you.

timsifive:
There are different reasons an access might fail:
1. The access may not be supported depending on the current target state. E.g. progbuf accesses won't work while the target is running. In that case we'd want to fall back to the next option.
2. The access might not work for that particular region of memory. E.g. an abstract access might fail because the current hart doesn't have permission to access the memory, but an sba access would succeed. Falling back would be helpful there also.
3. There may be some other problem that indicates a real issue, and in that case we should probably abort.

To handle the first problem, I'd introduce a new ERROR_ code (if necessary) that can be returned to indicate that this access mode won't work right now (but might at a later time).

To handle the second, the real solution is to let the user select the access method for different address regions. That's getting more complex, and I'm not sure if it would see much use. The downside of not doing that is now every time there's an access to a certain reason we first try a method that won't work before falling back on one that does. This PR is already large enough. Even if you do want to implement that, let's do it in a separate PR.

Differentiating between the 2nd and 3rd might also be tricky, so let's handle them together as follows:
When an access fails, log a warning and fall back to the next method. If all enabled methods fail, log an error and fail. To prevent tons of warnings in case 2 above, track a flag for each access method to only print the warning once per access method. The flags are reset when the new `set_mem_access` command is used.

How does that sound?

sobuch:
We have something (kind of) similar to the real solution for second problem implemented in https://github.com/Codasip/swerv-openocd, but I don't think we currently have the resources to do that properly.

Could you please explain what's the added value of new error code? it seems to me that if we won't differentiate between the last two kinds now, we may just as well process the first one the same way.

Some flags to prevent warnings definitely seem like a good idea.

JanMatCodasip:
@timsifive - Tim, thank you for your reply.

As for handling of memory access failures - 

> To handle the first problem, I'd introduce a new ERROR_ code (if necessary) that can be returned to indicate that this access mode won't work right now (but might at a later time).

For progbuf accesses, this is easy - we can skip this access method if the target is running. We'll add this to the merge request.

For abstract memory access, this is more complex - it may or may not be supported while the target runs. In order to detect what the target supports, we would have to try the abstract command, and then possibly remember the outcome. This would be similar logic to what we've implemented here for the "aampostincrement" bit. For simplicity & robustness, I recommend not to add further intelligence that would try to remember what abstract commands are supported under what circumstances, at least not for now. This can be further enhanced in the future, if needed. 

So our proposal for abstract memory commands is: simply try them every time, as it is in the current trunk code. 

> Differentiating between the 2nd and 3rd might also be tricky, so let's handle them together as follows: When an access fails, log a warning and fall back to the next method. If all enabled methods fail, log an error and fail.  (....)

This sounds very fine. We will adjust the code this way.

> To handle the second, the real solution is to let the user select the access method for different address regions. That's getting more complex, and I'm not sure if it would see much use.

We're also not sure about the value of different memory regions in upstream OpenOCD. Either way, we should be fine even if the need arises at some point, because the current implementation can be extended in a backward-compabible manner - e.g. command `set_mem_access_region` (or something along those lines) would be introduced to specify custom methods for a particular range of addresses. `set_mem_access` would remain in place to specify the defaults - the methods to use outside of any defined region.


We will adjust the code within the next week and will reach back when done.

Thanks and regards,
Jan

timsifive:
> I recommend not to add further intelligence that would try to remember what abstract commands are supported under what circumstances, at least not for now. This can be further enhanced in the future, if needed.

That sounds good.

Thanks for the update.

sobuch:
Hi @timsifive, we finished the implementation as specified here earlier. Riscv debug tests are passing, manual testing now also doesn't show any further problems. We look forward to any feedback from you, please let us known if there is anything more we should do before this pull request can be closed. Otherwise, this branch is rebased and changes squashed, so from our side it's ready to be merged :)

src/target/riscv/riscv-013.c
```diff
@@ -98,6 +98,9 @@ static int	riscv013_test_compliance(struct target *target);
 
 #define RISCV013_INFO(r) riscv013_info_t *r = get_info(target)
 
+#define LOG_WARNING_OR_DEBUG(use_warn, expr ...) \
```
timsifive: You only use this macro twice, and in both places you have almost the same logic to generate the string.

Can you make this a regular function, which is only used for memory accesses, and takes arguments for read/write (bool read), bool success, and method?

src/target/riscv/riscv-013.c
```diff
@@ -1831,27 +1836,41 @@ static int riscv013_hart_count(struct target *target)
...
+		} else if (method == RISCV_MEM_ACCESS_ABSTRACT) {
+			/* TODO: Once there is a spec for discovering abstract commands, we can
+			 * take those into account as well.  For now we assume abstract commands
+			 * support XLEN-wide accesses. */
+			return riscv_xlen(target);
+		} else if (method == RISCV_MEM_ACCESS_UNSPECIFIED)
+			/* No further mem access method to try. */
+			break;
+	}
+	LOG_DEBUG("Unable to determine supported data bits on this target. Assuming 32 bits.");
```
timsifive: Should this be an error? I don't see how we get here on a normal system.

src/target/riscv/riscv-013.c
```diff
@@ -2708,16 +2721,39 @@ static int read_memory_abstract(struct target *target, target_addr_t address,
...
+
+		if (info->has_aampostincrement == YNM_MAYBE) {
+			if (result == ERROR_OK) {
+				/* Safety: double-check that the address was really auto-incremented */
+				riscv_reg_t new_address = read_abstract_arg(target, 1, riscv_xlen(target));
+				if (new_address == address + size) {
+					LOG_DEBUG("aampostincrement is supported on this target.");
+					info->has_aampostincrement = YNM_YES;
+				} else {
+					LOG_DEBUG("Buggy aampostincrement! Address not incremented correctly.");
```
timsifive: A warning would be more appropriate. Especially since the user will only see it once per session (right?).

src/target/riscv/riscv-013.c
```diff
@@ -2708,16 +2721,39 @@ static int read_memory_abstract(struct target *target, target_addr_t address,
...
+				command = access_memory_command(target, false, width, false, false);
+				result = execute_abstract_command(target, command);
+				if (result == ERROR_OK) {
+					LOG_DEBUG("aampostincrement is not supported on this target.");
+					info->has_aampostincrement = YNM_NO;
+				}
+			}
+		}
+
 		if (result != ERROR_OK) {
```
timsifive: Remove the braces, per OpenOCD style guide. I'm surprised checkpatch didn't fail the build for this.

src/target/riscv/riscv-013.c
```diff
@@ -3219,30 +3268,137 @@ static int read_memory(struct target *target, target_addr_t address,
...
+				LOG_DEBUG("Skipping mem read via progbuf - unsupported size.");
+				progbuf_result = "skipped";
+				continue;
+			}
+			if ((sizeof(address) * 8 > riscv_xlen(target)) && (address >> riscv_xlen(target))) {
+				LOG_DEBUG("Skipping mem write via progbuf - progbuf only supports %u-bit address.",
+						riscv_xlen(target));
+				progbuf_result = "skipped";
+				continue;
+			}
```
timsifive: This looks identical to the code in write_memory_progbuf (except I suppose the read/write strings). Can you refactor this?

src/target/riscv/riscv-013.c
```diff
@@ -3219,30 +3268,137 @@ static int read_memory(struct target *target, target_addr_t address,
...
+				LOG_DEBUG("Skipping mem read via system bus - sba only supports %u-bit address.", sbasize);
+				sysbus_result = "skipped";
+				continue;
+			}
+			if (increment != size && (get_field(info->sbcs, DM_SBCS_SBVERSION) == 0 || increment != 0)) {
+				LOG_DEBUG("Skipping mem read via system bus - "
+						"sba reads only support size==increment or also size==0 for sba v1.");
+				sysbus_result = "skipped";
+				continue;
+			}
```
timsifive: Same.

src/target/riscv/riscv-013.c
```diff
@@ -3219,30 +3268,137 @@ static int read_memory(struct target *target, target_addr_t address,
...
+						riscv_xlen(target));
+				abstract_result = "skipped";
+				continue;
+			}
+			if (size != increment) {
+				LOG_ERROR("Skipping mem read via abstract access - "
+						"abstract command reads only support size==increment.");
+				abstract_result = "skipped";
+				continue;
+			}
```
timsifive: Same again.

src/target/riscv/riscv-013.c
```diff
@@ -3219,30 +3268,137 @@ static int read_memory(struct target *target, target_addr_t address,
...
+	char *sysbus_result = "disabled";
+	char *abstract_result = "disabled";
+
+	for (i = 0; i < RISCV_NUM_MEM_ACCESS_METHODS; i++) {
+		int method = r->mem_access_methods[i];
+
+		if (method == RISCV_MEM_ACCESS_PROGBUF) {
+			if (!has_sufficient_progbuf(target, 3)) {
+				LOG_DEBUG("Skipping mem read via progbuf - insufficient progbuf size.");
+				progbuf_result = "skipped";
```
timsifive: It looks like you're just tracking this string for the error messages. Can you give unique reasons instead of "skipped" for each one? E.g. "skipped (insufficient progbuf size)" etc.

src/target/riscv/riscv.c
```diff
@@ -2287,11 +2286,85 @@ COMMAND_HANDLER(riscv_test_compliance) {
 
 COMMAND_HANDLER(riscv_set_prefer_sba)
 {
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+	bool prefer_sba;
+	LOG_WARNING("`riscv set_prefer_sba` is deprecated. Please use `riscv set_mem_access` instead.");
```
timsifive, line 2292: Thanks for making this backwards compatible.

src/target/riscv/riscv.c
```diff
@@ -2287,11 +2286,85 @@ COMMAND_HANDLER(riscv_test_compliance) {
...
+	r->mem_access_abstract_warn = true;
+
+	return ERROR_OK;
+}
+
+COMMAND_HANDLER(riscv_set_mem_access)
+{
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+	unsigned int i;
```
timsifive: You're only using `i` inside loops. Can you declare it in each for loop instead? E.g.:
	for (unsigned int i = 0; i < CMD_ARGC; i++) {

timsifive (changes requested):
This structure looks great overall. I have some stylistic comments, but I like this design.

Also, please document the new command alongside the other riscv commands in https://github.com/riscv/riscv-openocd/blob/riscv/doc/openocd.texi

sobuch:
Thank you for looking into this and for the feedback. Looking at some of the read/write functions, probably more extensive refactor is in order. I tried at least not make it worse with this change, your suggestions were very helpful in this.

Otherwise, same as before - rebased, debug tests passing, checkpatch also passing

src/target/riscv/riscv-013.c
```diff
@@ -3234,30 +3413,65 @@ static int read_memory(struct target *target, target_addr_t address,
...
+	char *sysbus_result = "disabled";
+	char *abstract_result = "disabled";
+
+	for (unsigned i = 0; i < RISCV_NUM_MEM_ACCESS_METHODS; i++) {
+		int method = r->mem_access_methods[i];
+
+		if (method == RISCV_MEM_ACCESS_PROGBUF) {
+			bool skip = mem_should_skip_progbuf(target, address, size, true, &progbuf_result);
+			if (skip)
+				continue;
```
timsifive: The skip variable isn't used outside of here, right?

I'd just write:
```
if (mem_should_skip...)
   continue;
```

timsifive (dismissed):
This looks good. Thank you. I just have one last request below.

Also, instead of rebase+force push, can you `merge riscv` and regular push? That way when I'm reviewing code I can easily see just the changes you made since the last time I read it.

src/target/riscv/riscv-013.c
sobuch: of course, I missed this

sobuch:
Thanks for the review. Next time I won't rebase, if you prefer it that way. Now I just added a new commit.

timsifive (approved):
I'm happy with this. Thank you. I'll merge it tomorrow unless there are more comments.

PR #509, opened 2020-08-12
Account for impebreak in size requirements for progbuf

sobuch:
Although in several places impebreak is considered, in others it seems to be forgotten. With the latest spike (0ebb8a93) you can check that the current implementation is not correct. With combination of options --dm-no-impebreak and --dm-progsize=2 connected OpenOCD will try to access memory using program buffer, but this will produce error due to missing impebreak or not enough space in the buffer.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1285,6 +1285,14 @@ static unsigned register_size(struct target *target, unsigned number)
 		return riscv_xlen(target);
 }
 
+static bool has_sufficient_progbuf(struct target *target, unsigned size)
+{
+	RISCV013_INFO(info);
+	RISCV_INFO(r);
+
+	return info->progbufsize + r->impebreak >= size ;
+}
+
 /**
  * Immediately write the new value to the requested register. This mechanism
  * bypasses any caches.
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1830,10 +1830,11 @@ static int riscv013_hart_count(struct target *target)
 static unsigned riscv013_data_bits(struct target *target)
 {
 	RISCV013_INFO(info);
+	RISCV_INFO(r);
 	/* TODO: Once there is a spec for discovering abstract commands, we can
 	 * take those into account as well.  For now we assume abstract commands
 	 * support XLEN-wide accesses. */
-	if (info->progbufsize >= 2 && !riscv_prefer_sba)
+	if (info->progbufsize + r->impebreak >= 3 && !riscv_prefer_sba)
```
JanMatCodasip: @sobuch - Optionally, if you'd like, you can factor this out to a helper function to make the code even clearer:

```
if (has_sufficient_progbuf(target, 2) && !riscv_prefer_sba)
```

JanMatCodasip: Other than that, it looks fine.

timsifive (dismissed):
This is OK as is, but if you could implement @JanMatCodasip's suggestion that would make it really nice.

Do you have time to do that?

sobuch:
Done. Should be ok, debug tests are still passing.

timsifive (approved):
Looks great! Thank you.

I'll merge this on Monday, unless there is more discussion.

PR #510, opened 2020-08-14
Create `riscv repeat_read` command

timsifive:
SiFive uses to quickly read out an on-chip trace buffer, but there are probably other uses for this functionality. They probably all involve reading out a buffer that is memory-mapped at a single address.

Currently not supported using abstract commands. The bulk of the change is to rewrite `read_memory_progbuf_inner` to use a counter to track progress, instead of the address.


GregSavin (commented):
So, previously, it was implicit that increment==size?   And now increment is explicit, and it can be supplied as zero in some cases (SBA?)?   Or you can supply increment of zero for any progbuf-based memory read?

GregSavin:
OK, never mind, I see now.

GregSavin (dismissed):
Always fascinating to see the ripple effects that needed to be dealt with (e.g. the omission of address field in "md" output).  Looks good to me, though.

timsifive:
Thanks, @SiFiveGregS. Can you rereview? I had to resolve a conflict.

GregSavin (approved):

PR #515, opened 2020-08-24
Update debug_defines.h from riscv-debug-spec

timsifive:
A ton of constants got a new prefix, so I made a lot of changes to
match, but no functional changes.

I did define DTM_DMI_MAX_ADDRESS_LENGTH	in batch.c. That definition
never should have been in debug_defines.h, which I missed during code
review.

Change-Id: If5d86660f84bb0a3f2865fb36ef05d6630486d8b


GregSavin (approved):
OK so those architectural elements now prefixed with DM_ were always associated with the debug module rather than the bus used to access it from JTAG?  Maybe a "someday we should fix those prefixes" kind of situation, and then "someday" ended up arriving?

timsifive:
Yeah. I'm sure this was changed in the spec a long time ago, but I never bothered to regenerate the file.

PR #517, opened 2020-08-28
Check malloc/calloc return values.

timsifive:
Should not affect anything, but requested in
http://openocd.zylin.com/#/c/5821/5/src/target/riscv/batch.c@28

Change-Id: Ib7185bd93eeb918e72872416ab6364f8776cff88


GregSavin (dismissed):
One of those rare cases where judicious usage of "goto" can result in cleaner code than other options.

GregSavin (approved):

PR #521, opened 2020-09-01
Fix usage for our RISC-V commands.

timsifive:
Change-Id: Ia4e020637aae2384af223b0210ef2aef6a14b31a

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2708,35 +2708,34 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.name = "repeat_read",
 		.handler = handle_repeat_read,
 		.mode = COMMAND_ANY,
-		.usage = "riscv repeat_read count address [size=4]",
+		.usage = "count address [size=4]",
 		.help = "Repeatedly read the value at address."
 	},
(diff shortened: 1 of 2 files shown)


GregSavin (approved):

src/target/riscv/riscv.c
```diff
@@ -2708,43 +2708,42 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
 		.mode = COMMAND_ANY,
-		.usage = "riscv repeat_read count address [size=4]",
+		.usage = "count address [size=4]",
 		.help = "Repeatedly read the value at address."
 	},
 	{
 		.name = "test_compliance",
 		.handler = riscv_test_compliance,
 		.mode = COMMAND_EXEC,
-		.usage = "riscv test_compliance",
```
borneoa, line 2718: No, you cannot left 'usage' not initialized otherwise you get a run-time error at openocd startup.
Use instead
.usage = "",
This is valid also below.

timsifive: I assumed I'd get a crash. I didn't check for an error message, but I see they're there. I'll put them back in.

PR #523, opened 2020-09-04
Attempt to enforce signed-off-by lines in commits.

timsifive:
Requested by upstream OpenOCD maintainers.

The commit-msg script comes from OpenOCD, and ensures there's a Change-Id. This is meaningless to us, but matters when creating changes to be pushed upstream. Since it's harmless, I'm just including it here also.


GregSavin (approved):
Fascinating hook script, with an awk program embedded in shell script.

timsifive:
In my defense, I did not write the script. :-)

PR #525, opened 2020-09-08
Allow to put breakpoints in memories without 16 bit access

sobuch:
This change allows placing software breakpoints in memories which do not support half-word wide access (16-bit). This can be the case for various special memory regions intended solely for the program (instructions), and not for data. For instance various core-coupled memories for the program storage.

If the breakpoint placement via 16-bit access fails, OpenOCD will retry using 32-bit or 64-bit access (depending on the platform and memory access method used) and will perform read-modify-write operation.

Concrete example of a memory and processor where this is needed: ICCM (instruction closely-coupled memory) in SweRV EH1.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -818,8 +818,7 @@ static int write_by_any_size(struct target *target, target_addr_t address, uint3
 
 	/* On failure, try other access sizes.
 	   Minimize the number of accesses by trying first the largest size. */
-	for (unsigned access_size = 8; access_size > 0; access_size /= 2)
-	{
+	for (unsigned access_size = 8; access_size > 0; access_size /= 2) {
 		if (access_size == preferred_size)
 			/* Already tried this size. */
 			continue;
@@ -851,8 +850,7 @@ static int read_by_any_size(struct target *target, target_addr_t address, uint32
 
 	/* On failure, try other access sizes.
 	   Minimize the number of accesses by trying first the largest size. */
-	for (unsigned access_size = 8; access_size > 0; access_size /= 2)
-	{
+	for (unsigned access_size = 8; access_size > 0; access_size /= 2) {
 		if (access_size == preferred_size)
 			/* Already tried this size. */
 			continue;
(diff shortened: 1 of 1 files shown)


timsifive (changes requested):
This change looks OK in general, but I'd like to see if broken down slightly differently so that the helper function will be more useful in the future.

Instead of sw_breakpoint_*, instead can you implement:
```
static int write_by_any_size(struct target *target, target_addr_t address, uint32_t size,
                const uint8_t *buffer);
static int read_by_any_size(struct target *target, target_addr_t address,
		uint32_t size, uint8_t *buffer);
```
That way we can use the functions in places besides the breakpoint code.

Then inside that function first try target_read_memory() with the given size, and if that fails fall back to the code you already wrote. There's no need to handle cases (e.g. only 8-bit writes supported) beyond the ones you need right now. We can add those as needed.

sobuch:
Thanks for the review. We think it would be more reusable if the desired access size would also be passed as an argument. Besides that change, am I correct that you had something like https://github.com/sobuch/riscv-openocd/tree/breakpoints_v1 in mind? If so, we would also propose to break it into two functions; one of which would be general and just read/write `size` bits by `access_size` wide access, see https://github.com/sobuch/riscv-openocd/compare/breakpoints_v1...breakpoints_v2

What do you think about this? No need for thorough review, I will check the preferred version again and then get the changes in here

timsifive:
I had imagined this function would just take care of the read/write so the caller doesn't need to specify anything besides the size of the value to be written. So the write version would try, e.g.:
```
# Try the straightforward case.
if target_write_memory(target, address, size, 1, buffer):
    return ERROR_OK
for access_size in (1, 2, 4, 8):
   # Try writing using this size, using read-modify-write if necessary
   # No need to support complicated cases that you don't need right now. That can be added later.
```
Then the caller just calls the function and doesn't know nor care what access size is used, just that the read/write happened.

sobuch:
Alright, now I think I understand what you meant, thank you for the explanation @timsifive . I kept functions read/write_by_given_size to better decompose the logic. Do you see some more room for improvements? Thanks.

timsifive (approved):
This looks good. Thanks for improving it.

I'll probably merge this on Monday.

PR #526, opened 2020-09-08
Add empty usage strings back in.

timsifive:
Comment on #521 pointed out that they're required.


GregSavin (approved):
Ah, good to know that those fields are effectively required to be non-null.

PR #528, opened 2020-09-09
Make a couple variables static.

timsifive:


GregSavin (approved):
Looks good.  I vaguely recall that initially they were referenced from multiple compilation units, and at some point some code was refactored, and after refactoring, they were then referenced from only riscv.c.

PR #529, opened 2020-09-10
{read_from,write_to}_buf -> buf_[sg]et_u{32,64}

timsifive:
Requested in http://openocd.zylin.com/#/c/5821/7/src/target/riscv/riscv-013.c

Change-Id: I764d77192e902466eed8a6ccf1042b42016afb7e


GregSavin (approved):
Nice, so, some essentially duplicated code de-duplicated.

PR #530, opened 2020-09-11
Fix allowed mode for set_mem_access

sobuch:
OpenOCD fails with segfault when using newly introduced command `riscv set_mem_access` before calling init. 

This is caused by the option being set per target. To fix it, `set_mem_access` is to be allowed only after init; `set_prefer_sba`, which calls the same functions, will fail and warn the user to only call it after init.

This also relates to https://github.com/riscv/riscv-openocd/issues/527

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2286,6 +2286,11 @@ COMMAND_HANDLER(riscv_test_compliance) {
 
 COMMAND_HANDLER(riscv_set_prefer_sba)
 {
+	if (CMD_CTX->mode != COMMAND_EXEC) {
+		LOG_ERROR("Cannot use `riscv set_prefer_sba` in the config phase. "
+				"Please call it only in OpenOCD execution phase (after `init`).");
+		return ERROR_FAIL;
+	}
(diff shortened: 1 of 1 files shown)


JanMatCodasip:
After taking a closer look on this, it appears there is a better solution - to allocate `target->arch_info` in the target _creation_ phase, which happens before `init`, not only in the target initialization phase. 

That would place no restriction on `riscv set_mem_access` et. al. and would permit those calls even in the config phase because the needed structures for the config storage would already be allocated.

So please hold on with this pull request for a moment. I'll look into this on Monday and if this is doable, I'll post the code here.

JanMatCodasip:
As mentioned above, I have filed an alternate fix to this issue: #531. It is more flexible as it does not restrict configuration commands (e.g. `riscv set_prefer_sba`) from being called in the OpenOCD's configuration phase. Please if you can review that one.

If #531 is accepted, then this pull request can be closed.

sobuch:
We discussed this internally with @JanMatCodasip and I agree the other solution (https://github.com/riscv/riscv-openocd/pull/531) is better. Thus closing this one.

PR #531, opened 2020-09-14
Allocate RISC-V arch_info during target creation

JanMatCodasip:
* Ensured that target->arch_info is allocated as soon as the
  target is created. Needed so that per-target config commands
  (e.g. "riscv set_mem_access") can be executed also in the
  OpenOCD's config phase (before calling "init").

* Added several assert()'s for safety.

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -228,6 +228,8 @@ static int get_register(struct target *target, riscv_reg_t *value, int hartid,
 static riscv011_info_t *get_info(const struct target *target)
 {
 	riscv_info_t *info = (riscv_info_t *) target->arch_info;
+	assert(info);
+	assert(info->version_specific);
 	return (riscv011_info_t *) info->version_specific;
 }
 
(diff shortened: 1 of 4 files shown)


JanMatCodasip:
This change brings the ability to call per-target RISC-V configuration commands also in the OpenOCD's configuration phase (for example `riscv set_mem_access`). At the same time, it is a ground-work for https://github.com/riscv/riscv-openocd/issues/527.

This change supersedes #530. If this change is all right, please close #530.

This has been tested manually and also via `riscv-tests/debug`, all passed.

timsifive (changes requested):
Thanks for jumping on this. Can you take out that comment and move the discussion elsewhere?

JanMatCodasip:
Sure, I have removed the TODO comment and instead noted the idea under https://github.com/riscv/riscv-openocd/issues/534 so that it won't get lost.

timsifive (approved):
Looks good. I'll merge this tomorrow.

PR #532, opened 2020-09-14
Update Signed-Off-By Makefile changes to work for out of tree builds

simonpcook:
Change-Id: I617ca9f0eb6a473528646dfb80d75b6bafaa0635
Signed-off-by: Simon Cook <simon.cook@embecosm.com>


timsifive (approved):
That looks great. Thank you for figuring out what I broke and how to fix it.

PR #536, opened 2020-09-25
Improve riscv expose_[csrs|custom] commands

sobuch:
This pull request is intended as a general improvement for riscv expose_[csrs|custom] commands.
Changes in syntax improve usability of these commands, and maintain full backwards compatibility.

User will see the following features:
* Add option to specify custom name for registers.
* Allow to call commands multiple times without loss of previous data.
* Make sure the commands can only be used in the config phase (before "init").
* Validity checks and warnings.
* Change commands to be per target. (see https://github.com/riscv/riscv-openocd/issues/527)

Internal improvement:
* Fix memory leaks.
* Also fix unrelated memory leaks to keep valgrind happy.

Signed-off-by: Samuel Obuch <sobuch@codasip.com>

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -286,7 +286,6 @@ dm013_info_t *get_dm(struct target *target)
 	}
 	target_entry = calloc(1, sizeof(*target_entry));
 	if (!target_entry) {
-		free(info->dm);
 		info->dm = NULL;
 		return NULL;
 	}
@@ -1538,17 +1537,10 @@ int wait_for_authbusy(struct target *target, uint32_t *dmstatus)
 static void deinit_target(struct target *target)
 {
 	LOG_DEBUG("riscv_deinit_target()");
-	RISCV_INFO(r);
-	RISCV013_INFO(info);
-
-	target_list_t *entry, *tmp;
-	list_for_each_entry_safe(entry, tmp, &info->dm->target_list, list)
-		free(entry);
-
-	free(info->dm);
-	free(r->version_specific);
+	riscv_info_t *info = (riscv_info_t *) target->arch_info;
+	free(info->version_specific);
 	/* TODO: free register arch_info */
-	r->version_specific = NULL;
+	info->version_specific = NULL;
 }
 
 static int set_haltgroup(struct target *target, bool *supported)
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -286,6 +286,7 @@ dm013_info_t *get_dm(struct target *target)
 	}
 	target_entry = calloc(1, sizeof(*target_entry));
 	if (!target_entry) {
+		free(info->dm);
 		info->dm = NULL;
 		return NULL;
 	}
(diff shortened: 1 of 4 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -286,6 +286,7 @@ dm013_info_t *get_dm(struct target *target)
 	}
 	target_entry = calloc(1, sizeof(*target_entry));
 	if (!target_entry) {
+		free(info->dm);
```
timsifive: This was intentionally left out because it's possible that info->dm points at the dm that is actually in use by target 0 (assuming current target is pointing at target 1). We could reference count and solve this properly, but since this code is only hit when calloc fails (extremely unlikely) I didn't think it was worth the hassle.

src/target/riscv/riscv-013.c
```diff
@@ -1537,10 +1538,17 @@ int wait_for_authbusy(struct target *target, uint32_t *dmstatus)
...
-	riscv_info_t *info = (riscv_info_t *) target->arch_info;
-	free(info->version_specific);
+	RISCV_INFO(r);
+	RISCV013_INFO(info);
+
+	target_list_t *entry, *tmp;
+	list_for_each_entry_safe(entry, tmp, &info->dm->target_list, list)
+		free(entry);
+
+	free(info->dm);
```
timsifive: Same here. The same dm might be connected to multiple targets, so you risk freeing the same allocated memory multiple times.

src/target/riscv/riscv.c
```diff
@@ -2502,103 +2503,160 @@ COMMAND_HANDLER(riscv_set_enable_virtual)
 	return ERROR_OK;
 }
 
-void parse_error(const char *string, char c, unsigned position)
+int parse_ranges(struct list_head *ranges, const char *tcl_arg, const char *reg_type, unsigned max_val)
```
timsifive, line 2506: Can you add LOG_ERROR messages to the cases where you return error that don't already create a message?

src/target/riscv/riscv.c
```diff
@@ -2502,103 +2503,160 @@ COMMAND_HANDLER(riscv_set_enable_virtual)
...
+		char *id = NULL;
+
+		char *dash = strchr(arg, '-');
+		char *equals = strchr(arg, '=');
+		unsigned pos;
+
+		if (!dash && !equals) {
+			/* Expecting single register number. */
+			if (sscanf(arg, "%u%n", &low, &pos) != 1 || pos != strlen(arg)) {
+				free(args);
```
timsifive, line 2527: e.g.
LOG_ERROR("Failed to parse single register number from '%s'.", arg);

src/target/riscv/riscv.h
```diff
@@ -60,6 +60,12 @@ typedef struct {
 	unsigned custom_number;
 } riscv_reg_info_t;
 
+typedef struct {
+	struct list_head list;
+	uint16_t low, high;
+	char *id;
```
timsifive: I think "name" is more descriptive than "id".

timsifive (changes requested):
Looks good overall, thanks.

I'm curious what use case caused you to make this improvement.

src/target/riscv/riscv-013.c
sobuch: Interesting, thank you very much for the explanation. I reverted these changes; here, I would just note that this case is hit every time init is called, and not only in rare occasions.

src/target/riscv/riscv.c
sobuch: Done

src/target/riscv/riscv.h
sobuch: Changed ids to names

timsifive (approved):
Looks good. Thank you. I'll merge this tomorrow, unless there are more comments.

JanMatCodasip:
> I'm curious what use case caused you to make this improvement.

The main motivation for this code change was the ability to have textual names for custom registers (very convenient for the user when there is more than handful of them in the given target). The other improvements and clean-up were done just along the way.

PR #538, opened 2020-09-27
Improve support for GD32VF103 MCU

tchebb:
The GD32VF103 is a low-cost 32-bit RISC-V microcontroller with peripherals that are more-or-less compatible with the STM32F103 ARM microcontroller. It is available on several low-cost dev boards, such as the Sipeed Longan Nano. Unfortunately, its debug module has a number of quirks, including spec-violating reset behavior and an `ndmreset` line that doesn't seem to be hooked up to the core at all. Teach OpenOCD how to work around these quirks.

Tested working
 - Halt/resume
 - Register access
 - Single step
 - Hardware breakpoints
 - Reset (both `reset halt` and `reset run`)
 - Memory operations (both abstract and progbuf, SBA not supported by chip)
 - Flash read/write/erase

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2172,16 +2172,7 @@ static int deassert_reset(struct target *target)
 			index = r->current_hartid;
 		}
 
-		char *operation;
-		uint32_t expected_field;
-		if (target->reset_halt) {
-			operation = "halt";
-			expected_field = DM_DMSTATUS_ALLHALTED;
-		} else {
-			operation = "run";
-			expected_field = DM_DMSTATUS_ALLRUNNING;
-		}
-		LOG_DEBUG("Waiting for hart %d to %s out of reset.", index, operation);
+		LOG_DEBUG("Waiting for hart %d to come out of reset.", index);
 		while (1) {
 			int result = dmstatus_read_timeout(target, &dmstatus, true,
 					riscv_reset_timeout_sec);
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -2100,7 +2100,10 @@ static int assert_reset(struct target *target)
 
 	uint32_t control_base = set_field(0, DM_DMCONTROL_DMACTIVE, 1);
 
-	if (target->rtos) {
+	if (target_has_event_action(target, TARGET_EVENT_RESET_ASSERT)) {
+		/* Run the user-supplied script if there is one. */
+		target_handle_event(target, TARGET_EVENT_RESET_ASSERT);
+	} else if (target->rtos) {
```
timsifive, line 2106: This looks like if a user configures a reset-assert script, then none of the regular reset stuff happens. http://openocd.org/doc/html/CPU-Configuration.html#targetevents says reset-assert is "Issued as part of reset processing," not instead of.

I don't know how replacing reset usually happens. From the script it looks like you really do want to just do your own reset processing. There must be other targets where that's required.

The simplest solution I can think of is to add a riscv config option to disable the regular reset if a reset-assert handler exists. Do you have time to find out how reset on noncompliant targets is handled on different architectures?

timsifive (changes requested):
In addition to the testing you've done, can you confirm that the tests in https://github.com/riscv/riscv-tests/tree/master/debug still pass after this change?

src/target/riscv/riscv-013.c
tchebb: Replacing the default handler is consistent with what other architectures do. Most targets don't have a standard architectural reset mechanism and so by default use SRST if it's present. For those architectures, when a `reset-assert` handler exists they call it instead of (not in addition to) asserting SRST.

Furthermore, a couple architectures, Cortex-M and MIPS M4K, do have architectural reset mechanisms, and they too skip these mechanisms when the config specifies a `reset-assert` handler.

Cortex-M has the following block right at the beginning of its reset function, which checks for the handler, running it and exiting early if found:
https://github.com/riscv/riscv-openocd/blob/2c909f8faae052ccc3cec82ed0d98b043f871ce3/src/target/cortex_m.c#L1082-L1090

MIPS M4K's code structure is even more obvious, with an explicit three-way switch between using SRST, a `reset-assert` handler, or the default architectural reset sequence:
https://github.com/riscv/riscv-openocd/blob/2c909f8faae052ccc3cec82ed0d98b043f871ce3/src/target/mips_m4k.c#L342-L368

Notably, this behavior doesn't mean that you can't write a config which keeps the default sequence and just augments it: that's exactly what `reset-assert-pre` and `reset-assert-post` are for.

tchebb:
Thanks for the review! I'll run the tests tonight.

src/target/riscv/riscv-013.c
timsifive: Thanks. Sounds like the documentation is just confusing in this case.

tchebb:
I ran the `spike64` tests with OpenOCD from my branch and got no failures: https://gist.github.com/tchebb/e4719eba42bbc6c1ecb3a91a7d0ee3ea

I didn't run the others since they require a compiler with vector extensions, which I'd need to clone+build. I can do that if you think it's likely to turn up issues, but I'd prefer not to if this is sufficient. Let me know what you think.

timsifive (approved):
That will do testing-wise for now. Hopefully vector support in the compiler will merge around quickly. Otherwise I might have to change the tests to just skip those.

I'll merge this tomorrow, unless there are more comments.

PR #539, opened 2020-09-27
Remove unused riscv_error_t type

tchebb:
This seems to be completely unused in these two files. It was probably accidentally copied from `riscv-011.c`, where it is used.


timsifive (approved):
Thanks for cleaning this up.

PR #540, opened 2020-09-28
Remove outdated documentation in riscv.c

tchebb:
These comments appear to have been copied from `riscv-011.c`, for which they are accurate. However, it makes no sense to also have them in `riscv.c`, because 1) none of the things described are actually in `riscv.c`; and 2) `riscv-013.c` has an entirely different code structure, meaning everything in the comment is an implementation detail of `riscv-011.c`. Remove the copy in `riscv.c` and just leave the one in `riscv-011.c`.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -20,48 +20,6 @@
 #include "gdb_regs.h"
 #include "rtos/rtos.h"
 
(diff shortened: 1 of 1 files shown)


timsifive (approved):
Yes. Thank you.

PR #541, opened 2020-09-30
Add memory sample feature

timsifive:
The intended use for this is PC sampling where a PC is exposed in memory, but it can also be used to e.g. sample a variable's value (assuming it's in memory and not a register). This requires the target to support system bus access, while it is running.

The buffer can be dumped to the command line in human-readable or base64 format.

On a single-hart FPGA target I get 21k 32-bit samples/second, or 16k 64-bit samples/second.


src/target/riscv/riscv-013.c
```diff
@@ -2001,6 +2001,226 @@ static int riscv013_set_register_buf(struct target *target,
 	return result;
 }
 
+/**
+ * @par size in bytes
+ */
+static void write_to_buf(uint8_t *buffer, uint64_t value, unsigned size)
```
JanMatCodasip: Can already existing `buf_set_u64()` function be used instead?

src/target/riscv/riscv-013.c
```diff
@@ -2001,6 +2001,226 @@ static int riscv013_set_register_buf(struct target *target,
...
+						break;
+					case 4:
+						sbaccess = DM_SBCS_SBACCESS32;
+						break;
+					case 8:
+						sbaccess = DM_SBCS_SBACCESS64;
+						break;
+					default:
+						LOG_ERROR("Unsupported SBA access size: %d",
+								  config->bucket[i].size_bytes);
```
JanMatCodasip: The detection of supported SBACCESS size could be put to a separate helper function and reused also in `mem_should_skip_sysbus()`.

src/target/riscv/riscv.c
```diff
@@ -2697,7 +2768,158 @@ COMMAND_HANDLER(handle_repeat_read)
...
+}
+
+uint32_t sample_buf_read32(unsigned offset){
+	return sample_buf.buf[offset] |
+		   (((uint32_t)sample_buf.buf[offset + 1]) << 8) |
+		   (((uint32_t)sample_buf.buf[offset + 2]) << 16) |
+		   (((uint32_t)sample_buf.buf[offset + 3]) << 24);
+}
+
+uint64_t sample_buf_read64(unsigned offset){
```
JanMatCodasip: It appears `buf_get_u64()` and `buf_get_u32()` could be used here.

src/target/target.c
```diff
@@ -1699,8 +1697,9 @@ static int target_call_timer_callbacks_check_time(int checktime)
 
 	keep_alive();
 
-	struct timeval now;
-	gettimeofday(&now, NULL);
+	int64_t now = timeval_ms();
+	if (next_event)
+		*next_event = now + 1000;
```
JanMatCodasip, line 1702: Is this meant as a safe-guard to ensure this processing repeats at least once per second, even if no callback is scheduled?

src/target/riscv/riscv-013.c
```diff
@@ -2001,6 +2001,226 @@ static int riscv013_set_register_buf(struct target *target,
...
+			}
+		}
+
+		if (buf->used + result_bytes >= buf->size) {
+			riscv_batch_free(batch);
+			break;
+		}
+
+		/*int result =*/ batch_run(target, batch);
+		unsigned read = 0;
```
JanMatCodasip, line 2206: After the batch finishes, please consider checking for the possible error states: `sbcs.sberror`, `sbcs.sbbusyerror`. If they occur, the user should be notified, and the error state cleared. 

This is important because subsequent system bus accesses, as written now (e.g. `read_memory_bus_v1()`), assume that any previous sticky errors had been cleared.

Also the data won't be all valid in case of an error.

src/target/riscv/riscv.c
```diff
@@ -2043,11 +2062,51 @@ int set_debug_reason(struct target *target, enum riscv_halt_reason halt_reason)
...
+	if (!sample_buf.buf)
+		return ERROR_OK;
+
+    LOG_DEBUG("buf used/size: %d/%d", sample_buf.used, sample_buf.size);
+
+	uint64_t start = timeval_ms();
+	riscv_sample_buf_maybe_add_timestamp();
+	if (r->sample_memory) {
+		int result = r->sample_memory(target, &sample_buf, &sample_config,
+									  start + TARGET_DEFAULT_POLLING_INTERVAL);
```
JanMatCodasip, line 2159: Should the duration of the sampling reflect the current (configurable) polling interval instead of always using the default?

src/target/riscv/riscv.c
```diff
@@ -277,6 +278,9 @@ static enum {
 	RO_REVERSED
 } resume_order;
 
+static riscv_sample_buf_t sample_buf;
+static riscv_sample_config_t sample_config;
```
JanMatCodasip: Should the sample_config and sample_buf be per-target, instead of global?

src/target/riscv/riscv.c
```diff
@@ -2697,7 +2768,158 @@ COMMAND_HANDLER(handle_repeat_read)
...
+			} else {
+				command_print(CMD, "bucket %d; disabled", i);
+			}
+		}
+		return ERROR_OK;
+	}
+
+	if (CMD_ARGC < 2) {
+		LOG_ERROR("Command requires at least bucket and address arguments.");
+		return ERROR_FAIL;
```
JanMatCodasip: If you'd like, you can return `ERROR_COMMAND_SYNTAX_ERROR`, which makes OpenOCD print the syntax help for the command.

src/target/riscv/riscv.c
```diff
@@ -2697,7 +2768,158 @@ COMMAND_HANDLER(handle_repeat_read)
...
+	}
+
+	if (!sample_buf.buf) {
+		sample_buf.size = 1024 * 1024;
+		sample_buf.buf = malloc(sample_buf.size);
+	}
+	/* Clear the buffer when the configuration is changed. */
+	sample_buf.used = 0;
+
+	sample_config.enabled = true;
```
JanMatCodasip: Is it necessary to have a separate "enable" flag for the whole sampling mechanism? The other option would be to consider the sampling enabled if at least one bucket is enabled.

src/target/riscv/riscv.c
```diff
@@ -2697,7 +2768,158 @@ COMMAND_HANDLER(handle_repeat_read)
...
+	}
+
+	uint32_t bucket;
+	COMMAND_PARSE_NUMBER(u32, CMD_ARGV[0], bucket);
+	if (bucket > DIM(sample_config.bucket)) {
+		LOG_ERROR("Max bucket number is %d.", (unsigned) DIM(sample_config.bucket));
+		return ERROR_COMMAND_ARGUMENT_INVALID;
+	}
+
+	sample_config.bucket[bucket].enabled = true;
```
JanMatCodasip: How about disabling the memory sampling for the given buckets?

Or is it assumed that the user will configure the sampling once and will leave it that way for the whole OpenOCD session? If that is the case, then it would make sense to allow the `riscv memory_sample` command only in the configuration phase (i.e. `COMMAND_CONFIG` instead of `COMMAND_ANY`).

JanMatCodasip:
> The intended use for this is PC sampling where a PC is exposed in memory (...)

It is my understanding that trace specification should address this use case (I am not familiar with its current state, though).

src/target/riscv/riscv-013.c
timsifive: Done. Thanks.

src/target/riscv/riscv-013.c
timsifive: Good suggestion. Done.

src/target/riscv/riscv.c

src/target/riscv/riscv-013.c
timsifive: I knew there was something I forgot! Done.

src/target/riscv/riscv.c
timsifive: Where is the polling interval configurable? I don't see a command, and target.c defines
`static const int polling_interval = TARGET_DEFAULT_POLLING_INTERVAL;`

src/target/riscv/riscv.c
timsifive: Yes. Done.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
timsifive: Yes. It's primarily used to disable the sampling when an error occurs while sampling. That way the user only gets an error message once, and still has all the configuration that they can tweak if they typo'd a single entry.

src/target/riscv/riscv.c
timsifive: Disabling is a good point. Added that.

timsifive:
Trace will allow collection of PC samples (and even every single PC), but it's relatively expensive to implement in hardware. Making the PC accessible while running (either through DPC or memory mapped) is much cheaper, and still useful.

GregSavin (commented):
If the memory sample buffer fills up because the buffered contents are not being consumed in a timely manner, then subsequent samples aren't added to the buffer?

src/target/riscv/riscv.c
```diff
@@ -2918,7 +2994,151 @@ COMMAND_HANDLER(handle_repeat_read)
...
+
+	/* Clear the buffer when the configuration is changed. */
+	r->sample_buf.used = 0;
+
+	r->sample_config.enabled = true;
+
+	return ERROR_OK;
+}
+
+COMMAND_HANDLER(handle_dump_sample_buf_command)
```
GregSavin, line 3064: Is the sample buffer getting cleared in this function, or does that happen somewhere else?

src/target/riscv/riscv.c
JanMatCodasip: You're right, sorry for this misinformation. I was under the impression it is configurable but it is not. I think there was a merge request about this in the upstream, which was apparently not merged in the end.

src/target/riscv/riscv.c
JanMatCodasip: Understood, thanks for explaining.

timsifive:
> If the memory sample buffer fills up because the buffered contents are not being consumed in a timely manner, then subsequent samples aren't added to the buffer?

Yes. I added that to the documentation.

src/target/riscv/riscv.c
timsifive: Supposed to be here, but maybe I didn't implement that. :-) Let me fix it.

GregSavin (approved):
Looks good to me now, as relates to the couple of questions I had.

src/helper/base64.c
```diff
@@ -0,0 +1,157 @@
+/*
+ * Base64 encoding/decoding (RFC1341)
+ * Copyright (c) 2005-2011, Jouni Malinen <j@w1.fi>
+ *
+ * This software may be distributed under the terms of the BSD license.
+ * See README for more details.
```
borneoa, line 6: If you plan to upstream this in OpenOCD, please review the license text.
I think BSD license is ok, but where is the README ?
Maybe it would be enough to keep the copyright and replace the license text with a simple SPDX tag.
For an alternative, I have quickly searched for an implementation of base64 enc/dec under GPLv2-or-later, but nothing so far.

PR #542, opened 2020-10-02
Do not throw error if RISC-V tselect unimplemented

TobiasKaiser:
A RISC-V hart without Trigger Module may not implement any of the
associated CSRs such as tselect according to the specification.
riscv_enumerate_triggers previously threw an error in this case, but
only on the first invocation due to r->triggers_enumerated being set
regardless of this. Due to the propagation of this error condition to
disable_triggers and riscv_openocd_step, such a hart would remain
halted after the first 'step' (or 'continue') of a debug session.

This problem can be reproduced with the Ibex RISC-V CPU when
the DbgTriggerEn parameter is set to zero.

This commit changes the behavior of riscv_enumerate_triggers to
return ERROR_OK when tselect was not readable. This fixes the
described malfunction.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3623,8 +3623,11 @@ int riscv_enumerate_triggers(struct target *target)
 		riscv_reg_t tselect;
 		int result = riscv_get_register_on_hart(target, &tselect, hartid,
 				GDB_REGNO_TSELECT);
+		/* If tselect is not readable, the trigger module is likely not
+		 * implemented. There are no triggers to enumerate then and no error
+		 * should be thrown. */
 		if (result != ERROR_OK)
-			return result;
+			return ERROR_OK;
 
 		for (unsigned t = 0; t < RISCV_MAX_TRIGGERS; ++t) {
 			r->trigger_count[hartid] = t;
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3626,8 +3626,11 @@ int riscv_enumerate_triggers(struct target *target)
 		/* If tselect is not readable, the trigger module is likely not
 		 * implemented. There are no triggers to enumerate then and no error
 		 * should be thrown. */
-		if (result != ERROR_OK)
+		if (result != ERROR_OK) {
+			LOG_DEBUG("Cannot access tselect register on hart %d. "
+					"Assuming that triggers are not implemented.", hartid);
 			return ERROR_OK;
+		}
 
 		for (unsigned t = 0; t < RISCV_MAX_TRIGGERS; ++t) {
 			r->trigger_count[hartid] = t;
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3629,7 +3629,7 @@ int riscv_enumerate_triggers(struct target *target)
 		if (result != ERROR_OK) {
 			LOG_DEBUG("Cannot access tselect register on hart %d. "
 					"Assuming that triggers are not implemented.", hartid);
-			return ERROR_OK;
+			continue;
 		}
 
 		for (unsigned t = 0; t < RISCV_MAX_TRIGGERS; ++t) {


JanMatCodasip (commented):
Thank you for the fix! Looks all right to me.

timsifive (changes requested):
This seems like the wrong place for this fix. riscv_enumerate_triggers() is called from register_set(), add_trigger(), remove_trigger(), and disable_triggers(). In the first 3 cases the error really should be propagated. disable_triggers() when called from resume_prep() should also cause an error. 

I think the only case that you want to fail silently is when riscv_openocd_step() calls disable_triggers(), which fails in riscv_enumerate_triggers(). So the place to fix this is in riscv_openocd_step(), where it is acceptable if disable_triggers() fails because it's called "just in case."

JanMatCodasip:
> This seems like the wrong place for this fix. riscv_enumerate_triggers() is called from register_set(), add_trigger(), remove_trigger(), and disable_triggers(). In the first 3 cases the error really should be propagated. disable_triggers() when called from resume_prep() should also cause an error.

I am sorry to disagree. IMO, the fix placed exactly at the right spot. 

If access to `tselect` register fails, that indicates the trigger functionality is not implemented (see also https://github.com/riscv/riscv-debug-spec/pull/431). That means the result is trigger_count == 0 and the trigger enumeration process as such should be considered **successful**. In another words, there is no error to propagate anywhere.

All the follow-on code should handle trigger_count == 0 situation gracefully. If that was not the case, that would be a separate issue of that folĺow-on code, but this fix itself remains correct. 

Or, please correct me if I missed something. Thank you.

src/target/riscv/riscv.c
```diff
@@ -3623,8 +3623,11 @@ int riscv_enumerate_triggers(struct target *target)
 		riscv_reg_t tselect;
 		int result = riscv_get_register_on_hart(target, &tselect, hartid,
 				GDB_REGNO_TSELECT);
+		/* If tselect is not readable, the trigger module is likely not
+		 * implemented. There are no triggers to enumerate then and no error
+		 * should be thrown. */
 		if (result != ERROR_OK)
-			return result;
+			return ERROR_OK;
```
JanMatCodasip: At your own discretion, consider whether it'd be good to inform the user about the situation -

```
LOG_INFO("Cannot access 'tselect' register. Assuming that triggers are not implemented.");
```

timsifive:
> If access to tselect register fails, that indicates the trigger functionality is not implemented (see also riscv/riscv-debug-spec#431). That means the result is trigger_count == 0 and the trigger enumeration process as such should be considered successful. In another words, there is no error to propagate anywhere.

That depends on why the access failed. It might fail because the USB connection to your debug device had a glitch.

But in general such failures are rare. You are right that we can assume that this call only fails because the target does not implement any triggers, and in that case the change as written works fine.

I retract my earlier comment. I'll merge this in a day or two, so @TobiasKaiser has a chance to respond to your comment about an info message.

JanMatCodasip:
> >If access to tselect register fails, that indicates the trigger functionality is not implemented (see also riscv/riscv-debug-spec#431). That means the result is trigger_count == 0 and the trigger enumeration process as such should be considered successful. In another words, there is no error to propagate anywhere.
>
> That depends on why the access failed. It might fail because the USB connection to your debug device had a glitch.

Thank you for replying. Good point about the other possible error causes. We should be safe here, though. All calls to `jtag_execute_queue()` from the "riscv" target are checked for an error and a proper message printed. So such errors should not go unnoticed, assuming a well-behaving JTAG driver that reports the errors back, such as USB problems.

Of course there could be also signal integrity issues on the JTAG wires - e.g. random garbled bits due to poor physical connection. Yet there is no systemic way of detecting this in software as the JTAG does not have provisions for data integrity checking. So these low-level issues are out of scope of this discussion - not much we can do about it.

TobiasKaiser:
Thank you for the discussion. As suggested, a debug message makes sense, so I have added one.

Furthermore, I have changed the action when riscv_get_register_on_hart for tselect returns a failure to keep trying for the other harts instead of returning from riscv_enumerate_triggers. Other harts might implement triggers.

As a side node, I think the error handling in riscv_enumerate_triggers is questionable regardless of this pull request.  From a data model standpoint, when riscv_enumerate_triggers returns an error, r->triggers_enumerated has already been set already and subsequent calls will not retry enumeration.

I also find the statement "If triggers aren’t implemented, the CSRs may not exist at all [...]" in the current RISC-V debug spec draft a bit ambiguous. Would it be permissible for a hart to implement the tselect register but not the tdata1 register? In this case, the problem that I am addressing in riscv_enumerate_triggers would still pop up when trigger CSRs are only partially implemented. But this is likely rather an issue with the spec, or I may be missing something. I think the changes for now make sense and improve compatibility for cores that do not implement the Trigger Module.

src/target/riscv/riscv.c
TobiasKaiser: Debug message added, see blow.

timsifive (approved):
Looks good. I'll merge this tomorrow unless there's more discussion.

PR #544, opened 2020-10-09
Allow riscv_semihosting without 16 bit access to memory with instructions

sobuch:
Semihosting needs to read instructions from the memory, this patch enables to read also from memories without 16-bit access.

This is similar situation to https://github.com/riscv/riscv-openocd/pull/525, and we can reuse the same functions.

### DIFF
--- a/src/target/riscv/riscv.h
+++ b/src/target/riscv/riscv.h
@@ -414,4 +414,7 @@ semihosting_result_t riscv_semihosting(struct target *target, int *retval);
 void riscv_add_bscan_tunneled_scan(struct target *target, struct scan_field *field,
 		riscv_bscan_tunneled_scan_context_t *ctxt);
 
+int read_by_any_size(struct target *target, target_addr_t address, uint32_t size, uint8_t *buffer);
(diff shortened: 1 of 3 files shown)


src/target/riscv/riscv.h
```diff
@@ -414,4 +414,7 @@ semihosting_result_t riscv_semihosting(struct target *target, int *retval);
 void riscv_add_bscan_tunneled_scan(struct target *target, struct scan_field *field,
 		riscv_bscan_tunneled_scan_context_t *ctxt);
 
+int read_by_any_size(struct target *target, target_addr_t address, uint32_t size, uint8_t *buffer);
+int write_by_any_size(struct target *target, target_addr_t address, uint32_t size, uint8_t *buffer);
```
timsifive: Can you call this riscv_*_by_any_size(), to avoid stepping on functions that might be defined elsewhere in OpenOCD?

timsifive (changes requested):
Looks good, except for minor comment.

src/target/riscv/riscv.h
sobuch: done

timsifive (approved):
Looks good. Thanks!

PR #547, opened 2020-10-14
Clear sbcs.sbbusyerror without affecting other sbcs bits

JanMatCodasip:
This is a fix for an issue reported by Joe Stoy at:
https://sourceforge.net/p/openocd/mailman/message/37128537/

When clearing sbcs.sbbusyerror, preserve other bits in the sbcs
register, as needed for subsequent system bus transactions
to execute properly.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2192,7 +2192,7 @@ static int sample_memory_bus_v1(struct target *target,
 			/* Discard this batch (too much hassle to try to recover partial
 			 * data) and try again with a larger delay. */
 			info->bus_master_read_delay += info->bus_master_read_delay / 10 + 1;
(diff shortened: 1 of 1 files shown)


JanMatCodasip:
I am leaving this pull request as "draft" for the moment as I am yet to run `riscv-tests/debug` on the fixed code.

JanMatCodasip:
@timsifive - Please, take a look at this short & clear bugfix. Let me know if there is anything else to do prior to the merge.

I haven't gotten to run the `riscv-tests/debug` on this change, still struggling with picking "the right" toolchain+gdb versions under which all the tests would pass, providing a good baseline. But this change is simple and clear enough, so I believe it can go directly in.

Thank you.

timsifive (approved):
Looks good. Thanks.

PR #549, opened 2020-10-15
Merge upstream changes into this branches

timsifive:
As of change 4fc61a2f9d60e49b8134ab7e0929cfb2b3a4f771.

All tests against spike still pass.

Note to self: do not squash this merge.


JanMatCodasip:
> As of change 4fc61a2. 
> All tests against spike still pass.

@timsifive, could you please point me to proper toolchain version & spike version under which the tests (`riscv-tests/debug`, `make all`) are passing for you? In my setup, I have encountered failures on spike32 target:

* GDB complains about `bfd requires xlen 8, but target has xlen 4` (apparently mismatch between XLEN of the target and settings of GDB)
* Full log: [failing_test_spike32.txt](https://github.com/riscv/riscv-openocd/files/5390354/failing_test_spike32.txt)

I am currently using:
* riscv-gnu-toolchain `5842fde` (latest from branch rvv-0.9.x)
* riscv-isa-sim `036aacb` (latest master)

Are these the right versions I should be using? Are all `riscv-tests/debug` (`make all`) expected to pass or are there currently any known issues with the tests?

Thank you.

JanMatCodasip:
Quick update - 

After further tinkering with the tests ([rv32-workaround-diff.txt](https://github.com/riscv/riscv-openocd/files/5392927/rv32-workaround-diff.txt)) , it seems that sending `set arch riscv:rv32` to GDB on `spike32*` targets resolves the issue `bfd requires xlen 8`. This was not needed before, what changed in the system?

Even with that workaround, few vector tests still fail:
```
spike32_2-VectorTest.log
spike32-VectorTest.log
spike64_2-VectorTest.log
```

Are these known broken tests?

Thank you.

timsifive:
> GDB complains about bfd requires xlen 8, but target has xlen 4 (apparently mismatch between XLEN of the target and settings of GDB)

This used to be a problem in older GDBs, but then it was fixed (certainly over a year ago). The workaround you found does work, but shouldn't be required.

I seem to be using gdb 85f783647061e58968ecdc516137d8df9f2d5e16 from git://sourceware.org/git/binutils-gdb.git. For my testing I don't see any benefit using a rvv branch gdb.

The fact that the vector tests are failing for you is probably related. They pass for me. The only test that fails intermittently for me is Sv48Test. For spike I'm on c7f1505cacc2b6ddb9f22da96410a0622862fb17.

For both of these, I don't make much effort to always update, unless there's a compelling reason. We could really use some automated build and test of this. There used to be a travis job for riscv-tools which played this role, but it took 2 hours or so to run because it built the entire toolchain from scratch every time.

JanMatCodasip:
Thank you for clarifying the current state. For now, I will use the tool versions that work for you when checking other (unrelated) OpenOCD pull requests, then. I would also be happy if the test process could be automated (if time permits, of course).

Either way, none of my previous comments should block this pull request anyhow.

GregSavin (approved):
Yeah, I do remember the rationale for not "squashing" these (hopefully we never have to bisect, but it's definitely nice to retain the option if needed).

PR #555, opened 2020-11-10
rename and cleanup, add tiny-h-cjtag

sifivekevin:
- rename and cleanup to arm-usb-ocd-h-cjtag file to be consistent
- add copy of olimex-arm-jtag-cjtag.cfg as arm-usb-tiny-h-cjtag.cfg
- did not remove olimex-arm-jtag-cjtag.cfg (should be deprecated)

### DIFF
--- a/tcl/interface/ftdi/olimex-arm-usb-ocd-h-cjtag.cfg
+++ b/tcl/interface/ftdi/olimex-arm-usb-ocd-h-cjtag.cfg
@@ -1,0 +1,22 @@
+#
+# Olimex ARM-USB-OCD-H (using cJTAG)
+#
+# http://www.olimex.com/dev/arm-usb-ocd-h.html
+#
+
+interface ftdi
+ftdi_oscan1_mode on
+ftdi_device_desc "Olimex OpenOCD JTAG ARM-USB-OCD-H"
+ftdi_vid_pid 0x15ba 0x002b
+
+ftdi_layout_init 0x0808 0x0a1b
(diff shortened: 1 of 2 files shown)


TommyMurphyTM1234:
I think the configuration name still needs consideration - what Olimex device(s?) is it intended to support with cjtag?

https://github.com/riscv/riscv-openocd/pull/554#issuecomment-724944413

That should dictate the name?

sifivekevin:
I need to support the ARM-USB-OCD-H and ARM-USB-TINY-H devices (then JTAG and cJTAG variants of each, so four scripts). I'm not familiar with, nor have, the other Olimex devices (the non-H ones and the arm-jtag device).

ARM-USB-TINY-H: olimex-arm-usb-tiny-h.cfg, and olimex-arm-usb-tiny-h-cjtag.cfg
ARM-USB-OCD-H: olimex-arm-usb-ocd-h.cfg, and olimex-arm-usb-ocd-h-cjtag.cfg

If we want to go further and address the other devices:

Rename olimex-jtag-tiny.cfg --> olimex-arm-usb-tiny.cfg
Add cJTAG variants for TINY, and OCD: olimex-arm-usb-tiny-cjtag.cfg, olimex-arm-usb-ocd-cjtag.cfg

That should make everything consistent.

timsifive:
Adding new scripts that we need is fine, but I'd rather not go renaming existing scripts. That kind of cleanup should be done upstream.

timsifive (approved):

PR #558, opened 2020-12-08
Add `riscv info` command.

timsifive:
This command displays some basic information that OpenOCD has detected
about the target. The output is displayed in YAML so it can easily be
parsed. Example of current output:
```
Hart:
  XLEN: 32
  trigger count: 4
Debug Module:
  abits: 6
  progbufsize: 2
  sbversion: 0
  sbasize: 0
  sbaccess128: 0
  sbaccess64: 0
  sbaccess32: 0
  sbaccess16: 0
  sbaccess8: 0
```

Change-Id: If920c083ff6ec9f482c50f913cd8ceaa62461217


timsifive:
@cgsfv does this work for you?

cgsfv:
Wouldn't it be nicer to have it as a tcl mix of arrays and lists for easy access in tcl?

timsifive:
What would the output look like in that case? I don't have the option of returning a TCL data structure. I just get to make print statements. Looks like TCL's "array set" command takes a sequence of whitespace-separated key-value pairs. So maybe this would work with that:
```
  XLEN 32
  trigger_count 4
  abits 6
  progbufsize 2
  sbversion 0
  sbasize 0
  sbaccess128 0
  sbaccess64 0
  sbaccess32 0
  sbaccess16 0
  sbaccess8 0
```

That looks worse for a human reader, though.

Is there a better option?

cgsfv:
Yep, that output works perfect with the tcl "array set" command.
I you would like it to be a bit more verbose, you could prefix the different entries with hart, debug, ...

```
hart.xlen 32
hart.trigger_count 4
debug.abits 6
debug.progbufsize 2
debug.sbversion 0
debug.sbasize 0
debug.sbaccess128 0
debug.sbaccess64 0
debug.sbaccess32 0
debug.sbaccess16 0
debug.sbaccess8 0
```

JanMatCodasip:
> I don't have the option of returning a TCL data structure. I just get to make print statements. 

For completeness, creating a TCL variable (array) in a command handler is possible, but that would require a bit more code, in the end perhaps not worth the effort. An example of how it can be done is in `mem2array`: https://github.com/riscv/riscv-openocd/blob/riscv/src/target/target.c#L4384

* The first option (YAML) seems fine to me as it is both user readable and machine parseable.
* The third option is also fine - still readable and can be directly converted to a TCL array.

timsifive:
I went with Carsten's proposed format. It's not quite as readable as YAML, but close enough.

@JanMatCodasip, thanks for pointing out how I can make real TCL objects, if I need that some day.

timsifive:
Sample output:
```
 hart.xlen            32
 hart.trigger_count    4
   dm.abits            6
   dm.progbufsize      2
   dm.sbversion        0
   dm.sbasize          0
   dm.sbaccess128      0
   dm.sbaccess64       0
   dm.sbaccess32       0
   dm.sbaccess16       0
   dm.sbaccess8        0
```

cgsfv:
LGTM

JanMatCodasip:
> Sample output:
> 
> ```
>  hart.xlen            32
>  hart.trigger_count    4
>    dm.abits            6
>  ...
> ```

Just a little thing. I would prefer if the "dm" items were not indented. It may make someone incorrectly assume there is a hierarchy.

timsifive:
New sample output:
```
hart.xlen              64
hart.trigger_count      4
dm.abits                6
dm.progbufsize          2
dm.sbversion            0
dm.sbasize              0
dm.sbaccess128          0
dm.sbaccess64           0
dm.sbaccess32           0
dm.sbaccess16           0
dm.sbaccess8            0
```

GregSavin (approved):
looks like the Tcl compatible output will work well

PR #560, opened 2020-12-10
Fix error handling in read_memory_progbuf_one().

### LINKED ISSUE #559: src/target/riscv/riscv-013.c:read_memory() failure corrupts x8/s0/fp register
During debugging I found that the frame pointer (x8/s0/fp), which is also used by the debugger for certain operations, was getting corrupted. I added some logging to the above file and see this:

```
Info : read_memory:3560 Begin         x8/s0/fp = 0x8013000
Warn : Failed to read memory via program buffer.
Warn : Failed to read memory via abstract access.
Error: Target mpfs.hart0_e51: Failed to read memory (addr=0x500000000000000)
Error:   progbuf=failed, sysbus=skipped (unsupported size), abstract=failed
Info : read_memory:3621 End (failure) x8/s0/fp = 0x500000000000000
```

So it seems that a failed memory read leaves the frame pointer corrupted with obvious concomitant negative impact on further execution of the program being debugged.

I guess that at least one code path fails to restore the x8/s0/fp register used by openocd to do some work. There are several places where it saves and restores the original value but in this case I guess that the restore never happens.

One obvious solution would be to save it at the start of read_memory() and restore it at the end (success and failure cases?) but maybe this is not the appropriate place?

@timsifive - any suggestions before I submit a patch/PR? Thanks.

Thanks a lot
Cheers
Tommy

Edit: the reason for the attempted read of memory at 0x0500_0000_0000_0000 is that the stack contains a pointer variable which has not been initialized yet and happens to assume that value on entry and (in Eclipse) when the Variables view is active or the user mouses over the variable in the Editor view the debugger attempts to read that memory.

Edit 2: I don't know if the same or a similar issue might also affect riscv-011.c?

timsifive:
Be sure to restore mstatus/s0 even if there is a failure during the
operation.

Fixes #559.

Change-Id: Ib86ca2c7455bad4a668f34703566060a782116db

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3397,7 +3397,7 @@ restore_s0:
 restore_mstatus:
 	if (mstatus != mstatus_old)
 		if (register_write_direct(target, GDB_REGNO_MSTATUS, mstatus_old))
-			return ERROR_FAIL;
+			result = ERROR_FAIL;
 
 	return result;
 }
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3339,9 +3339,10 @@ static int read_memory_progbuf_one(struct target *target, target_addr_t address,
 		return ERROR_FAIL;
 
 	uint64_t s0;
+	int result = ERROR_FAIL;
 
 	if (register_read(target, &s0, GDB_REGNO_S0) != ERROR_OK)
-		return ERROR_FAIL;
+		goto restore_mstatus;
 
 	/* Write the program (load, increment) */
 	struct riscv_program program;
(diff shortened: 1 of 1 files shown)


JanMatCodasip (dismissed):
Thank you, the code looks all right.

src/target/riscv/riscv-013.c
```diff
@@ -3363,40 +3364,42 @@ static int read_memory_progbuf_one(struct target *target, target_addr_t address,
...
+restore_s0:
 	if (riscv_set_register(target, GDB_REGNO_S0, s0) != ERROR_OK)
-		return ERROR_FAIL;
+		result = ERROR_FAIL;
 
-	/* Restore MSTATUS */
+restore_mstatus:
 	if (mstatus != mstatus_old)
 		if (register_write_direct(target, GDB_REGNO_MSTATUS, mstatus_old))
 			return ERROR_FAIL;
```
JanMatCodasip: This is a little thing, but you may still use `result = ERROR_FAIL;` here. For consistency with the rest of the function and to avoid mistakes if anyone later adds more code below this line.

timsifive: Makes sense. I'll do that.

GregSavin (approved):

PR #566, opened 2021-01-13
Fix comment in read_memory_progbuf

craigblackmore:
The comment should refer to reading rather than writing.

Change-Id: I72937bb48053233ab5e48d343c4bd1e394f77bda
Signed-off-by: Craig Blackmore <craig.blackmore@embecosm.com>


timsifive (approved):
Thanks!

PR #567, opened 2021-01-13
Remove `-rtos riscv`

timsifive:
Clean up a bunch of code as a result. There are probably things left to do, but this is the bulk of it.


GregSavin (approved):
Nice how this set of changes simplifies the target/riscv APIs

PR #568, opened 2021-01-14
Don't write to zero.

### LINKED ISSUE #564: x0  actually is NOT cacheable, because it will always read a "ZERO", not the value writen into it
Hi， sir, I  used the "riscv_set_register/riscv_get_register" to test on gpr "x0" , and get the wrong  result !
because,  the openocd treats the "x0" as a cacheable register and returns the value I wrote by "riscv_set_register" which is not zero
so we can see the log below  read the x0 : 12345678 (cached) , it's a little bit weird, isn't it ?

Image 45

I  think this is because openocd treat  "x0" as a cacheable register below
https://github.com/riscv/riscv-openocd/blob/d9c73e231e5144e05fe35496975c64817ac54555/src/target/riscv/riscv.c#L3676
and "riscv_get_register->riscv_get_register_on_hart" will return the cached value before the value can be read from the hardware when the register is cacheable
https://github.com/riscv/riscv-openocd/blob/d9c73e231e5144e05fe35496975c64817ac54555/src/target/riscv/riscv.c#L3767
so we get a  non-zero value from "x0"
Image 45

timsifive:
It doesn't have any effect on real hardware, and by caching the value we
pretended it did.

Fixes #564

Change-Id: I9f4e2cc8abddee61435bbd8d992cbff971a0c28d

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -4165,6 +4165,11 @@ static int register_set(struct reg *reg, uint8_t *buf)
 			riscv_current_hartid(target), str, reg->name, reg->valid);
 	free(str);
 
+	/* Exit early for writing x0, which on the hardware would be ignored, and we
+	 * don't want to update our cache. */
+	if (reg->number == GDB_REGNO_ZERO)
+		return ERROR_OK;
+
 	memcpy(reg->value, buf, DIV_ROUND_UP(reg->size, 8));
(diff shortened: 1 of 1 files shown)


JanMatCodasip:
This looks all right.

I've come across only a minor comment aside, which is submitted separately: https://github.com/riscv/riscv-openocd/pull/569

GregSavin (approved):

PR #569, opened 2021-01-15
riscv-013: Fix comment for register_read()

JanMatCodasip:
Updated comment for `register_read()` to reflect what that function really does.


timsifive (approved):
Thanks.

PR #570, opened 2021-01-20
Document how vector registers are exposed to gdb.

timsifive:
Change-Id: Ie7cdef3717e107a9df0b48316cfbc547dea9a7fd


timsifive:
@bruceable, FYI.

GregSavin (approved):
Looks good to me, as this is information outside the scope of RISC-V specs yet could be useful details (as an example) for the RISC-V tools ecosystem beyond the OpenOCD/GDB case.

PR #575, opened 2021-01-27
Add authdata_read/authdata_write support to 0.11.

timsifive:
AFAIK there is no hardware that implements this, but it should be a
close-to-done starting point in case it is ever required.

Also adds `dm.authenticated` entry to `riscv info` output.

Change-Id: I49e3082e8629b1d70b12e8a847c2848e75b04508

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10129,12 +10129,14 @@ set challenge [riscv authdata_read]
 riscv authdata_write [expr $challenge + 1]
 @end example
 
-@deffn Command {riscv authdata_read}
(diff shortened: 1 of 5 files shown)


GregSavin (approved):
OK so there are spec level differences between 0.11 and 0.13, and the feature hadn't been implemented in OpenOCD for 0.11 because of lack of hardware; makes sense to me and looks good.

RobertChyla:
I am aware about one implementation - we (IAR) added psuedo-registers (in out debugger, not OPENOCD ...) to allow user to play with authentication from script (before we connect to core).

PR #576, opened 2021-01-28
Add memory access while running to `riscv info`

timsifive:
This way people can write TCL scripts that rely on these more abstract
properties instead of having to check for the existence of sbaccess,
which is not part of 0.11. (There is a similar feature but things are
named differently.)

Change-Id: I5c95a29ef43cb40c3a73b904f11fa7ca38d87b21

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1878,6 +1878,19 @@ COMMAND_HELPER(riscv013_print_info, struct target *target)
 {
 	RISCV013_INFO(info);
 
+	/* Abstract description. */
+	riscv_print_info_line(CMD, "target", "memory.read_while_running8", get_field(info->sbcs, DM_SBCS_SBACCESS8));
+	riscv_print_info_line(CMD, "target", "memory.write_while_running8", get_field(info->sbcs, DM_SBCS_SBACCESS8));
+	riscv_print_info_line(CMD, "target", "memory.read_while_running16", get_field(info->sbcs, DM_SBCS_SBACCESS16));
+	riscv_print_info_line(CMD, "target", "memory.write_while_running16", get_field(info->sbcs, DM_SBCS_SBACCESS16));
+	riscv_print_info_line(CMD, "target", "memory.read_while_running32", get_field(info->sbcs, DM_SBCS_SBACCESS32));
+	riscv_print_info_line(CMD, "target", "memory.write_while_running32", get_field(info->sbcs, DM_SBCS_SBACCESS32));
+	riscv_print_info_line(CMD, "target", "memory.read_while_running64", get_field(info->sbcs, DM_SBCS_SBACCESS64));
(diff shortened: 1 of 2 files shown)


GregSavin (approved):
Interesting; somehow I wasn't aware that system bus access availability could depend on the specific access width; good to know.

src/target/riscv/riscv-013.c
```diff
@@ -1878,6 +1878,19 @@ COMMAND_HELPER(riscv013_print_info, struct target *target)
...
+	riscv_print_info_line(CMD, "target", "memory.read_while_running8", get_field(info->sbcs, DM_SBCS_SBACCESS8));
+	riscv_print_info_line(CMD, "target", "memory.write_while_running8", get_field(info->sbcs, DM_SBCS_SBACCESS8));
+	riscv_print_info_line(CMD, "target", "memory.read_while_running16", get_field(info->sbcs, DM_SBCS_SBACCESS16));
+	riscv_print_info_line(CMD, "target", "memory.write_while_running16", get_field(info->sbcs, DM_SBCS_SBACCESS16));
+	riscv_print_info_line(CMD, "target", "memory.read_while_running32", get_field(info->sbcs, DM_SBCS_SBACCESS32));
+	riscv_print_info_line(CMD, "target", "memory.write_while_running32", get_field(info->sbcs, DM_SBCS_SBACCESS32));
+	riscv_print_info_line(CMD, "target", "memory.read_while_running64", get_field(info->sbcs, DM_SBCS_SBACCESS64));
+	riscv_print_info_line(CMD, "target", "memory.write_while_running64", get_field(info->sbcs, DM_SBCS_SBACCESS64));
+	riscv_print_info_line(CMD, "target", "memory.read_while_running128", get_field(info->sbcs, DM_SBCS_SBACCESS128));
+	riscv_print_info_line(CMD, "target", "memory.write_while_running128", get_field(info->sbcs, DM_SBCS_SBACCESS128));
```
JanMatCodasip, line 1891: Just an idea/minor comment - could this all be replaced by a single boolean `1/0` item, e.g. `sbaccess_while_running`? 

This way, anyone needing this information in a script would still have it available for use (as a combination of `sbaccess_while_running` and `sbaccess##`).  And the output of "riscv info" would not get too verbose, so that it remains easily readable also for users.

timsifive: I don't want to mention "sbaccess" here because it implies a method that the access happens. RISC-V could also do this using the quick access method (which OpenOCD doesn't support, and I'm not aware of any hardware implementations of either). I also want to use the same term in 0.11 and 0.13, and the language changed between specs.

I thought about just having a single option, but then what does OpenOCD report for hardware where e.g. reading works while running but writing does not? The goal is to make it easy to write scripts that check if they can access memory in some way, and then do so (e.g. by setting up memory sampling).

I'm not a fan of how verbose this ended up. Does that make sense? Is there a better way that I overlooked?

JanMatCodasip: Thank you for explaining the reasons behind what the `riscv info` entries look like here. That makes perfect sense.

PR #577, opened 2021-01-29
Fix write_memory_bus_v1: incorrect getting sbcs state.

dmitryryzhov:
Function dmi_op with given "ensure_success == false" does not read back the result of DMI read operation. Thus data_in variable won't be updated and actual register's value won't be loaded. It will cause false positive decision and result to write failure without detection of an error.
Over here it causes false positive decision and results to write failure when DM_SBCS_SBBUSY occurs. Using dmi_op with "ensure_success == false" leads to outdated sbcs state.
And further "get_field(sbcs, DM_SBCS_SBBUSY) || dmi_busy" will result wrong decision.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3701,7 +3701,7 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
 
 		bool dmi_busy_encountered;
 		if (dmi_op(target, &sbcs, &dmi_busy_encountered, DMI_OP_READ,
-				DM_SBCS, 0, false, false) != ERROR_OK)
+				DM_SBCS, 0, false, true) != ERROR_OK)
 			return ERROR_FAIL;
 
 		time_t start = time(NULL);


JanMatCodasip:
Hi @dmitryryzhov,
Thanks for the fix. It seems we have been independently looking at the same issue (https://github.com/riscv/riscv-openocd/pull/578). Perhaps we can combine the patches.

src/target/riscv/riscv-013.c
```diff
@@ -3701,7 +3701,7 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
 
 		bool dmi_busy_encountered;
 		if (dmi_op(target, &sbcs, &dmi_busy_encountered, DMI_OP_READ,
-				DM_SBCS, 0, false, false) != ERROR_OK)
+				DM_SBCS, 0, false, true) != ERROR_OK)
```
JanMatCodasip, line 3704: If this `dmi_op` call has `ensure_success` set, then the follow-up while loop should no longer check for `dmi_busy`.

Also both the occurences of `dmi_op()` ([1](https://github.com/riscv/riscv-openocd/pull/577/files#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63L3703), [2](https://github.com/riscv/riscv-openocd/pull/577/files#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63R3717)) can be replaced by `dmi_read()`.

Other than that, the fix appears all right, and simpler than what I proposed under #578.

JanMatCodasip (changes requested):
-

src/target/riscv/riscv-013.c
dmitryryzhov: Hi @JanMatCodasip, thanks for your comments.
It seems that a better solution is in between our patches: you can revert `read_sbcs_nonbusy` call with a while loop, replace `dmi_op()` with `dmi_read()` in this loop and remove `dmi_busy` from loop clause as well. It will allow to avoid unnecessary `dmi_op` call in sbbusy-free case. 
Or I can do the same in my patch: replace loop's `dmi_op()` with `dmi_read()` and remove `dmi_busy` from cycle clause. But still I can't replace first `dmi_op` because we need to track `dmi_busy_encountered`.
What do you think about it?

JanMatCodasip: @dmitryryzhov - Thank you for your feedback. 
Agreed, the first dmi_op() call cannot be replaced by dmi_read(), I overlooked that one.

I have updated the patch at #578 per our discussion. Please let me know if you see anything else there that should be improved.

dmitryryzhov: @JanMatCodasip - your update looks good for me.
So I will close my pr and wait for #578 merge.

PR #578, opened 2021-01-29
Fix in write_memory_bus_v1: Read sbcs properly

JanMatCodasip:
Fixed an issue with system bus write: system bus access errors (sbbusyerror, sberror)
were not properly detected because of an incomplete read of "sbcs". The read of sbcs was 
initiated but not completed (value not acquired - missing a second DMI scan).

Minor changes:
- Added debug prints in case sberror != 0.
- Added few comments to explain the algorithm.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3700,25 +3700,31 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
 		if (result != ERROR_OK)
 			return result;
 
-		/* Deal with DMI busy state that could have arisen during the batch execution.
-		 * In the process, the following is handled:
-		 * - clear DMI busy sticky error, if it occurred
-		 * - increment of dmi_busy_delay (slow down)
-		 */
-		bool dmi_busy_encountered = false;
-		if (dmi_op(target, NULL, &dmi_busy_encountered, DMI_OP_NOP, 0, 0, false, true) != ERROR_OK)
+		/* Read sbcs value.
+		 * At the same time, detect if DMI busy has occurred during the batch write. */
+		bool dmi_busy_encountered;
+		if (dmi_op(target, &sbcs, &dmi_busy_encountered, DMI_OP_READ,
+				DM_SBCS, 0, false, true) != ERROR_OK)
 			return ERROR_FAIL;
 		if (dmi_busy_encountered)
-			LOG_DEBUG("DMI busy encountered during system bus write. Will retry.");
+			LOG_DEBUG("DMI busy encountered during system bus write.");
 
-		/* Wait until SBBUSY goes low */
-		if (read_sbcs_nonbusy(target, &sbcs) != ERROR_OK)
-			return ERROR_FAIL;
+		/* Wait until sbbusy goes low */
+		time_t start = time(NULL);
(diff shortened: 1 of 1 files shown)


JanMatCodasip:
Further comments:

1) I am leaving this pull request as "draft" for the moment as it should undergo at least riscv-tests/debug.

2) I realize that the fix slows down the system bus write a bit as it introduces additional DMI scans compared to the original code. It would be possible to minimize the DMI scan number by:
- testing for DMI busy during batch write and at the same time initiating the read of SBCS (by a single DMI transaction)
- reading the SBCS in a loop by `dmi_scan()`, testing for both DMI busy and sbcs.sbbusy != 0
This however requires more complicated code and direct use of `dmi_scan()`, as in the [sysbus read algorithm](https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv-013.c#L2712).

I am definitely open to suggestions how to lay out the code - whether this part of the code should be fine-tuned for performance, at the cost of a bit more complex code. 

Anyway, this sbcs read issue should be fixed one way or the other.

Thank you.

JanMatCodasip:
Updated per review discussion at #577.

JanMatCodasip:
I have run riscv-tests/debug on this fix (targets "spike64" and "spike64-2") - the tests are passing.

src/target/riscv/riscv-013.c
```diff
@@ -3694,54 +3694,76 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
-			 * something we did wrong. */
+		unsigned sberror = get_field(sbcs, DM_SBCS_SBERROR);
+		if (sberror != 0) {
+			/* Sberror indicates the bus access failed, but not because we issued the writes
+			 * too fast. Cannot recover. Sbaddress holds the address where the error occurred
+			 * (unless sbautoincrement in the HW is buggy).
+			 */
+			target_addr_t sbaddress = sb_read_address(target);
+			LOG_DEBUG("System bus access failed with sberror=%u (sbaddress=0x%" TARGET_PRIxADDR ")",
+					  sberror, sbaddress);
```
timsifive, line 3757: Should this be INFO or ERROR?

timsifive (approved):
This looks good. If I understand it right, it adds one scan per batch, which should not affect performance much.

JanMatCodasip:
> If I understand it right, it adds one scan per batch, which should not affect performance much.

Yes, compared to the previous code, this should add one scan.

src/target/riscv/riscv-013.c
JanMatCodasip: I left this as LOG_DEBUG as there are already another error prints that cover a failed memory access:

* Error if the memory read/write fails as whole (all enabled methods fail): [riscv-013.c#L3595](https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv-013.c#L3595)
* Warning in case the given method alone (system bus) fails for the first time: [riscv-013.c#L3589](https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv-013.c#L3589)

I'll change this print to higher level if you'd like, however the already present errors/warnings seem sufficient.

PR #580, opened 2021-02-10
From upstream

timsifive:
I'm trying to do this more often so the changes are smaller and everything is more manageable.


JanMatCodasip (approved):
I've skimmed through the changes from upstream - everything looks all right.

GregSavin (approved):
Looks good to me.  I notice target/riscv/riscv.c had a diff, with a zero initializer added.  I'm mildly curious whether that was to satisfy some lint check and/or coding standards, or if the previous code without the initializer was functionally misbehaving somehow.

timsifive:
@SiFiveGregS, that change was in https://github.com/riscv/riscv-openocd/commit/404993b29f1304fb53202fb88147298465bfb525 to fix a compiler warning.

PR #582, opened 2021-02-17
RISC-V Freertos support

timsifive:
On single-hart RV32, this exposes all FreeRTOS threads to gdb, and allows gdb to read all the registers that are stored on the stack for those threads. Other registers are read from the hardware directly.

There's certainly more to do, but this seems like a plenty big change to review, so let's do that before making it any larger.


timsifive:
This is still very early. I'm just making this PR to make sure it still builds on various compilers.

configure.ac
```diff
@@ -17,14 +17,10 @@ AC_SUBST([MAKEINFO])
 AM_INIT_AUTOMAKE([-Wall -Wno-portability dist-bzip2 dist-zip subdir-objects])
 
 AC_CONFIG_HEADERS([config.h])
-AH_BOTTOM([
-#include <helper/system.h>
-#include <helper/types.h>
-#include <helper/replacements.h>
-])
```
timsifive, line 24: Note: This is the change that required me to touch a bazillion .h files.

This change was required because otherwise compiling gnulib files gave me errors about not finding .h files. I couldn't see how to change things in the Makefiles to add the appropriate -I there without manually editing the auto-generated Makefile inside gnulib. These lines feel like a hack regardless, so removing them should be an improvement.

src/rtos/FreeRTOS.c
```diff
@@ -58,15 +57,15 @@ struct FreeRTOS_params {
...
-	struct list_head list;
 	threadid_t threadid;
 	target_addr_t tcb;
 };
 
 struct FreeRTOS {
 	const struct FreeRTOS_params *param;
 	unsigned thread_count;
-	struct list_head thread_list;
+	/* Map from tcb to FreeRTOS_thread_entry. */
```
GregSavin, line 69: So this is the motivation for pulling in the gnulib stuff, to get optimized performance on the lookups and maintenance of this thread_map?

timsifive: Basically. The real motivation is not the performance so much as not having to write the code myself. (There are actually lots of places within OpenOCD that could benefit from using well-implemented data structures, and assuming this gets accepted upstream I plan to use these structures more in the future.)

GregSavin: Makes sense.  I guess if upstream community balks at gnulib for some reason, there's a remedy at that time, but leveraging gnulib more widely seems like a good direction to try to move toward.

src/rtos/rtos.c
```diff
@@ -625,13 +628,58 @@ int rtos_generic_stack_read(struct target *target,
...
+			address -= stacking->stack_registers_size;
+
+		if (target_read_buffer(
+				target, address + offsets->offset,
+				width_bytes, reg->value) != ERROR_OK)
+			return ERROR_FAIL;
+		LOG_DEBUG("register %d has value 0x%" PRIx64, reg->number,
+				  buf_get_u64(reg->value, 0, 64));
+	} else {
+		memset(reg->value, 0, width_bytes);
```
GregSavin, line 677: Ah, so this must be where x0 reported value comes from.

src/rtos/rtos_standard_stackings.c
```diff
@@ -249,41 +291,50 @@ static int64_t rtos_standard_Cortex_M4F_FPU_stack_align(struct target *target,
...
+	.register_offsets = rtos_standard_NDS32_N1068_stack_offsets
+};
+
+const struct rtos_register_stacking rtos_standard_RV32_stacking = {
+	.stack_registers_size = (32 + 2) * 4,
+	.stack_growth_direction = -1,
+	.num_output_registers = 33,
+	.calculate_process_stack = rtos_generic_stack_align8,
+	.register_offsets = rtos_standard_RV32_stack_offsets,
+	.total_register_count = ARRAY_SIZE(rtos_standard_RV32_stack_offsets)
```
GregSavin, line 339: So is the distinction between num_output_registers and total_register_count that num_output_registers pertains to the registers that apply to the GDB 'g' packet, and total_register_count can also include additional registers that aren't part of the 'g' packet?

GregSavin (approved):
This is looking good, to me.  I'm just assuming that the gnulib code itself is very well tested.

src/rtos/rtos_standard_stackings.c
timsifive: Yes. It's a bit annoying that this level of the code needs to know about gdb packets, but that's the existing API and I don't want to go changing even more stuff right now.

src/rtos/rtos.c
timsifive: Right.

configure.ac
JanMatCodasip: @timsifive  - Hi Tim, it seems this change has further ripple-effect in other files. E.g. in usb_common.{c,h}, as reported here: https://github.com/riscv/riscv-openocd/issues/589. The issue can be reproduced with `./configure --enable-jlink [...]`. There may be more.

I wonder if there is a different solution that would allow to have gnulib and at the same time less impact on the existing code/headers. Lower amount of changes would also make the upstream acceptance easier.

timsifive: I would love it if there is a better solution, but I couldn't find one.

While this change is large, it's also very simple. I hadn't considered building with different configure options, but it's a simple matter to add the appropriate includes where they're needed.

PR #585, opened 2021-03-12
Support writing to FreeRTOS registers in suspended threads

timsifive:
Change-Id: Ie2a8bef5e9aa24fc85e9b6c9093021731c58c3e7

### DIFF
--- a/src/rtos/FreeRTOS.c
+++ b/src/rtos/FreeRTOS.c
@@ -190,6 +190,7 @@ static int FreeRTOS_get_thread_reg_list(struct rtos *rtos, threadid_t thread_id,
 		struct rtos_reg **reg_list, int *num_regs);
 static int FreeRTOS_get_thread_reg(struct rtos *rtos, threadid_t thread_id,
 		uint32_t reg_num, struct rtos_reg *reg);
+static int FreeRTOS_set_reg(struct rtos *rtos, uint32_t reg_num, uint8_t *reg_value);
 static int FreeRTOS_get_symbol_list_to_lookup(symbol_table_elem_t *symbol_list[]);
 
 struct rtos_type FreeRTOS_rtos = {
@@ -200,6 +201,7 @@ struct rtos_type FreeRTOS_rtos = {
 	.update_threads = FreeRTOS_update_threads,
 	.get_thread_reg_list = FreeRTOS_get_thread_reg_list,
 	.get_thread_reg = FreeRTOS_get_thread_reg,
+	.set_reg = FreeRTOS_set_reg,
 	.get_symbol_list_to_lookup = FreeRTOS_get_symbol_list_to_lookup,
 };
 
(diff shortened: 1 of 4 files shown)


src/rtos/FreeRTOS.c
```diff
@@ -585,6 +587,24 @@ static int FreeRTOS_get_thread_reg(struct rtos *rtos, threadid_t thread_id,
 	return rtos_generic_stack_read_reg(rtos->target, stacking_info, stack_ptr, reg_num, reg);
 }
 
+static int FreeRTOS_set_reg(struct rtos *rtos, uint32_t reg_num, uint8_t *reg_value)
+{
+	LOG_DEBUG("[%" PRId64 "] reg_num=%" PRId32, rtos->current_threadid, reg_num);
+
+	/* Let the caller write registers directly for the current thread. */
+	if (rtos->current_threadid == rtos->current_thread)
+		return ERROR_FAIL;
```
GregSavin, line 596: Ah, OK, so this works because OpenOCD's gdb stub code gives this RTOS handler first chance at it, and returning non-OK status is essentially the RTOS handler saying "I didn't handle this"

src/rtos/rtos.c
```diff
@@ -680,6 +680,58 @@ int rtos_generic_stack_read_reg(struct target *target,
...
+			return ERROR_FAIL;
+	} else if (offsets->offset == -1) {
+		/* This register isn't on the stack, but is listed as one of those. We
+		 * read it as 0, and ignore writes. */
+	} else if (offsets->offset == -2) {
+		/* This register requires computation when we "read" it. I'm not sure
+		 * how to handle writes. We can't simply return error here because then
+		 * the higher level code will end up writing the register in the halted
+		 * core, which is definitely not the same as writing it for a thread. */
+		LOG_ERROR("Don't know how to write register %d with offset -2 in a thread.",
```
GregSavin, line 724: One thing that isn't clear to me is why some of the non-RISC-V stackings use encoding -2 for the stack pointer offset, and others, including RV32, use a normal offset for stack pointer.  Is it that some architectures don't keep the stack pointer in a normal accessible register, and for those cases, within OpenOCD it's exposed a "pseudo" register, or something like that?   But RISC-V does have a literal stack pointer register, so normally for RISC-V we shouldn't ever see this particular code getting reached?

GregSavin (approved):

src/rtos/FreeRTOS.c
timsifive: Yes. It would be better if there was a special ERROR code to indicate this (e.g. ERROR_NOT_IMPLEMENTED), but there are only so many target-independent changes I want to make at once.

src/rtos/rtos.c
timsifive: Exactly.

> But RISC-V does have a literal stack pointer register, so normally for RISC-V we shouldn't ever see this particular code getting reached?

Funny you mention that. Just yesterday I discovered that mainline FreeRTOS puts registers on the stack in a different order than the Freedom Metal FreeRTOS fork, and mainline FreeRTOS in fact does not put the stack pointer on the stack, but stores it in the task control block instead. I haven't figured out yet how I want to handle this.

PR #586, opened 2021-03-17
Support RV32/RV64 mainline/metal stackings

timsifive:
Add an option that lets a user choose between metal and mainline FreeRTOS stackings (defaults to mainline).

Add support for RV64.

### DIFF
--- a/src/rtos/FreeRTOS.c
+++ b/src/rtos/FreeRTOS.c
@@ -279,8 +279,7 @@ static int FreeRTOS_read_struct_value(
 }
 
 typedef struct {
-	enum
-	{
+	enum {
 		TYPE_POINTER,
 		TYPE_UBASE,
 		TYPE_TICKTYPE,
@@ -327,8 +326,7 @@ static unsigned populate_offset_size(struct FreeRTOS *freertos,
 
 		largest = MAX(largest, align);
 
-		if (offset & (align - 1))
-		{
+		if (offset & (align - 1)) {
 			offset = offset & ~(align - 1);
 			offset += align;
 		}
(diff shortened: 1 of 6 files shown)


t-8ch:
FYI this works for me (FreeRTOS 10.3.1)!

At least after having adapted a few offsets, too:
```diff
@@ -176,12 +179,12 @@ static const struct FreeRTOS_params FreeRTOS_params_list[] = {
        "riscv",                        /* target_name */
        4,                                              /* thread_count_width; */
        4,                                              /* pointer_width; */
-       16,                                             /* list_next_offset; */
-       20,                                             /* list_width; */
+       4,                                              /* list_next_offset; */
+       24,                                             /* list_width; */
        8,                                              /* list_elem_next_offset; */
-       12,                                             /* list_elem_content_offset */
+       16,                                             /* list_elem_content_offset */
        0,                                              /* thread_stack_offset; */
-       52,                                             /* thread_name_offset; */
+       64,                                             /* thread_name_offset; */
```

t-8ch:
One issue that cropped up with this applied are with breakpoints.
Setting the breakpoint works correctly and it is correctly hit it will instantly rehit the same breakpoint over and over again.
Removing the breakpoint allows the code to continue.

timsifive:
@t-8ch, these offsets don't all make sense to me. How did you arrive at these new values? Did you use build options like configUSE_LIST_DATA_INTEGRITY_CHECK_BYTES that would affect the layout of these structures? (What is TickType_t in your project, and UBaseType_t?)

t-8ch:
@timsifive No that specifc value is not set, but other integrity options are.
( https://github.com/Ralim/IronOS/blob/master/source/Core/BSP/Pine64/Vendor/OS/FreeRTOS/Source/portable/GCC/portmacro.h )
I got these values manually by looking at the structure definitions in gdb.

I also wrote to the gdb mailing list to have a way of directly retrieving the offsets from the debug info via gdb:
https://sourceware.org/pipermail/gdb/2021-March/049329.html

src/rtos/FreeRTOS.c
```diff
@@ -33,26 +33,13 @@
...
-
-/* FIXME: none of the _width parameters are actually observed properly!
- * you WILL need to edit more if you actually attempt to target a 8/16/64
- * bit target!
- */
+#define FREERTOS_THREAD_NAME_STR_SIZE 200
 
 struct FreeRTOS_params {
 	const char *target_name;
-	const unsigned char thread_count_width;
```
GregSavin, line 46: I'm curious of the rationale for moving these data members to struct FreeRTOS; is it because those fields get assigned by querying RTOS data structures at runtime?  (As opposed to being parameters provided in OpenOCD init scripts etc?

GregSavin (approved):

src/rtos/FreeRTOS.c
timsifive: That's exactly why. Specifically, FreeRTOS_params is defined as const, and it didn't seem right to make it writable.

PR #588, opened 2021-03-31
Update version check error message from 0.14 to 1.0.

ernie-sifive:
The latest Debug Spec draft has been promoted to "stable" status and is now called 1.0 instead of 0.14.


timsifive (approved):
Thanks!

PR #593, opened 2021-04-09
Don't check target_address_bits() until we need it.

### LINKED ISSUE #591: openocd: src/rtos/FreeRTOS.c:324: populate_offset_size: Assertion `info[i].size > 0' failed
Since 737f013ada404fdd620c2c0d4f1cb318a0ff64b2 I receive the error from the subject directly after starting when using `target/gd32vf103.cfg`.

It seems that
* `freertos->ubasetype_size == -1`
* because `riscv_data_bits() == -1`
* because `riscv_xlen() == -1`
* because it's not overwritten after `riscv_info_init()`.

Steps to reproduce:
```tcl
# test.cfg
source [find interface/dummy.cfg]
source [find target/gd32vf103.cfg]
$_TARGETNAME configure -rtos FreeRTOS
```
```sh
$ openocd -f test.cfg
Open On-Chip Debugger 0.11.0-rc2+dev-00994-g50a5971be (2021-04-09-17:15)
Licensed under GNU GPL v2
For bug reports, read
Info : only one transport option; autoselect 'jtag'
openocd: src/rtos/FreeRTOS.c:324: populate_offset_size: Assertion `info[i].size > 0' failed.
Aborted (core dumped)
```

timsifive:
Fixes #591, and also makes `-rtos FreeRTOS` work (in addition to `-rtos
auto`).

Change-Id: Ieafe703a2929a0959fcc81997096e8beeeb42496

### DIFF
--- a/src/rtos/FreeRTOS.c
+++ b/src/rtos/FreeRTOS.c
@@ -345,6 +345,74 @@ static unsigned populate_offset_size(struct FreeRTOS *freertos,
 	return offset;
 }
 
+static void FreeRTOS_compute_offsets(struct rtos *rtos)
+{
+	struct FreeRTOS *freertos = (struct FreeRTOS *) rtos->rtos_specific_params;
+
+	if (freertos->pointer_size != 0)
+		return;
+
+	freertos->pointer_size = DIV_ROUND_UP(target_address_bits(rtos->target), 8);
+	freertos->ubasetype_size = DIV_ROUND_UP(target_data_bits(rtos->target), 8);
+
+	/*
+	 * FreeRTOS can be compiled with configUSE_LIST_DATA_INTEGRITY_CHECK_BYTES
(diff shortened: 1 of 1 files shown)


t-8ch:
Looks good.

Is this new offset computation work intended to be upstreamed?
I'd like to extend it with a mechanism to read this information also from GDB (as outline at https://sourceware.org/pipermail/gdb/2021-March/049329.html ) and would like to know on which tree to base this on.

GregSavin (approved):

timsifive:
@t-8ch, I will definitely upstream this, but historically I haven't been very fast about it. I realize I've rewritten most of the FreeRTOS code at this point, so I'm going to try to be better about it. If you're making changes in mainline we're guaranteed large conflicts. Hopefully the way I wrote this makes it easier for you to tie in the values you can query out of gdb than in the original design.

I will do my best to start upstreaming some of this this week.

PR #594, opened 2021-04-13
Merge mainline down again.

timsifive:


GregSavin (approved):
These merges tend to seem like a leap of faith to some extent, but necessary (and always interesting to see what has been going on upstream)

PR #595, opened 2021-04-16
Fix handeling of ebreak

ewft:
According to RISC-V Spec 0.13 - 4.8.2, value of `dpc `when debug cause is an `ebreak` is the address of the `ebreak`.
Actually when an `ebreak` cause a debug halt the `dpc` is not modified so when execution is resumed the `ebreak` cause again a debug halt. ( Except when semihosting without semihosting_fileio activated ) 
So in order to continue execution after an `ebreak`, `dpc` need to be set to the next instruction.

This patch set `dpc` for the next instruction when the debug cause is `ebreak` and remove it from the semihosting special case.

This also fix #433 in a more general and cleaner way.
Concerning semi-hosting the issue is not detected by riscv-test because `semihosting_fileio` is not enabled.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1356,7 +1356,39 @@ static int resume_prep(struct target *target, int current,
 
 		if (enable_triggers(target, trigger_state) != ERROR_OK)
 			return ERROR_FAIL;
+
+	} else if (target->debug_reason == DBG_REASON_BREAKPOINT) {
+
+		riscv_reg_t dpc;
+		riscv_get_register(target, &dpc, GDB_REGNO_DPC);
+		const uint8_t length = 4;
+		LOG_DEBUG("dpc is 0x%" PRIx64, dpc);
+
+		/* fetch the instruction at dpc */
+		uint8_t buffer[length];
+		if (target_read_buffer(target, dpc, length, buffer) != ERROR_OK) {
+			LOG_ERROR("Failed to read instruction at dpc 0x%" PRIx64, dpc);
+			return ERROR_FAIL;
+		}
+
+		uint32_t instruction = 0;
+
+		for (int i = 0; i < length; i++) {
+			LOG_DEBUG("Next byte is %x", buffer[i]);
+			instruction += (buffer[i] << 8 * i);
+		}
+		LOG_DEBUG("Full instruction is %x", instruction);
+
+		if (instruction == MATCH_EBREAK)
+			dpc = dpc + 4;
+		else if ((instruction & MASK_C_EBREAK) == MATCH_C_EBREAK)
+			dpc = dpc + 2;
+
+
+		riscv_set_register(target, GDB_REGNO_DPC, dpc);
+
 	}
+
 
 	if (r->is_halted) {
 		if (riscv_resume_prep_all_harts(target) != ERROR_OK)
--- a/src/target/riscv/riscv_semihosting.c
+++ b/src/target/riscv/riscv_semihosting.c
@@ -159,9 +159,6 @@ semihosting_result_t riscv_semihosting(struct target *target, int *retval)
 	 */
 	if (semihosting->is_resumable && !semihosting->hit_fileio) {
 		/* Resume right after the EBREAK 4 bytes instruction. */
-		*retval = riscv_set_register(target, GDB_REGNO_PC, pc + 4);
-		if (*retval != ERROR_OK)
-			return SEMI_ERROR;
 
 		LOG_DEBUG("   -> HANDLED");
 		return SEMI_HANDLED;


src/target/riscv/riscv.c
```diff
@@ -1356,8 +1356,40 @@ static int resume_prep(struct target *target, int current,
...
+			instruction += (buffer[i] << 8 * i);
+		}
+		LOG_DEBUG("Full instruction is %x", instruction);
+
+		if (instruction == MATCH_EBREAK)
+			dpc = dpc + 4;
+		else if ((instruction & MASK_C_EBREAK) == MATCH_C_EBREAK)
+			dpc = dpc + 2;
+
+
```
JanMatCodasip, line 1387: Please do not use more than one blank line (per OpenOCD's upstream coding style). 
Also do not forget to run `tools/checkpatch.sh` on the committed changes.

JanMatCodasip (commented):
Hi Ben,

Thank you for the patch.

>  So in order to continue execution after an ebreak, dpc need to be set to the next instruction.

I am afraid this is not a good solution. It all boils down to why the ebreak is in the code in the first place. There are multiple purposes that ebreaks can have:

1) Ebreaks due to syscalls (recognized as such because they are part of the "magic sequence") 

- These must be handled transparently within OpenOCD and the core resumed, so the dpc+4 logic is correct here.
- It is already part of the current OpenOCD code, no need to change anything.

2) Ebreaks placed as software breakpoints

- These are typically managed by the debuggers (GDB, LLDB)
- If debugger wants to resume the core after hitting a SW break, it already contains the logic that: removes the ebreak (during which OpenOCD restores the original instruction), issues one step, restores the breakpoint, then finally resumes the processor (all transparent to the user).
- So no need to change anything here, either.
- Even if the user works with OpenOCD directly (using low-level TCL commands, without debugger), it is their responsibility to remove the breakpoint (`rbp` coimmand) before continuing.

3) Ebreaks in the code for other reason.

- If the programmer puts ebreaks into the program that are not syscalls nor SW breaks managed by OpenOCD, let's not step over them automatically. If the user runs into any such ebreak during debugging, it is their responsibility to analyze the situation and react on it accordingly (e.g. manual increment of DPC / "set PC to cursor" in visual IDEs). 

The main risk with this automation is the fact that SW breaks could be stepped-over without executing the original instruction.

I am sorry, but I cannot recommend this for a merge.

Maybe if you can show us the use case that led to this patch, we can come up with a different solution.

ewft:
Hi Jan,

Thanks for the detailed comment 

The use case here is SW breakpoint in flash code. If I am not wrong it cannot use the OpenOCD SW breakpoints mechanism because it require to modify the content of the memory which is not really feasible with flash memory. 

With limited number of HW breakpoint (trigger), we use the ebreak instruction in our code to increase the number of breakpoint available.

Maybe a dedicated option can be used to enable or disable this behavior ?

Concerning the syscalls (1), the actual OpenOCD code only works when the semihosting_fileio is not enabled. If the semihosting_fileio is enable the code is stuck on the ebreak.

JanMatCodasip:
> The use case here is SW breakpoint in flash code. If I am not wrong it cannot use the OpenOCD SW breakpoints mechanism because it require to modify the content of the memory which is not really feasible with flash memory.
> 
> With limited number of HW breakpoint (trigger), we use the ebreak instruction in our code to increase the number of breakpoint available.

Correct, SW breakpoints in flash are not supported by OpenOCD. (As a note, [third-party tools](https://www.segger.com/products/debug-probes/j-link/technology/flash-breakpoints/) exist that do support SW breaks in flash memories, at the cost of erasing and re-writing a whole flash sector.)

To step over custom hand-made ebreaks (not SW breaks that OpenOCD or debugger know of), maybe custom event handlers in OpenOCD could help? These are the events for which custom hooks can be registered in OpenOCD: http://openocd.org/doc/html/CPU-Configuration.html#Target-Events For your case, the `step-start` or `resume-start` events could work. The detection of ebreak + modification of DPC can be done within the event hook, just prior step or resume.

> Concerning the syscalls (1), the actual OpenOCD code only works when the semihosting_fileio is not enabled. If the semihosting_fileio is enable the code is stuck on the ebreak.

Ok, thank you for explanation. I haven't yet tried the fileio part of semihosting. So I take it the original issue #433 is still valid?

src/target/riscv/riscv.c
ewft: I ran the script, so I assume it was good. I will be more careful next time :-) Or it can be added to to the script .

ewft:
Thanks for the hint on the hook, didn't thought about it. Will look about it.

For #433, yes still valid. but don't actually have the time/resources to implement proper test to show it. ( activating `semihosting_fileio` may be enougth to reproduce in the actual risc-v test suite)

I think we can close this PR.

JanMatCodasip:
> I think we can close this PR.

Closing this PR, as discussed previously.

PR #600, opened 2021-05-03
From upstream

timsifive:
More changes from upstream.

Change https://sourceforge.net/p/openocd/code/ci/7dd323b26d93e49e409e02053e30f53ac8138cd5/ cut remote bitbang performance (when talking to the spike RISC-V simulator) approximately in half. This change "removes the file write stream, replaces it with socket write calls." Previously performance was better because the individual byte-size writes were buffered, and didn't result in a system call until fflush() was called.


GregSavin (approved):
There's an ftdi.c adjustment that looks interesting.  From logs, it seems that change was made to accommodate ST On Chip Emulator.  The change adds a JTAG state traversal path optimization case that also acts as a workaround for some issue in that emulator.  It looks like the change shouldn't cause problems with any spec-compliant JTAG TAP.  I'm hoping that the proprietary BSCAN tunneling hardware does not rely on the previous unoptimized state traversal path (reliance would represent a bug in the tunneling hardware or the way it is being driven).  I'll make a note to look into that question (I'm fuzzy on the details of the proprietary BSCAN tunnel but hopefully it copes just fine with this -- off-hand I'm not aware of a specific reason it would cause breakage in that area).

timsifive:
Thanks for reviewing, and for looking into that.

PR #603, opened 2021-05-12
Keep alive when accessing registers.

timsifive:
With the bitbang performance decrease and talking to slower targets
(daisy chain spike) we more often don't meet the expected keepalive time
for gdb. This addresses the cases I ran into.

Change-Id: Ie69c2c602c3be9c156e508fdfa6d0178f104e1d8

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3513,6 +3513,8 @@ int riscv_set_register(struct target *target, enum gdb_regno regid, riscv_reg_t 
 	LOG_DEBUG("[%s] %s <- %" PRIx64, target_name(target), gdb_regno_name(regid), value);
 	assert(r->set_register);
 
+	keep_alive();
+
 	/* TODO: Hack to deal with gdb that thinks these registers still exist. */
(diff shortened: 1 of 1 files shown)


JanMatCodasip (approved):
The changes look fine. I guess the timeout was exceeded when processing `g` message from GDB which reads multiple registers in one batch, was that the case?

GregSavin:
I suppose increasing the "set remotetimeout" on the GDB side is a workaround for the daisy-chain case without adding keep_alive calls, but from studying OpenOCD code, it appears that OpenOCD culture tries to keep things working when GDB's default "set remotetimeout" value is in force.  Guess it could still be a brittle solution if it's possible to conjure up a topology with even larger daisy chains and slower TCKs to once again push it over the time threshold, but hopefully that's unlikely to happen in practice (and if it did happen then there's still the workaround of bumping up "set remotetimeout")

GregSavin (approved):

timsifive:
Thanks for the review.

The timeout I ran into actually occurred when OpenOCD first needed to enumerate the triggers (which involves a bunch of CSR accesses). I think the 'g' command already has a keepalive() call somewhere.

PR #604, opened 2021-05-14
Remote bitbang buf

timsifive:
Implement a write buffer in remote_bitbang.c. This restores most of the performance that was lost when the file was changed to add Windows support.

My testcase has been the ChecKMisa test against dual-hart spike64. Before the Windows change this took 2.2s. After it became 5.78s. With this change the time is down to 2.92s again. I'd like to get all the performance back but don't want to spend the time right now.

### DIFF
--- a/src/jtag/drivers/remote_bitbang.c
+++ b/src/jtag/drivers/remote_bitbang.c
@@ -203,7 +203,9 @@ static int remote_bitbang_write(int tck, int tms, int tdi)
 static int remote_bitbang_reset(int trst, int srst)
 {
 	char c = 'r' + ((trst ? 0x2 : 0x0) | (srst ? 0x1 : 0x0));
-	return remote_bitbang_queue(c, false);
+	/* Always flush the send buffer on reset, because the reset call need not be
+	 * followed by jtag_execute_queue(). */
+	return remote_bitbang_queue(c, true);
 }
 
 static int remote_bitbang_blink(int on)
@@ -354,6 +356,10 @@ static const struct command_registration remote_bitbang_command_handlers[] = {
 
 static int remote_bitbang_execute_queue(void)
 {
+	/* safety: the send buffer must be empty, no leftover characters from
+	 * previous transactions */
+	assert(remote_bitbang_send_buf_used == 0);
+
 	/* process the JTAG command queue */
 	int ret = bitbang_execute_queue();
 	if (ret != ERROR_OK)
--- a/src/jtag/drivers/remote_bitbang.c
+++ b/src/jtag/drivers/remote_bitbang.c
@@ -36,6 +36,8 @@ static char *remote_bitbang_host;
 static char *remote_bitbang_port;
 
 static int remote_bitbang_fd;
+static uint8_t remote_bitbang_send_buf[512];
+static unsigned int remote_bitbang_send_buf_used;
 
 /* Circular buffer. When start == end, the buffer is empty. */
 static char remote_bitbang_buf[64];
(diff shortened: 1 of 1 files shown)


src/jtag/drivers/remote_bitbang.c
```diff
@@ -36,6 +36,8 @@ static char *remote_bitbang_host;
 static char *remote_bitbang_port;
 
 static int remote_bitbang_fd;
+static uint8_t remote_bitbang_send_buf[512];
+static unsigned int remote_bitbang_send_buf_used;
 
 /* Circular buffer. When start == end, the buffer is empty. */
 static char remote_bitbang_buf[64];
```
JanMatCodasip: Now that there is a new send_buf, you may want to rename this one to `remote_bitbang_recv_buf` or similar.

src/jtag/drivers/remote_bitbang.c
```diff
@@ -90,20 +92,37 @@ static int remote_bitbang_fill_buf(void)
 	return ERROR_OK;
 }
 
-static int remote_bitbang_putc(int c)
+static int remote_bitbang_flush(void)
```
JanMatCodasip, line 95: Currently, this flush function is called:
- 1. on read_sample()
- 2. when remote_bitbang_queue() is called and the send buffer is full
- 3. when remote_bitbang_queue(ch, true) is used

IMO, one more case should be covered - if there are still pending characters in the send buffer after the whole JTAG command queue has been processed.

This could be done e.g. by defining own `remote_bitbang_execute_queue()` which would call the underlying function in bitbang.c and then flush the buffer when everything is done:

```c
int remote_bitbang_execute_queue(void)
{
    /* safety: the send buffer must be empty, no leftover characters from previous transactions */
    assert(remote_bitbang_send_buf_used == 0);
    
    /* process the JTAG command queue */
    int ret = bitbang_execute_queue();
    if (ret != ERROR_OK)
        return ret;
        
    /* flush not-yet-sent characters, if any */
    return remote_bitbang_flush();
}

```

JanMatCodasip (changes requested):
The changes look all right. I just believe the flush should be performed in one more spot, in addition to what is already implemented.

src/jtag/drivers/remote_bitbang.c
timsifive: Done.

src/jtag/drivers/remote_bitbang.c
timsifive: Done. (I removed the assert because it always fired. I did not take the time to investigate why.)

JanMatCodasip: I am sorry, but the failing assert should not be dropped just like that. It means there is still a hidden bug in this merge request. 

I took a moment to look at it. The catch is that OpenOCD can call the `.reset` callback of `struct adapter_driver` directly, without the subsequent call to `jtag_execute_queue()`. (See: https://github.com/riscv/riscv-openocd/blob/riscv/src/jtag/core.c#L814) As a result, the reset command ('r' character) is buffered and not sent out, causing `remote_bitbang_send_buf_used==1`, which fires the assert when the JTAG queue is next executed. 

Flushing the queue on reset should fix it:
```
diff --git a/src/jtag/drivers/remote_bitbang.c b/src/jtag/drivers/remote_bitbang.c
index 214fa43fc..43452766b 100644
--- a/src/jtag/drivers/remote_bitbang.c
+++ b/src/jtag/drivers/remote_bitbang.c
@@ -203,7 +203,9 @@ static int remote_bitbang_write(int tck, int tms, int tdi)
 static int remote_bitbang_reset(int trst, int srst)
 {
        char c = 'r' + ((trst ? 0x2 : 0x0) | (srst ? 0x1 : 0x0));
-       return remote_bitbang_queue(c, false);
+       /* Always flush the send buffer on reset, because the reset call
+          need not be followed by jtag_execute_queue(). */
+       return remote_bitbang_queue(c, true);
 }
```

The assert can then stay in place.

Hope this helped.

timsifive: Fixed. Thank you for tracking that down.

JanMatCodasip (approved):
Thank you for fixing the last remaining pieces. The code looks fine.

PR #606, opened 2021-05-16
ARRAY_SIZE and unused variable

borneoa:
I have split the patch for ARRAY_SIZE in two parts:
the first covers only code not upstreamed and should be applied on riscv repository only
the second covers the code already upstreamed and should be applied on both riscv and upstream repositories.


borneoa:
Two of these patches already pushed on upstream gerrit as:
http://openocd.zylin.com/6257
http://openocd.zylin.com/6258
but are both waiting for review/approval here

wxjstz (approved):

wxjstz (approved):

wxjstz (approved):

JanMatCodasip (approved):
Thank you, the changes are fine.

timsifive (approved):
Thank you for doing this cleanup.

PR #607, opened 2021-05-18
riscv: Add assert for debug buffer

wxjstz:
The current debug buffer uses a fixed maximum length, and the new version of
the debugging protocol may be changed in the future. Here, add assert for
error detection.

Change-Id: I7187b489173b8b74e8483a892dbc9a46edf46157
Signed-off-by: Xiang W <wxjstz@126.com>

### DIFF
--- a/src/target/riscv/program.c
+++ b/src/target/riscv/program.c
@@ -23,6 +23,7 @@ int riscv_program_init(struct riscv_program *p, struct target *target)
 	for (size_t i = 0; i < RISCV_REGISTER_COUNT; ++i)
 		p->writes_xreg[i] = 0;
 
+	assert(RISCV_MAX_DEBUG_BUFFER_SIZE >= riscv_debug_buffer_size(target));
 	for (size_t i = 0; i < RISCV_MAX_DEBUG_BUFFER_SIZE; ++i)
 		p->debug_buffer[i] = -1;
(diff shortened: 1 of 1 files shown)


wxjstz:
The same patch at http://openocd.zylin.com/#/c/6265/

timsifive:
This doesn't look right. RISCV_MAX_DEBUG_BUFFER_SIZE really refers to the size of the debug_buffer[] inside OpenOCD. When the target has a larger debug buffer than that we should not assert, but simply debug it as usual while using no more than RISCV_MAX_DEBUG_BUFFER_SIZE entries in the debug buffer. The check at the start of riscv_program_insert() should probably be updated to also take RISCV_MAX_DEBUG_BUFFER_SIZE into account.

wxjstz:
> This doesn't look right. RISCV_MAX_DEBUG_BUFFER_SIZE really refers to the size of the debug_buffer[] inside OpenOCD. When the target has a larger debug buffer than that we should not assert, but simply debug it as usual while using no more than RISCV_MAX_DEBUG_BUFFER_SIZE entries in the debug buffer. The check at the start of riscv_program_insert() should probably be updated to also take RISCV_MAX_DEBUG_BUFFER_SIZE into account.

Yes! You are right.

PR #608, opened 2021-05-19
remove needless function declaration from src/target/riscv/program.h

wxjstz:


JanMatCodasip (approved):
Ok, so these are leftover declarations for functions that no longer exist. Thanks for the cleanup.

PR #611, opened 2021-05-19
Add keepalive for vector register access.

timsifive:
(And whatever else does a lot of register writes.)

Change-Id: I86a1a784fb7b9430aa470dbb39a495b89f56d8c9


GregSavin (approved):
Curious whether GDB protocol timeouts have been seen involving vector register accesses in single-DM topologies, or just in chained TAP scenarios.  This looks like the right thing to do, in either case though.

timsifive:
I've only seen it in simulated multi-TAP setup. But I don't think there's really any cost to calling keep_alive() since that function checks the time to see if it's actually necessary to communicate with gdb.

PR #613, opened 2021-05-24
Refactor opcodes.h

wxjstz:
Simplify the code and prevent some field of the instruction from overflowing

Change-Id: I07f3526a35ca87cc58e24851832c0512b472044e

### DIFF
--- a/src/target/riscv/opcodes.h
+++ b/src/target/riscv/opcodes.h
@@ -17,214 +17,202 @@ static uint32_t bit(uint32_t value, unsigned int b)
 	return (value >> b) & 1;
 }
 
+static uint32_t inst_rd(uint32_t r) __attribute__ ((unused));
+static uint32_t inst_rd(uint32_t r)
+{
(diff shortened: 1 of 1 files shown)


src/target/riscv/opcodes.h
```diff
@@ -17,214 +17,163 @@ static uint32_t bit(uint32_t value, unsigned int b)
 	return (value >> b) & 1;
 }
 
+#define INST_RD(r)	(bits(r, 4, 0) << 7)
```
timsifive: Can you make these static functions instead of macros? Modern compilers will optimize regardless, and provide better type checking etc.

timsifive (changes requested):
That make the opcode functions much more readable. Thank you.

src/target/riscv/opcodes.h
wxjstz: Done

timsifive (approved):
Looks good, and more readable than before. Thank you.

PR #615, opened 2021-05-27
Remove gnulib.

timsifive:
Just use regular lists in FreeRTOS.c. O(n) lookups should be fine.

Change-Id: I5dc7b2fd209e0214a87d73bd6c4b63fd664bf49a


GregSavin (approved):
Ah, yes I remember this, so upstream balked at gnulib.  Yeah I guess the number of FreeRTOS tasks would have to be very large for the linear search to get human-discernably slow.  Seems unlikely.

PR #616, opened 2021-06-03
Call keep_alive() more often.

timsifive:
When it doesn't do anything (most of the time) it has negligible
performance impact. With slower remote bitbang, and multiple spike
instances being tested in a single chain, things are sufficiently slow
that if a computer is busy then this is required to pass
riscv-tests/debug.

Change-Id: I8816efedaa0cc3b25734ba8fdc979ee4502284a1


GregSavin (approved):
Explanation makes perfect sense to me.

PR #619, opened 2021-06-04
batch: Not retrieving DR when writing to DMI

cinlyooi-intel:
Indicate to the JTAG driver developer that he does not need to read and return the DR register value after scanning the JTAG chain.

riscv_batch_run(), calls jtag_add_dr_scan() to schedule a DR scan operation.  Eventually, this will result in the JTAG driver performing a JTAG scan to write to or read from the DR. The decision on whether to write to and/or read from the JTAG register is determined by the second parameter to jtag_add_dr_scan(), i.e.  a "struct scan_field". Of particular interest here is if batch->fields[i]->in_value is not NULL, the JTAG developer must return the DR value collected from the JTAG DR scan operation.

When creating the DR scan operation instruction with riscv_batch_add_dmi_write(), batch->fields[i]->in_value is pointed to a location in batch->data_in buffer, meaning batch->field[i]->in_value is not NULL, and the JTAG developer must read and return the DR value collected. The returning on the DR value is redundant in a write operation.

This patch set batch->fields[i]->in_value to NULL to indicate the DR value need not be returned. This allows the  JTAG developer to optimize away any code associated with returning the DR value.

Normally, the extra work to return the DR value is negligible. However, in some use case it introduces delays that can be significant. There is a real use case where this makes a different. In this use case a JTAG driver forwards all JTAG scan  to a server on a network. If the server has to return the DR value, it has to perform the JTAG scan before replying to the JTAG driver, and only then the JTAG driver can send the next request. However, if there is no need to return the DR value, the server can just acknowledge the JTAG operation and the JTAG driver is free to send the next JTAG scan operation.  The server is able to receive the next JTAG operation while at the same time process the original JTAG scan. This saves time and mitigate network delay. Also, not having to include the DR value in resulting in smaller eply packet from server to JTAG driver and save on network traffic.

I could had deleted "riscv_fill_dmi_nop_u64(batch->target, (char *)field->in_value);"  right before the new code introduced in the patch. I did not because I am not full clear about the consequence of doing so even though I am expecting it to be benign. I don't find it to have much impact to the overall speed of the JTAG driver yet, so I decided to keep it.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -122,6 +122,7 @@ void riscv_batch_add_dmi_write(struct riscv_batch *batch, unsigned address, uint
 	field->in_value  = (void *)(batch->data_in  + batch->used_scans * DMI_SCAN_BUF_SIZE);
 	riscv_fill_dmi_write_u64(batch->target, (char *)field->out_value, address, data);
 	riscv_fill_dmi_nop_u64(batch->target, (char *)field->in_value);
+	field->in_value = NULL;
 	batch->last_scan = RISCV_SCAN_TYPE_WRITE;
 	batch->used_scans++;
 }


cinlyooi-intel:
Oops, a lot of mistakes in the commit message. I will redo it when the review is complete. Sorry

timsifive (commented):
Please do remove the call to riscv_fill_dmi_nop_u64() and simply init in_value to NULL a few lines above.

Can you share some before/after numbers? You say the impact is small, and that is fine. If all we gain is removing a line of code the PR is still worth it. :-)

What testing have you done?

cinlyooi-intel:
Hi @timsifive ,
> Please do remove the call to riscv_fill_dmi_nop_u64() and simply init in_value to NULL a few lines above.

Will do.

> Can you share some before/after numbers? You say the impact is small, and that is fine. If all we gain is removing a line of code the PR is still worth it. :-)

The reason why I said the impact is small is except for people testing my driver, I haven't seen anyone complaining that OpenOCD is slow to load any RISC V image. Also, since most users are likely to be using local USB connection to the board, I do expect they are achieving great speed. However, truth be told that I actually do not have any figure to back up my claim :-(

However, on our network, with aji_client driver (https://github.com/intel/aji_openocd), the transfer speed for gdb "load" command was very slow. A typical value on our worst performing board was like
```
...
Start address 0x6ad, load size 119016
Transfer rate: 433 bytes/sec, 9155 bytes/write.
shutdown command invoked
Remote connection closed

real    4m46.030s
user    0m0.013s
sys     0m0.019s
```
i.e., hovering around 4:30. After disabling returning the DR value as in this patch, we are getting typical value hovering around 0:50 (3KB/s). The fastest observed on our fastest board is 0:07 (17KB/s) at around 20% of the time. We are currently using the write_memory_progbuf() route.
 
 One limitation to note is the test image is very small. I am obviously hoping it is representative of throughput speed.

> What testing have you done?

(1) Created two copies of aji_openocd, one set `in_value` to NULL, the other did not. 
(2) Create a gdb test script (gdbtest) that do "load <image>" command. The test script does the following
```
set arch riscv:rv32
file image.elf
target extended-remote localhost:3333
load
quit
```
(3) Start one copy of aji_openocd in (1), then `time gdb -x gdbtest` 
(4) Repeat (3) on the other copy of openocd in (1).
(5) Repeat (3)-(4) 5 times
(6) Visually examine `real` time result.

Because network traffic varies with time, (3)-(5) are one after another in quick succession using  a batch script  to minimize the influence of network traffic on the transfer speed comparison. Running them successively one after another also give us a feel of whether any spike or drop in transfer speed is traffic-related.

Repeated the test for 3 days to confirm the result.

timsifive:
So it sounds like the speedup is significant for your use-case. It might also help with remote bitbang, which I use for testing, so I'd love a speedup there also.

Are you able to run the tests from https://github.com/riscv/riscv-tests/tree/master/debug? #563 is trying to automate that process, but until we're there it's best to run those tests manually.

cinlyooi-intel:
@timsifive 

Haven't been running test in the debug directory.

cinlyooi-intel:
I am having some trouble in branch management on my forked directory. This means a new pull request has to be created to replace this one. See #621

PR #625, opened 2021-06-21
Fix merge

timsifive:
Should result in no changes, but now I will properly merge the PR so git knows the merge happened. (I accidentally did a squash merge before.)


GregSavin (approved):
So that's how a mistaken squash merge is fixed, just merge again without squash.  That's new to me; makes sense though.

PR #626, opened 2021-06-30
From upstream

timsifive:
Trying to stay relatively up-to-date.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1704,7 +1704,7 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 	}
 
 	/* Save registers */
-	struct reg *reg_pc = register_get_by_name(target->reg_cache, "pc", 1);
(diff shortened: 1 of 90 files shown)


src/target/riscv/riscv.c
```diff
@@ -3296,6 +3308,7 @@ struct target_type riscv_target = {
 	.mmu = riscv_mmu,
 	.virt2phys = riscv_virt2phys,
 
+	.get_gdb_arch = riscv_get_gdb_arch,
```
GregSavin, line 3311: Off-hand do we know what user-visible effects this addition has?

GregSavin (approved):

src/target/riscv/riscv.c
timsifive: It might allow people to connect to a RISC-V target from gdb without first telling gdb which binary is being debugged, or whether it's riscv32/riscv64. It's been on my TODO list forever, so I'm glad somebody did it without me having to do anything. (You can read the upstream PR at http://openocd.zylin.com/#/c/6322/)

PR #627, opened 2021-07-01
Use github actions for automated builds

timsifive:
Travis died. We need something that checks OpenOCD builds.

Change-Id: I7c018caaa3a13884fc526733271697bc90edc891

### DIFF
--- a/.github/workflows/checkpatch.yml
+++ b/.github/workflows/checkpatch.yml
@@ -11,6 +11,9 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
+        with:
+            fetch-depth: 30
+      - run: sudo apt-get install filterdiff
       - name: Run checkpatch
         run: |
             git diff -U20 HEAD~40 | \
--- a/.github/workflows/checkpatch.yml
+++ b/.github/workflows/checkpatch.yml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@v2
         with:
             fetch-depth: 30
-      - run: sudo apt-get install filterdiff
+      - run: sudo apt-get install patchutils
       - name: Run checkpatch
         run: |
             git diff -U20 HEAD~40 | \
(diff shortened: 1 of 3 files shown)


.github/workflows/checkpatch.yml
```diff
@@ -0,0 +1,19 @@
...
+      DL_DIR: ../downloads
+      BUILD_DIR: ../build
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+      - name: Run checkpatch
+        run: |
+            git diff -U20 HEAD~40 | \
+                    filterdiff -x "a/src/jtag/drivers/libjaylink/*" -x "a/tools/git2cl/*" \
+                    -x "b/src/gnulib/*" |
```
JanMatCodasip: It seems this part can be dropped as `src/gnulib` is no longer present.

.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,109 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Copyright (C) 2020 by Tarek BOUCHKATI <tarek.bouchkati@gmail.com>
+
+on: push
+
+name: OpenOCD Snapshot
```
JanMatCodasip, line 7: This workflow runs an elaborate cross build for Windows under Linux, including build of dependencies. 

Could you also add a workflow that would make just a plain simple Linux build? (`./bootstrap` + `./configure` + `make` + `make install`) Thanks.

.github/workflows/checkpatch.yml
timsifive: Good point.

.github/workflows/snapshot.yml
timsifive: Yes, I want to add 32-bit and 64-bit Linux builds. This one is here because it exists in mainline OpenOCD and we might as well use it.

GregSavin (approved):

PR #631, opened 2021-07-20
Add support for hypervisor bits.

timsifive:
Expose them in the virtual priv register, and set them on mcontrol6
hardware triggers. I have no good way to test this right now, so it's
all untested. But this change doesn't break anything, at least.

Change-Id: I0343a6169a0b2b1f0cc0abf687c6bdc560d99b1b


src/target/riscv/riscv-013.c
```diff
@@ -4088,7 +4085,8 @@ static int riscv013_get_register(struct target *target,
 		uint64_t dcsr;
 		/* TODO: move this into riscv.c. */
 		result = register_read(target, &dcsr, GDB_REGNO_DCSR);
-		*value = get_field(dcsr, CSR_DCSR_PRV);
+		*value = set_field(0, VIRT_PRIV_V, get_field(dcsr, CSR_DCSR_V));
```
GregSavin, line 4088: Ah, so "V" is a new DCSR bit in the draft debug spec, and "priv" is a pseudo register abstraction with just the privilege mode and virtualization mode from DCSR.

src/target/riscv/riscv.c
```diff
@@ -626,6 +625,58 @@ static int maybe_add_trigger_t2(struct target *target,
...
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	}
+
+	/* address/data match trigger */
+	tdata1 |= MCONTROL_DMODE(riscv_xlen(target));
+	tdata1 = set_field(tdata1, CSR_MCONTROL6_ACTION,
+			MCONTROL_ACTION_DEBUG_MODE);
+	tdata1 = set_field(tdata1, CSR_MCONTROL6_MATCH, MCONTROL_MATCH_EQUAL);
+	tdata1 |= CSR_MCONTROL6_M;
+	if (r->misa & (1 << ('H' - 'A')))
```
GregSavin, line 645: and trigger type 6 is new in spec too, and includes new VS and VU fields (along with other small bitwise adjustments); got it.  Looks good to me.

GregSavin (approved):

PR #634, opened 2021-07-22
Fix clobbering s2 in read_memory_progbuf(increment=0)

### LINKED ISSUE #632: read_memory_progbuf restores wrong value for s2 on increment=0
riscv-013.c::read_memory_progbuf()

At the beginning: Read of reg s1 into var s2 (content of reg s1 is backed up into s2):
```
	if (increment == 0 && register_read(target, &s2, GDB_REGNO_S1) != ERROR_OK)
		return ERROR_FAIL;
```

On exit restore:
```
	if (increment == 0)
		riscv_set_register(target, GDB_REGNO_S2, s2);  // yes: s2 into s2
```

Proposed fix for the first part: read-out s2 into var s2:
```
	if (increment == 0 && register_read(target, &s2, **GDB_REGNO_S2**) != ERROR_OK)
		return ERROR_FAIL;
```

timsifive:
Fixes #632.

Change-Id: Ic884823faf67749f1ac8fbd91fe67ff9ebdd8fd0

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3457,7 +3457,7 @@ static int read_memory_progbuf(struct target *target, target_addr_t address,
 		return ERROR_FAIL;
 	if (register_read(target, &s1, GDB_REGNO_S1) != ERROR_OK)
 		return ERROR_FAIL;
-	if (increment == 0 && register_read(target, &s2, GDB_REGNO_S1) != ERROR_OK)
+	if (increment == 0 && register_read(target, &s2, GDB_REGNO_S2) != ERROR_OK)
 		return ERROR_FAIL;
 
 	/* Write the program (load, increment) */


timsifive:
The package action is failing because it can't install cmake. It feels like something somewhere cached a negative result. Anyway, I'm going to ignore that failure (but still wait for a review).

JanMatCodasip (approved):

JanMatCodasip:
@timsifive - The fix itself looks all right.

> The package action is failing because it can't install cmake. It feels like something somewhere cached a negative result.

This is due to the local package database being out of date. Adding `apt-get update` into the workflow fixes the issue: https://github.com/riscv/riscv-openocd/pull/635.

I have also submitted that change upstream since that is where the snapshot workflow comes from: http://openocd.zylin.com/#/c/6378/.

GregSavin (approved):
Nice catch by the person who found that issue.

PR #635, opened 2021-07-23
Added 'apt-get update' to snapshot workflow

JanMatCodasip:
This change fixes the snapshot workflow by adding 'apt-get update':

Before installing any packages via apt-get, the local package database needs to be updated.

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -14,7 +14,9 @@ jobs:
       BUILD_DIR: ../build
     steps:
       - name: Install needed packages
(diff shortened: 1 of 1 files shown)


timsifive:
This looks right, but why aren't any automated checks running on this PR?

timsifive:
I think the checks aren't running because "on:" should also include pull_request.

timsifive (approved):

JanMatCodasip:
> This looks right, but why aren't any automated checks running on this PR?
> I think the checks aren't running because "on:" should also include pull_request.

It might also be related to the fact that the pull request came from a different (forked) repository, but I haven't investigated further.

PR #636, opened 2021-07-23
Perform these actions on pull requests, too.

timsifive:
Otherwise they don't happen if somebody makes a PR from a different
fork.

Change-Id: Ic09eb8a556e2dfbc0900e3df9f07cf7aff3d5309

### DIFF
--- a/.github/workflows/checkpatch.yml
+++ b/.github/workflows/checkpatch.yml
@@ -1,4 +1,4 @@
-on: push
+on: pull_request
 
 name: Check Code Style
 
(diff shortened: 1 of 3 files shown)


timsifive:
Now they happen twice. What?

timsifive:
OK, this looks better.

GregSavin (commented):
Any recommendations on where to read a 10000 foot level introduction to github actions?

timsifive:
When I try to figure out I usually google "github action <foo>". The documentation at https://docs.github.com/en/actions seems pretty complete, but doesn't really have a 10k foot view.

GregSavin (approved):

PR #638, opened 2021-08-11
Try to improve remote bitbang speed again

timsifive:
For whatever reason my benchmarks all show this doesn't make a difference, and I can't even replicate the speed regression.

However, without a doubt the code now performs fewer system calls, so it must be faster, right?

### DIFF
--- a/src/jtag/drivers/remote_bitbang.c
+++ b/src/jtag/drivers/remote_bitbang.c
@@ -23,6 +23,7 @@
 #ifndef _WIN32
 #include <sys/un.h>
 #include <netdb.h>
+#include <netinet/tcp.h>
 #endif
 #include "helper/system.h"
 #include "helper/replacements.h"
@@ -270,6 +271,12 @@ static int remote_bitbang_init_tcp(void)
 
 		close(fd);
 	}
+
+	/* We work hard to collapse the writes into the minimum number, so when
+	 * we write something we want to get it to the other end of the
+	 * connection as fast as possible. */
+	int one = 1;
+	setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, &one, sizeof(one));
 
 	freeaddrinfo(result); /* No longer needed */
 
--- a/src/jtag/drivers/remote_bitbang.c
+++ b/src/jtag/drivers/remote_bitbang.c
@@ -276,7 +276,8 @@ static int remote_bitbang_init_tcp(void)
 	 * we write something we want to get it to the other end of the
 	 * connection as fast as possible. */
 	int one = 1;
-	setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, &one, sizeof(one));
+	/* On Windows optval has to be a const char *. */
+	setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, (const char *) &one, sizeof(one));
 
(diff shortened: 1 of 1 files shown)


src/jtag/drivers/remote_bitbang.c
```diff
@@ -188,16 +189,14 @@ static int remote_bitbang_sample(void)
 static bb_value_t remote_bitbang_read_sample(void)
 {
 	if (remote_bitbang_recv_buf_start == remote_bitbang_recv_buf_end) {
```
JanMatCodasip: This could be put to a separate function `remote_bitbang_recv_buf_is_empty()` for readability.

src/jtag/drivers/remote_bitbang.c
```diff
@@ -51,10 +51,48 @@ static bool remote_bitbang_buf_full(void)
...
+
+	if (block == BLOCK) {
+		if (remote_bitbang_flush() != ERROR_OK)
+			return ERROR_FAIL;
+		socket_block(remote_bitbang_fd);
+	}
+
+	bool first = true;
 	while (!remote_bitbang_buf_full()) {
 		unsigned int contiguous_available_space;
```
JanMatCodasip: The computation of contiguous_available_space could also go to a separate function (for code readability).

src/jtag/drivers/remote_bitbang.c
```diff
@@ -188,16 +189,14 @@ static int remote_bitbang_sample(void)
...
-			(remote_bitbang_recv_buf_start + 1) % sizeof(remote_bitbang_recv_buf);
-		return char_to_int(c);
+		if (remote_bitbang_fill_buf(BLOCK) != ERROR_OK)
+			return BB_ERROR;
 	}
-	return remote_bitbang_rread();
+	assert(remote_bitbang_recv_buf_start != remote_bitbang_recv_buf_end);
+	int c = remote_bitbang_recv_buf[remote_bitbang_recv_buf_start];
+	remote_bitbang_recv_buf_start =
+		(remote_bitbang_recv_buf_start + 1) % sizeof(remote_bitbang_recv_buf);
```
JanMatCodasip, line 210: Another snippet that should go to its own function - `remote_bitbang_recv_buf_pop_char()` or similar name.

JanMatCodasip (commented):
Checked visually, the code looks all right. I have just few suggestions towards better code readability.

JanMatCodasip:
One more idea that may potentially boost the performance: `TCP_NODELAY`. In my opinion, it is worth a try.

It is explained e.g. here: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_for_real_time/7/html/tuning_guide/tcp_nodelay_and_small_buffer_writes

See also: https://github.com/riscv/riscv-openocd/blob/riscv/src/jtag/drivers/jtag_vpi.c#L573

timsifive:
> One more idea that may potentially boost the performance: TCP_NODELAY. In my opinion, it is worth a try.

That was a huge improvement! Thanks!

JanMatCodasip (approved):

GregSavin (approved):
I wonder if the slowdown that was reported was specific to a particular operating system.  In any case, that was a good idea to try TCP_NODELAY.

PR #640, opened 2021-08-25
Update gd32vf103.cfg

groencode:
After compiling OpenOCD from source (commit: 6edf98db7), OpenOCD could not connect with the target board which is a GD32VF103T_START development board,
when I used the command: openocd -f "interface/cmsis-dap.cfg" -f "target/gd32vf103.cfg"

Issue 1: The clock speed hasn't been set, the error message given is:
Error: CMSIS-DAP command CMD_DAP_SWJ_CLOCK failed.
Which can be solved by adding a clock speed setting to the gd32vf103.cfg file:
adapter speed 1000

Issue 2: The default transport is not set to jtag, which it requires, the error shown is:
Open On-Chip Debugger 0.11.0+dev-01861-g6edf98db7-dirty (2021-08-25-10:18)
Licensed under GNU GPL v2
For bug reports, read
Info : auto-selecting first available session transport "swd". To override use 'transport select <transport>'.
embedded:startup.tcl:128: Error: session transport is "swd" but your config requires JTAG
in procedure 'script' 
at file "embedded:startup.tcl", line 26
in procedure 'jtag' called at file "/home/user/riscv-openocd-tool/usr/local/bin/../share/openocd/scripts/target/gd32vf103.cfg", line 9
in procedure 'default_to_jtag' called at file "embedded:startup.tcl", line 133
at file "embedded:startup.tcl", line 128
This issue can be solved by adding the line:
transport select jtag

After adding these two lines, OpenOCD could connect to the development board successfully. Programming the flash also works.

### DIFF
--- a/tcl/target/gd32vf103.cfg
+++ b/tcl/target/gd32vf103.cfg
@@ -1,3 +1,6 @@
+adapter speed 1000
+transport select jtag
+
 reset_config srst_nogate
 
 set _CHIPNAME gd32vf103


tcl/target/gd32vf103.cfg
```diff
@@ -1,3 +1,6 @@
+adapter speed 1000
```
timsifive, line 1: This feels slow. Does it not work at 10000?

timsifive (commented):
This looks right to me, but I don't have a board to test it on.

tcl/target/gd32vf103.cfg
groencode: Since the command is: adapter speed [kHz], the speed of 1000 equals 1MHz.
And it is the fastest I have seen in a few examples, like this one: 
https://github.com/riscv-mcu/GD32VF103_Demo_Suites/blob/master/openocd_gdlink.cfg

I have tried running the programming sequence with the speed of 10MHz, the flash worked, but the time consumption is the same with approximately 12 seconds. I would say 1000 is the best option for now.

timsifive: Generally I do all my debugging at 10MHz, but I've never used this target. If it works, why not leave it at the faster speed?

timsifive (approved):

PR #642, opened 2021-08-31
From upstream

timsifive:
Time to get a load of changes from upstream again...

Tested against spike, HiFive1, HiFive Unleashed


timsifive:
I'm ignoring the code style check for this merge, since it's in a change that came from upstream.

src/jtag/drivers/ftdi.c
```diff
@@ -1288,67 +1288,67 @@ COMMAND_HANDLER(ftdi_handle_oscan1_mode_command)
 }
 #endif
 
-static const struct command_registration ftdi_command_handlers[] = {
+static const struct command_registration ftdi_subcommand_handlers[] = {
```
GregSavin, line 1291: So there's a new "ftdi" command with subcommands intended to take the place of a bunch of separate commands with names prefixed with "ftdi_".  But are the old "ftdi_" style commands still supported somehow?  Or do preexisting user scripts that have the old command names need to be updated?

GregSavin (approved):
Looks like the RISC-V related changes were mostly eliminating explicit comparisons against literal NULL

src/jtag/drivers/ftdi.c
GregSavin: I've answered my own question by trying it.  It appears the old ftdi_ commands still work, for now, but apparently any user scripts or auto-generated scripts that use "ftdi_" commands will need to be changed at some point. 

DEPRECATED! use 'ftdi device_desc' not 'ftdi_device_desc'
DEPRECATED! use 'ftdi vid_pid' not 'ftdi_vid_pid'
DEPRECATED! use 'ftdi layout_init' not 'ftdi_layout_init'
DEPRECATED! use 'ftdi layout_signal' not 'ftdi_layout_signal'
DEPRECATED! use 'ftdi layout_signal' not 'ftdi_layout_signal'
DEPRECATED! use 'ftdi layout_signal' not 'ftdi_layout_signal'

PR #644, opened 2021-09-10
Support more socket connections on Windows

sifivekevin:


timsifive (approved):
Looks good to me. I'm sure there's a cleaner way to do this, but this is at least a very small change.

timsifive:
Ignore the style check. I understand the problem, and it's nothing to do with this PR.

Makefile.am
```diff
@@ -31,14 +31,16 @@ DIST_SUBDIRS += jimtcl
...
+			  -DFD_SETSIZE=128
 
 AM_CPPFLAGS = $(HOST_CPPFLAGS)\
 			  -I$(top_srcdir)/src \
 			  -I$(top_builddir)/src \
 			  -I$(top_srcdir)/src/helper \
 			  -DPKGDATADIR=\"$(pkgdatadir)\" \
-			  -DBINDIR=\"$(bindir)\"
+			  -DBINDIR=\"$(bindir)\"\
+			  -DFD_SETSIZE=128
```
JanMatCodasip, line 43: I am curious what was the reason behind this change.

In what scenario was the original socket limit reached? (I assume that the [default limit for connections](https://docs.microsoft.com/en-us/windows/win32/winsock/maximum-number-of-sockets-supported-2) prior this change was 64.)

When OpenOCD is used in a typical way, I would expect one socket for the GDB connection, possibly one socket for the connection towards the target (e.g. remote_bitbang), plus maybe one connection for the telnet/tcl console. What consumed the 64 connection slots?

TommyMurphyTM1234: I'm curious too.

In normal use I'd expect one each for gdb server/remote serial protocol interface, Tcl interface and telnet interface. Maybe some "JTAG" interfaces might use another? But 3-4 seems like it should be the max for a single instance of openocd? Unless the change is to allow more than one instance of openings to use a wider range of ports of something?

sifivekevin:
I have a target system (for a customer) that has 37 JTAG taps. Without this change, I could only open 26 gdb sessions before winsock went off the rails (but reported no errors, just infinite waits on new sockets, and the existing telnet socket). [Linux worked fine.] Four of the taps get grouped into an SMTP group, so I need 33 sockets available for gdb connections.

I'm not sure if 64 is the "real" default ([though this article implies it is](https://docs.microsoft.com/en-us/windows/win32/winsock/maximum-number-of-sockets-supported-2)). It seems like the default might be closer to 32 (or 28, see below). 

Having 34 gdb + 1 telnet + 1 tcl sockets is only 36 sockets, but that clearly is not working. Using netstat reveals only these 36 sockets opened by OpenOCD.  

```
C:\WINDOWS\system32> netstat -anob | grep openocd | wc -l
36
C:\WINDOWS\system32>
```

My "working" case uses 26 gdb + 1 telnet + 1 tcl, so 28 sockets actually connected (still with OpenOCD listening on all 36 sockets, only connecting with 28 of them.)

Setting FD_SETSIZE to 128 allowed everything to work (I did not try to find the minimum FD_SETSIZE that would work for this target, and, in fact, am wanting to future-proof our OpenOCD to work with even larger taps counts). Although this might be an interesting exercise to validate/prove the default value.

My recipe for triaging this failure was:

1. Start OpenOCD (with all 37 taps enumerated).  This opened the telnet and tcl ports.
2. Open a telnet session and issue a 'targets' command to see all enumerate targets listed.
3. Spawn and connect 27 gdb sessions.  This takes about 4 minutes. (Windows is slow!)
4. Session 27 would connect, but was basically hung, could not debug the target. The telnet session is also hung at this point.

Repeat the steps above, but use 26 gdb sessions and everything works fine.

timsifive:
Kevin is supporting somebody using 34 cores, who wants a separate gdb connection for each, and empirically he was only able to connect 29 gdb instances. So presumably the real limit was 32, and this changes it to be 128.

I'm not planning to upstream this, because really this change should only be applied to Windows and who knows what the automake voodoo for that it is.

TommyMurphyTM1234:
Ah, of course - I forgot about the case of different gdb server ports for different harts (and/or TAPs?) and was fixated on the single gdb server connection scenario. Thanks for the explanation @timsifive .

JanMatCodasip:
> Tim: Kevin is supporting somebody using 34 cores, who wants a separate gdb connection for each [...]

@sifivekevin, @timsifive - Thank you for the explanation.

> Kevin: I'm not sure if 64 is the "real" default (though this article implies it is). It seems like the default might be closer to 32 (or 28, see below).

It may be that the default limit is really 64 but one connection actually consumes two sockets on the OpenOCD side. One is the server socket (the one that listens for incoming connections), the other socket represents the actual established connection (as yielded by the accept() call). But that's not important - just a side note.

JanMatCodasip (approved):

PR #645, opened 2021-09-16
Properly cache s0/s1

timsifive:
Read those registers when we're about to use them for scratch, and then don't write them back until we resume the target. This saves a lot of needless saving/restoring of s0/s1. It only makes my tests run a few percent faster, though. I'm not sure why the difference isn't more clear there.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3644,7 +3644,7 @@ int riscv_save_register(struct target *target, enum gdb_regno regid)
 {
 	riscv_reg_t value;
 	if (!target->reg_cache) {
-		LOG_ERROR("TODO: Properly save %s", gdb_regno_name(regid));
+		assert(!target_was_examined(target));
 		return ERROR_OK;
 	}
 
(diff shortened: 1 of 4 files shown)


src/target/riscv/riscv.c
```diff
@@ -3625,14 +3628,40 @@ int riscv_get_register(struct target *target, riscv_reg_t *value,
...
+	if (!target->reg_cache) {
+		assert(!target_was_examined(target));
+		return ERROR_OK;
+	}
+
+	struct reg *reg = &target->reg_cache->reg_list[regid];
+	LOG_DEBUG("[%s] save %s", target_name(target), reg->name);
+	if (riscv_get_register(target, &value, regid) != ERROR_OK)
+		return ERROR_FAIL;
+
```
GregSavin, line 3655: So, the variable named "value" never gets used after riscv_get_register call, because what we care about when calling riscv_get_register here is the side effect of priming the register value into the underlying cache.  Makes sense.  And I guess the compiler isn't flagging that as a warning, which is OK because "we know what we're doing" :)

GregSavin (approved):

src/target/riscv/riscv.c
timsifive: I need that on a t-shirt. :-)

PR #647, opened 2021-09-30
Dynamically allocate memory for RTOS registers.

timsifive:
This makes things work on RISC-V cores with large vector registers
(which can be up to kilobytes in size).

Change-Id: Ie53cb43a88e2a475f695cd5c1e28605569926817


GregSavin (approved):
So the previous code could potentially be driven in such a way that bad things could happen if the underlying register value was wider than 16 bytes?  This fix makes sense to me.

timsifive:
Yes. Previously it would assert, which is not a great user experience.

PR #648, opened 2021-10-04
Clear type 6 triggers on connecting.

timsifive:
I missed this when I first add mcontrol6 support.

Change-Id: I1a2706c7ea3a6757ed5083091cd2c764a8b0267c

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3791,6 +3791,10 @@ int riscv_enumerate_triggers(struct target *target)
 				if (tdata1 & MCONTROL_DMODE(riscv_xlen(target)))
 					riscv_set_register(target, GDB_REGNO_TDATA1, 0);
 				break;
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv.c
```diff
@@ -3791,6 +3791,10 @@ int riscv_enumerate_triggers(struct target *target)
 				if (tdata1 & MCONTROL_DMODE(riscv_xlen(target)))
 					riscv_set_register(target, GDB_REGNO_TDATA1, 0);
 				break;
+			case 6:
```
GregSavin, line 3794: Looks correct to me.  (FWIW, the leading whitespace on some of the pre-existing lines within that switch statement seem to be different than the newly added lines.  At least in the way that GitHub's diff shows that area of code.  Not a big deal, obviously)

GregSavin (approved):

PR #649, opened 2021-10-05
Fix flashing on HiFive1.

timsifive:
Broken by #645. It probably broke some other uses as well, but it was
reported (and easy to reproduce) as an issue with flashing.

Change-Id: Ic1b579c1361442479ced14156102ce68ab232396

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1113,6 +1113,9 @@ static int execute_resume(struct target *target, bool step)
 
 	LOG_DEBUG("step=%d", step);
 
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-011.c
```diff
@@ -1113,6 +1113,9 @@ static int execute_resume(struct target *target, bool step)
 
 	LOG_DEBUG("step=%d", step);
 
+	if (riscv_flush_registers(target) != ERROR_OK)
```
GregSavin, line 1116: Ah yes, I guess that is indeed a crucial detail to include even for version 011 support

GregSavin (approved):

PR #650, opened 2021-10-06
From upstream

timsifive:


src/target/riscv/riscv-013.c
```diff
@@ -1615,11 +1615,7 @@ static int examine(struct target *target)
 		LOG_ERROR("Debugger is not authenticated to target Debug Module. "
 				"(dmstatus=0x%x). Use `riscv authdata_read` and "
 				"`riscv authdata_write` commands to authenticate.", dmstatus);
-		/* If we return ERROR_FAIL here, then in a multicore setup the next
-		 * core won't be examined, which means we won't set up the
-		 * authentication commands for them, which means the config script
-		 * needs to be a lot more complex. */
-		return ERROR_OK;
+		return ERROR_FAIL;
```
GregSavin, line 1618: Did this change come from the RISC-V community?  Just wondering if the issue mentioned in the comment that used to be there (saying that returning ERROR_FAIL could be problematic for multi-core, and ERROR_OK was being returned for that reason) is no longer an issue.

GregSavin (approved):

src/target/riscv/riscv-013.c
timsifive: No, this one came from me.

Upstream changed how the return of this value is handled. Previously it would not affect whether the target is considered examined or not, but now return OK marks the target examined, and return ERROR marks it unexamined.

In this case we definitely do not want to mark the target examined, and the comment made no sense to me. So I changed the code to be "right" and all my tests still pass, so assuming that is fine.

PR #651, opened 2021-10-13
Use submodules

timsifive:


GregSavin (approved):
That c284a9f commit comment explains the rationale clearly.

PR #654, opened 2021-10-18
Properly handle held-in-reset targets.

timsifive:
1. Let OpenOCD continue into some pre-existing code that will periodically
   call examine() until it passes.
2. Fix crash in riscv_openocd_poll()
   When SMP is configured, it's not guaranteed that all targets have been
   examine()d when poll is called on one of them.

Change-Id: Ic6c1d217dc766ea69b67bb2e9a4898e37ee94927


GregSavin (approved):

timsifive:
This change might be unnecessary if you use `-defer-examine` in the config. See discussion at https://review.openocd.org/c/openocd/+/6964

PR #655, opened 2021-10-25
From upstream

timsifive:
This primarily contains the large upstreaming of RISC-V changes, so lots more RISC-V changes than usual.


src/target/riscv/riscv-013.c
```diff
@@ -1828,7 +1828,7 @@ static unsigned riscv013_data_bits(struct target *target)
 	RISCV013_INFO(info);
 	RISCV_INFO(r);
 
-	for (unsigned i = 0; i < RISCV_NUM_MEM_ACCESS_METHODS; i++) {
+	for (unsigned int i = 0; i < RISCV_NUM_MEM_ACCESS_METHODS; i++) {
```
GregSavin, line 1831: So I guess in OpenOCD coding standards, there's a preference for spelling out "unsigned int" rather than "unsigned"?  I wonder what the story is on that.  I don't think I've ever seen that preference elsewhere.

GregSavin (dismissed):

src/target/riscv/riscv-013.c
timsifive: There is. I don't know the story. I just try to always stick to the coding standard. (Where I'm aware of it. This standard was actually introduced somewhat recently.)

GregSavin (approved):

PR #656, opened 2021-10-26
Flush register cache when disconnecting or polling

timsifive:
This makes things work as expected when OpenOCD disconnects and then
connects again. (This became a problem in #645, which started using the
cache.)

Change-Id: I764e5b410a1a68ca47d2ec39968085618ee363c2

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -489,6 +489,9 @@ static void riscv_deinit_target(struct target *target)
 	riscv_info_t *info = (riscv_info_t *) target->arch_info;
 	struct target_type *tt = get_target_type(target);
 
+	if (riscv_flush_registers(target) != ERROR_OK)
+		LOG_ERROR("[%s] Failed to flush registers. Ignoring this error.", target_name(target));
+
 	if (tt && info->version_specific)
 		tt->deinit_target(target);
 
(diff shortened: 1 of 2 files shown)


mgfarmer:
Tested and working!

GregSavin (approved):
I'm curious what scenario was causing problems, without the auto-flushing after 100 ms.  Is it a case that OpenOCD sometimes gets terminated in a less-than-fully-complete-and-graceful exit path when it's being controlled by an integrated development environment?

mgfarmer:
The specific failure case was:

- Start a debug session, stopping at main()
- Disconnect the debug session (leaving openocd running)
- Reattach the debug session (reattaching to the running openocd, and target core
- Upon reattaching the $fp (aka $s0) register was "corrupt" (presumably because it was not restored from the openocd cache prior to disconnecting (Tim can proably explain this bit better))
- So the first instruction that stored to the frame stack goes sideways to the early_trap_vector.

GregSavin:
Ah, OK, detaching and reattaching from GDB, I see.  I guess we don't have hooks that notify about that level of attachment/detachment (clearly the server.c code level sees those disconnections/connections happen but maybe it doesn't fire corresponding notifications down to target/riscv*.c code, otherwise we'd have a perfect time/place to do any necessary flushing at that point, but instead we're now handling any needed flushing in the polling loop (and also in target_deinit that gets called when OpenOCD exits normally), makes sense

timsifive:
I thought the bug happened when OpenOCD itself exited and started again. Detaching/reattaching gdb should never have caused this problem. The 100ms flush is just in case OpenOCD gets SIGKILL or something and doesn't get a chance to do its regular clean-up.

timsifive:
But what @mgfarmer describes does really happen. Let me understand what's going on there.

sifivekevin:
(Now that I'm signed into github with the correct account, i.e. not mgfarmer)  Interesting that your fix works...

timsifive:
#657 should fix the actual bug @sifivekevin reported. This change is still valuable because it addresses a different use case.

PR #657, opened 2021-10-28
Don't invalidate register cache of halted target.

timsifive:
This causes trouble if you disconnect/reconnect gdb, because in that
case a halt is issued again. (It probably would be possible to run into
this problem in other ways as well.)

Change-Id: Id5815f9d1dc2c2dd627770e001c03874a307c279


GregSavin (approved):

sifivekevin:
Confirmed that this works on my test system.

PR #660, opened 2021-11-03
Revive `riscv resume_order`

timsifive:
This functionality was lost in #567. Now it works as expected again.

Change-Id: I11a36e076867c0268034ceee763e28b2d4e6ff0f


GregSavin (approved):
Ah, I now see from #567 where the resume_order code was removed.  The tlist_reverse/tlist_free helper routines make it easier to understand what is going on, too.

PR #664, opened 2021-11-19
Deal with halt race.

timsifive:
What could happen:
1. hart 1 halts due to a breakpoint
2. OpenOCD polls and notices hart 1 has halted.
3. hart 0 halts because it is in a halt group with halt 1
4. OpenOCD decides to halt hart 0 also.
5. OpenOCD discovers hart 0 is already halted, and doesn't update the
debug reason.

In that case OpenOCD would tell gdb that hart 0 halted, while the
interesting event is that hart 1 halted.

Fix this by updating the debug reason when we discover a hart is already
halted when we try to halt it.

This race was exposed in the Sv* address translation tests, but the bug
has nothing to do with address translation.

Change-Id: I59d871e8545bd644bf42581266f15234b93e9900

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1189,6 +1189,33 @@ int riscv_flush_registers(struct target *target)
 	return ERROR_OK;
 }
 
+int set_debug_reason(struct target *target, enum riscv_halt_reason halt_reason)
+{
+	switch (halt_reason) {
+		case RISCV_HALT_BREAKPOINT:
+			target->debug_reason = DBG_REASON_BREAKPOINT;
+			break;
+		case RISCV_HALT_TRIGGER:
+			target->debug_reason = DBG_REASON_WATCHPOINT;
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv.c
```diff
@@ -1189,6 +1189,33 @@ int riscv_flush_registers(struct target *target)
 	return ERROR_OK;
 }
 
+int set_debug_reason(struct target *target, enum riscv_halt_reason halt_reason)
```
JanMatCodasip, line 1193: I know this is not new code (just moved), but maybe add a note what this function does:

```suggestion
/* Convert: RISC-V hart's halt reason --> OpenOCD's generic debug reason */
int set_debug_reason(struct target *target, enum riscv_halt_reason halt_reason)
```

JanMatCodasip (dismissed):

JanMatCodasip (approved):

PR #665, opened 2021-11-30
Don't reexamine targets until it's time.

### LINKED ISSUE #663: Custom JTAG setup sequence before target examine
Hello,

After last openocd update from this repo I've found that my custom JTAG setup procedure is broken. It seems that 
after this change https://github.com/riscv/riscv-openocd/pull/654 target examine is called before JTAG setup.
Is there any way to setup JTAG TAPs and to override reset procedure after this change? Currently I've reverted this change locally.

The problem is that we have the system with two RISC-V subsystems with
individual debug module each connected in JTAG chain. Both TAPs are online after power up.
One of the subsystems is kept in reset while another one is doing system reset sequence.
ndmreset signal of only one of them is connected to the rest of the system (and to the cores of the second subsystem)

I use overridden jtag setup  procedure to make reset sequence before targets examine.

With that changes targets examine is called before jtag setup and that breaks functionality.
Do you know another mechanism to make initial reset sequence before targets examine?
The mechanism which can be applied with that changes.

I attach openocd config (sys_cfg.txt) and logs with and without that patch.

The problem as I see it is in target.c "handle_target" function.
In my case it runs examine target before jtag setup is called.

Thank you,
Fedor Borisovskii

[sys_run_with_patch.log](https://github.com/riscv/riscv-openocd/files/7564589/sys_run_with_patch.log)
[sys_run_without_patch.log](https://github.com/riscv/riscv-openocd/files/7564590/sys_run_without_patch.log)
[sys_cfg.txt](https://github.com/riscv/riscv-openocd/files/7564592/sys_cfg.txt)

timsifive:
Don't reexamine targets until we're past the examine stage of init().
Earlier than that, examine() will likely fail because the scan chain
hasn't been examined yet.

This will likely fix #663.

Change-Id: I76ee9181f35cedcdb1a3e0f8ac33ab361c68d3af

### DIFF
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -770,6 +770,12 @@ static int jtag_enable_callback(enum jtag_event event, void *priv)
 	return target_examine_one(target);
 }
 
+/* When this is true, it's OK to call examine() again in the hopes that this time
+ * it will work.  Earlier than that there is probably other initialization that
+ * needs to happen (like scanning the JTAG chain) before examine should be
+ * called. */
+static bool examine_attempted;
+
 /* Targets that correctly implement init + examine, i.e.
  * no communication with target during init:
  *
@@ -779,6 +785,8 @@ int target_examine(void)
 {
 	int retval = ERROR_OK;
 	struct target *target;
+
+	examine_attempted = true;
 
 	for (target = all_targets; target; target = target->next) {
 		/* defer examination, but don't skip it */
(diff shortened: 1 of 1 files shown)


src/target/target.c
```diff
@@ -770,6 +770,12 @@ static int jtag_enable_callback(enum jtag_event event, void *priv)
 	return target_examine_one(target);
 }
 
+/* When this is true, it's OK to call examine() again in the hopes that this time
+ * it will work.  Earlier than that there is probably other initialization that
+ * needs to happen (like scanning the JTAG chain) before examine should be
+ * called. */
+static bool examine_attempted;
```
GregSavin, line 777: Out of curiosity, is there any plausible scenario where examine_attempted might need to be tracked on a per-target basis (e.g. as a data member of "struct target")?

GregSavin (approved):

src/target/target.c
timsifive: I don't believe so. Really I just want to make sure we don't call examine() from poll() until after the TAP scan chain has been examined. Hopefully the scan chain will always be a single thing, and examine() won't be called outside of poll() until after all chains (I don't think OpenOCD supports multiple chains even) have been examined. There isn't a clear place where that ordering happens that I could find, so I'm using this work-around.

PR #666, opened 2021-11-30
From upstream

timsifive:


GregSavin (approved):
I saw the JTAG related refactoring (addition of adapter.h etc); but I guess that was basically moving equivalent code from one place to another (and one of those things that, if there was breakage introduced there, it'd probably be obvious immediately on any basic smoke test)

PR #669, opened 2021-12-13
--enable-ftdi-oscan1 in Linux build.

timsifive:
Change-Id: I531cd283a1020d94eaeee36e81140c8c4801c69e

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -12,9 +12,9 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


GregSavin (approved):
I seem to recall --enable-ftdi-oscan1 defaulting to enabled originally, and maybe that got turned off upstream (which seems reasonable)

timsifive:
Maybe it's only enabled in SiFive builds? It's definitely not enabled by default in this branch.

None of the oscan1 changes have merged upstream yet, so that wouldn't be where the change comes from. Although perhaps I did it accidentally when resolving a merge conflict.

GregSavin:
Ah, OK, that makes sense, it must've gotten turned off incidentally, somehow.

PR #671, opened 2021-12-28
From upstream

timsifive:


src/target/riscv/riscv-013.c
```diff
@@ -1194,10 +1194,7 @@ static int scratch_reserve(struct target *target,
 static int scratch_release(struct target *target,
 		scratch_mem_t *scratch)
 {
-	if (scratch->area)
-		return target_free_working_area(target, scratch->area);
-
-	return ERROR_OK;
+	return target_free_working_area(target, scratch->area);
```
GregSavin, line 1197: Ah OK so the scratch->area == NULL case is now handled within target_free_working_area()

GregSavin (approved):

timsifive:
Ugh. I messed this up by doing a squash commit where a merge commit was needed. I'll fix it no later than tomorrow.

PR #672, opened 2022-01-26
Fix small memory leak.

timsifive:
Change-Id: Ia11ab9bcf860f770ea64ad867102c74b898f6b66


GregSavin (approved):
Ah, I guess init_target can be called more than once.  Good catch.  Curious if that leak was found by static inspection of the code or by using valgrind or something.

timsifive:
It can be called multiple times, if you have a target which you need to authenticate to. It feels a little ugly to fix it this way, but it's effective.

Found the leak using valgrind.

PR #673, opened 2022-01-28
Ask the RTOS which target to set swbp on.

timsifive:
This lets the RTOS pick the "current" target, which matters if address
translation differs between threads.

Change-Id: I5b5510ab6a06621589c902f42a91562055817dc4


GregSavin (approved):
This looks promising as far as possibly also fixing an issue reported on multicore debug in bare metal environment I'm looking at (I'll try this patch in that context, but it looks good).

timsifive:
This fixes the following problem:
gdb wants to step, and does so by setting a breakpoint on the next instruction. Regardless of which thread is currently selected, gdb will set this breakpoint on the first thread. If the first target and the target being stepped have the MMU configured differently, then this is not going to work.

PR #674, opened 2022-01-28
From upstream

timsifive:


GregSavin (approved):
I wonder what the story is with the addition of log_exit(); I guess maybe lost log output if a crash happened before final exit() or something.  LGTM though.

timsifive:
Looks like log_exit() was added to fix a memory leak: https://github.com/riscv/riscv-openocd/commit/492ac453ab6b97280030fe6161dfcac3d84b36ce

PR #675, opened 2022-02-03
fix for missing thread ID in stop reply when smp-configured harts other than hart 0 single-stepped

GregSavin:
To replicate the issue that this fixes:

1) checkout the riscv branch, build, connect to a multi-hart target configured as an SMP group.  
2) Start a GDB instance against the running OpenOCD.  
3) Observe that GDB might display "warning: multi-threaded target stopped without sending a thread-id, using first non-exited thread."
4) Set a breakpoint in code that any non-hart-0 hart is expected to reach (but hart 0 is not expected to reach).
5)  Allow a non-hart-0 hart to reach the breakpoint.
6) Remove the breakpoint.
7) Do a few sequential "stepi" commands in GDB.
8) Observe that GDB displays "Switching to Thread 1" even though the thread that was just single stepped was not Thread 1 in GDB.  Also observe that the register values in GDB correspond to the thread that was single-stepped, not Thread 1.  Basically GDB erroneously starts to consider thread 1 to be current, when in fact the thread that was single-stepped is still current.

The changes in this pull request are intended to avoid the erroneous "Switching to Thread 1" described in (8) above.

What was happening was that, in a couple areas of code, non-hart-0 harts weren't seen as belonging to an rtos module, and this had the effect of (1) bypassing hwthread_update_threads() being called after a halt; (2) omitting a thread ID in a stop reply over GDB remote protocol connection (requiring GDB to take an arbitrary guess of current thread id, a guess that is wrong unless the current thread happens to be hart 0).

### DIFF
--- a/src/rtos/rtos.c
+++ b/src/rtos/rtos.c
@@ -789,10 +789,34 @@ static int rtos_try_next(struct target *target)
 	return 1;
 }
 
-int rtos_update_threads(struct target *target)
+static struct rtos * rtos_of_target(struct target *target)
 {
 	if ((target->rtos) && (target->rtos->type))
+	{
+	   return target->rtos;
+	}
+	else if (target->smp)
+	{
+	   // find target in SMP group that has ->rtos and ->rtos->type non-zero
+	   struct target_list *cur = target->head;
+	   while (cur) {
+	      if (cur->target->rtos && cur->target->rtos->type)
+		 return cur->target->rtos;
+	      cur = cur->next;
+	   }
+	}
+	return NULL;
+}
+
+int rtos_update_threads(struct target *target)
+{
+/*   
+	if ((target->rtos) && (target->rtos->type))
 		target->rtos->type->update_threads(target->rtos);
+*/
+	struct rtos *rtos = rtos_of_target(target);
+	if (rtos)
+		rtos->type->update_threads(rtos);	   
 	return ERROR_OK;
 }
 
--- a/src/rtos/rtos.c
+++ b/src/rtos/rtos.c
@@ -789,7 +789,7 @@ static int rtos_try_next(struct target *target)
 	return 1;
 }
 
-struct rtos * rtos_of_target(struct target *target)
+struct rtos *rtos_of_target(struct target *target)
 {
 	if ((target->rtos) && (target->rtos->type))
 	{
--- a/src/rtos/rtos.c
+++ b/src/rtos/rtos.c
@@ -797,7 +797,7 @@ struct rtos *rtos_of_target(struct target *target)
 		Otherwise NULL return means that no associated non-zero rtos field could be found. */
 
 	struct target_list *pos;
-   
+
 	if ((target->rtos) && (target->rtos->type))
 		return target->rtos;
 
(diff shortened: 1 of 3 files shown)


GregSavin:
I ran riscv-tests/debug/make on this branch, and didn't see any failing tests.  Note that there is not yet a test in riscv-tests/debug that covers this specific current issue.

timsifive:
> I ran riscv-tests/debug/make on this branch, and didn't see any failing tests. Note that there is not yet a test in riscv-tests/debug that covers this specific current issue.

Did you run `make` or `make all`? The latter runs more tests, and covers a few more corner cases that happen in different multicore setups.

src/rtos/rtos.c
```diff
@@ -789,10 +790,29 @@ static int rtos_try_next(struct target *target)
 	return 1;
 }
 
-int rtos_update_threads(struct target *target)
+struct rtos *rtos_of_target(struct target *target)
 {
+	/* Primarily consider the rtos field of the target itself, secondarily consider
+		rtos field SMP leader target, then consider rtos field of any other target in the SMP group.
+		Otherwise NULL return means that no associated non-zero rtos field could be found. */
```
timsifive: Something is wonky with the indenting here.

timsifive (dismissed):
The change you made is definitely right for targets configured with SMP and `-rtos hwthread`. I'm not sure it's right for every combination of those options, but I think that the intent behind those two features is to be used together. E.g. while technically you could have a target with SMP cores and ThreadX running on only one of them, that is not the intended use case behind the features. I suspect you're going to have a hard time with that in gdb regardless.

So I think this change is good.

GregSavin:
> Did you run `make` or `make all`? The latter runs more tests, and covers a few more corner cases that happen in different multicore setups.

I ran "make".  I'll check again with "make all"

GregSavin:
@timsifive "make all" looked OK.

[riscv-tests/debug]$ make all
(voluminous output not shown)
[riscv-tests/debug]$ grep "Result: pass" logs/* | wc -l
323
[riscv-tests/debug]$ grep "Result: fail" logs/* | wc -l
0
[riscv-tests/debug]$ grep "Result: exception" logs/* | wc -l
0
[riscv-tests/debug]$ grep "Result: not_applicable" logs/* | wc -l
52

src/rtos/rtos.c
GregSavin: Latest revision fixes the indentation.  Looks like my commit had the side-effect of github invalidating your prior approval, so I guess this needs an approval again, before I'm allowed to press the merge button.

GregSavin:
@timsifive Do you think it'd be worthwhile if I add a new test case to riscv-tests to get regression coverage for the scenario that this patch is a fix for?

timsifive:
> @timsifive Do you think it'd be worthwhile if I add a new test case to riscv-tests to get regression coverage for the scenario that this patch is a fix for?

I think that'd be great. I really should have made a testcase for the software breakpoint with different harts having different address translation, but I'm busy, <insert worthless excuse>. Having coverage for this kind of thing is important because it's exactly the kind of thing that could easily break while making some other improvement.

timsifive (approved):

PR #676, opened 2022-02-09
riscv: fix remove_trigger return code for unavailable hw bp slot

erhankur:
If there is no available slot for the hardware breakpoints, it should return with different error code to allow vendors to do specific operation like switching to flash breakpoint's add/remove functions. 

Actually it is working as expected for the `add_trigger`, but not for the `remove_trigger`


timsifive:
That looks like an improvement. Do any of the higher layers actually try to set a different kind of breakpoint if the hardware breakpoint returns ERROR_TARGET_RESOURCE_NOT_AVAILABLE?

erhankur:
Yes in Espressif fork we decide whether we will use a flash breakpoint or not based on the return value.  I thought it would be better to keep same return code with `add_trigger`

JanMatCodasip (approved):

PR #677, opened 2022-02-10
.gitmodules: switch away from repo.or.cz

timsifive:
Manually applying https://review.openocd.org/c/openocd/+/6834 (which
hasn't merged yet) since repo.or.cz is down often enough that it's
affecting development.

Change-Id: Idb337c799662f3e0fa72379c59c52a61a048044e

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
+[submodule "tools/git2cl"]
+	path = tools/git2cl
+	url = https://git.savannah.nongnu.org/git/git2cl.git
 [submodule "jimtcl"]
 	path = jimtcl
 	url = https://github.com/msteveb/jimtcl.git
-[submodule "tools/git2cl"]
-	path = tools/git2cl
-	url = https://repo.or.cz/git2cl.git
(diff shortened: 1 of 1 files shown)


TommyMurphyTM1234:
> Manually applying https://review.openocd.org/c/openocd/+/6834 (which hasn't merged yet) since repo.or.cz is down often enough that it's affecting development.

Would getting it from the master repo at SF.net instead help?
(Although I recall SF.net being terribly slow in the past so I don't know if that's still a problem?)

https://openocd.org/pages/getting-openocd.html

timsifive:
These modules don't get the source for OpenOCD. They get the source for a few other components. I did not look where all they may be hosted. I simply went with what is likely to merge into upstream soon in any case.

TommyMurphyTM1234:
> These modules don't get the source for OpenOCD. They get the source for a few other components. I did not look where all they may be hosted. I simply went with what is likely to merge into upstream soon in any case.

Ah, ok. Fair enough. :-)

GregSavin (approved):

PR #678, opened 2022-02-11
fix progbuf cache: invalidate it when needed

JanMatCodasip:
This commit relates to progbuf cache,
implemented in https://github.com/riscv/riscv-openocd/pull/381

Make sure the cache gets invalidated when the progbuf
contents change via other means. I've identified two
such cases where the invalidation is required:

1) When the user manually tinkers with the progbuf registers
   (TCL command "riscv dmi_write")

2) When program buffer is used as a scratch memory
   (scratch_write64())

Change-Id: Ie7ffb0fccda63297de894ab919d09082ea21cfae

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1568,6 +1568,9 @@ static int examine(struct target *target)
 		dmi_write(target, DM_DMCONTROL, 0);
 		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_DMACTIVE);
 		dm->was_reset = true;
+
+		/* The DM gets reset, so forget any cached progbuf entries. */
+		riscv013_invalidate_cached_debug_buffer(target);
 	}
 
 	dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_HARTSELLO |
@@ -2371,7 +2374,7 @@ static int assert_reset(struct target *target)
 	/* The DM might have gotten reset if OpenOCD called us in some reset that
 	 * involves SRST being toggled. So clear our cache which may be out of
 	 * date. */
-	memset(dm->progbuf_cache, 0, sizeof(dm->progbuf_cache));
+	riscv013_invalidate_cached_debug_buffer(target);
 
 	return ERROR_OK;
 }
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -52,6 +52,7 @@ static int riscv013_write_debug_buffer(struct target *target, unsigned index,
 		riscv_insn_t d);
 static riscv_insn_t riscv013_read_debug_buffer(struct target *target, unsigned
 		index);
+static int riscv013_invalidate_cached_debug_buffer(struct target *target);
 static int riscv013_execute_debug_buffer(struct target *target);
 static void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a, uint64_t d);
 static void riscv013_fill_dmi_read_u64(struct target *target, char *buf, int a);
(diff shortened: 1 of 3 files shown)


JanMatCodasip:
I have addressed two more scenarios where the progbuf cache needs to be invalidated:

- reset of the debug module (dmactive=0) during examination
- reset of the debug module (dmactive=0) triggered manually by the user via `riscv dmi_write` TCL command

@timsifive - What is the performance gain of the progbuf cache (alone, not counting the the other improvements introduced in #381)? What scenario was used to measure it? If the speed-up is very small (units of percent), please consider removing the progbuf cache feature entirely due to its low benefit/complexity ratio. If you decide to keep it, please look at the fixes I have submitted. Thank you.

timsifive:
> @timsifive - What is the performance gain of the progbuf cache (alone, not counting the the other improvements introduced in https://github.com/riscv/riscv-openocd/pull/381)? What scenario was used to measure it? If the speed-up is very small (units of percent), please consider removing the progbuf cache feature entirely due to its low benefit/complexity ratio. If you decide to keep it, please look at the fixes I have submitted. Thank you.

Against spike, the speedup is immeasurable, although the cache hit rate is 34% on the Registers test. (spike is not a great target for testing wall clock time due to how its scheduling works. I should have grabbed a hardware target.) I imagine against some FPGA it would have been noticeable. If we continue to have problems due to the program buffer cache then I'll look again. It's a feature that has had no problems in 2.5 years. I assume your change here addresses most of the other ones.

timsifive (approved):
Looks good to me.

A broader point: It seems like you're using dmi writes in the TCL script to execute custom command sequences. Would it make more sense to add a custom command that takes RISC-V opcodes as argument and executes them directly, possibly with data input argument as well?

JanMatCodasip:
@timsifive - Thank you for your reply and reviewing the code.

> Against spike, the speedup is immeasurable, although the cache hit rate is 34% on the Registers test. (spike is not a great target for testing wall clock time due to how its scheduling works. I should have grabbed a hardware target.) I imagine against some FPGA it would have been noticeable.

I haven't made thorough measurements of the progbuf cache either, but from observation, the cache gets mostly hit on the ebreak instruction in the progbuf, which probably roughly corresponds to the observed ~30 % hit rate.

In hardware targets (e.g. FPGAs) the prominent speed factor is the USB round-trip. So the goal is to optimize the number of transfers (batches), not necessarily the amount of transferred data. For memory accesses, this has been well taken care of via the dmi batches. I imagine that to optimize the register access, similar approaches could be taken instead of the progbuf cache, for example: 
- read cacheable registers in a batch when halt of the target is detected and populate the register cache - this reduces the number of USB transfers, too
- or when populating progbuf, write all the instructions to the progbuf in one batch (again, this may reduce the number of USB transfers even more than the progbuf cache)

If I find the time to prototype the ideas and measure it, I'll reach back with the numbers :)

> It's a feature that has had no problems in 2.5 years. 

I understand that my use cases (fixed in this pull request) are not very common, but I ran into them within one week of working with this OpenOCD version, and they were not easy to diagnose. Even with the fixes, I am still little uneasy whether there are other (so far missed) scenarios that would cause the progbuf cache to get out of date.

I hope to replace the cache in the future with one of the solutions above (or similar) that have the potential to be faster and more robust at the same time.

> A broader point: It seems like you're using dmi writes in the TCL script to execute custom command sequences. Would it make more sense to add a custom command that takes RISC-V opcodes as argument and executes them directly, possibly with data input argument as well?

Good point, I'll add this to my list. I have used `riscv dmi_{read,write}` commands for testing only, but I imagine low-level interface for progbuf execution could be useful to advanced users in special cases.

timsifive:
> I understand that my use cases (fixed in this pull request) are not very common, but I ran into them within one week of working with this OpenOCD version, and they were not easy to diagnose. Even with the fixes, I am still little uneasy whether there are other (so far missed) scenarios that would cause the progbuf cache to get out of date.

Another option for keeping the program buffer cache up-to-date might be to update them in dmi_op_timeout(), which I think is at the bottom of almost every scan done. (The exception being the batch stuff in the memory access code.)

PR #681, opened 2022-02-15
riscv: implement maskisr steponly command

erhankur:
With this PR,  `riscv maskisr <off | steponly>` command implemented. Now it will be possible to disable/restore interrupts within the step function.  

ARM implementation of this command has also `auto and on` option but it is not ready for the riscv yet.  

I will also run and share the test results to make sure at least nothing broken. 

Change-Id: I1a3b666d466b064460c3acc307a36485ce165601

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10425,8 +10425,10 @@ tunneled DR scan consists of:
 @end deffn
 
 @deffn {Command} {riscv set_maskisr} [@option{off}|@option{steponly}]
-Selects whether interrupts will be disabled when single stepping. The default configuration is
-@option{off}.
+Selects whether interrupts will be disabled when single stepping. The default configuration is @option{off}.
+This feature is only useful on hardware that always steps into interrupts and doesn't support dcsr.stepie=0.
+Keep in mind, disabling the option does not guarantee that single stepping will go into interrupt handlers.
+To make that happen, dcsr.stepie would have to be written to 1 as well.
 @end deffn
 
 @deffn {Command} {riscv set_ebreakm} on|off
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10424,6 +10424,11 @@ tunneled DR scan consists of:
 
 @end deffn
 
+@deffn {Command} {riscv set_maskisr} [@option{off}|@option{steponly}]
+Selects whether interrupts will be disabled when single stepping. The default configuration is
+@option{off}.
+@end deffn
+
 @deffn {Command} {riscv set_ebreakm} on|off
 Control dcsr.ebreakm. When on (default), M-mode ebreak instructions trap to
 OpenOCD. When off, they generate a breakpoint exception handled internally.
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2350,7 +2350,7 @@ int riscv_openocd_poll(struct target *target)
 }
 
 int riscv_openocd_step(struct target *target, int current,
-		target_addr_t address, int handle_breakpoints)
+	target_addr_t address, int handle_breakpoints)
 {
 	LOG_DEBUG("stepping rtos hart");
 
@@ -2361,38 +2361,48 @@ int riscv_openocd_step(struct target *target, int current,
 	if (disable_triggers(target, trigger_state) != ERROR_OK)
 		return ERROR_FAIL;
 
+	bool success = true;
 	uint64_t current_mstatus;
 	RISCV_INFO(info);
+
 	if (info->isrmask_mode == RISCV_ISRMASK_STEPONLY) {
 		/* Disable Interrupts before stepping. */
 		uint64_t irq_disabled_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
-		if (riscv_interrupts_disable(target, irq_disabled_mask, &current_mstatus) != ERROR_OK)
-			return ERROR_FAIL;
+		if (riscv_interrupts_disable(target, irq_disabled_mask,
+				&current_mstatus) != ERROR_OK) {
+			success = false;
+			LOG_ERROR("unable to disable interrupts");
+			goto _exit;
+		}
 	}
 
-	int out = riscv_step_rtos_hart(target);
-	if (out != ERROR_OK) {
+	if (riscv_step_rtos_hart(target) != ERROR_OK) {
+		success = false;
 		LOG_ERROR("unable to step rtos hart");
-		return out;
-	}
-
-	if (info->isrmask_mode == RISCV_ISRMASK_STEPONLY) {
-		if (riscv_interrupts_restore(target, current_mstatus) != ERROR_OK)
-			return ERROR_FAIL;
 	}
 
 	register_cache_invalidate(target->reg_cache);
 
-	if (enable_triggers(target, trigger_state) != ERROR_OK)
-		return ERROR_FAIL;
+	if (info->isrmask_mode == RISCV_ISRMASK_STEPONLY)
+		if (riscv_interrupts_restore(target, current_mstatus) != ERROR_OK) {
+			success = false;
+			LOG_ERROR("unable to restore interrupts");
+		}
 
-	target->state = TARGET_RUNNING;
-	target_call_event_callbacks(target, TARGET_EVENT_RESUMED);
-	target->state = TARGET_HALTED;
-	target->debug_reason = DBG_REASON_SINGLESTEP;
-	target_call_event_callbacks(target, TARGET_EVENT_HALTED);
+_exit:
+	if (enable_triggers(target, trigger_state) != ERROR_OK) {
+		success = false;
+		LOG_ERROR("unable to enable triggers");
+	}
 
-	return out;
+	if (success) {
+		target->state = TARGET_RUNNING;
+		target_call_event_callbacks(target, TARGET_EVENT_RESUMED);
+		target->state = TARGET_HALTED;
+		target->debug_reason = DBG_REASON_SINGLESTEP;
+		target_call_event_callbacks(target, TARGET_EVENT_HALTED);
+	}
+	return success ? ERROR_OK : ERROR_FAIL;
 }
 
 /* Command Handlers */
@@ -2956,7 +2966,7 @@ COMMAND_HANDLER(riscv_set_maskisr)
 		n = jim_nvp_value2name_simple(nvp_maskisr_modes, info->isrmask_mode);
 		command_print(CMD, "riscv interrupt mask %s", n->name);
 	}
-	
+
 	return ERROR_OK;
 }
 
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1953,26 +1953,11 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 		}
 	}
 
-
 	/* Disable Interrupts before attempting to run the algorithm. */
 	uint64_t current_mstatus;
-	uint8_t mstatus_bytes[8] = { 0 };
-
-	LOG_DEBUG("Disabling Interrupts");
-	struct reg *reg_mstatus = register_get_by_name(target->reg_cache,
-			"mstatus", true);
-	if (!reg_mstatus) {
-		LOG_ERROR("Couldn't find mstatus!");
+	uint64_t irq_disabled_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
+	if (riscv_interrupts_disable(target, irq_disabled_mask, &current_mstatus) != ERROR_OK)
 		return ERROR_FAIL;
-	}
-
-	reg_mstatus->type->get(reg_mstatus);
-	current_mstatus = buf_get_u64(reg_mstatus->value, 0, reg_mstatus->size);
-	uint64_t ie_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
-	buf_set_u64(mstatus_bytes, 0, info->xlen, set_field(current_mstatus,
-				ie_mask, 0));
-
-	reg_mstatus->type->set(reg_mstatus, mstatus_bytes);
 
 	/* Run algorithm */
 	LOG_DEBUG("resume at 0x%" TARGET_PRIxADDR, entry_point);
@@ -2028,9 +2013,8 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 	}
 
 	/* Restore Interrupts */
-	LOG_DEBUG("Restoring Interrupts");
-	buf_set_u64(mstatus_bytes, 0, info->xlen, current_mstatus);
-	reg_mstatus->type->set(reg_mstatus, mstatus_bytes);
+	if (riscv_interrupts_restore(target, current_mstatus) != ERROR_OK)
+		return ERROR_FAIL;
 
 	/* Restore registers */
 	uint8_t buf[8] = { 0 };
@@ -2377,10 +2361,24 @@ int riscv_openocd_step(struct target *target, int current,
 	if (disable_triggers(target, trigger_state) != ERROR_OK)
 		return ERROR_FAIL;
 
+	uint64_t current_mstatus;
+	RISCV_INFO(info);
+	if (info->isrmask_mode == RISCV_ISRMASK_STEPONLY) {
+		/* Disable Interrupts before stepping. */
+		uint64_t irq_disabled_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
+		if (riscv_interrupts_disable(target, irq_disabled_mask, &current_mstatus) != ERROR_OK)
+			return ERROR_FAIL;
+	}
+
 	int out = riscv_step_rtos_hart(target);
 	if (out != ERROR_OK) {
 		LOG_ERROR("unable to step rtos hart");
 		return out;
+	}
+
+	if (info->isrmask_mode == RISCV_ISRMASK_STEPONLY) {
+		if (riscv_interrupts_restore(target, current_mstatus) != ERROR_OK)
+			return ERROR_FAIL;
 	}
 
 	register_cache_invalidate(target->reg_cache);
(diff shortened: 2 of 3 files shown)


src/target/riscv/riscv.c
```diff
@@ -2916,6 +2915,31 @@ COMMAND_HANDLER(riscv_use_bscan_tunnel)
...
+
+	if (CMD_ARGC > 0) {
+		n = jim_nvp_name2value_simple(nvp_maskisr_modes, CMD_ARGV[0]);
+		if (!n->name)
+			return ERROR_COMMAND_SYNTAX_ERROR;
+		info->isrmask_mode = n->value;
+	}
+
+	n = jim_nvp_value2name_simple(nvp_maskisr_modes, info->isrmask_mode);
+	command_print(CMD, "riscv interrupt mask %s", n->name);
```
timsifive: Is that expected? I don't think any of the riscv commands echo their result. Should they?

src/target/riscv/riscv.h
```diff
@@ -56,6 +56,13 @@ enum riscv_halt_reason {
 	RISCV_HALT_ERROR
 };
 
+enum riscv_isrmasking_mode {
+	RISCV_ISRMASK_AUTO,	/* not supported yet */
```
timsifive: Let's not include them in the enum at all then. If you want them listed as a future reminder, then comment them out.

timsifive (changes requested):
Overall the change looks OK to me. Can you also add documentation for the command in doc/openocd.texi?

src/target/riscv/riscv.h
erhankur: I left comment for the future.

src/target/riscv/riscv.c
erhankur: Printing result allows us to get the current value without setting.  It can be useful for the tests like `store_existing_setting, change_it, restore_previous_value` 

`riscv maskisr` --> will return existing value

Also I think it is good to see the result in the telnet command line.

JanMatCodasip: Another option for your consideration: You may display the current setting only if the command is called without any parameters. For example OpenOCD's commands `tcl_port`, `telnet_port` (and some others) work this way. Unfortunately, OpenOCD commands do not behave consistently and we would find exceptions. 

In any case, it would be good to remain consistent at least within the riscv target. I would like to learn what is the preferred behavior from the folks in upstream, then update the riscv commands to follow that.

src/target/riscv/riscv.c
```diff
@@ -3500,6 +3533,55 @@ static int riscv_resume_go_all_harts(struct target *target)
 	return ERROR_OK;
 }
 
+int riscv_interrupts_disable(struct target *target, uint64_t ie_mask, uint64_t *old_mstatus)
+{
+	RISCV_INFO(info);
+	uint64_t current_mstatus;
+	uint8_t mstatus_bytes[8] = { 0 };
```
JanMatCodasip: Just a little thing: You can move the variable definitions closer to the point of their first use. 

OpenOCD's code style in upstream no longer requires (nor encourages) the old ANSI C style where variables had to be defined at the beginning of a block.

src/target/riscv/riscv.c
```diff
@@ -2372,12 +2356,26 @@ int riscv_openocd_step(struct target *target, int current,
...
 	if (out != ERROR_OK) {
 		LOG_ERROR("unable to step rtos hart");
 		return out;
 	}
 
+	if (info->isrmask_mode == RISCV_ISRMASK_STEPONLY) {
+		if (riscv_interrupts_restore(target, current_mstatus) != ERROR_OK)
+			return ERROR_FAIL;
+	}
+
```
JanMatCodasip: Even if riscv_step_rtos_hart() fails, I believe we shall still try to restore mstatus to the original value, to leave the hart in as unchanged as possible state.

src/target/riscv/riscv.c
```diff
@@ -1952,26 +1952,11 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
...
-	current_mstatus = buf_get_u64(reg_mstatus->value, 0, reg_mstatus->size);
 	uint64_t ie_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
-	buf_set_u64(mstatus_bytes, 0, info->xlen, set_field(current_mstatus,
-				ie_mask, 0));
-
-	reg_mstatus->type->set(reg_mstatus, mstatus_bytes);
+	if (riscv_interrupts_disable(target, ie_mask, &current_mstatus) != ERROR_OK)
+		return ERROR_FAIL;
 
 	/* Run algorithm */
```
JanMatCodasip, line 1962: There is potential issue in riscv_run_algorithm() - if the algorithm execution fails at some point below this line (`return <whatever>;`), the saved register contents won't be restored - not even attempted.

This risk was already present in the original code, so I believe the pull request need not be blocked by that. The issue can be addressed in a separate patch.

src/target/riscv/riscv.c
```diff
@@ -3325,6 +3349,13 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 			"(optional) to indicate Bscan Tunnel Type {0:(default) NESTED_TAP , "
 			"1: DATA_REGISTER}"
 	},
+	{
+		.name = "maskisr",
```
JanMatCodasip: The other option-setting TCL commands so far follow the `set_<something>` naming convention.

@timsifive - I do not have any preference about the naming convention of the commands but I'd like them to be unified. Similarly, it would be good to have a unified way how to just display the current value without setting a new one (see my earlier comment in this pull request). What is your opinion on that?

src/target/riscv/riscv.c
```diff
@@ -3325,6 +3349,13 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
+		.handler = riscv_mask_interrupts,
```
JanMatCodasip: Please consider renaming the handler function to `riscv_maskisr` to follow the naming convention already used for the other handlers.

src/target/riscv/riscv.c
```diff
@@ -1952,26 +1952,11 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
...
-	struct reg *reg_mstatus = register_get_by_name(target->reg_cache,
-			"mstatus", true);
-	if (!reg_mstatus) {
-		LOG_ERROR("Couldn't find mstatus!");
-		return ERROR_FAIL;
-	}
-
-	reg_mstatus->type->get(reg_mstatus);
-	current_mstatus = buf_get_u64(reg_mstatus->value, 0, reg_mstatus->size);
 	uint64_t ie_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
```
JanMatCodasip: Just a suggestion: Could you rename `ie_mask` to `irq_disabled_mask`, `disabled_irqs` (or something similar) - to make it clear that the bits indicate which irqs get disabled, not enabled. 

There are more occurences of `ie_mask`. Preferably, address them all.

JanMatCodasip (commented):
@erhankur - Thank you for your contribution.

Overall, the change looks all right. Please look at my feedback:

- two items are important: restoring of registers when a certain operation fails (single-step or algorithm)
- the remaining comments relate to cosmetic stuff like naming conventions

src/target/riscv/riscv.c
erhankur: @JanMatCodasip thanks to point out this behavior. Looking at the other parts,  I was in between restore/return or just return. So what about enabling triggers if there is step or irq_restore  error?

src/target/riscv/riscv.c
erhankur: Yes right. I just changed the existing code block with the function. Do you have a backlog to follow todo issues?

src/target/riscv/riscv.c
timsifive: I agree on both points. Consistency is more important than the specific solution. Unfortunately we're pretty far away from that.

It looks like most riscv commands that set an option use set_, so let's stick with that for this PR.

As far as displaying a value, I prefer to show the value when there are no arguments, and to be silent when an option is set.

src/target/riscv/riscv.c
erhankur: @timsifive your comments are also welcome.

src/target/riscv/riscv.c
JanMatCodasip: I have recorded the issue here: https://github.com/riscv/riscv-openocd/issues/682

src/target/riscv/riscv.c
JanMatCodasip: > Looking at the other parts, I was in between restore/return or just return.

You're right - the risk of not restoring everything has already been present in the original code (restore of triggers would not be attempted if step fails).

> So what about enabling triggers if there is step or irq_restore error?

IMO, an ideal solution would be to rework this function so that in case of any failure in any of the steps, whatever got changed gets restored.  I understand that this is additional work, not directly related to the new "maskisr" command.

Please create an issue for that so that we don't forget to return to this.

src/target/riscv/riscv.c
erhankur: Ok. changed the command as `set_maskisr` and did not print information for set.

src/target/riscv/riscv.c
erhankur: Didn't want to create another technical debt.  I did some restore work. Please check it again.

src/target/riscv/riscv.c
erhankur: Done

src/target/riscv/riscv.c
```diff
@@ -2372,22 +2356,38 @@ int riscv_openocd_step(struct target *target, int current,
...
-		return ERROR_FAIL;
+	if (info->isrmask_mode == RISCV_ISRMASK_STEPONLY)
+		out |= riscv_interrupts_restore(target, current_mstatus);
 
-	target->state = TARGET_RUNNING;
-	target_call_event_callbacks(target, TARGET_EVENT_RESUMED);
-	target->state = TARGET_HALTED;
-	target->debug_reason = DBG_REASON_SINGLESTEP;
-	target_call_event_callbacks(target, TARGET_EVENT_HALTED);
+_exit_riscv_openocd_step:
```
JanMatCodasip: Cosmetic stuff: The goto labels have function scope, so you can safely use a shorter name - e.g. `done` or `exit`. There won't be any conflicts with labels in different functions.

erhankur: Yes. Right. Used _exit.

src/target/riscv/riscv.c
```diff
@@ -2372,22 +2356,38 @@ int riscv_openocd_step(struct target *target, int current,
...
+	if (info->isrmask_mode == RISCV_ISRMASK_STEPONLY)
+		out |= riscv_interrupts_restore(target, current_mstatus);
 
-	target->state = TARGET_RUNNING;
-	target_call_event_callbacks(target, TARGET_EVENT_RESUMED);
-	target->state = TARGET_HALTED;
-	target->debug_reason = DBG_REASON_SINGLESTEP;
-	target_call_event_callbacks(target, TARGET_EVENT_HALTED);
+_exit:
+	out |= enable_triggers(target, trigger_state);
```
JanMatCodasip: Please, do not use `|=` operator on the error codes. Although it works in this case (because ERROR_OK equals to zero), it is not clean. It may also lead to returning a non-sense error code, created as a bitwise mixture of two different error codes (`ERROR_SOMETHING1 | ERROR_SOMETHING2`).

One possible way how to fix it: use `bool success = true;` instead of the `int out` variable, set it to false when any operation fails. Then at the end of function: `return success ? ERROR_OK : ERROR_FAIL`.

If you decide to remember the full error code of the first failed call (instead of just success flag), that is also perfectly fine, but requires few more lines of code.

JanMatCodasip (commented):
Thank you for the changes. I inspected the code visually and it looks all right.

I have one last comment related to the code style (the use of `|=`).

src/target/riscv/riscv.c
erhankur: @JanMatCodasip thanks for feedbacks and reviews. Changed as you suggested. Please review again.

timsifive:
Broader question about this PR. The debug spec defines dcsr.stepie, which defaults to 0. When it is 0 "Interrupts (including NMI) are disabled during single stepping." Why is this change needed? Is there hardware that has the bit wired to 1?

JanMatCodasip (commented):
The code looks all right, thanks.

Please, discuss also with Tim whether there is a specific reason why `dcsr.stepie` bit cannot be used. It did not occur to me initially that there is already this dedicated bit in `dcsr` for disabling the interrupts during step.

erhankur:
@JanMatCodasip thanks. I will clarify `stepie` bit usage. 

Can you also help me to pass the tests? No idea why `MemorySampleSingle` test fails. 

./gdbserver.py ./targets/RISC-V/spike32-2.py MemorySampleSingle

[20220218-084040-spike32_2-MemorySampleMixed.log](https://github.com/riscv/riscv-openocd/files/8095980/20220218-084040-spike32_2-MemorySampleMixed.log)

Warn : Error writing to GDB socket. Dropping the connection. 

It is not related to my branches. Same with `riscv`

erhankur:
> Broader question about this PR. The debug spec defines dcsr.stepie, which defaults to 0. When it is 0 "Interrupts (including NMI) are disabled during single stepping." Why is this change needed? Is there hardware that has the bit wired to 1?

Thanks @timsifive to point out this bit. Simple answer we just missed that bit. We use `maskisr` command for the `xtensa` architecture and implemented it for `riscv` as the same way. 

So, in this case, what is your suggestion? Correct me if I am wrong, since hardwiring  to 1 or 0 is implementation defined, vendors can still need this command if it is wired to 1.

timsifive:
The current OpenOCD behavior is to write 0 to stepie, the result being that you never single step into an interrupt handler, unless the hardware hard-wires stepie to 1. (I'm not aware of any hardware that does that.)

For hardware that hard-wires stepie to 1, your new option would be useful. But if we don't know of any hardware like that then it seems like this change just makes things more complicated, and risks users setting it when they don't need to. That both slows down single step, and changes the behavior of code when a user steps over an instruction that reads/writes mstatus.

erhankur:
OK. I will remove `riscv set_maskisr` command and dependencies. Can we keep the other parts? I can make another PR for the remaining part.

timsifive:
The other part being the refactor that created riscv_interrupts_disable()? Yes, that is a nice improvement.

erhankur:
@timsifive As I just learned we implemented v0.13 for all riscv chips. Unfortunately `stepie` bit is not exist at v0.13. 

So, can we add this command to support earlier version of v0.13.2?

timsifive:
I thought that on specs before stepie was added, interrupts are disabled during single stepping. 0.13 is a work-in-progress version, and there are probably multiple variants of it floating around. Can you share the PDF or a URL of the exact document you implemented?

erhankur:
I have asked for the exact document. In the meantime, I found this https://static.dev.sifive.com/riscv-debug-spec-0.13.b4f1f43.pdf

timsifive:
Thanks. In the link you posted, the definition of the `step` bit says:

> When set and not in Debug Mode, the hart will
> only execute a single instruction and then enter
> Debug Mode. *Interrupts are disabled when this
> bit is set.* If the instruction does not complete due
> to an exception, the hart will immediately enter
> Debug Mode before executing the trap handler,
> with appropriate exception registers set.

erhankur:
Confirmed with the team. We implemented the url I sent above.

As a conclusion;
1 - We didn't implement `stepie` since it is not exist in v0.13
2 - Interrupts are kept `enabled` during a single step to give control to the debugger software. 

In my opinion, safest place to disable/enable interrupts is inside `riscv_openocd_step`.  Changes will stay at local,  no events will be fired without restoring, tested for the long time, working well etc.

I moved disable/enable routines to the upper layer  and run some of our tests but not worked as expected. Need to look at deeply. 

@timsifive sorry for bringing this on the table again.  I understand your concern but don't think so this will bring much overhead since `isrmask_mode` will be disabled by default.  Ordinary users will not touch this setting.  

So, can we add this command to the upstream? Otherwise looks like things will be getting more complex.

timsifive:
So to recap, you have hardware that does not implement stepie (because it's based on an old spec that doesn't have it) and keeps interrupts enabled during single step, even though the spec says they should be disabled.

stepie was defined so that older implementations would appear to have stepie hard-wired to 0, and everything just works. Except your hardware looks like stepie is hard-wired to 0 but interrupts are enabled in any case.

Right now OpenOCD doesn't allow the user to set stepie to 1. So long-term the functionality we want is:
1. Give the user a setting that says "step into interrupts" or "don't step into interrupts." OpenOCD can set stepie based on this setting. OpenOCD can detect whether the setting is supported, and give an error if the desired behavior isn't supported.
2. Give the user a setting that says "mask interrupts while stepping" which is orthogonal to the first one. This is really only useful if you do not want to step into interrupts, and the hardware doesn't support stepie=0.

This change implements the second part of that, so that's a good improvement.

Can you update the documentation to make it clear that this feature is only useful on hardware that always steps into interrupts, and also that disabling the new option does not guarantee that single stepping will go into interrupt handlers? (To make that happen, stepie would have to be written to 1 as well.)

erhankur:
@timsifive thank you for your support. Document is updated.

timsifive (dismissed):

doc/openocd.texi
```diff
@@ -10424,6 +10424,13 @@ tunneled DR scan consists of:
 
 @end deffn
 
+@deffn {Command} {riscv set_maskisr} [@option{off}|@option{steponly}]
+Selects whether interrupts will be disabled when single stepping. The default configuration is @option{off}.
+This feature is only useful on hardware that always steps into interrupts and doesn't support stepie=0.
+Keep in mind, disabling the option does not guarantee that single stepping will go into interrupt handlers.
+To make that happen, stepie would have to be written to 1 as well.
```
JanMatCodasip: Just one last cosmetic suggestion: For clarity, it would be good to mention also the register name - `stepie` --> `dcsr.stepie`.

When dealing with RISC-V debug, it is important to be very clear about which registers we talk about, because there are (at least?) three different register spaces involved:

* JTAG registers (IR, DRs)
* Debug module registers (dmstatus, dmcontrol, ...)
* Hart's achitectural registers (GPRs, CSRs, ...)

Thanks.

```suggestion
This feature is only useful on hardware that always steps into interrupts and doesn't support dcsr.stepie=0.
Keep in mind, disabling the option does not guarantee that single stepping will go into interrupt handlers.
To make that happen, dcsr.stepie would have to be written to 1 as well.
```

erhankur: Done

JanMatCodasip (approved):

doc/openocd.texi
JanMatCodasip: Thanks.

PR #683, opened 2022-02-21
add callback for target reset

erhankur:
In this MR  `on_reset` callback function added to inform upper layers when reset happened. 

Reset can be triggered by user request or in an unexpected time. Before displaying an "unexpected reset" warning message, target state checked to inform users correctly.

Change-Id: Ife0244f31db06a93a264e797501d5266243364cd

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -4263,6 +4263,7 @@ static int riscv013_on_halt(struct target *target)
 static bool riscv013_is_halted(struct target *target)
 {
 	uint32_t dmstatus;
+
 	if (dmstatus_read(target, &dmstatus, true) != ERROR_OK)
 		return false;
 	if (get_field(dmstatus, DM_DMSTATUS_ANYUNAVAIL))
@@ -4271,7 +4272,10 @@ static bool riscv013_is_halted(struct target *target)
 		LOG_ERROR("Hart %d doesn't exist.", riscv_current_hartid(target));
 	if (get_field(dmstatus, DM_DMSTATUS_ANYHAVERESET)) {
 		int hartid = riscv_current_hartid(target);
-		LOG_INFO("Hart %d unexpectedly reset!", hartid);
+		if (target->state != TARGET_RESET) {
+			/* warn for "unexpected" reset when it is not requested by user */
+			LOG_INFO("Hart %d unexpectedly reset!", hartid);
+		}
 		/* TODO: Can we make this more obvious to eg. a gdb user? */
 		uint32_t dmcontrol = DM_DMCONTROL_DMACTIVE |
 			DM_DMCONTROL_ACKHAVERESET;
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -4284,6 +4288,10 @@ static bool riscv013_is_halted(struct target *target)
 		if (target->state == TARGET_HALTED)
 			dmcontrol |= DM_DMCONTROL_HALTREQ;
 		dmi_write(target, DM_DMCONTROL, dmcontrol);
+
+		RISCV_INFO(r);
+		if (r->on_reset)
+			r->on_reset(target);
```
JanMatCodasip, line 4294: May I ask what the purpose (use case) of this new `on_reset` callback is?

There is no implementation of the callback added. Is a concrete implementation of the callback something you would like to add later?

Also note that OpenOCD already has multiple "on reset" events implemented, for which you can write own TCL callbacks and they will be invoked during target reset. Maybe those events can cover your use cases:

https://openocd.org/doc/html/CPU-Configuration.html#Target-Events

erhankur: I think, reset events are designed to use before/after user triggered resets. (Reset from OpenOCD)
We need that callback to detect reset in case of HW reset button pressed. (Unexpected reset)

In the `on_reset` implementation we are doing some rtos cleaning and displaying the last reset reason.

JanMatCodasip: > I think, reset events are designed to use before/after user triggered resets. (Reset from OpenOCD)
> We need that callback to detect reset in case of HW reset button pressed. (Unexpected reset)

Thank you for explaining your use case.

> In the on_reset implementation we are doing some rtos cleaning and displaying the last reset reason.

Two broader points:

(1) The on-reset actions will generally be different for each RISC-V platform - for example reading the reset reason from a platform-specific memory-mapped register. So it would be good to have the "on_reset" event available on the TCL level, and the users could write their own handlers in TCL. We can't expect an ordinary user to edit the `on_reset` handler directly in OpenOCD source :). 

If you're interested how that is done, you can find an example in `arm_tpiu_swo.c` ([registration of custom event handlers](https://github.com/riscv/riscv-openocd/blob/d97da0eb74330bdeaca0991967451c17f279d062/src/target/arm_tpiu_swo.c#L487), [invocation of a handler](https://github.com/riscv/riscv-openocd/blob/d97da0eb74330bdeaca0991967451c17f279d062/src/target/arm_tpiu_swo.c#L163)).

But that would be a task for a separate merge request (future).

(2) Also note that the on-reset processing may not be reliable unless the `dmcontrol.setresethaltreq` bit is set to halt the target immediately after the reset.

If that bit is not set (or the target does not support it), then the processor will execute an unknown (undetermined) number of instructions before OpenOCD detects the reset, and may be in undetermined state at that point. Or even worse, the target may even reset multiple times before OpenOCD detects that the reset happened.

erhankur: @JanMatCodasip thanks for the valuable points.

For our case it is important to reset FreeRTOS settings when reset happens. We can reset multiple times or whenever OpenOCD detects. We have not seen an issue so far with this implementation. But it is a really good point.

I am open to add features to the more generic layers. Will have a look at your custom event suggestion in another MR.

erhankur:
Hi @JanMatCodasip, is this PR OK for you? Anything from my side? 
@timsifive can you also please have a look?

timsifive:
What exactly is the use case for this? Why does the FreeRTOS layer need to be told when a reset happens?

I share Jan's concern that this is only useful to people modifying the OpenOCD source code, and generally people shouldn't need to do that.

timsifive:
Inactive.

PR #684, opened 2022-03-01
From upstream

timsifive:


src/target/smp.h
```diff
@@ -19,10 +19,14 @@
...
 #define OPENOCD_TARGET_SMP_H
 
+#include <helper/list.h>
 #include "server/server.h"
 
 #define foreach_smp_target(pos, head) \
-	for (pos = head; (pos); pos = pos->next)
+	list_for_each_entry(pos, head, lh)
+
+#define foreach_smp_target_direction(forward, pos, head) \
```
GregSavin, line 28: Did somebody upstream add foreach_smp_target_direction just to accommodate riscv-specific code?  Looks like riscv code is the only place that ever walks the list in reverse order.  Well, I guess after the change was made from target->head to target->smp_targets, the riscv-specific code wouldn't compile, so it made sense to add foreach_smp_target_direction to solve that.

GregSavin (approved):

src/target/smp.h
timsifive: Yes, in https://review.openocd.org/c/openocd/+/6783. This replaces my code where I made a new list in memory and iterated through that.

PR #686, opened 2022-03-16
Fix & improvements in Linux Build workflow

JanMatCodasip:
Fix: 

- added missing apt-get update (otherwise the subsequent apt-get install may end with HTTP 404 errors)

Minor improvements:

- Use parallel build (use -j for make) .. reduces the total CI run time by approx 25%
- Check that the resulting OpenOCD executable can actually be launched (call openocd --version)


timsifive (approved):
Thank you!

PR #687, opened 2022-03-16
Fix: Set proper debug_reason in deassert_reset()

JanMatCodasip:
The issue was visible for example when user's .cfg file ended
with "reset halt" command:

In such case, the hart would remain halted but the debug_reason would not be
updated and may retain an incorrect value, e.g. DBG_REASON_NOTHALTED.
As a result, gdb_last_signal() would provide an incorrect reply to GDB.

Change-Id: Ie6f050295fb5cbe9db38b189c4bc385662acf5b4

### DIFF
--- a/.github/workflows/checkpatch.yml
+++ b/.github/workflows/checkpatch.yml
@@ -1,6 +1,6 @@
 on: pull_request
 
-name: Check Code Style
+name: Check Code Style (checkpatch)
 
 jobs:
   check:
@@ -16,6 +16,9 @@ jobs:
       - run: sudo apt-get install patchutils
       - name: Run checkpatch
         run: |
-            git diff -U20 HEAD~40 | \
-                    filterdiff -x "a/src/jtag/drivers/libjaylink/*" -x "a/tools/git2cl/*" | \
-                    ./tools/scripts/checkpatch.pl --no-signoff -
+            git diff -U20 HEAD~40 \
+            | filterdiff \
+                -x "a/src/jtag/drivers/libjaylink/*" \
+                -x "a/tools/git2cl/*" \
+                -x "a/.github/*" \
+            | ./tools/scripts/checkpatch.pl --no-signoff -
--- a/.github/workflows/checkpatch.yml
+++ b/.github/workflows/checkpatch.yml
@@ -13,7 +13,10 @@ jobs:
         uses: actions/checkout@v2
         with:
             fetch-depth: 50
-      - run: sudo apt-get install patchutils
+      - name: Install required packages (apt-get)
+        run: |
+            sudo apt-get update
+            sudo apt-get install patchutils
       - name: Run checkpatch
         run: |
             git diff -U20 HEAD~40 \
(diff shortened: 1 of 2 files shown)


JanMatCodasip:
Riscv-tests/debug have all passed (launched via CI in https://github.com/JanMatCodasip/riscv-openocd/pull/5).

The failing workflow `Linux Build / build32` is an unrelated issue, and addressed in a separate pull request: https://github.com/riscv/riscv-openocd/pull/686

timsifive (commented):
This looks right. Have you run all the debug tests?

Assuming so, let's wait until I merge #686 (tomorrow, unless anybody comments), then rebase this change so all the checks pass.

JanMatCodasip:
> Have you run all the debug tests?

Yes, I have executed `make all` in riscv-tests/debug and all the tests passed. To launch the testing, I have re-used the (so far dormant) CI automation proposal and the results can be found here: https://github.com/JanMatCodasip/riscv-openocd/pull/5/checks#step:9:415

Broader point: 

I have tried re-running the debug tests multiple times and no longer see any sporadic fails which I occasionally encountered in the past. The testing now appears solid. (I am not sure which specific changes in OpenOCD and/or the tests addressed the remaining issues, though.) 

Since riscv-tests and all the required components are not straightforward to set up, I would like to revive the idea of having an automation. The original approach in https://github.com/riscv/riscv-openocd/pull/563 was to run the riscv-tests in CI via the github actions. Because the tests are complex and have many dependencies, I consider a different approach - create a script to run the testing locally in an isolated environment (docker container), which developers would manually execute and this way have more control over the testing process (version of spike, version of riscv-tests, ..., ...). @timsifive, does this approach seem OK to you? Alternate suggestions welcome.

timsifive:
I would love something automated using github actions. Having a script/docker image would be helpful for contributors who don't have spike/gcc set up, but there aren't that many contributors. I probably wouldn't use it since I already have a setup. Without regular use it would bitrot quickly. github actions have the benefit that it can't bitrot, because we would notice right away.

JanMatCodasip:
> let's wait until I merge https://github.com/riscv/riscv-openocd/pull/686 (tomorrow, unless anybody comments), then rebase this change so all the checks pass.

I have rebased this change but had to add one more commit to avoid false alarms from checkpatch - prevent checkpatch from looking at changes in the .github/ directory. Checkpatch had found "CC:" substrings in the *.yml files and apparently mis-interpreted them as signatures in commit messages.

JanMatCodasip:
> I would love something automated using github actions. Having a script/docker image would be helpful for contributors who don't have spike/gcc set up, but there aren't that many contributors. I probably wouldn't use it since I already have a setup. Without regular use it would bitrot quickly. github actions have the benefit that it can't bitrot, because we would notice right away.

Thank you for replying. So it seems one set of scripts that can be run both locally and as part of the github actions could be the way to go.

timsifive (dismissed):

timsifive:
> Thank you for replying. So it seems one set of scripts that can be run both locally and as part of the github actions could be the way to go.

That would be ideal.

JanMatCodasip:
Sorry for one last quick change. The `apt-get update` was missing also in the checkpatch workflow, so I have added it there, too. It hasn't been an issue now but might pop up in the future.

timsifive (approved):

PR #689, opened 2022-04-08
riscv: Increase batch allocation size to improve transfer speed.

Dolu1990:
Hi,

I did a speed test for file transfer, looking at the scope, there was quite some dead time between each batch of commands, so, this PR increase the maximal size of jtag batches in order to improve the speed.

Here is a few test i did based on a FT2232, targetting a core running at 100 Mhz, JTAG at the maximum of the FTDI was able to (probably around 20 Mhz).

Batch size = 32 (as it is currently in upstream)
downloaded 1152044 bytes in 2.192316s (513.176 KiB/s)
dumped 1152044 bytes in 3.017069s (372.893 KiB/s)

Batch size = 128 (as it is in this patch)
downloaded 1152044 bytes in 1.357257s (828.909 KiB/s)
dumped 1152044 bytes in 2.227927s (504.973 KiB/s)

Batch size = 1024
downloaded 1152044 bytes in 1.151690s (976.863 KiB/s)
dumped 1152044 bytes in 2.018675s (557.318 KiB/s)

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3151,7 +3151,7 @@ static int read_memory_progbuf_inner(struct target *target, target_addr_t addres
 		 * dm_data0 contains[read_addr-size*2]
 		 */
 
-		struct riscv_batch *batch = riscv_batch_alloc(target, 32,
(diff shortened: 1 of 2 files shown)


timsifive (approved):
That's a nice speedup!

Dolu1990:
It help loading kernel images ^^

Memory access implemented via progbuf

PR #690, opened 2022-04-08
riscv: Add a option to specify the JTAG TAP tunnel IR 

Dolu1990:
Hi,

This patch fix :
https://github.com/riscv/riscv-openocd/blob/0a70e59cb86deed71fd92ceffdd233478e237ebb/src/target/riscv/riscv.c#L454

The TAP IR used when tunneling was hardcoded to 0x23 << IR_WIDTH-6, which refer to some Xilinx FPGA USER4 BSCANE.

This PR allows to select which IR is used in the TAP to access the tunnel entry, allowing targetting other FPGA vendors / device families.

This PR It is backward compatible.

TCL Usage example : 
riscv set_bscan_tunnel_ir 0x23

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10494,7 +10494,7 @@ tunneled DR scan consists of:
 
 @deffn {Command} {riscv set_bscan_tunnel_ir} value
 Allows the use_bscan_tunnel feature to target non Xilinx device by
-specifing the JTAG TAP IR used to access the bscan tunnel.
+specifying the JTAG TAP IR used to access the bscan tunnel.
 @end deffn
 
 @deffn {Command} {riscv set_maskisr} [@option{off}|@option{steponly}]
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10492,6 +10492,11 @@ tunneled DR scan consists of:
 
 @end deffn
 
+@deffn {Command} {riscv set_bscan_tunnel_ir} value
+Allows the use_bscan_tunnel feature to target non Xilinx device by
+specifing the JTAG TAP IR used to access the bscan tunnel.
+@end deffn
+
 @deffn {Command} {riscv set_maskisr} [@option{off}|@option{steponly}]
 Selects whether interrupts will be disabled when single stepping. The default configuration is @option{off}.
 This feature is only useful on hardware that always steps into interrupts and doesn't support dcsr.stepie=0.
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -451,8 +451,12 @@ static int riscv_init_target(struct command_context *cmd_ctx,
 	select_idcode.num_bits = target->tap->ir_length;
 
 	if (bscan_tunnel_ir_width != 0) {
-		assert(target->tap->ir_length >= 6);
-		uint32_t ir_user4_raw = bscan_tunnel_ir_id != 0 ? bscan_tunnel_ir_id : 0x23 << (target->tap->ir_length - 6);
+		uint32_t ir_user4_raw = bscan_tunnel_ir_id;
+		/* Provide a default value which target some Xilinx FPGA USER4 IR */
+		if (ir_user4_raw == 0) {
+			assert(target->tap->ir_length >= 6);
+			ir_user4_raw = 0x23 << (target->tap->ir_length - 6);
+		}
 		ir_user4[0] = (uint8_t)ir_user4_raw;
 		ir_user4[1] = (uint8_t)(ir_user4_raw >>= 8);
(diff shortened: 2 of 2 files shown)


src/target/riscv/riscv.c
```diff
@@ -129,6 +129,7 @@ struct scan_field select_idcode = {
 
 bscan_tunnel_type_t bscan_tunnel_type;
 int bscan_tunnel_ir_width; /* if zero, then tunneling is not present/active */
+int bscan_tunnel_ir_id; /* IR ID of the JTAG TAP to access the tunnel. Valid when not 0 */
```
```suggestion
static int bscan_tunnel_ir_id = 0x23; /* IR ID of the JTAG TAP to access the tunnel. Valid when not 0 */
```

src/target/riscv/riscv.c
```diff
@@ -451,7 +452,7 @@ static int riscv_init_target(struct command_context *cmd_ctx,
 
 	if (bscan_tunnel_ir_width != 0) {
 		assert(target->tap->ir_length >= 6);
-		uint32_t ir_user4_raw = 0x23 << (target->tap->ir_length - 6);
+		uint32_t ir_user4_raw = bscan_tunnel_ir_id != 0 ? bscan_tunnel_ir_id : 0x23 << (target->tap->ir_length - 6);
```
```suggestion
		uint32_t ir_user4_raw = bscan_tunnel_ir_id << (target->tap->ir_length - 6);
```

timsifive (changes requested):
Can you also document this command in doc/openocd.texi?

src/target/riscv/riscv.c
Dolu1990: The original code was "0x23 << (target->tap->ir_length - 6);" I may be wrong but I think the intend of that shift was to cover some other device from xilinx  which would follow the rule  0x23 << (ir_length - 6), which is very specific.

If it is the case, i think that shift should not be applied to user specified bscan_tunnel_ir_id.

So, 
```c
uint32_t ir_user4_raw = bscan_tunnel_ir_id << (target->tap->ir_length - 6);
```
would be problematic.

Dolu1990:
> Can you also document this command in doc/openocd.texi?

Sure

src/target/riscv/riscv.c
Dolu1990: > static int bscan_tunnel_ir_id = 0x23;

I had it to 0 (disabled by default) to ensure backward compatibility (no argument => Ok for some Xilinx devices). 


The following : 
```c
static int bscan_tunnel_ir_id = 0x23;
+ 
uint32_t ir_user4_raw = bscan_tunnel_ir_id != 0 ? bscan_tunnel_ir_id : 0x23 << (target->tap->ir_length - 6);
```
would be problematic, as it would never trigger the current upstream behaviour, right ?

src/target/riscv/riscv.c
timsifive: Thanks. I missed that detail.

I had hoped there was a way to default to the current behavior without the extra logic, but if the current behavior depends on ir_length then that's really not doable and your implementation is our best option.

Dolu1990:
Documentation pushed.

Once the PR is merged, i will take a look at the bscan_tunnel_xxx variable. They are currently defined as global variable, instead of being attribute of a given target. So currently, in a JTAG chain with multiple RISC-V TAP, settings will collide.

RetroHQ:
The assert (was at line 450) --
assert(target->tap->ir_length >= 6)

also needs removing or changing, there are chips (such as Efinix Trion) which support a BSCAN style USER TAP interface but have an ilen of less than 6 (4 in this case).

Dolu1990:
> assert(target->tap->ir_length >= 6)

@RetroHQ Hooo yes right ^^, i just moved it in the proper place. Thanks

GregSavin (dismissed):
Tim asked me to review this because I wrote the initial implementation that was narrowly targeted.  These changes look good to me.

doc/openocd.texi
```diff
@@ -10492,6 +10492,11 @@ tunneled DR scan consists of:
 
 @end deffn
 
+@deffn {Command} {riscv set_bscan_tunnel_ir} value
+Allows the use_bscan_tunnel feature to target non Xilinx device by
+specifing the JTAG TAP IR used to access the bscan tunnel.
```
timsifive:
```suggestion
specifying the JTAG TAP IR used to access the bscan tunnel.
```

src/target/riscv/riscv.c
```suggestion
static int bscan_tunnel_ir_id; /* IR ID of the JTAG TAP to access the tunnel. Valid when not 0 */
```

timsifive (changes requested):

timsifive (approved):
Looks good!

PR #691, opened 2022-04-11
From upstream

timsifive:


doc/openocd.texi
```diff
@@ -8491,6 +8540,78 @@ Debug and trace infrastructure:
 @end example
 @end deffn
 
+@deffn {Command} {set_reg} dict
```
GregSavin, line 8543: These new Tcl interfaces (set_reg, write_memory etc) are a great addition from upstream.

src/target/startup.tcl
```diff
@@ -206,6 +206,32 @@ proc init_target_events {} {
 proc init_board {} {
 }
 
+proc mem2array {arrayname bitwidth address count {phys ""}} {
```
GregSavin, line 209: Ah, good, there will be a grace period before current mem2array usages in script will need to be converted.

GregSavin (approved):
Looks like no big RISC-V related changes.  Some interesting enhancements for Tcl scripts.

PR #693, opened 2022-04-30
Use updated macros, which take arguments as expected.

timsifive:
This is more to test the debug changes than to improve OpenOCD in any
way.

Change-Id: If2c3424af89f93952049a519108762a0644405ec


timsifive:
I ended up with two branches with the same changes.

PR #694, opened 2022-04-30
Look at trigger hit bits to see which trigger was hit.

timsifive:
If that doesn't work (the hit bit is optional), then disassemble as we used to do.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1252,11 +1252,15 @@ int riscv_flush_registers(struct target *target)
 /* Convert: RISC-V hart's halt reason --> OpenOCD's generic debug reason */
 int set_debug_reason(struct target *target, enum riscv_halt_reason halt_reason)
 {
+	RISCV_INFO(r);
+	r->trigger_hit = -1;
 	switch (halt_reason) {
 		case RISCV_HALT_BREAKPOINT:
 			target->debug_reason = DBG_REASON_BREAKPOINT;
 			break;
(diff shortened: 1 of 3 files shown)


src/target/riscv/riscv.c
```diff
@@ -1017,6 +1017,67 @@ int riscv_remove_watchpoint(struct target *target,
...
+			LOG_DEBUG("Trigger %d (unique_id=%d) has hit bit set.", i, r->trigger_unique_id[i]);
+			if (riscv_set_register(target, GDB_REGNO_TDATA1, tdata1 & ~hit_mask) != ERROR_OK)
+				return ERROR_FAIL;
+
+			*unique_id = r->trigger_unique_id[i];
+			break;
+		}
+	}
+
+	if (riscv_set_register(target, GDB_REGNO_TSELECT, tselect) != ERROR_OK)
```
GregSavin, line 1075: If reached, any of the occurrences of "return ERROR_FAIL;" within the loop would appear to skip the restoration of the original tselect value.  Might that cause a problem in practice?

timsifive: Not in practice. The only way we get an error is if we're unable to read/write a register. If that's the case, it seems unlikely that writing tselect is going to work. Reading/writing a register (that we know exists) should never fail unless the target is in some weird state, the cable got unplugged, etc.

Still wouldn't hurt to try, of course.

GregSavin (approved):

PR #696, opened 2022-05-04
From upstream

timsifive:
Get code from upstream, which includes support for gd32vf103 using the stm32f1x driver. So I removed the dedicated gd32vf103 code in this branch since it's no longer needed.

Then brought in change https://review.openocd.org/c/openocd/+/6886 to fix a bug that had been introduced in upstream.


.github/workflows/snapshot.yml
```diff
@@ -55,6 +55,7 @@ jobs:
           LIBUSB1_CONFIG: --enable-shared --disable-static
           HIDAPI_CONFIG: --enable-shared --disable-static --disable-testgui
           CAPSTONE_CONFIG: "CAPSTONE_BUILD_CORE_ONLY=yes CAPSTONE_STATIC=yes CAPSTONE_SHARED=no"
+          CAPSTONE_CFLAGS: -I$(CAPSTONE_SRC)/include/capstone
```
borneoa, line 58: From github build log, capstone includes get installed in /home/runner/work/riscv-openocd/build/i686-w64-mingw32-root/usr/include/capstone
Why using the source code here instead of the installation?
Apparently the issue is in some crap commands run by github on the file capstone.pc AFTER it is installed.
From line 2042 to 2049
https://github.com/riscv/riscv-openocd/runs/6293338772?check_suite_focus=true#step:8:2042
The file BEFORE this crap works and
 pkg-config --cflags capstone
 -I/usr/include/capstone
but after the crap the same command gives no output!
No idea how to fix it

timsifive: This is definitely a hack, but all I really care about is that the Windows build passes so it'll catch problems with new code. If this is fixed properly upstream then that will come down here in time.

timsifive:
@SiFiveGregS can you take a look at this?

GregSavin (approved):
I don't see anything obviously concerning.  The gdb_server.c change slightly piques my interest as far as what's going on there (guess it's some replacement for the command_run_line() call; maybe the old way was problematic somehow)

timsifive:
> The gdb_server.c change slightly piques my interest as far as what's going on there

From https://review.openocd.org/c/openocd/+/6886:

> server/gdb: fix return of gdb remote monitor command
> 
> Current implementation for gdb remote monitor command uses the
> command_run_line() to execute the command.
> While command_run_line() has several advantages, it unfortunately
> hides the error codes and outputs the result of the command
> through LOG_USER(), which is not what gdb requires. See 'qRcmd' in
> https://sourceware.org/gdb/onlinedocs/gdb/General-Query-Packets.html
> 
> Replace command_run_line() with Jim_EvalObj() and parse the output
> to provide the proper result to gdb.

PR #697, opened 2022-05-04
Change set_haltgroup() to more general set_group()

timsifive:
Change-Id: Ib91a252ac63604e54b756f70c549ccd47241fd10

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1503,15 +1503,20 @@ static void deinit_target(struct target *target)
 	info->version_specific = NULL;
 }
 
-static int set_haltgroup(struct target *target, bool *supported)
+typedef enum {
+	HALTGROUP,
+	RESUMEGROUP
+} grouptype_t;
+static int set_group(struct target *target, bool *supported, unsigned group, grouptype_t grouptype)
 {
-	uint32_t write = set_field(DM_DMCS2_HGWRITE, DM_DMCS2_GROUP, target->smp);
+	uint32_t write = set_field(DM_DMCS2_HGWRITE, DM_DMCS2_GROUP, group);
+	write = set_field(write, DM_DMCS2_GROUPTYPE, 1);
 	if (dmi_write(target, DM_DMCS2, write) != ERROR_OK)
 		return ERROR_FAIL;
 	uint32_t read;
 	if (dmi_read(target, &read, DM_DMCS2) != ERROR_OK)
 		return ERROR_FAIL;
-	*supported = get_field(read, DM_DMCS2_GROUP) == (unsigned)target->smp;
+	*supported = get_field(read, DM_DMCS2_GROUP) == group;
 	return ERROR_OK;
 }
 
@@ -1754,7 +1759,7 @@ static int examine(struct target *target)
 
 	if (target->smp) {
 		bool haltgroup_supported;
-		if (set_haltgroup(target, &haltgroup_supported) != ERROR_OK)
+		if (set_group(target, &haltgroup_supported, target->smp, HALTGROUP) != ERROR_OK)
 			return ERROR_FAIL;
 		if (haltgroup_supported)
 			LOG_INFO("Core %d made part of halt group %d.", target->coreid,
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1510,7 +1510,7 @@ typedef enum {
 static int set_group(struct target *target, bool *supported, unsigned group, grouptype_t grouptype)
 {
 	uint32_t write = set_field(DM_DMCS2_HGWRITE, DM_DMCS2_GROUP, group);
-	write = set_field(write, DM_DMCS2_GROUPTYPE, 1);
+	write = set_field(write, DM_DMCS2_GROUPTYPE, (grouptype == HALTGROUP) ? 1 : 0);
 	if (dmi_write(target, DM_DMCS2, write) != ERROR_OK)
 		return ERROR_FAIL;
 	uint32_t read;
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1503,15 +1503,20 @@ static void deinit_target(struct target *target)
...
+	RESUMEGROUP
+} grouptype_t;
+static int set_group(struct target *target, bool *supported, unsigned group, grouptype_t grouptype)
 {
-	uint32_t write = set_field(DM_DMCS2_HGWRITE, DM_DMCS2_GROUP, target->smp);
+	uint32_t write = set_field(DM_DMCS2_HGWRITE, DM_DMCS2_GROUP, group);
+	write = set_field(write, DM_DMCS2_GROUPTYPE, 1);
 	if (dmi_write(target, DM_DMCS2, write) != ERROR_OK)
 		return ERROR_FAIL;
 	uint32_t read;
```
JanMatCodasip: Just a cosmetic thing: I'd recommend to use different names for the variables than `read` and `write`. These are  already used for system calls (`man 2 read`, `man 2 write`). 

E.g. `read_val`, `write_val`.

src/target/riscv/riscv-013.c
```diff
@@ -1503,15 +1503,20 @@ static void deinit_target(struct target *target)
 	info->version_specific = NULL;
 }
 
-static int set_haltgroup(struct target *target, bool *supported)
+typedef enum {
+	HALTGROUP,
+	RESUMEGROUP
+} grouptype_t;
+static int set_group(struct target *target, bool *supported, unsigned group, grouptype_t grouptype)
```
JanMatCodasip, line 1510: Cosmetic stuff: Blank line between the typedef and the following function.

src/target/riscv/riscv-013.c
```diff
@@ -1503,15 +1503,20 @@ static void deinit_target(struct target *target)
...
 {
-	uint32_t write = set_field(DM_DMCS2_HGWRITE, DM_DMCS2_GROUP, target->smp);
+	uint32_t write = set_field(DM_DMCS2_HGWRITE, DM_DMCS2_GROUP, group);
+	write = set_field(write, DM_DMCS2_GROUPTYPE, 1);
```
JanMatCodasip: I believe this should be:

```suggestion
	write = set_field(write, DM_DMCS2_GROUPTYPE, (grouptype == HALTGROUP) ? 1 : 0);
```

src/target/riscv/riscv-013.c
```diff
@@ -1503,15 +1503,20 @@ static void deinit_target(struct target *target)
...
 
-static int set_haltgroup(struct target *target, bool *supported)
+typedef enum {
+	HALTGROUP,
+	RESUMEGROUP
+} grouptype_t;
+static int set_group(struct target *target, bool *supported, unsigned group, grouptype_t grouptype)
 {
-	uint32_t write = set_field(DM_DMCS2_HGWRITE, DM_DMCS2_GROUP, target->smp);
+	uint32_t write = set_field(DM_DMCS2_HGWRITE, DM_DMCS2_GROUP, group);
```
JanMatCodasip: It may be more readable to split this one into two lines, to make it clear to the reader that we're setting two fields, not just one. (And the compiler optimizes the assignments to `write` anyway.)

Also a good place to put an assertion for the `group` parameter - must not exceed 5 bits, and if it does, it is a logic bug that will be caught here.

```suggestion
	assert(group <= 31);
	
	uint32_t write = 0;
	uint32_t write = set_field(write, DM_DMCS2_GROUP, group);	
	uint32_t write = set_field(write, DM_DMCS2_HGWRITE, 1);
```

src/target/riscv/riscv-013.c
timsifive: I intentionally didn't do that since this typedef is only relevant for this function. I really wanted to put the definition in the prototype itself, but then it's not accessible to callers.

src/target/riscv/riscv-013.c
timsifive: OK

src/target/riscv/riscv-013.c
timsifive: OK

src/target/riscv/riscv-013.c
JanMatCodasip: Ok, understood. I just made the remark because I don't recall seeing this style in other parts of OpenOCD code. I hope that folks in upstream will be OK with that.

src/target/riscv/riscv-013.c
```diff
@@ -1754,7 +1761,7 @@ static int examine(struct target *target)
 
 	if (target->smp) {
 		bool haltgroup_supported;
-		if (set_haltgroup(target, &haltgroup_supported) != ERROR_OK)
+		if (set_group(target, &haltgroup_supported, target->smp, HALTGROUP) != ERROR_OK)
```
JanMatCodasip, line 1764: Broader point that goes beyond this merge request: 

What is please the meaning of `target->smp`? Is it a bool, denoting if SMP is enabled for that target? Or is it an integer, containing the number of the SMP cluster? 

It seems those two approaches are mixed:
* OpenOCD (outside of the riscv target) code apparently treats target->smp just as a bool.
* On the other hand, the riscv target expects takes it directly as the cluster group --> halt group number in this function.

And in this piece of code, there is apparently a merge mistake (target->smp is always set to 1):
https://github.com/riscv/riscv-openocd/blob/riscv/src/target/target.c#L6473

My recommendation would be to make target->smp a real bool. For the SMP cluster number, another struct member could be introduced (`target->smp_group_num`). Preferably, this should be submitted also upstream to avoid this split and have the SMP numbers handled consistently.

JanMatCodasip (approved):
The code of set_group() looks fine. Thank you for reflecting my earlier suggestions.

Please take a look at my comment on `target->smp` (which is a broader issue, to be addressed in a separate merge request).

src/target/riscv/riscv-013.c
timsifive: I added this in https://github.com/riscv/riscv-openocd/pull/353. It's not yet upstreamed, so no other targets do anythign with it.

PR #698, opened 2022-05-05
fix: semihosting_fileio does not work

wzgpeter:
the PC did not plus 4 before resume, which cause this
feature failure

Change-Id: I15c1e7426f1925e78f607c566976f9352216506f
Signed-off-by: Wu Zhigang <zhigang.wu@starfivetech.com>


wzgpeter:
update the comments

PR #699, opened 2022-05-05
fix: semihosting_fileio display the unsupported info

wzgpeter:
the abnormal info display below:
semihosting: unsupported call 0
semihosting: unsupported call 0
semihosting: unsupported call 0
semihosting: unsupported call 0
semihosting: unsupported call 0
semihosting: unsupported call 0
semihosting: unsupported call 0
semihosting: unsupported call 0
semihosting: unsupported call 0
semihosting: unsupported call 0
semihosting: unsupported call 0

the PC did not plus 4 before resume, which cause this
abnormal info popup.

Change-Id: I15c1e7426f1925e78f607c566976f9352216506f
Signed-off-by: Wu Zhigang <zhigang.wu@starfivetech.com>

### DIFF
--- a/src/target/riscv/riscv_semihosting.c
+++ b/src/target/riscv/riscv_semihosting.c
@@ -155,16 +155,16 @@ semihosting_result_t riscv_semihosting(struct target *target, int *retval)
 		}
 	}
 
+	/* Resume right after the EBREAK 4 bytes instruction. */
+	*retval = riscv_set_register(target, GDB_REGNO_PC, pc + 4);
+	if (*retval != ERROR_OK)
+		return SEMI_ERROR;
+
 	/*
 	 * Resume target if we are not waiting on a fileio
 	 * operation to complete.
 	 */
 	if (semihosting->is_resumable && !semihosting->hit_fileio) {
-		/* Resume right after the EBREAK 4 bytes instruction. */
-		*retval = riscv_set_register(target, GDB_REGNO_PC, pc + 4);
-		if (*retval != ERROR_OK)
-			return SEMI_ERROR;
-
 		LOG_DEBUG("   -> HANDLED");
 		return SEMI_HANDLED;
 	}


wzgpeter:
for example: when the 'write' operation comes, it does not increase the PC before resume, then 'ebreak' will be executed again and cause the halt immediately. then in the second halt time, it will get the '0' CMD from CPU side, the abnormal info report over here, it will set the 'pc' register and resume again, this time the resume successfully. based on this, whole semihosting_fileio process is not affected, but it is definitely a bug.

timsifive:
When exactly are you seeing this problem? I have a semihosting testcase that works in https://github.com/riscv-software-src/riscv-tests/blob/master/debug/gdbserver.py#L937

The program that runs is https://github.com/riscv-software-src/riscv-tests/blob/master/debug/programs/semihosting.c

wzgpeter:
tested it on the spike environment.

wzgpeter:
I tested it on the spike environment. the step like below:
1.  spike --rbb-port=1234 --isa=rv64gcv_Zba_Zbb_Zbc_Zbs    [elf-file]
2.  openocd -f  spike.cfg
3.  gdb
In the gdb console, i input the commands likes below:
1. target remote :3333
2. monitor arm semihosting enable
3. monitor arm semihosting_fileio enable
4. load  [elf-file]
5. file [elf-file]
6. continue
then in the OpenOCD console you will see the error info pop up.
Surely, the GDB console will display the printf correctly.

timsifive:
What program are you running in this example?

wzgpeter:
[hello.zip](https://github.com/riscv/riscv-openocd/files/8644112/hello.zip)
I attached the elf file here, which is compiled included libsemihost.a library.

JanMatCodasip (commented):
Although I currently don't have the setup to test this, the changes looks correct: The e-break should be always skipped (PC moved by 4), regardless of who is performing the resume operation:
- a) gdb (if semihosting_fileio is enabled)
- b) openocd itself (if semihosting_fileio is disabled)

timsifive (approved):
I'm not convinced I understand all the details about this code, but the change looks right and passes all of riscv-tests/debug.

TommyMurphyTM1234:
> I'm not convinced I understand all the details about this code, but the change looks right and passes all of riscv-tests/debug.

Do you know why your existing test(s?) work without this change?

https://github.com/riscv/riscv-openocd/pull/699#issuecomment-1118885449

Wouldn't it be more prudent to understand this and if/why the change is needed in the general case before accepting it?

wzgpeter:
I described the issue more detailed here, a little complexity. Sorry.

1. 	when use command "monitor arm semihosting_fileio enable", the flag "semihosting->is_fileio" is set.

2. 	when the elf code running on the target CPU, it executes the printf() function.
   	the "ebreak" instruction will be executed, CPU halted is detected by OpenOCD. 
   	the function "semihosting_common()" will be called to process the different semihosting operation type.

3.  take "SEMIHOSTING_SYS_WRITE" type processed for example, because of the "semihosting->is_fileio" is true,
	then the flag "semihosting->hit_fileio" is set with true. this will cause the "PC + 4" operation is skipped.

4.	the resume operation will be executed by openocd immediately. but because of PC is NOT increased in CPU.
	the "ebreak" instruction will be executed by CPU again!

5.  the OpenOCD detected this halt again! the function "semihosting_common()" will be entered again after passing checking the three op-codes. But this time, the operation type is different from before, I verified this point on spike and FPGA platform, the A0 register value retrieved by OpenOCD is "zero" now! It is NOT a reasonable operation type!!!  the "default" case will be executed in "semihosting_common()" function!

6.  in the "default" case of the "semihosting_common()" function, the "semihosting->hit_fileio" is NOT set with true (it is cleared each time in the end of halt process)!!!

7.  so the "PC + 4" operation will be executed in the "riscv_semihosting()" function. after this, the CPU can be resumed correctly.

the CPU must halt two times for one print operation, it is incorrect.
That is why the existing case can work without this change.
but the abnormal info is displayed in the OpenOCD console.
thank you!

wzgpeter:
openocd-error-log
I attached the snapshot for the error log of openocd, please help check it.

timsifive:
Thanks for explaining this more.

The reason my tests work is because they don't set semihosting_fileio. What does that feature do? The documentation doesn't say, and the name is also confusing. My [test program](https://github.com/riscv-software-src/riscv-tests/blob/master/debug/programs/semihosting.c) does write to a file, so it's not required for file I/O.

wzgpeter:
the feature "monitor arm semihosting_fileio enable" is to display the printf() output (just take print function for example) in the GDB console. just like the feature "monitor arm semihosting enable" display the print log in OpenOCD console. that is my experiment result. the more details, i suggest you can refer to "File-I/O Remote Protocol Extension" part in GDB manual.
but one point we have to pay attention to, when enable the fileio feature, we should not trigger the GDB with "-tui" option. if not, the log will not display over there.
Thanks!

JanMatCodasip:
> The reason my tests work is because they don't set semihosting_fileio. What does that feature do? The documentation doesn't say, and the name is also confusing. 

File-I/O is a GDB feature - an implementation of syscalls in GDB. If `semihosting_fileio` is enabled in OpenOCD, then OpenOCD won't execute the syscalls on its own but instead sends them to GDB for execution via the gdb-remote protocol. The File-I/O feature is further described here: https://sourceware.org/gdb/onlinedocs/gdb/File_002dI_002fO-Overview.html#File_002dI_002fO-Overview

timsifive:
Thanks for the extra info and pointers. I've written a test and confirmed that it only passes with this change.

PR #700, opened 2022-05-11
flash: nor: add hpm_xpi flash programming support

jhqian:
- Add qspi nor flash driver for HPM6700 series

Signed-off-by: Ryan QIAN <jianghao.qian@hpmicro.com>


timsifive:
Can you submit this change to upstream (as described in https://openocd.org/doc/doxygen/html/HACKING_source.html) instead of here? The flash code should be the same in both, and that way I won't have the burden later of upstreaming code I did not write. Just let me know once it's accepted upstream and I'll pull it down here quickly.

jhqian:
> Can you submit this change to upstream (as described in https://openocd.org/doc/doxygen/html/HACKING_source.html) instead of here? The flash code should be the same in both, and that way I won't have the burden later of upstreaming code I did not write. Just let me know once it's accepted upstream and I'll pull it down here quickly.

sure thing, thank you.

AlanCui4080:
Support for hpm_xpi is still not contributed to the upstream beacuse of build failure, see https://review.openocd.org/c/openocd/+/6987. And the cause is those code depends on riscv-openocd i guess (i just paid little time so may not accurately), so if you can reopen and merge this PR and resync it to upstream?

@timsifive

en-sc:
@AlanCui4080, I believe the most recent version of the patch in question is in https://review.openocd.org/c/openocd/+/6988.
In this version the build has passed -- I doubt the change depends on anything from RISC-V OpenOCD fork.
Please, consider pinging the author and reviewers in the mainline review.

PR #702, opened 2022-05-24
Add non-functional gd32vf103 driver.

timsifive:
Now users of the old driver will get an error message that points them
in the right direction.

Change-Id: I1654db1a36c7b7cca8a496a272717decd948672e


fanghuaqi:
Maybe rediect gd32vf103 to stm32f1x could be better, and throw out a warning or information about the changes.

timsifive:
Can you play with this a little and find something that works for your existing users, and is just a small change that we can phase out over the next 12 months?

fanghuaqi:
I opened a new PR #704 , could you help to review my changes?

timsifive:
#704 is better.

PR #703, opened 2022-05-24
Use new debug_defines.h

timsifive:
This gives us:
1. Macros that take e.g. XLEN as argument, where necessary.
2. Macros for all the field values, which we don't need to replicate ourselves anymore.


GregSavin (approved):
That's convenient that the RISC-V ISA and/or Debug spec repositories offer a mechanism to generate corresponding C header file content.

timsifive:
> That's convenient that the RISC-V ISA and/or Debug spec repositories offer a mechanism to generate corresponding C header file content.

Yes, give whoever made that happen a bonus! :-)

PR #704, opened 2022-05-26
flash: redirect gd32vf103 driver to stm32f1x

fanghuaqi:
Hi @timsifive , I modify the flash driver like this, is it ok?

Most of the code of stm32f1x are reused, I just add struct of gd32vf103, so `flash gd32vf103` can be still recognized.

### DIFF
--- a/src/flash/nor/stm32f1x.c
+++ b/src/flash/nor/stm32f1x.c
@@ -1752,11 +1752,36 @@ const struct flash_driver stm32f1x_flash = {
 	.free_driver_priv = default_flash_free_driver_priv,
 };
 
+/* flash bank gd32vf103 <base> <size> 0 0 <target#>
+ */
+FLASH_BANK_COMMAND_HANDLER(gd32vf103_flash_bank_command)
+{
+	struct stm32x_flash_bank *stm32x_info;
+
+	LOG_WARNING("DEPRECATED: The gd32vf103 flash target will be removed in June of 2023, please use stm32f1x instead.");
+	/* The reset code are just copy from stm32x_flash_bank_command function */
+	if (CMD_ARGC < 6)
+		return ERROR_COMMAND_SYNTAX_ERROR;
+
+	stm32x_info = malloc(sizeof(struct stm32x_flash_bank));
+
+	bank->driver_priv = stm32x_info;
+	stm32x_info->probed = false;
+	stm32x_info->has_dual_banks = false;
+	stm32x_info->can_load_options = false;
+	stm32x_info->register_base = FLASH_REG_BASE_B0;
+	stm32x_info->user_bank_size = bank->size;
+
+	/* The flash write must be aligned to a halfword boundary */
+	bank->write_start_alignment = bank->write_end_alignment = 2;
+	return ERROR_OK;
+}
+
 static const struct command_registration gd32vf103_command_handlers[] = {
 	{
 		.name = "gd32vf103",
 		.mode = COMMAND_ANY,
-		.help = "gd32vf103 flash command group(reuse stm32f1x)",
+		.help = "gd32vf103 flash command group (identical to flash commands from stm32f1x)",
 		.usage = "",
 		.chain = stm32f1x_exec_command_handlers,
 	},
@@ -1766,7 +1791,7 @@ static const struct command_registration gd32vf103_command_handlers[] = {
 const struct flash_driver gd32vf103_flash = {
 	.name = "gd32vf103",
 	.commands = gd32vf103_command_handlers,
-	.flash_bank_command = stm32x_flash_bank_command,
+	.flash_bank_command = gd32vf103_flash_bank_command,
 	.erase = stm32x_erase,
 	.protect = stm32x_protect,
 	.write = stm32x_write,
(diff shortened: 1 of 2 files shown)


timsifive (changes requested):
This looks like a good framework.

You should make a gd32vf103_flash_bank_command that prints an error, and then calls stm32x_flash_bank_command, that way at least users will be notified.

I propose we then fully remove this temporary solution in June of 2023.

fanghuaqi:
Hi @timsifive , I have add a `gd32vf103_flash_bank_command` function to print warning message to tell user to use stm32f1x, I want to call `stm32x_flash_bank_command` this function in the `gd32vf103_flash_bank_command` implementation, but didn't find a way, so I just copy the code from `stm32x_flash_bank_command`.

bootstrap
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/env sh
```
timsifive: This has nothing to do with this change, and doesn't have anything to do with RISC-V either. Can you submit this upstream instead?

src/flash/nor/stm32f1x.c
```diff
@@ -1751,3 +1751,54 @@ const struct flash_driver stm32f1x_flash = {
...
 	.free_driver_priv = default_flash_free_driver_priv,
 };
+
+/* flash bank gd32vf103 <base> <size> 0 0 <target#>
+ */
+FLASH_BANK_COMMAND_HANDLER(gd32vf103_flash_bank_command)
+{
+	struct stm32x_flash_bank *stm32x_info;
+
+	LOG_WARNING("The gd32vf103 might be removed in future, please use stm32f1x instead.");
```
timsifive:
```suggestion
	LOG_WARNING("The gd32vf103 flash target will be removed in June of 2023, please use stm32f1x instead.");
```

src/flash/nor/stm32f1x.c
```diff
@@ -1751,3 +1751,54 @@ const struct flash_driver stm32f1x_flash = {
...
+	if (CMD_ARGC < 6)
```
timsifive, line 1763: Can you add a comment that the rest of this function was copied from stm32x...?

timsifive (changes requested):

bootstrap
fanghuaqi: Sorry, this is mistaken added. I will remove it.

src/flash/nor/stm32f1x.c
fanghuaqi: Done

src/flash/nor/stm32f1x.c
fanghuaqi: Comment message added in source code.

fanghuaqi:
Thanks @timsifive for such detailed review, I forced pushed the commit, could you help to review it again.

timsifive (dismissed):
Thanks! I'll merge this tomorrow unless somebody else comments.

src/flash/nor/stm32f1x.c
```diff
@@ -1751,3 +1751,55 @@ const struct flash_driver stm32f1x_flash = {
...
+	/* The flash write must be aligned to a halfword boundary */
+	bank->write_start_alignment = bank->write_end_alignment = 2;
+	return ERROR_OK;
+}
+
+static const struct command_registration gd32vf103_command_handlers[] = {
+	{
+		.name = "gd32vf103",
+		.mode = COMMAND_ANY,
+		.help = "gd32vf103 flash command group(reuse stm32f1x)",
```
JanMatCodasip: Suggestion for minor language correction:

```suggestion
		.help = "gd32vf103 flash command group (identical to flash commands from stm32f1x)",
```

src/flash/nor/stm32f1x.c
```diff
@@ -1751,3 +1751,55 @@ const struct flash_driver stm32f1x_flash = {
...
 	.free_driver_priv = default_flash_free_driver_priv,
 };
+
+/* flash bank gd32vf103 <base> <size> 0 0 <target#>
+ */
+FLASH_BANK_COMMAND_HANDLER(gd32vf103_flash_bank_command)
+{
+	struct stm32x_flash_bank *stm32x_info;
+
+	LOG_WARNING("The gd32vf103 flash target will be removed in June of 2023, please use stm32f1x instead.");
```
JanMatCodasip: Here I'd suggest a minor change to improve the visibility of the message.

The word `DEPRECATED` is commonly used in such messages other parts of OpenOCD, too.

```suggestion
	LOG_WARNING("DEPRECATED: The gd32vf103 flash target will be removed in June of 2023, please use stm32f1x instead.");
```

JanMatCodasip (commented):
I have posted two minor suggestions for the log messages.

src/flash/nor/stm32f1x.c
fanghuaqi: Fixed

src/flash/nor/stm32f1x.c
fanghuaqi: Fixed.

fanghuaqi:
Forced pushed changes requested by @JanMatCodasip

JanMatCodasip (approved):
Thanks, looks good.

PR #706, opened 2022-06-01
tcl: add Espressif riscv targets (ESP32-C2 & ESP32-C3)

erhankur:
Currently Espressif has 2 different RiscV chips. 
This PR adds configuration files for ESP32-C2 and ESP32-C3

openocd -s tcl -f board/esp32c3-ftdi.cfg
openocd -s tcl -f board/esp32c2-ftdi.cfg

Change-Id: I48fead33f5fd5890a7724cd5f500f2d14e2a5ffa

### DIFF
--- a/tcl/target/esp32c2.cfg
+++ b/tcl/target/esp32c2.cfg
@@ -1,0 +1,111 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# The ESP32-C2 only supports JTAG.
+transport select jtag
+
+# Source the ESP common configuration file
+source [find target/esp_common.cfg]
+
+if { [info exists CHIPNAME] } {
+	set _CHIPNAME $CHIPNAME
+} else {
+	set _CHIPNAME esp32c2
+}
+
+if { [info exists CPUTAPID] } {
+	set _CPUTAPID $CPUTAPID
+} else {
+	set _CPUTAPID 0x0000cc25
+}
+
+set _TARGETNAME $_CHIPNAME
+set _CPUNAME cpu
+set _TAPNAME $_CHIPNAME.$_CPUNAME
+
+jtag newtap $_CHIPNAME $_CPUNAME -irlen 5 -expected-id $_CPUTAPID
+
+proc esp32c2_wdt_disable { } {
+    # Halt event can occur during config phase (before "init" is done).
+    # Ignore it since mww commands don't work at that time.
+    if { [string compare [command mode] config] == 0 } {
+        return
+    }
+
+    # Timer Group 0 WDT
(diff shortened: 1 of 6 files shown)


tcl/target/esp32c2.cfg
```diff
@@ -0,0 +1,106 @@
...
+$_TARGETNAME configure -event halted {
+    esp32c2_wdt_disable
+}
+$_TARGETNAME configure -event examine-end {
+    # Need this to handle 'apptrace init' syscall correctly because semihosting is not enabled by default
+    arm semihosting enable
+    arm semihosting_resexit enable
+    if { [info exists _SEMIHOST_BASEDIR] } {
+        if { $_SEMIHOST_BASEDIR != "" } {
+            # TODO: cherry-pick from upstream
```
timsifive, line 91: Can you link to something specific, so it's easier to determine later whether this has happened or not?

tcl/target/esp32c3.cfg
```diff
@@ -0,0 +1,108 @@
...
+    mww 0x60020048 0
+    # RTC WDT
+    mww 0x600080a8 0x50D83AA1
+    mww 0x60008090 0
+    # SWD
+    mww 0x600080b0 0x8F1D312A
+    mww 0x600080ac 0x84B00000
+}
+
+proc esp32c3_soc_reset { } {
```
timsifive, line 49: This appears identical to the one above. Is it possible to put this in some shared cfg file that both of them source?

Likewise for wdt_disable(), and maybe some of the other setup as well?

timsifive (changes requested):
Looks good overall.

Does this replace https://review.openocd.org/c/openocd/+/6942?

tcl/target/esp32c2.cfg
erhankur: Done.

erhankur:
> Looks good overall.
> 
> Does this replace https://review.openocd.org/c/openocd/+/6942?

Yes. Starting from zero is the best way to see if our custom changes are still needed.

tcl/target/esp32c3.cfg
erhankur: Can we keep this as is? I need to refactor all config files for all chips (xtensa, riscv) Need some time. Don't want to do double work here.

timsifive: OK. In that case can you add a comment saying that this is identical and will be refactored in the future?

erhankur: Yes sure. Comment added.

timsifive (approved):
Looks good. I'll merge this Monday unless there are more comments.

PR #707, opened 2022-06-14
Remove unused code.

timsifive:
Change-Id: I3961a62c0c51200ff3241265d9ea3e69492bdc4e

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2201,7 +2201,6 @@ int riscv_openocd_poll(struct target *target)
 	enum target_state old_state = target->state;
 
 	if (target->smp) {
-		unsigned halts_discovered = 0;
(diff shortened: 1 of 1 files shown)


GregSavin (approved):
Would there be value in adding a "halts_discovered" field to the riscv target struct?

timsifive:
I cannot think of why I would want to see that as a persistent counter. Do you have a use case in mind?

GregSavin:
I don't have a specific use case in mind.  I've seen unrelated projects (e.g. in networking stack code) where a set of various low-level/internal statistics are maintained (dropped link frames or packets, etc) which aren't normally looked at once things are running correctly, but could be useful in (internal) troubleshooting later.  And for some reason I thought about that when seeing the halts_discovered bit of state.  I guess maintaining info like that evolves out of specific practical need (which doesn't exist in this case).

PR #714, opened 2022-07-26
Fix overflow issue in write_memory_progbuf

en-sc:
If range's upper bound was equal to 2^64 or the range was wrapping around 0
(which is perfectly legal), writes were not performed due to riscv_addr_t
overflow.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3834,10 +3834,10 @@ static int write_memory_progbuf(struct target *target, target_addr_t address,
 	riscv_program_write(&program);
 
 	riscv_addr_t cur_addr = address;
-	riscv_addr_t fin_addr = address + (count * size);
+	riscv_addr_t distance = (riscv_addr_t)count * size;
 	bool setup_needed = true;
-	LOG_DEBUG("writing until final address 0x%016" PRIx64, fin_addr);
-	while (cur_addr < fin_addr) {
+	LOG_DEBUG("writing until final address 0x%016" PRIx64, cur_addr + distance);
+	while (cur_addr - address < distance) {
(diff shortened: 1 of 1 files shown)


timsifive:
This looks right. What testing have you done? Are you able to run the tests in https://github.com/riscv-software-src/riscv-tests/tree/master/debug against spike to check for regressions?

en-sc:
I've tested my changes on spike. Though as it is - spike requires some fixes too (I'll link the PR to risc-isa-sim) soon. Once all the changes are done I can provide a regression test for that configuration.

en-sc:
I've ran tests in riscv-tests/debug with and without proposed patches (both to spike and to riscv-openocd),  running default target (not all-tests). Regardless of configuration I've gotten two exceptions on DisconnectTest and CustomRegisterTest. Is it an issue with my environment (like gcc version), or is it an expected behavior?

timsifive:
DisconnectTest and CustomRegisterTest pass for me, so the problem is somehow related to your setup. It could be a gdb version thing, maybe. In either case, if there are no regressions anywhere and the various memory tests pass, then I think that's enough coverage for this change. Please do rerun that with `all-tests` however, which catches more corner cases.

en-sc:
I've changed gdb to 12.1 and ran `all-tests`. There were no exceptions (make's output is attached, can provide specific logs, if needed)
[make-all-tests.log](https://github.com/riscv/riscv-openocd/files/9217316/make-all-tests.log)

timsifive (approved):
Great! Thank you for running all the tests.

en-sc:
Could you merge the PR or is there something else I need to do first?

timsifive:
Yup. Usually I wait a day or weekend in case anyone else wants to comment.

PR #720, opened 2022-08-30
Code for 'SIZE'  of type 2 trigger

wxjstz:
Define CSR_MCONTROL_SIZEHI.

watchpoint set SIZE. 
Code will always execute at a fixed address, but not data.
For example, setting a watchpoint to a struct
```c
struct point {
    uint32_t x;
    uint32_ty;
} pp;
```
The address cannot match the address of pp when accessing pp.y

### DIFF
--- a/src/target/riscv/debug_defines.h
+++ b/src/target/riscv/debug_defines.h
@@ -695,6 +695,13 @@
  * \FcsrMcontrolTiming of 1 matching as well.
  */
 /*
+ * This field contains the 2 high bits of the access size. The low bits come
+ * from \FcsrMcontrolSizelo.
+ */
+#define CSR_MCONTROL_SIZEHI_OFFSET          0x15
+#define CSR_MCONTROL_SIZEHI_LENGTH          2
+#define CSR_MCONTROL_SIZEHI                 0x600000
+/*
  * This field contains the 2 low bits of the access size. The high bits come
  * from \FcsrMcontrolSizehi. The combined value is interpreted as follows:
  */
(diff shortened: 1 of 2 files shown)


src/target/riscv/debug_defines.h
```diff
@@ -694,6 +694,13 @@
  * implementation-dependent whether that prevents a trigger with
  * \FcsrMcontrolTiming of 1 matching as well.
  */
+/*
```
JanMatCodasip, line 697: This file is auto-generated (see the [comment on top of the file](https://github.com/riscv/riscv-openocd/blob/ac9ec8df854b672a27174d76a4d331528a08e1d1/src/target/riscv/debug_defines.h#L2)). For that reason, please do not make any modifications to this file directly. 

If needed, consider making your changes in the riscv-debug-spec repository.

src/target/riscv/riscv.c
```diff
@@ -544,6 +544,37 @@ static int maybe_add_trigger_t2(struct target *target,
...
+			9, /* 128bits 16bytes */
+		};
+		unsigned size = trigger->length > 16 ? 0 : stab[trigger->length];
+		if (size == 0) {
+			LOG_DEBUG("Trigger doesn't support watchpoint "
+				"with a length of %u bytes", trigger->length);
+			return ERROR_FAIL;
+		}
+		tdata1 = set_field(tdata1, CSR_MCONTROL_SIZEHI, (size >> 2) & 0x3);
+		tdata1 = set_field(tdata1, CSR_MCONTROL_SIZELO, (size >> 0) & 0x3);
```
JanMatCodasip, line 575: Here the right shift is redundant and can be dropped:

```suggestion
		tdata1 = set_field(tdata1, CSR_MCONTROL_SIZELO, size & 0x3);
```

src/target/riscv/riscv.c
```diff
@@ -544,6 +544,37 @@ static int maybe_add_trigger_t2(struct target *target,
 	if (trigger->write)
 		tdata1 |= CSR_MCONTROL_STORE;
 
+	if (trigger->execute == false) {
+		/* set size for watchpoint */
+		const unsigned stab[] = {
+			0, /* 0 bits  0 bytes */
```
JanMatCodasip, line 550: Value `tdata1.size=0` does not mean "0 bits" but "match against access of any size".

src/target/riscv/riscv.c
```diff
@@ -544,6 +544,37 @@ static int maybe_add_trigger_t2(struct target *target,
...
+			8, /* 112bits 14bytes */
+			0, /* 120bits 15bytes */
+			9, /* 128bits 16bytes */
+		};
+		unsigned size = trigger->length > 16 ? 0 : stab[trigger->length];
+		if (size == 0) {
+			LOG_DEBUG("Trigger doesn't support watchpoint "
+				"with a length of %u bytes", trigger->length);
+			return ERROR_FAIL;
+		}
```
JanMatCodasip, line 573: This code does not allow to set watchpoint with `size=0` --> an error would be produced. 

However, `size=0` is a valid value ("any access size") and it should be allowed.

src/target/riscv/riscv.c
```diff
@@ -544,6 +544,37 @@ static int maybe_add_trigger_t2(struct target *target,
...
+			7, /* 96bits  12bytes */
+			0, /* 104bits 13bytes */
+			8, /* 112bits 14bytes */
+			0, /* 120bits 15bytes */
+			9, /* 128bits 16bytes */
+		};
+		unsigned size = trigger->length > 16 ? 0 : stab[trigger->length];
+		if (size == 0) {
+			LOG_DEBUG("Trigger doesn't support watchpoint "
+				"with a length of %u bytes", trigger->length);
```
JanMatCodasip, line 571: I'd recommend the following message: `RISC-V triggers do not support watchpoint length of %u bytes. Supported lengths are 1, 2, 4, 8, 16 bytes or 0 (access of any size).`

Consider making this at least a warning (or maybe even error?).

JanMatCodasip: These `mcontrol.size` values should also be dis-allowed because there are no loads/stores of the corresponding widths: 4, 6, 7, 8

JanMatCodasip (changes requested):
Thank you for the contribution. I am in favor of adding the support for watchpoint lengths, however there are few items in the code that need to be improved. Please see my comments.

src/target/riscv/riscv.c
wxjstz: The size of the structure may not be 1, 2, 4, 8 bytes in length.

src/target/riscv/riscv.c
wxjstz: I will use -1 replace

src/target/riscv/riscv.c
wxjstz: I used 0 as an illegal value because a watchpoint of size 0 should not be present. I will use -1 to identify illegal values, new patch will be submitted later

src/target/riscv/riscv.c
wxjstz: Yes, I will send a new commit

src/target/riscv/debug_defines.h
wxjstz: Ok! I shouldn't need this modification. I will cancel this patch.

wxjstz:
please review #721

PR #721, opened 2022-08-31
Set SIZE for watchpoint of type2 trigger

wxjstz:
Instruction fetch take place at fixed addresses, so ANY can be used to
set SIZE. But for a watchpoint, the exact SIZE needs to be set. For
example, to set a watchpoint for a structure, the address of accessing
the structure object in the code is not always the starting address of
the structure.

Change-Id: I8b514ca1ba097a5fe481d132a342f07641bee598
Signed-off-by: Xiang W <wxjstz@126.com>

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -544,6 +544,38 @@ static int maybe_add_trigger_t2(struct target *target,
 	if (trigger->write)
 		tdata1 |= CSR_MCONTROL_STORE;
 
+	if (trigger->execute == false) {
+		/* set size for watchpoint */
+		const int stab[] = {
+			-1, /* 0 bits  0 bytes */
+			1,  /* 8 bits  1 bytes */
+			2,  /* 16bits  2 bytes */
+			-1, /* 24bits  3 bytes */
+			3,  /* 32bits  4 bytes */
+			-1, /* 40bits  5 bytes */
+			4,  /* 48bits  6 bytes */
+			-1, /* 56bits  7 bytes */
+			5,  /* 64bits  8 bytes */
+			-1, /* 72bits  9 bytes */
+			6,  /* 80bits  10bytes */
+			-1, /* 88bits  11bytes */
+			7,  /* 96bits  12bytes */
+			-1, /* 104bits 13bytes */
+			8,  /* 112bits 14bytes */
+			-1, /* 120bits 15bytes */
+			9,  /* 128bits 16bytes */
+		};
+		int size = trigger->length > 16 ? -1 : stab[trigger->length];
+		if (size == -1) {
+			LOG_DEBUG("RISC-V trigger do not support watchpoint length %u bytes. "
+				"Supported lengths are 1,2,4,6,8,10,12,14,16 bytes or "
+				"0(access of any size).", trigger->length);
+			return ERROR_FAIL;
+		}
+		tdata1 = set_field(tdata1, CSR_MCONTROL_SIZEHI, (size >> 2) & 0x3);
+		tdata1 = set_field(tdata1, CSR_MCONTROL_SIZELO, size & 0x3);
+	}
+
 	riscv_set_register(target, GDB_REGNO_TDATA1, tdata1);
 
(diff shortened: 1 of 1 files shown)


JanMatCodasip:
> But for a watchpoint, the exact SIZE needs to be set. For
example, to set a watchpoint for a structure, the address of accessing
the structure object in the code is not always the starting address of
the structure.

I am afraid the purpose of of `mcontrol.size` field has been misunderstood here.

`Mcontrol.size` is says which sizes of load/store access (e.g. 1-byte store, half-word store, word store ...) are going to trigger the watchpoint. It is not used to cover a range of addresses (for example whole C structure).

To set up a watchpoint on a certain range of addresses, `mcontrol.match` field shall be used. It may be even be needed to chain two triggers (`mcontrol.chain`) to accomplish the goal - depending on what the address range looks like. 

Note that not all hardware will support this advanced (and optional) trigger functionality - trigger chaining & advanced mcontrol.match options may not be available.

JanMatCodasip (changes requested):
The main issue here is that `mcontrol.size` cannot be used to go set up watchpoint on a range of addresses (see my previous comment for explanation).

wxjstz:
> > But for a watchpoint, the exact SIZE needs to be set. For
> > example, to set a watchpoint for a structure, the address of accessing
> > the structure object in the code is not always the starting address of
> > the structure.
> 
> I am afraid the purpose of of `mcontrol.size` field has been misunderstood here.
> 
> `Mcontrol.size` is says which sizes of load/store access (e.g. 1-byte store, half-word store, word store ...) are going to trigger the watchpoint. It is not used to cover a range of addresses (for example whole C structure).
> 
> To set up a watchpoint on a certain range of addresses, `mcontrol.match` field shall be used. It may be even be needed to chain two triggers (`mcontrol.chain`) to accomplish the goal - depending on what the address range looks like.
> 
> Note that not all hardware will support this advanced (and optional) trigger functionality - trigger chaining & advanced mcontrol.match options may not be available.

You mean need to set the watchpoint for each object of the base data type of the structure.
```c
struct point {
  uint32_t x;
  uint32_t y;
} pp; 
```
Such a structure requires two triggers?

This is unreasonable, if I set a 32-bit watchpoint to the variable a
((char*) &a)[1] = 11
Such code should be catchable by trigger.
So setting a 64bit watchpint for pp can catch the changes of pp.x and pp.y

timsifive:
For mcontrol (type 2) triggers, size refers to the *access size*. Not an address range. In the example of the struct given above, your options are:
2 chained triggers, one using ge (2) match, and the other using lt (3) match.
A single trigger using napot (1) match. This can work if the struct has a size that is a power of 2 of number of bytes, and the actual address has the same alignment.

wxjstz:
> For mcontrol (type 2) triggers, size refers to the _access size_. Not an address range. In the example of the struct given above, your options are: 2 chained triggers, one using ge (2) match, and the other using lt (3) match. A single trigger using napot (1) match. This can work if the struct has a size that is a power of 2 of number of bytes, and the actual address has the same alignment.

thk. RISC-V trigger can't catch the following data changes.
```c
uint32_t a = 0;
((char *) &a)[1] = 1; /* can't catch*/
```
This is a spec bug?

JanMatCodasip:
This is not a spec bug. RISC-V triggers are able to catch the changes from your example - if you use the approach suggested by Tim and me above.

wxjstz:
> This is not a spec bug. RISC-V triggers are able to catch the changes from your example - if you use the approach suggested by Tim and me above.

Use two chained triggers, one use ge match, and other use lt match?
This consumes too many triggers.

wxjstz:
> > This is not a spec bug. RISC-V triggers are able to catch the changes from your example - if you use the approach suggested by Tim and me above.
> 
> Use two chained triggers, one use ge match, and other use lt match? This consumes too many triggers.

match(1) may be best. thk @JanMatCodasip @timsifive

wxjstz:
please review #725 @JanMatCodasip @timsifive

PR #725, opened 2022-09-12
Use match field for trigger

wxjstz:
The watchpoint cannot capture all data modifications only through the trigger of ANY SIZE and EQUAL, and an error will occur. This patch accommodates watchpoints by adding more types of matches

Change-Id: I5c3c908dbd49ca47755b06f5cdbe451be3a81c8b
Co-authored-by: Jan Matyas \<50193733+JanMatCodasip@users.noreply.github.com\>
Co-authored-by: Tim Newsome \<tim@sifive.com\>


src/target/riscv/riscv.c
```diff
@@ -462,11 +462,92 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
+	if (found) {
+		if (idx)
+			*idx = i - (num - 1);
+		return ERROR_OK;
+	}
+
+	return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+}
+
+static int set_trigget(struct target *target, int idx, riscv_reg_t tdata1, riscv_reg_t tdata2)
```
```suggestion
static int set_trigger(struct target *target, int idx, riscv_reg_t tdata1, riscv_reg_t tdata2)
```

src/target/riscv/riscv.c
```diff
@@ -462,11 +462,92 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
+			break;
+		}
+		riscv_set_register(target, GDB_REGNO_TDATA1, tdata2);
+	} while (0);
+	riscv_set_register(target, GDB_REGNO_TSELECT, tselect);
+
+	return ret;
+}
+
+static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
```
timsifive, line 531: There's no need to modify this function. Trigger type t1 is a legacy trigger (per the debug spec). It is only implemented by the HiFive1 AFAIK. The code is here to support existing functionality, but there's no need to add anything new. (Unless you're working with hardware that supports type 1 and some of the new ones.)

timsifive (changes requested):
Overall this looks headed in the right direction.
What testing have you done for this? Specifically, have you run https://github.com/riscv-software-src/riscv-tests/tree/master/debug against spike?
What happens on hardware that doesn't support NAPOT or ranges? It should fall back on other options. I'm a little concerned that the existing straightforward use of triggers doesn't work with these changes. Is that what the trigger->length == 1 check is for (line 636)?

src/target/riscv/riscv.c
```diff
@@ -462,11 +462,109 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
+		(size & (size - 1)) == 0 &&
+		(addr & (size - 1)) == 0) {
+		if (tdata2)
+			*tdata2 = addr | ((size - 1) >> 1);
+		return true;
+	}
+	return false;
+}
+
+static int find_trigger(struct target *target, int type, int num, int *idx)
```
timsifive: I assume this function needs to find contiguous triggers (e.g. 1 and 2, not 1 and 3) but the logic doesn't check for that.

src/target/riscv/riscv.c
```diff
@@ -462,11 +462,109 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
+	riscv_reg_t tselect;
+	if (riscv_get_register(target, &tselect, GDB_REGNO_TSELECT) != ERROR_OK)
+		return ERROR_FAIL;
+
+	unsigned int i;
+	int found = 0, cnt = 0;
+	for (i = 0; i < r->trigger_count; i++) {
+		if (r->trigger_unique_id[i] == -1) {
+			riscv_reg_t tdata1 = 0;
+			riscv_set_register(target, GDB_REGNO_TSELECT, i);
```
timsifive: Please check the return value of every call. (It's not necessary to pass the value received. Simply doing cleanup and returning ERROR_FAIL if the call fails is good enough.)

src/target/riscv/riscv.c
```diff
@@ -462,11 +462,109 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
+	if (found) {
+		if (idx)
+			*idx = i - (num - 1);
+		return ERROR_OK;
+	}
+
+	return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+}
+
+static int get_trigger(struct target *target, int idx, riscv_reg_t *tdata1, riscv_reg_t *tdata2)
```
timsifive: This function, set_trigger, and find_trigger, all save/restore tselect. That's a lot of wasted time. Instead, can you comment that these functions clobber tselect, and rely on add_trigger()/remove_trigger() to do the save/restore?

src/target/riscv/riscv.c
```diff
@@ -477,203 +575,205 @@ static int maybe_add_trigger_t1(struct target *target,
...
+	tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_NAPOT);
+	tdata1 = set_field(tdata1, CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_DISABLED);
+	ret = set_trigger(target, idx, tdata1, tdata2);
+	if (ret != ERROR_OK)
+		goto F1;
+	r->trigger_unique_id[idx] = trigger->unique_id;
+	return ERROR_OK;
 
-	riscv_set_register(target, GDB_REGNO_TDATA2, trigger->address);
+F1:
```
timsifive: Please give F1 and F2 here and in the other functions meaningful names.
E.g. here F1 could be called match_ge_lt and F2 match_equal.

src/target/riscv/riscv.c
```diff
@@ -477,203 +575,205 @@ static int maybe_add_trigger_t1(struct target *target,
 	const uint32_t bpcontrol_bpmatch = 0xf << 7;
 	const uint32_t bpcontrol_bpaction = 0xff << 11;
 
+	ret = find_trigger(target, 1, 1, &idx);
```
timsifive: Everywhere that find_trigger() is called, please use the macro name for the type, e.g. CSR_TDATA1_TYPE_LEGACY in this instance.

src/target/riscv/riscv.c
```diff
@@ -849,27 +949,30 @@ static int remove_trigger(struct target *target, struct trigger *trigger)
...
+	bool done = false;
 	for (i = 0; i < r->trigger_count; i++) {
-		if (r->trigger_unique_id[i] == trigger->unique_id)
-			break;
+		if (r->trigger_unique_id[i] == trigger->unique_id) {
+			riscv_set_register(target, GDB_REGNO_TSELECT, i);
+			riscv_set_register(target, GDB_REGNO_TDATA1, 0);
+			r->trigger_unique_id[i] = -1;
+			LOG_DEBUG("[%d] Stop using resource %d for bp %d",
+				target->coreid, i, trigger->unique_id);
```
```suggestion
			LOG_TARGET_DEBUG(target, "Stop using resource %d for bp %d",
				i, trigger->unique_id);
```

timsifive (changes requested):
Note that you're assuming that trigger types are fixed. The spec actually allows a single trigger to support multiple types, which the debugger can change by writing. OpenOCD has no support for this right now, and I don't think you should add it because that will make this change even larger. But it's good to be aware.

This still looks like headed in the right direction. Can you add some comments here and there to explain what is going on? (Say, at least 2 per maybe_... function?)

src/target/riscv/riscv.c
```diff
@@ -462,9 +462,88 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
+				found = (++cnt == num);
+				if (found) {
+					if (idx)
+						*idx = i - (num - 1);
+					return ERROR_OK;
+				}
+				continue;
+			}
+		}
+		cnt = 0;
```
timsifive: If you set cnt to 0 here, how will it ever become 2 on line 502?

src/target/riscv/riscv.c
```diff
@@ -477,203 +556,227 @@ static int maybe_add_trigger_t1(struct target *target,
...
+	tdata1 = set_field(tdata1, CSR_MCONTROL_M, 1);
+	tdata1 = set_field(tdata1, CSR_MCONTROL_S, !!(r->misa & (1 << ('S' - 'A'))));
+	tdata1 = set_field(tdata1, CSR_MCONTROL_U, !!(r->misa & (1 << ('U' - 'A'))));
+	tdata1 = set_field(tdata1, CSR_MCONTROL_EXECUTE, trigger->execute);
+	tdata1 = set_field(tdata1, CSR_MCONTROL_LOAD, trigger->read);
+	tdata1 = set_field(tdata1, CSR_MCONTROL_STORE, trigger->write);
+	tdata1 = set_field(tdata1, CSR_MCONTROL_SIZELO, CSR_MCONTROL_SIZELO_ANY & 3);
+	tdata1 = set_field(tdata1, CSR_MCONTROL_SIZEHI, (CSR_MCONTROL_SIZELO_ANY >> 2) & 3);
+
+	if (trigger->execute || trigger->length == 1)
```
timsifive, line 597: Why do execute triggers have to use MATCH_EQUAL? I could imagine somebody wanting to trigger when an instruction anywhere on the stack is executed, for instance. Maybe gdb/OpenOCD has no way to set a breakpoint like that?

src/target/riscv/riscv.c
```diff
@@ -462,9 +462,88 @@ static void trigger_from_breakpoint(struct trigger *trigger,
 	trigger->unique_id = breakpoint->unique_id;
 }
 
-static int maybe_add_trigger_t1(struct target *target,
-		struct trigger *trigger, uint64_t tdata1)
+static bool match_napot_checker(struct trigger *trigger, riscv_reg_t *tdata2)
```
```suggestion
/* Return true iff the trigger can be set using MATCH_NAPOT but not using MATCH_EQUAL. */
static bool match_napot_checker(struct trigger *trigger, riscv_reg_t *tdata2)
```

src/target/riscv/riscv.c
```diff
@@ -477,203 +556,227 @@ static int maybe_add_trigger_t1(struct target *target,
...
+	ret = find_trigger(target, CSR_TDATA1_TYPE_MCONTROL, true, &idx);
+	if (ret != ERROR_OK)
+		return ret;
+	tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_GE);
+	tdata1 = set_field(tdata1, CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_ENABLED);
+	tdata2 = trigger->address;
+	ret = set_trigger(target, idx, tdata1, tdata2);
+	if (ret != ERROR_OK) {
+		if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+			goto MATCH_EQUAL; /* Fallback to EQUAL MATCH */
```
timsifive, line 626: Should there be some kind of warning that the requested trigger length isn't supported?
How often does this happen? If it happens for every watchpoint (when ge/lt aren't supported) it would only be annoying to print a message.

src/target/riscv/riscv.c
```diff
@@ -477,203 +556,227 @@ static int maybe_add_trigger_t1(struct target *target,
...
+	ret = set_trigger(target, idx, tdata1, tdata2);
+	if (ret != ERROR_OK)
+		return ret;
+	r->trigger_unique_id[idx] = trigger->unique_id;
 	return ERROR_OK;
 }
 
-static int maybe_add_trigger_t6(struct target *target,
-		struct trigger *trigger, uint64_t tdata1)
+static int maybe_add_trigger_t6(struct target *target, struct trigger *trigger)
```
timsifive, line 661: I assume this function is identical to maybe_add_trigger_t1() except for the constants.

Ideally that code would be refactored to a function that takes all the CSR_... constants as arguments to avoid the code duplication. I realize that you're just following the pattern of the existing code here. This change doesn't have to be done as part of this PR. We can do it later.

timsifive (changes requested):
This is getting pretty close. Thanks for addressing all my feedback.

src/target/riscv/riscv.c
TommyMurphyTM1234: OpenOCD can set breakpoints on an address range:

* https://openocd.org/doc/html/General-Commands.html#index-breakpoint

I don't think that gdb can:

* https://ftp.gnu.org/old-gnu/Manuals/gdb/html_node/gdb_28.html#SEC29

src/target/riscv/riscv.c
wxjstz: A continue on line 509 will skip the setting of 0 on line 511, and two consecutive t==type will make cnt equal to 2

src/target/riscv/riscv.c
wxjstz: A fallback is required on machines that do not support MATCH_LT or MATCH_GE. Fallback may result in some data modifications not triggering watchpoint.

src/target/riscv/riscv.c
timsifive: Got it. Thanks. Can you add a small comment to make it more obvious?

src/target/riscv/riscv.c
```diff
@@ -477,203 +557,227 @@ static int maybe_add_trigger_t1(struct target *target,
...
+	do {
+		ret = maybe_add_trigger_t1(target, trigger);
+		if (ret == ERROR_OK)
+			break;
+		ret = maybe_add_trigger_t2(target, trigger);
+		if (ret == ERROR_OK)
+			break;
+		ret = maybe_add_trigger_t6(target, trigger);
+		if (ret == ERROR_OK)
+			break;
```
timsifive, line 763: This leads to some pretty poor behavior if only type 6 triggers are implemented. The code will end up writing tselect, and reading tdata2 twice for every trigger (once for t1 and once for t2). And that happens every time a trigger is set, which means every time you resume execution from gdb.

The best solution is probably to store the supported types in riscv_enumerate_triggers(), then there's no need to access the target to check the supported type at all.

timsifive (changes requested):
I ran riscv-tests/debug and all tests still pass.

src/target/riscv/riscv.c
wxjstz: The new patch adds type records in riscv_enumerate_triggers, so finding triggers no longer requires writing tselect and reading tdata1

timsifive (commented):
This last change nicely fixes the number of CSR reads/writes. Thanks.

The code looks good. All the existing tests in riscv-tests/debug still pass.

One remaining item: How have you tested this? I'm mostly interested in how you set triggers that cause the new code to be run.

src/target/riscv/riscv.c
```diff
@@ -462,11 +462,76 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
 
-static int maybe_add_trigger_t1(struct target *target,
-		struct trigger *trigger, uint64_t tdata1)
+static bool match_napot_checker(struct trigger *trigger, riscv_reg_t *tdata2)
+{
+	riscv_reg_t addr = trigger->address;
+	riscv_reg_t size = trigger->length;
+	if (size > 1 &&
+		(size & (size - 1)) == 0 &&
+		(addr & (size - 1)) == 0) {
```
JanMatCodasip: If you'd like, you can further improve the readability:

```suggestion
	bool sizePowerOf2 = (size & (size - 1)) == 0;
	bool addrAligned = (addr & (size - 1)) == 0;	
	if (size > 1 && sizePowerOf2 && addrAligned) {
```

src/target/riscv/riscv.c
```diff
@@ -849,27 +938,30 @@ static int remove_trigger(struct target *target, struct trigger *trigger)
...
 		return ERROR_FAIL;
 
+	riscv_reg_t tselect;
+	int result = riscv_get_register(target, &tselect, GDB_REGNO_TSELECT);
+	if (result != ERROR_OK)
+		return result;
+
 	unsigned int i;
+	bool done = false;
 	for (i = 0; i < r->trigger_count; i++) {
```
JanMatCodasip: Just a detail: While at it, could you please move the `i` variable definition inside the loop:

```suggestion
	bool done = false;
	for (unsigned int i = 0; i < r->trigger_count; i++) {
```

src/target/riscv/riscv.h
```diff
@@ -118,6 +118,9 @@ typedef struct {
 	/* The number of triggers per hart. */
 	unsigned int trigger_count;
 
+	/* record the type of each trigger */
+	unsigned int trigger_type[RISCV_MAX_TRIGGERS];
+
```
JanMatCodasip: I am afraid that this storing of the `trigger_type`s together with how the trigger enumeration is currently implemented would break triggers on some RISC-V cores.

The algorithm in `riscv_enumerate_triggers()` only supports one trigger type per trigger, and it is not looking at `tinfo` register to discover other supported trigger types. Before this change, that limitation was a minor issue because the functions `maybe_add_trigger_XX()` always tried to add the trigger - and only then checked if it was successful. Now that the trigger types are discovered at the beginning (but the enumeration is incomplete), this will be a larger issue.

I'd recommend:

1) Either do not store the trigger types (which won't reduce the number of CSR reads/writes, but at least won't break triggers on cores where it worked before)

2) Or, extend the implementation in `riscv_enumerate_triggers()` to discover all the trigger types using the full trigger discovery algorithm from the spec [1]. This requires bit more coding, but this is IHMO clean and preferred solution. In that case, the saved trigger types could later be used safely.

[1] https://raw.githubusercontent.com/riscv/riscv-debug-spec/master/riscv-debug-stable.pdf, section 5.1

src/target/riscv/riscv.c
```diff
@@ -462,11 +462,76 @@ static void trigger_from_breakpoint(struct trigger *trigger,
 	trigger->unique_id = breakpoint->unique_id;
 }
 
-static int maybe_add_trigger_t1(struct target *target,
-		struct trigger *trigger, uint64_t tdata1)
+static bool match_napot_checker(struct trigger *trigger, riscv_reg_t *tdata2)
```
JanMatCodasip: I would recommend a more descriptive name:

```suggestion
static bool can_use_napot_match(struct trigger *trigger, riscv_reg_t *tdata2)
```

src/target/riscv/riscv.c
```diff
@@ -462,11 +462,76 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
+	return false;
+}
+
+static int find_trigger(struct target *target, int type, bool chained, int *idx)
 {
 	RISCV_INFO(r);
 
+	int found;
+	int cnt = 0;
+	int num = chained ? 2 : 1;
```
JanMatCodasip: Please give the variables a more descriptive names, for example:

```suggestion
	unsigned int num_found = 0;
	unsigned int num_required = chained ? 2 : 1;
```

src/target/riscv/riscv.c
```diff
@@ -462,11 +462,76 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
+		if (r->trigger_unique_id[i] == -1) {
+			int t = r->trigger_type[i];
+			if (type == t) {
+				found = (++cnt == num);
+				if (found) {
+					if (idx)
+						*idx = i - (num - 1);
+					return ERROR_OK;
+				}
+				continue;
```
JanMatCodasip: Please define the `found` variable close to the first place of use. Also use `bool` and a more descriptive name.

Avoid the use of pre-increment and post-increment in expressions. Please put it to a separate statement, where possible.

```suggestion
				num_found++;
				bool done = (num_required == num_found);
				if (done) {
					/* Found num_required consecutive free triggers - success, done. */
					if (idx)
						*idx = i - (num - 1);
					return ERROR_OK;
				}
				/* Found a trigger but need more consecutive ones */
				continue;
```

src/target/riscv/riscv.c
```diff
@@ -462,11 +462,76 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
+			}
+		}
+		cnt = 0;
```
```suggestion
		/* Trigger already occupied or incompatible type. 
		 * Reset the counter of found consecutive triggers */
		cnt = 0;
```

JanMatCodasip (changes requested):
Thank you for the changes. Overall the code looks all right.

I have one important point regarding the trigger enumeration & storing of the types.

Then there are few coding/readability suggestions.

wxjstz:
> This last change nicely fixes the number of CSR reads/writes. Thanks.
> 
> The code looks good. All the existing tests in riscv-tests/debug still pass.
> 
> One remaining item: How have you tested this? I'm mostly interested in how you set triggers that cause the new code to be run.

I did not test. I just think that with Match equal, can't set breakpoints for structs. After viewing the datasheet, #721 was sent. This patch was sent after your discussion #721.

src/target/riscv/riscv.h
wxjstz: please review https://github.com/riscv/riscv-openocd/pull/725/commits/0faf40ebbd1dc444fbcebfe9cc88aa469e92eb49

src/target/riscv/riscv.c
```diff
@@ -3916,6 +4010,13 @@ int riscv_enumerate_triggers(struct target *target)
 					riscv_set_register(target, GDB_REGNO_TDATA1, 0);
 				break;
 		}
+
+		uint64_t tinfo;
+		result = riscv_get_register(target, &tinfo, GDB_REGNO_TINFO);
+		if (result == ERROR_OK)
+			r->trigger_tinfo[t] = tinfo;
+		else
+			r->trigger_tinfo[t] = 1 << type;
```
JanMatCodasip: Per the already mentioned spec [1]:
- the trigger enumeration algorithm should first read and use `tinfo` 
- `tdata1.type` should only be used as a secondary source if `tinfo` is not available 

Please, reverse the order of the actions to match the spec. The enumeration algorithm should follow this pseudo-code. Thank you.

```
tselect = 0
for (i in 0 .. RISCV_MAX_TRIGGERS-1) {
    try write tselect;
    read back tselect;
    if (not tselect written OK)
         break;

    read tinfo;
    if (tinfo read successfully) {
        if (tinfo == 1)
             # nonexistent trigger, enumeration done
             break;
        if (tinfo == 0)
             # invalid tinfo value - should never be equal to 0
             break;
        # success, trigger found
        store trigger types found in tinfo;
    }
    else 
    {
        # tinfo does not exist
        read tdata1.type;
        if (read of tdata1 failed)
            # error, enumeration ends
            break;
        if (tdata1.type == 0)
            # no trigger at this tselect index, enumeration ends
            break;
        # success, trigger found            
        store the only trigger type found in tdata1.type;
    }

    disable the found trigger;

    print the supported types in the trigger;

    tselect++;
}
```


[1] https://raw.githubusercontent.com/riscv/riscv-debug-spec/master/riscv-debug-stable.pdf, section 5.1

src/target/riscv/riscv.c
```diff
@@ -3916,6 +4010,13 @@ int riscv_enumerate_triggers(struct target *target)
...
 	}
```
JanMatCodasip, line 4026: Please, print the types supported in each trigger. This will be very useful when debugging issues.

For example:

```suggestion

		LOG_TARGET_DEBUG(target, "Trigger %u: supported types (mask) = 0x%08x", t, r->trigger_tinfo[t]);
	}
```

JanMatCodasip (changes requested):

src/target/riscv/riscv.c
wxjstz: please review https://github.com/riscv/riscv-openocd/pull/725/commits/20ac4c213f5a3e6c257599a94644fb94f55104ad

src/target/riscv/riscv.c
```diff
@@ -3987,36 +3987,42 @@ int riscv_enumerate_triggers(struct target *target)
...
+						riscv_set_register(target, GDB_REGNO_TDATA1, 0);
+					break;
+				case 6:
+					if (tdata1 & CSR_MCONTROL6_DMODE(riscv_xlen(target)))
+						riscv_set_register(target, GDB_REGNO_TDATA1, 0);
+					break;
+			}
 			r->trigger_tinfo[t] = 1 << type;
+		}
+		LOG_TARGET_DEBUG(target, "Trigger %u: supported types (mask) = 0x%08x", t, r->trigger_tinfo[t]);
```
JanMatCodasip, line 4025: Thank you for the changes. Overall, the code looks good.

One last point: 
Shouldn't we unconditionally disable any trigger found (by writing `0` to `tdata1`)? If so, the code would look as shown below. @timsifive, what is your opinion, please?

```suggestion
		} else {
			uint64_t tdata1;
			result = riscv_get_register(target, &tdata1, GDB_REGNO_TDATA1);
			if (result != ERROR_OK)
				return result;

			int type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
			if (type == 0)
				break;

			r->trigger_tinfo[t] = 1 << type;
		}
		
		/* Successfully enumerated one trigger */
		LOG_TARGET_DEBUG(target, "Trigger %u: supported types (mask) = 0x%08x", t, r->trigger_tinfo[t]);
		
		/* Disable the found trigger. Writing 0 to tdata1 guarantees the trigger will get disabled. */
		riscv_set_register(target, GDB_REGNO_TDATA1, 0);
```

timsifive: > Shouldn't we unconditionally disable any trigger found (by writing 0 to tdata1)? If so, the code would look as shown below. @timsifive, what is your opinion, please?

I don't know. OpenOCD's capability of specifying triggers is fairly limited, so I can see how users would want to set triggers directly using other tools. We should be wary of stomping on them. On the other hand, what happens if OpenOCD crashes and we connect again to an important debug session. Now any triggers that OpenOCD had set will persist and the user won't have a good UI to remove them.

I'd say print a message using LOG_TARGET_INFO when OpenOCD finds a trigger that is set (and owned by the debugger). Then we could add (in a different PR) a command to easily disable triggers that OpenOCD didn't set.

JanMatCodasip: @timsifive - Thank you for replying.

> I'd say print a message using LOG_TARGET_INFO when OpenOCD finds a trigger that is set (and owned by the debugger).

This sounds fine. However, detecting an active trigger (for all supported trigger types) sounds like enough work for a separate merge request. For that reason, it seems OK to me to accept this merge request as is. And further improve the logic around triggers in subsequent merge requests, as needed.

JanMatCodasip (approved):
Checked visually, looks OK. (All my suggestions have been addressed.)

timsifive:
This looks good. I just want you to explain in a comment here how you actually set a trigger that exercises the new code path. Just example OpenOCD or gdb commands is all I'm looking for.

timsifive:
@wxjstz, can you provide some example commands on how to set a hardware breakpoint that uses the new code?

I'm not sure why you haven't responded to this yet. I think I'm asking for something very easy, since you must have set a few triggers just for manual testing. Can you explain what I don't understand?

src/target/riscv/riscv.c
wxjstz: > OpenOCD can set breakpoints on an address range:
> 
>     * https://openocd.org/doc/html/General-Commands.html#index-breakpoint
> 
> 
> I don't think that gdb can:
> 
>     * https://ftp.gnu.org/old-gnu/Manuals/gdb/html_node/gdb_28.html#SEC29

gcc should get the length information by analyzing the debugging information
gdb and openocd communicate via GDB Remote Serial Protocol

‘z0,addr,kind’
‘z1,addr,kind’
‘z2,addr,kind’
‘z3,addr,kind’
‘z4,addr,kind’

‘Z0,addr,kind’
‘Z1,addr,kind’
‘Z2,addr,kind’
‘Z3,addr,kind’
‘Z4,addr,kind’

The default kind is the length
https://github.com/bminor/binutils-gdb/blob/master/gdb/remote.c#L10665-L10700

wxjstz:
> @wxjstz, can you provide some example commands on how to set a hardware breakpoint that uses the new code?
> 
> I'm not sure why you haven't responded to this yet. I think I'm asking for something very easy, since you must have set a few triggers just for manual testing. Can you explain what I don't understand?

sorry for the late reply

https://github.com/riscv/riscv-openocd/pull/725#discussion_r1014611853

```c
char d0;

int d1;

struct {
	int a;
	char b;
} __attribute__((packed)) d2;
```
Setting a watchpoint on d0 or setting a breakpoint will trigger MATCH_EQUAL
Set watchpoint on d1 to trigger MATCH_NAPOT
Set watchpoint on d2 to trigger MATCH_GE_LT

src/target/riscv/riscv.c
JanMatCodasip: In my opinion, if this fallback gets hit, it would be good to:

* print a `LOG_WARNING()` message when this occurs for the first time for a given target
* print `LOG_DEBUG()` every time

timsifive (approved):

PR #726, opened 2022-09-13
Add error message when dmstatus read times out.

timsifive:
Otherwise OpenOCD simply doesn't work without giving any indication why.

Change-Id: I21703fc1a0d9bed2f59da95f8a8395fe139484a4

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -701,8 +701,14 @@ int dmstatus_read_timeout(struct target *target, uint32_t *dmstatus,
 int dmstatus_read(struct target *target, uint32_t *dmstatus,
 		bool authenticated)
 {
-	return dmstatus_read_timeout(target, dmstatus, authenticated,
+	int result = dmstatus_read_timeout(target, dmstatus, authenticated,
 			riscv_command_timeout_sec);
+	if (result == ERROR_TIMEOUT_REACHED)
+		LOG_TARGET_ERROR(target, "DMSTATUS read didn't complete in %d seconds. The target is "
+					 "either really slow or broken. You could increase the "
+					 "timeout with `riscv set_command_timeout_sec`.",
+				 riscv_command_timeout_sec);
+	return result;
 }
 
 static void increase_ac_busy_delay(struct target *target)


GregSavin (approved):
If someone had a core clocked at a "normal" speed, but was intentionally using a slow JTAG clock to reach the DM (for whatever reason), would this timeout be likely to happen?

TommyMurphyTM1234:
> If someone had a core clocked at a "normal" speed, but was intentionally using a slow JTAG clock to reach the DM (for whatever reason), would this timeout be likely to happen?

Slightly tangential to that but, I presume that the JTAG TCK is totally asynchronous to the target system clock speed and there are no dependencies between them? I just recall an older Arm core (probably ARM7TDMI-S) where there was a dependency between TCK and the target system clock (something like the former could be a maximum of 25% of the latter). I presume that no such dependencies/constraints apply in the RISC-V debug infrastructure?

timsifive:
> If someone had a core clocked at a "normal" speed, but was intentionally using a slow JTAG clock to reach the DM (for whatever reason), would this timeout be likely to happen?

This timeout will happen if TCK is less than 40 Hz or so, and DMI returns `busy` when reading dmstatus. Someone ran into it when running OpenOCD against a simulated target.

> Slightly tangential to that but, I presume that the JTAG TCK is totally asynchronous to the target system clock speed and there are no dependencies between them? I just recall an older Arm core (probably ARM7TDMI-S) where there was a dependency between TCK and the target system clock (something like the former could be a maximum of 25% of the latter). I presume that no such dependencies/constraints apply in the RISC-V debug infrastructure?

That is an implementation detail. If you implement clock domain crossing simply by sampling the signals on the fast side then you end up with behavior like that. I think it was only ever an issue on ARM because some of those cores would power up using a 32kHz clock or something, so you had to set TCK really slow and poke the bits that sped up the core clock before you could set TCK to something reasonable.

TommyMurphyTM1234:
> That is an implementation detail. If you implement clock domain crossing simply by sampling the signals on the fast side then you end up with behavior like that. I think it was only ever an issue on ARM because some of those cores would power up using a 32kHz clock or something, so you had to set TCK really slow and poke the bits that sped up the core clock before you could set TCK to something reasonable.

Thanks @timsifive.

PR #727, opened 2022-09-16
Make poll backoff time based.

timsifive:
Requested in https://review.openocd.org/c/openocd/+/6964. I'm making the change here so I have a chance to test it properly before pushing it upstream.

A nice effect of this change is that we avoid some unnecessary polling early on when gdb is connecting and we would poll once every time we send gdb a qXfer packet.

Change-Id: I4bdb9f05839e8c1e01ff6dde49d6589595418095

### DIFF
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -3059,6 +3059,7 @@ static int handle_target(void *priv)
 		/* polling may fail silently until the target has been examined */
 		retval = target_poll(target);
 		if (retval == ERROR_OK) {
+			/* Polling succeeded, reset the back-off interval */
 			target->backoff.interval = polling_interval;
 		} else {
 			/* Increase interval between polling up to 5000ms */
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -3046,44 +3046,41 @@ static int handle_target(void *priv)
 			is_jtag_poll_safe() && target;
 			target = target->next) {
 
-		if (!target->tap->enabled)
+		/* This function only gets called every polling_interval, so
+		 * allow some slack in the time comparison. Otherwise, if we
+		 * schedule for now+polling_interval, the next poll won't
+		 * actually happen until a polling_interval later. */
+		if (!target->tap->enabled ||
+		    power_dropout ||
+		    srst_asserted ||
+		    timeval_ms() + polling_interval / 2 < target->backoff.next_attempt)
 			continue;
 
-		if (target->backoff.times > target->backoff.count) {
-			/* do not poll this time as we failed previously */
-			target->backoff.count++;
-			continue;
+		/* polling may fail silently until the target has been examined */
+		retval = target_poll(target);
+		if (retval == ERROR_OK) {
+			target->backoff.interval = polling_interval;
+		} else {
+			/* Increase interval between polling up to 5000ms */
+			target->backoff.interval = MAX(polling_interval,
+					MIN(target->backoff.interval * 2 + 1, 5000));
+			/* Tell GDB to halt the debugger. This allows the user to run
+			* monitor commands to handle the situation. */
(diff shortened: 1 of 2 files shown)


src/target/target.c
```diff
@@ -3046,44 +3046,41 @@ static int handle_target(void *priv)
...
+		/* polling may fail silently until the target has been examined */
+		retval = target_poll(target);
+		if (retval == ERROR_OK) {
+			target->backoff.interval = polling_interval;
+		} else {
+			/* Increase interval between polling up to 5000ms */
+			target->backoff.interval = MAX(polling_interval,
+					MIN(target->backoff.interval * 2 + 1, 5000));
+			/* Tell GDB to halt the debugger. This allows the user to run
+			* monitor commands to handle the situation. */
```
JanMatCodasip: Cosmetic detail - indent. 

```suggestion
			/* Tell GDB to halt the debugger. This allows the user to run
			 * monitor commands to handle the situation. */
```

src/target/target.c
```diff
@@ -3046,44 +3046,41 @@ static int handle_target(void *priv)
...
 			continue;
 
-		if (target->backoff.times > target->backoff.count) {
-			/* do not poll this time as we failed previously */
-			target->backoff.count++;
-			continue;
+		/* polling may fail silently until the target has been examined */
+		retval = target_poll(target);
+		if (retval == ERROR_OK) {
+			target->backoff.interval = polling_interval;
```
JanMatCodasip, line 3061: Just a little favor to whoever will try to understand the code in the future.

```suggestion
			/* Polling succeeded, reset the back-off interval */
			target->backoff.interval = polling_interval;
```

src/target/target.c
```diff
@@ -3046,44 +3046,41 @@ static int handle_target(void *priv)
...
-		if (!target->tap->enabled)
+		/* This function only gets called every polling_interval, so
+		 * allow some slack in the time comparison. Otherwise, if we
+		 * schedule for now+polling_interval, the next poll won't
+		 * actually happen until a polling_interval later. */
+		if (!target->tap->enabled ||
+		    power_dropout ||
+		    srst_asserted ||
+		    timeval_ms() + polling_interval / 2 < target->backoff.next_attempt)
 			continue;
```
JanMatCodasip: An attempt at improving readability:

```suggestion
                const bool time_to_poll = (timeval_ms() + polling_interval / 2) >= target->backoff.next_attempt;
		if (!target->tap->enabled ||
		    power_dropout ||
		    srst_asserted ||
		    !time_to_pool) {
			/* Do not poll this target now */
			continue;
		}
```

JanMatCodasip (dismissed):
Checked visually - the changes look all right.

src/target/target.c
timsifive: Done, but slightly differently.

JanMatCodasip (dismissed):

src/target/target.h
```diff
@@ -117,8 +117,8 @@ struct gdb_service {
 
 /* target back off timer */
 struct backoff_timer {
-	int times;
-	int count;
+	int64_t next_attempt;
+	unsigned int interval;
```
GregSavin, line 121: With next_attempt and the result from timeval_ms() being signed, and interval field being unsigned, I was wondering if there is the potential for signed/unsigned arithmetic sorts of issues with the calculation "bool poll_needed = timeval_ms() + polling_interval / 2 >= target->backoff.next_attempt;"    But I guess timeval_ms() is never going to return a value with a sign bit of 1, so no problem.

GregSavin (approved):

src/target/target.h
timsifive: Agreed that it's a bit odd, but timeval_ms() returns int64_t and that's why next_attempt is also signed.

PR #729, opened 2022-09-19
Support for custom TMSC buffer: cJTAG OScan1 improvement

mrv96:
Allow to use a custom TMSC buffer instead using TMS signal to handle the direction of TMSC.

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -2583,10 +2583,23 @@ and initially asserted reset signals.
 @end deffn
 
 @deffn {Command} {ftdi oscan1_mode} on|off
-Enable or disable OSCAN1 mode. This mode is intended for use with an adapter,
+Enable or disable OScan1 mode. This mode is intended for use with an adapter,
 such as the ARM-JTAG-SWD by Olimex, that sits in between the FTDI chip and the
-target. The adapter uses the normal JTAG signals to control TCKC and TMSC
-(bidirectional) signals used in 2-wire cJTAG.
+target. The cJTAG prococol is composed of two wires: TCKC (clock) and TMSC (data).
+TMSC is a bidirectional signal which is time-multiplexed alternating TDI, TMS and
+TDO. The multiplexing is achieved by a tri-state buffer which puts TMSC in Hi-Z
+when the device is supposed to take the control of the line (TDO phase).
+
+The ARM-JTAG-SWD adapter uses standard TRST and TMS signals to control TMSC
+direction. TRST is used by the adapter as selector for the multiplexers which set
+the JTAG probe in 2-wire mode. Whatever signal is used for this purpose, it must
+be defined with the name JTAG_SEL using @command{ftdi layout_signal}. JTAG_SEL is
+set to 0 during OScan1 initialization.
+
+Some JTAG probes like the Digilent JTAG-HS2, support cJTAG by using a
+separate pin to control when TMS is driven onto TMSC. You can use such
+probes by defining the signal TMSC_EN using
+@command{ftdi layout_signal TMSC_EN -data <mask>}.
 @end deffn
 
 @deffn {Command} {ftdi layout_signal} name [@option{-data}|@option{-ndata} data_mask] [@option{-input}|@option{-ninput} input_mask] [@option{-oe}|@option{-noe} oe_mask] [@option{-alias}|@option{-nalias} name]
--- a/src/jtag/drivers/ftdi.c
+++ b/src/jtag/drivers/ftdi.c
@@ -808,6 +808,8 @@ static void oscan1_mpsse_clock_data(struct mpsse_ctx *ctx, const uint8_t *out, u
 	static const uint8_t zero;
 	static const uint8_t one = 1;
 
+	struct signal *tmsc_en = find_signal_by_name("TMSC_EN");
+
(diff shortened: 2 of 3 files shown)


timsifive (commented):
This doesn't appear to be RISC-V-specific at all. Can you submit this upstream instead (see https://openocd.org/doc-release/html/Developers.html)? If it's urgent, let me know and I will merge it down as soon as it is accepted there.

mrv96:
I just pushed again to remove trailing spaces.

This is not strictly related to RISC-V, but OSCAN1 is present only in this version of OpenOcd thanks to SiFive. Moreover I'm working and trying OpenOcd on a RISC-V microcontroller.

I think that this is the right place to do the PR. Do you have a better place to suggest?

timsifive:
You're right. This code was written here and never pushed up stream, so changes belong here also.

TMSC_EN doesn't seem to be a standard thing. I can see the use for it. Can you find a suitable place in doc/openocd.texi and document this new behavior? There seems to be not enough documentation of oSCAN1 there currently, but let's not fall even further behind.

mrv96:
I've just added a little bit of documentation. I hope that it's clear.

doc/openocd.texi
```diff
@@ -2585,8 +2585,21 @@ and initially asserted reset signals.
...
+The ARM-JTAG-SWD adapter uses standard TRST and TMS signals to control TMSC
+direction. TRST is used by the adapter as selector for the multiplexers which set
+the JTAG probe in 2-wire mode. Whatever signal is used for this purpose, it must
+be defined with the name JTAG_SEL using @command{ftdi layout_signal}. JTAG_SEL is
+set to 0 during OSCAN1 initialization.
+
+Some JTAG probes like the Digilent JTAG-HS2, support cJTAG but don't handle
+the TMSC buffer using TMS. In these cases, it is sufficient to define the
+signal TMSC_EN placing the buffer output enable GPIO into the data mask of
+@command{ftdi layout_signal}.
```
timsifive: This last paragraph is confusing.

Do you mean "using TRST" instead of "using TMS"?

timsifive (commented):
This looks OK to me. Thanks for adding some documentation.

@SiFiveGregS can you take a look at this as well?

doc/openocd.texi
mrv96: Use this picture extracted form [here](https://www.olimex.com/Products/ARM/JTAG/ARM-JTAG-SWD/resources/ARM-JTAG-SWD-schematic.pdf) as reference:

image

You can see that TMSC (alias TMS_SWDIO) is actually driven by a combination of both TRST and TMS.

TRST is set to 0 at the beginning exploiting JTAG_SEL signal while, during the JTAG command executions, TMSC buffer is then driven dinamically by TMS.

If you can explain this stuff in a better way, feel free to update my raught documentation.

mrv96: If you update my documentation commit, it could be a good chance to apply also this correction: https://github.com/riscv/riscv-openocd/commit/0e9d3e76492eb8d622be53de91954034bf987e0a#commitcomment-84473606

GregSavin (dismissed):
I'm not familiar with the Digilent JTAG-HS2 adapter, but it makes sense that different adapter vendor+model combinations have differing signaling details for 2-wire operation, and this code looks reasonable to me.

doc/openocd.texi
```diff
@@ -2585,8 +2585,21 @@ and initially asserted reset signals.
 @deffn {Command} {ftdi oscan1_mode} on|off
 Enable or disable OSCAN1 mode. This mode is intended for use with an adapter,
```
timsifive:
```suggestion
Enable or disable OScan1 mode. This mode is intended for use with an adapter,
```

doc/openocd.texi
```diff
@@ -2585,8 +2585,21 @@ and initially asserted reset signals.
...
+target. The cJTAG prococol is composed of two wires: TCKC (clock) and TMSC (data).
+TMSC is a bidirectional signal which is time-multiplexed alternating TDI, TMS and
+TDO. The multiplexing is achieved by a tri-state buffer which puts TMSC in Hi-Z
+when the device is supposed to take the control of the line (TDO phase).
+
+The ARM-JTAG-SWD adapter uses standard TRST and TMS signals to control TMSC
+direction. TRST is used by the adapter as selector for the multiplexers which set
+the JTAG probe in 2-wire mode. Whatever signal is used for this purpose, it must
+be defined with the name JTAG_SEL using @command{ftdi layout_signal}. JTAG_SEL is
+set to 0 during OSCAN1 initialization.
```
timsifive:
```suggestion
set to 0 during OScan1 initialization.
```

doc/openocd.texi
timsifive:
```suggestion
Some JTAG probes like the Digilent JTAG-HS2, support cJTAG by using a
separate pin to control when TMS is driven onto TMSC. You can use such
probes by defining the signal TMSC_EN using
@command{ftdi layout_signal TMSC_EN -data <mask>}.
```

timsifive (commented):
I think this is more clear.

mrv96:
Done, documentation updated

timsifive (approved):

PR #732, opened 2022-09-26
Propagate error code in register_read/write_direct

en-sc:
In some cases error code returned by riscv_program_insert was ignored


timsifive (approved):
Thanks for this cleanup.

PR #733, opened 2022-09-26
Remove incorrect debug_log in wait_for_idle

en-sc:
According to RISC-V External Debug Support Version 0.13.2 (paragraph 3.12.6), cmderr field contains a valid value only if busy is 0, so it is incorrect to analize it on timeout.


timsifive (approved):
That sounds right.

PR #734, opened 2022-09-26
Fix breackpoint_add for rtos swbp

en-sc:
breakpoint_add should use rtos only if request is done by gdb. If multiple targets are specified as -rtos <rtos_type>, the rtos_update_threads was called only if the last target was specified as rtos, which is inconcistent with other checks of whether or not smp target is an rtos one.

### DIFF
--- a/src/server/gdb_server.c
+++ b/src/server/gdb_server.c
@@ -1785,7 +1785,13 @@ static int gdb_breakpoint_watchpoint_packet(struct connection *connection,
 		case 0:
 		case 1:
 			if (packet[0] == 'Z') {
-				retval = breakpoint_add(target, address, size, bp_type);
+				struct target *bp_target = target;
+				if (target->rtos && bp_type == BKPT_SOFT) {
+					bp_target = rtos_swbp_target(target, address, size, bp_type);
+					if (!bp_target)
+						return ERROR_FAIL;
+				}
+				retval = breakpoint_add(bp_target, address, size, bp_type);
 				if (retval != ERROR_OK) {
 					retval = gdb_error(connection, retval);
 					if (retval != ERROR_OK)
--- a/src/target/breakpoints.c
+++ b/src/target/breakpoints.c
@@ -218,19 +218,10 @@ int breakpoint_add(struct target *target,
 	uint32_t length,
 	enum breakpoint_type type)
 {
-	if (target->smp) {
-		struct target_list *head;
-
-		if (type == BKPT_SOFT) {
-			head = list_first_entry(target->smp_targets, struct target_list, lh);
-			struct target *curr = head->target;
-			if (target->rtos)
-				curr = rtos_swbp_target(target, address, length, type);
-			return breakpoint_add_internal(curr, address, length, type);
-		}
-
-		foreach_smp_target(head, target->smp_targets) {
-			struct target *curr = head->target;
+	if (target->smp && type == BKPT_HARD) {
+		struct target_list *list_node;
+		foreach_smp_target(list_node, target->smp_targets) {
+			struct target *curr = list_node->target;
 			int retval = breakpoint_add_internal(curr, address, length, type);
 			if (retval != ERROR_OK)
 				return retval;


timsifive:
Is this something it's possible to add a test case for in https://github.com/riscv-software-src/riscv-tests/tree/master/debug? This seems like exactly the kind of issue where regressions can easily happen with nobody noticing otherwise.

Why are you submitting this here and abandoned https://review.openocd.org/c/openocd/+/7222?

en-sc:
This PR uses changes local to this fork (like 52ca5d198e3b4a0565e810ff4b5545dfac39cec9). That is why I have abandoned https://review.openocd.org/c/openocd/+/7222

src/target/target.c
```diff
@@ -6433,16 +6433,31 @@ static int jim_target_names(Jim_Interp *interp, int argc, Jim_Obj *const *argv)
 	return JIM_OK;
 }
 
+static struct target_list *
+__attribute__((warn_unused_result))
```
JanMatCodasip: Curious, why is this `__attribute__((warn_unused_result))` used here?

src/target/target.c
```diff
@@ -6433,16 +6433,31 @@ static int jim_target_names(Jim_Interp *interp, int argc, Jim_Obj *const *argv)
...
+	if (!new) {
+		LOG_ERROR("Out of memory");
+		return new;
+	}
+
+	new->target = target;
+	return new;
+}
+
 static int jim_target_smp(Jim_Interp *interp, int argc, Jim_Obj *const *argv)
```
JanMatCodasip: Could the changes in `jim_target_smp()` be submitted separately, since they are RISC-V independent?

These could even go to the upstream directly.

JanMatCodasip (commented):
Checked visually. Overall the code looks all right.

src/server/gdb_server.c
```diff
@@ -1785,7 +1785,13 @@ static int gdb_breakpoint_watchpoint_packet(struct connection *connection,
 		case 0:
 		case 1:
 			if (packet[0] == 'Z') {
-				retval = breakpoint_add(target, address, size, bp_type);
+				struct target *bp_target = target;
+				if (target->rtos && bp_type == BKPT_SOFT) {
+					bp_target = rtos_swbp_target(target, address, size, bp_type);
+					if (!bp_target)
+						return ERROR_FAIL;
```
JanMatCodasip, line 1792: If `rtos_swbp_target` returns NULL (for example because the given RTOS does not have this callback set), I don't believe we should fail. In that case, the original `target` should be used: `breakpoint_add(target, ....);`.

JanMatCodasip: It would be good to re-test the code with a conventional RTOS (something else than `hwthread`, for example FreeRTOS).

There is at least one FreeRTOS-based test in the `riscv-tests` repo, @timsifive will know the details: 
- https://github.com/riscv-software-src/riscv-tests/blob/master/debug/gdbserver.py#L1810

en-sc:
> Is this something it's possible to add a test case for in https://github.com/riscv-software-src/riscv-tests/tree/master/debug? This seems like exactly the kind of issue where regressions can easily happen with nobody noticing otherwise.
> 
> Why are you submitting this here and abandoned https://review.openocd.org/c/openocd/+/7222?

What exactly would you like to test for? If it is the change in the behavior of breakpoint_add - it can't be tested when the gdb connection is established, which is the case for `riscv-tests/debug`. When gdb_connection is established, `rtos->update_threads` is called, so `rtos_swbp_target` would return a correct target even if breakpoint is added via `monitor bp`

src/server/gdb_server.c
timsifive: If an RTOS doesn't define swbp_target(), then rtos_swbp_target() will return the original target, so you only get NULL if the RTOS implemented swbp_target() and decided to return NULL.

timsifive:
> What exactly would you like to test for?

Hmm... ideally that breakpoints set from the CLI go to the correct target. Maybe you can do that using the `monitor` command from gdb. Then I guess you can use TranslateTest as a starting point to set up different address translation on different harts, and verify that the software breakpoint you set modifies the correct physical address.

That's a pretty involved testcase though. So probably not necessary (although it would be appreciated).

timsifive (approved):

en-sc:
> Hmm... ideally that breakpoints set from the CLI go to the correct target. Maybe you can do that using the `monitor` command from gdb. Then I guess you can use TranslateTest as a starting point to set up different address translation on different harts, and verify that the software breakpoint you set modifies the correct physical address.
> 
> That's a pretty involved testcase though. So probably not necessary (although it would be appreciated).

Thank you for the guidance. I'll try to add the test

timsifive:
> Thank you for the guidance. I'll try to add the test

Cool. I look forward to seeing the test.

PR #738, opened 2022-09-29
Remove riscv_info_t.current_hartid

timsifive:
This was used to track which hart a given operation must apply to. But we already have a target associated with each operation, and from there we can find the desired hart id. dm013_info_t already tracks current_hartid (meaning which hart ID is currently selected by the DM).

This makes the code simpler to understand. Also it turns out we don't need to make sure the correct hart ID is currently selected because there are only a few real entry points.

Change-Id: Ibe8d5e156523397f245edd6ec0a5df3239b717bf

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -38,7 +38,7 @@ static void riscv013_clear_abstract_error(struct target *target);
 static int riscv013_get_register(struct target *target,
 		riscv_reg_t *value, int rid);
 static int riscv013_set_register(struct target *target, int regid, uint64_t value);
-static int riscv013_select_current_hart(struct target *target);
+static int dm013_select_hart(struct target *target, int hart_index);
 static int riscv013_halt_prep(struct target *target);
 static int riscv013_halt_go(struct target *target);
 static int riscv013_resume_go(struct target *target);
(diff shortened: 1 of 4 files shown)


src/target/riscv/riscv.c
```diff
@@ -2205,8 +2177,7 @@ int riscv_openocd_poll(struct target *target)
 			struct target *t = list->target;
 			if (!target_was_examined(t))
 				continue;
-			riscv_info_t *r = riscv_info(t);
-			enum riscv_poll_hart out = riscv_poll_hart(t, r->current_hartid);
+			enum riscv_poll_hart out = riscv_poll_hart(t, target->coreid);
```
JanMatCodasip: Curious: Why do we use `target->coreid` in some places and `info->index` on others? For simplicity, could we use one and drop the other? Thanks.

src/target/riscv/riscv-013.c
```diff
@@ -1678,7 +1690,7 @@ static int examine(struct target *target)
 		LOG_DEBUG("Detected %d harts.", dm->hart_count);
 	}
 
-	r->current_hartid = target->coreid;
+	info->index = target->coreid;
```
JanMatCodasip: Do we need this assignment? The value is already assigned above on line 1574.

src/target/riscv/riscv.c
timsifive: Good point. target->coreid is an internal OpenOCD thing. info->index contains the hart index in the DM, which could be different. info->index is generally the one we want.

src/target/riscv/riscv-013.c
timsifive: Fixed. Thanks.

src/target/riscv/riscv.c
timsifive: Although info->index is only in the riscv013, so I can't easily fix that in riscv.c. I'm planning a future cleanup where we stop passing hart id everywhere and instead just pass the target. E.g. in this call there's no need to pass the hart id at all.

JanMatCodasip (approved):
Checked visually - looks all right.

Thank you for the cleanup.

Issue #737, opened 2022-09-29
OScan1 mode doesn't work


mrv96:
After PR #727 OScan1 mode doesn't work anymore, init command fails. In the next days I'll try to figure out the issue and fix it in PR #736.

timsifive:
That is very curious. Thank you for taking a look at it.

mrv96:
The issue is due to this line: https://github.com/riscv/riscv-openocd/blob/550a66e72094bc59b33fd2f020bf78acf731f60c/src/target/target.c#L3053

Increasing `polling_interval` to at least 400 solves the issue.

I didn't understood well the code involved in PR #727, @timsifive what do you suggest to fix this issue?

timsifive:
In my test case, Oscan1 still works after #727. #727 just changes the timing of when some functions are called. Sounds like you're running into a race condition. Can you collected `openocd -d` output from before the change (which works) and after the change (where it doesn't work)? Then we can sift through that and hopefully get an idea of what's going wrong.

TommyMurphyTM1234:
Apologies if this is a dumb question/comment, but is a race condition actually possible in OpenOCD given that, apart from asynchronous signal/Ctrl-C handling, it is single threaded? As far as I know all OpenOCD side JTAG interface/driver and related logic is also synchronous and controlled by the host? Or am I mistaken?

timsifive:
If OpenOCD looks at the wall time to decide what to call next, then the speed of the host, connection, etc. can all affect the behavior of OpenOCD. That might not technically be a race condition, but that's what I mean.

sifiverobert:
I agree with Tim - IMO changing time from 100ms(?) to 400ms is just making problem less probable to happen.
I do not see how polling interval may help with Oscan problem ...

mrv96:
Output when OpenOCD works: 
[working.txt](https://github.com/riscv/riscv-openocd/files/9699493/working.txt)

Output when OpenOCD doesn't work: 
[not_working.txt](https://github.com/riscv/riscv-openocd/files/9699495/not_working.txt)

TommyMurphyTM1234:
> If OpenOCD looks at the wall time to decide what to call next, then the speed of the host, connection, etc. can all affect the behavior of OpenOCD. That might not technically be a race condition, but that's what I mean.

Thanks @timsifive - apologies for the pedantry. I understand your point now. 👍

mrv96:
[...]

PR #739, opened 2022-10-04
target: Try to examine unexamined target in halt()

timsifive:
Some targets fail to examine during `init`, e.g. for timing reasons. The right solution is a better config script, but that ends up being complicated to figure out. This is a bit of a hack, but hopefully silently fixes the problem so users won't have to deal with it at all.

Change-Id: Ib04d0680eb4136f06c383caa6775dd2581a08ce0

### DIFF
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -589,8 +589,14 @@ int target_halt(struct target *target)
 	int retval;
 	/* We can't poll until after examine */
 	if (!target_was_examined(target)) {
-		LOG_ERROR("Target not examined yet");
-		return ERROR_FAIL;
+		/* Try to examine the target right now, in case the target we're
+		 * talking to didn't examine correctly during `init`. */
+		LOG_TARGET_INFO(target, "Try to examine unexamined target in target_halt().");
+		target_examine();
+		if (!target_was_examined(target)) {
+			LOG_ERROR("Target not examined yet");
+			return ERROR_FAIL;
+		}
 	}
 
 	retval = target->type->halt(target);
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -591,10 +591,10 @@ int target_halt(struct target *target)
 	if (!target_was_examined(target)) {
 		/* Try to examine the target right now, in case the target we're
 		 * talking to didn't examine correctly during `init`. */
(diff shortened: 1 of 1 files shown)


sifiverobert:
Hi @timsifive - this fix is not bad (as I said before). If we know that target was not examined it should not hurt to examine it.

src/target/target.c
```diff
@@ -589,8 +589,14 @@ int target_halt(struct target *target)
...
-		LOG_ERROR("Target not examined yet");
-		return ERROR_FAIL;
+		/* Try to examine the target right now, in case the target we're
+		 * talking to didn't examine correctly during `init`. */
+		LOG_TARGET_INFO(target, "Try to examine unexamined target in target_halt().");
+		target_examine();
+		if (!target_was_examined(target)) {
+			LOG_ERROR("Target not examined yet");
+			return ERROR_FAIL;
+		}
```
JanMatCodasip: I'd recommend to not mention concrete functions (`target_halt()`) since the messages will be read mostly by users that do not know these OpenOCD's internals.

For instance:

```suggestion
		LOG_TARGET_INFO(target, "Trying to examine unexamined target before halt attempt.");
		target_examine();
		if (!target_was_examined(target)) {
			LOG_TARGET_ERROR(target, "Re-examination before halt failed. Target not examined yet.");
			return ERROR_FAIL;
		}
```

JanMatCodasip (commented):
> Some targets fail to examine during init, e.g. for timing reasons. 

Which targets do you please have in mind where the initial examine could fail due to the target being slow?

I don't object to this change, even tough I'd prefer to handle this via:
- configurable timeouts for operations which can be adjusted if the target is really slow)
- and/or detect & re-examine the target in TCL scripts (or in the interactive telnet console)

GregSavin (dismissed):
Looks good to me (including the review comment from @JanMatCodasip )

timsifive:
> Which targets do you please have in mind where the initial examine could fail due to the target being slow?

Someone's internal target. This came from a discussion in #737.

timsifive:
The problem we encountered was fixed in #744, so right now there is no known case where this change would make things better.

PR #741, opened 2022-10-05
[riscv] step operation handler should respect handle_breakpoints parameter

aap-sc:
When step operation is requested the OpenOCD frontend (like gdb server
or TCL server) has an option to control how existing breakpoints are
handled upon step.
    
Some OpenOCD frontends (like gdbserver) may choose to disable special
handling of existing breakpoints - thus handle_breakpoints is set to 0,
while others (like TCL server) expect target handler to temporary
disable the matching breakpoint to allow the step operation to complete
successfully.
    
In the current implementation handle_breakpoints parameter was ignored
by target-specific handler. Thus, the following sequence of commands:
    
```
halt
bp <current_pc> 4
step
```
    
Resulted in *step* operation to not change PC because of bp match.
    
This commit addresses this issue.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2289,10 +2289,24 @@ int riscv_openocd_poll(struct target *target)
 int riscv_openocd_step(struct target *target, int current,
 	target_addr_t address, int handle_breakpoints)
 {
-	LOG_DEBUG("stepping rtos hart");
+	LOG_DEBUG("stepping hart");
 
-	if (!current)
-		riscv_set_register(target, GDB_REGNO_PC, address);
+	if (!current) {
+		if (riscv_set_register(target, GDB_REGNO_PC, address) != ERROR_OK)
+			return ERROR_FAIL;
+	}
+
+	struct breakpoint *breakpoint = NULL;
+	/* the front-end may request us not to handle breakpoints */
+	if (handle_breakpoints) {
+		if (current) {
+			if (riscv_get_register(target, &address, GDB_REGNO_PC) != ERROR_OK)
+				return ERROR_FAIL;
+		}
+		breakpoint = breakpoint_find(target, address);
+		if (breakpoint && (riscv_remove_breakpoint(target, breakpoint) != ERROR_OK))
+			return ERROR_FAIL;
+	}
 
 	riscv_reg_t trigger_state[RISCV_MAX_HWBPS] = {0};
 	if (disable_triggers(target, trigger_state) != ERROR_OK)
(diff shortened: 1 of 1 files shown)


timsifive:
How does this problem show up? gdb already clears/sets hardware breakpoints for a step.

aap-sc:
> How does this problem show up? gdb already clears/sets hardware breakpoints for a step.

For example - if you connect to openocd via telnet and try to do some debug without gdb.

timsifive:
That seems like it should be solved at the general target level, not inside the RISC-V-specific code. E.g. in target_step() of target.c. I'd e-mail the OpenOCD mailing list about this problem. It's possible that there already is a general solution to this problem, for instance using callbacks.

aap-sc:
> That seems like it should be solved at the general target level, not inside the RISC-V-specific code. E.g. in target_step() of target.c. I'd e-mail the OpenOCD mailing list about this problem. It's possible that there already is a general solution to this problem, for instance using callbacks.

Oh. I was not aware of that. For whatever reason I thought that it was only riscv-specific problem.  Will take a look at the mailing list then.

aap-sc:
@timsifive , regarding this one:

> That seems like it should be solved at the general target level, not inside the RISC-V-specific code. E.g. in target_step() of target.c. I'd e-mail the OpenOCD mailing list about this problem. It's possible that there already is a general solution to this problem, for instance using callbacks.

I've looked at how existing targets handle similar problem. Turns out:
1. The issue with riscv target is not limited to only hwbp. Any breakpoints is affected.
2. You were right - the original proposal was completely incorrect.
3. There is this **handle_breakpoints** parameter passed to the target-specific handler. It controls if  target-specific handler should disable the existing breakpoints. gdbserver sets it to "0", while TCL server sets is to "1".

I've edited the original description and updated PR. Could you please take a look once again?

src/target/riscv/riscv.c
```diff
@@ -2289,10 +2289,24 @@ int riscv_openocd_poll(struct target *target)
 int riscv_openocd_step(struct target *target, int current,
 	target_addr_t address, int handle_breakpoints)
 {
-	LOG_DEBUG("stepping rtos hart");
+	LOG_DEBUG("stepping hart");
```
```suggestion
	LOG_TARGET_DEBUG(target, "stepping hart");
```

src/target/riscv/riscv.c
```diff
@@ -2332,6 +2346,11 @@ int riscv_openocd_step(struct target *target, int current,
 		LOG_ERROR("unable to enable triggers");
 	}
 
+	if (breakpoint && (riscv_add_breakpoint(target, breakpoint) != ERROR_OK)) {
+		success = false;
+		LOG_ERROR("unable to restore the disabled breakpoint");
```
```suggestion
		LOG_TARGET_ERROR(target, "unable to restore the disabled breakpoint");
```

timsifive (changes requested):
This looks good overall. Did you run the tests in [riscv-tests/debug](https://github.com/riscv-software-src/riscv-tests/tree/master/debug)?

aap-sc:
> Did you run the tests in [riscv-tests/debug](https://github.com/riscv-software-src/riscv-tests/tree/master/debug)?

yep. I've run `make all-tests` from the riscv-tests/debug directory. It looks like everything pass.

timsifive (approved):
Thank you!

PR #743, opened 2022-10-05
Properly track selecting multiple harts at once.

timsifive:
use_hasel is a bit of a hack.

Change-Id: Ia589ebc16bca32038d915df9988361b88e940917

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -281,20 +281,20 @@ dm013_info_t *get_dm(struct target *target)
 	return dm;
 }
 
-static uint32_t set_hartsel(uint32_t initial, int index)
+static uint32_t set_dmcontrol_hartsel(uint32_t initial, int hart_index)
 {
 	/* Use the provided dmcontrol value but clear HARTSEL & HASEL fields */
 	initial &= ~DM_DMCONTROL_HASEL;
 	initial &= ~DM_DMCONTROL_HARTSELLO;
 	initial &= ~DM_DMCONTROL_HARTSELHI;
 
-	if (index > 0) {
-		uint32_t index_lo = index & ((1 << DM_DMCONTROL_HARTSELLO_LENGTH) - 1);
+	if (hart_index > 0) {
+		uint32_t index_lo = hart_index & ((1 << DM_DMCONTROL_HARTSELLO_LENGTH) - 1);
 		initial |= index_lo << DM_DMCONTROL_HARTSELLO_OFFSET;
-		uint32_t index_hi = index >> DM_DMCONTROL_HARTSELLO_LENGTH;
+		uint32_t index_hi = hart_index >> DM_DMCONTROL_HARTSELLO_LENGTH;
 		assert(index_hi < 1 << DM_DMCONTROL_HARTSELHI_LENGTH);
 		initial |= index_hi << DM_DMCONTROL_HARTSELHI_OFFSET;
-	} else if (index < 0) {
+	} else if (hart_index < 0) {
 		initial |= DM_DMCONTROL_HASEL;
 	}
 
@@ -1692,7 +1692,7 @@ static int examine(struct target *target)
 
 			if (get_field(s, DM_DMSTATUS_ANYHAVERESET))
 				dmi_write(target, DM_DMCONTROL,
-						set_hartsel(DM_DMCONTROL_DMACTIVE | DM_DMCONTROL_ACKHAVERESET, i));
+						set_dmcontrol_hartsel(DM_DMCONTROL_DMACTIVE | DM_DMCONTROL_ACKHAVERESET, i));
 		}
 
 		LOG_DEBUG("Detected %d harts.", dm->hart_count);
@@ -2366,7 +2366,7 @@ static int assert_reset(struct target *target)
 		/* Set haltreq for each hart. */
 		uint32_t control = control_base;
 
-		control = set_hartsel(control_base, info->index);
+		control = set_dmcontrol_hartsel(control_base, info->index);
 		control = set_field(control, DM_DMCONTROL_HALTREQ,
 				target->reset_halt ? 1 : 0);
 		dmi_write(target, DM_DMCONTROL, control);
@@ -2377,7 +2377,7 @@ static int assert_reset(struct target *target)
 
 	} else {
 		/* Reset just this hart. */
-		uint32_t control = set_hartsel(control_base, info->index);
+		uint32_t control = set_dmcontrol_hartsel(control_base, info->index);
 		control = set_field(control, DM_DMCONTROL_HALTREQ,
 				target->reset_halt ? 1 : 0);
 		control = set_field(control, DM_DMCONTROL_NDMRESET, 1);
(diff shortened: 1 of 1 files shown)


GregSavin (dismissed):
Is this hasel method of selection relevant when running a "rtos-hwthread" style configuration?

timsifive:
> Is this hasel method of selection relevant when running a "rtos-hwthread" style configuration?

That's right. Whenever you see code talking about SMP that is what it is about.

src/target/riscv/riscv-013.c
```diff
@@ -279,16 +281,21 @@ dm013_info_t *get_dm(struct target *target)
 	return dm;
 }
 
-static uint32_t set_hartsel(uint32_t initial, uint32_t index)
+static uint32_t set_hartsel(uint32_t initial, int index)
```
JanMatCodasip: Can we call it the argument `hart_index`? For readability and also to avoid naming conflicts. `index()` is a function in one of the standard header files (`man 3 index`).

I'd also rename it to `set_dmcontrol_hartsel`.

```suggestion
static uint32_t set_dmcontrol_hartsel(uint32_t dmcontrol_val, int hart_index)
```

src/target/riscv/riscv-013.c
```diff
@@ -279,16 +281,21 @@ dm013_info_t *get_dm(struct target *target)
...
 {
+	initial &= ~DM_DMCONTROL_HASEL;
```
JanMatCodasip: A little favor for whoever reads the code:

```suggestion
	/* Use the provided dmcontrol value but clear HARTSEL & HASEL fields */
	initial &= ~DM_DMCONTROL_HASEL;
```

src/target/riscv/riscv-013.c
```diff
@@ -279,16 +281,21 @@ dm013_info_t *get_dm(struct target *target)
...
-	initial |= index_hi << DM_DMCONTROL_HARTSELHI_OFFSET;
+	if (index > 0) {
+		uint32_t index_lo = index & ((1 << DM_DMCONTROL_HARTSELLO_LENGTH) - 1);
+		initial |= index_lo << DM_DMCONTROL_HARTSELLO_OFFSET;
+		uint32_t index_hi = index >> DM_DMCONTROL_HARTSELLO_LENGTH;
+		assert(index_hi < 1 << DM_DMCONTROL_HARTSELHI_LENGTH);
+		initial |= index_hi << DM_DMCONTROL_HARTSELHI_OFFSET;
+	} else if (index < 0) {
+		initial |= DM_DMCONTROL_HASEL;
+	}
```
JanMatCodasip: Please, use an easier-to-read assertion.

I'd also recommend to use `index >= 0` in the first condition (even though in case of `index==0` the operations are redundant). But still, clarity of the code here has higher value than the few extra spent CPU cycles :)

Can we also use `set_field()` here instead of the bitmanip operators?

There is also `BIT_MASK` macro in helpers/bits.h that can be used here.

```suggestion
	const int MAX_HART_ID = (1 << (DM_DMCONTROL_HARTSELLO_LENGTH + DM_DMCONTROL_HARTSELHI_LENGTH)) - 1;
	/* -1 means multiple harts are selected */	
	assert(index == -1 || (index >= 0 && index <= MAX_HART_ID));
	
	/* Fill new values to the fields */
	if (index >= 0) {
		uint32_t index_lo = index & BIT_MASK(DM_DMCONTROL_HARTSELLO_LENGTH);
		uint32_t index_hi = index >> DM_DMCONTROL_HARTSELLO_LENGTH;				
		initial = set_field(initial, DM_DMCONTROL_HARTSELLO, index_lo);
		initial = set_field(initial, DM_DMCONTROL_HARTSELHI, index_hi);
	} else if (index < 0) {
		initial |= DM_DMCONTROL_HASEL;
	}
```

The `MAX_HART_ID` constant could also go to one of the common header files.

src/target/riscv/riscv-013.c
```diff
@@ -279,16 +281,21 @@ dm013_info_t *get_dm(struct target *target)
...
-	assert(index_hi < 1 << DM_DMCONTROL_HARTSELHI_LENGTH);
-	initial |= index_hi << DM_DMCONTROL_HARTSELHI_OFFSET;
+	if (index > 0) {
+		uint32_t index_lo = index & ((1 << DM_DMCONTROL_HARTSELLO_LENGTH) - 1);
+		initial |= index_lo << DM_DMCONTROL_HARTSELLO_OFFSET;
+		uint32_t index_hi = index >> DM_DMCONTROL_HARTSELLO_LENGTH;
+		assert(index_hi < 1 << DM_DMCONTROL_HARTSELHI_LENGTH);
+		initial |= index_hi << DM_DMCONTROL_HARTSELHI_OFFSET;
+	} else if (index < 0) {
+		initial |= DM_DMCONTROL_HASEL;
```
JanMatCodasip: Caution: If `index==-1`, hart #0 will always get selected - because in that case `dmcontrol.hartsel` will contain value `0`. 

Will this work as expected in all cases? What if we have SMP+hwthread configuration on harts 1,3,5, for instance, and hart 0 is not part of the SMP group?

src/target/riscv/riscv-013.c
```diff
@@ -279,16 +281,21 @@ dm013_info_t *get_dm(struct target *target)
...
-	uint32_t index_hi = index >> DM_DMCONTROL_HARTSELLO_LENGTH;
-	assert(index_hi < 1 << DM_DMCONTROL_HARTSELHI_LENGTH);
-	initial |= index_hi << DM_DMCONTROL_HARTSELHI_OFFSET;
+	if (index > 0) {
+		uint32_t index_lo = index & ((1 << DM_DMCONTROL_HARTSELLO_LENGTH) - 1);
+		initial |= index_lo << DM_DMCONTROL_HARTSELLO_OFFSET;
+		uint32_t index_hi = index >> DM_DMCONTROL_HARTSELLO_LENGTH;
+		assert(index_hi < 1 << DM_DMCONTROL_HARTSELHI_LENGTH);
+		initial |= index_hi << DM_DMCONTROL_HARTSELHI_OFFSET;
+	} else if (index < 0) {
```
JanMatCodasip: Suggestion for clarity - to make it clear to the reader what values of `index` are expected here.

```suggestion
	} else if (index == -1) {
```

JanMatCodasip (changes requested):

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
timsifive: Mostly done.

src/target/riscv/riscv-013.c
timsifive: You're right. This will not always work. This bug would have existed before this commit as well. I don't want to fix this right now, but I filed #748 to track it.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
JanMatCodasip: Thank you.

JanMatCodasip (approved):
Checked visually - looks OK. Thank you.

PR #744, opened 2022-10-07
Handle correctly cJTAG reset

mrv96:
This PR aims to handle correctly resets when IEEE 1149.7 (cJTAG) is used.

This should also fix the strange behavior carried out in issue https://github.com/riscv/riscv-openocd/issues/737

### DIFF
--- a/src/jtag/drivers/ftdi.c
+++ b/src/jtag/drivers/ftdi.c
@@ -676,18 +676,19 @@ static void ftdi_execute_stableclocks(struct jtag_command *cmd)
 static void ftdi_execute_command(struct jtag_command *cmd)
 {
 	switch (cmd->type) {
-		case JTAG_RESET:
 #if BUILD_FTDI_CJTAG == 1
-			cjtag_reset_online_activate(); /* put the target back into selected cJTAG mode */
-#endif
+		case JTAG_RESET:
+			if (cmd->cmd.reset->trst)
+				cjtag_reset_online_activate(); /* put the target (back) into selected cJTAG mode */
 			break;
+#endif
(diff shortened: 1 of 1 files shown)


timsifive (changes requested):
This change removes `case CJTAG_RESET` altogether when BUILD_FTDI_CJTAG isn't 1. I bet that leads to a compiler warning or even error. Can you leave the original structure in place, so `case CJTAG_RESET` is defined but doesn't do anything when BUILD_FTDI_CJTAG isn't 1?

mrv96:
@timsifive the removed case is `case JTAG_RESET` not `case CJTAG_RESET`. It doesn't produce neither warnings nor errors, otherwire CI/CD pipeline of this PR would fails in building steps.

I took the decision to remove `case JTAG_RESET` because it has been removed from the upstream. You can check it here: https://github.com/openocd-org/openocd/blob/ae937791d35b0820e0bc9bf0ab134c2bebd113e4/src/jtag/drivers/ftdi.c#L597-L625

Are you sure that I should leave the original case structure?

timsifive:
I would have expected you to get an error because not every value in the enum is accounted for. But testing it, that is not the case. I guess the default clause saves you. So this is OK.

timsifive (approved):
Looks good to me. @SiFiveGregS can you confirm this looks fine?

GregSavin (approved):
This change looks right, in the sense that the old code would have apparently tried to move the JTAG state before any post-reset cJTAG setup was done, which looks like it was wrong.  I guess that particular situation wasn't universally seen; evidently it's been lurking.

PR #745, opened 2022-10-07
Fix incorrect braces caused by #732

dmitryryzhov:
Actually we want to compare result of riscv_program_insert() with ERROR_OK, not its second argument


timsifive (approved):
Thanks for catching that. Looks like this code doesn't get exercised by the tests because spike always supports abstract access to FPRs.

PR #747, opened 2022-10-10
target/riscv: Clean up halt_go for multiple harts.

timsifive:
Also add an early exit for if any harts are unavailable.

Change-Id: I0875d4d213c9faf87b219d8d57e440881366c8f8


JanMatCodasip (approved):
Checked visually - the changes look fine.

PR #750, opened 2022-10-19
target/riscv: Correctly set target->state in deassert_reset

### LINKED ISSUE #749: Error: [riscv.cpu0] Hart is not halted! in riscv-013 examine
Hello,

Seems after changes https://github.com/riscv/riscv-openocd/pull/743
examine function in riscv-013.c causes error message for harts with no halt on reset:

```
static int examine(struct target *target)
...
	if (!halted)
```
The function riscv013_step_or_resume_current_hart prints error when target is not in HALTED state.
```
static int riscv013_step_or_resume_current_hart(struct target *target,
		bool step)
{
	if (target->state != TARGET_HALTED) {
		LOG_TARGET_ERROR(target, "Hart is not halted!");
		return ERROR_FAIL;
	}
...
```

And target->state = TARGET_HALTED is set later in deassert_reset function

I've combined old and new approaches together to overcome this issue
(and not calling riscv_is_halted every time):
```
static int riscv013_step_or_resume_current_hart(struct target *target,
		bool step)
{
	if (target->state != TARGET_HALTED && !riscv_is_halted(target)) {
		LOG_TARGET_ERROR(target, "Hart is not halted!");
		return ERROR_FAIL;
	}
...
```

Please check it out.
Log with error message is in attachment.
[ocd.log](https://github.com/riscv/riscv-openocd/files/9812566/ocd.log)

Regards,
Fedor

timsifive:
This bug didn't lead to problems that I can see, but it would with some upcoming changes. It might fix #749.

Change-Id: I552acbae9977150c4c9e573f8852033bc80fcebb

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2452,11 +2452,13 @@ static int deassert_reset(struct target *target)
 				return ERROR_FAIL;
 			}
 		}
-		if (target->reset_halt)
+		if (target->reset_halt) {
 			target->state = TARGET_HALTED;
-		else
+			target->debug_reason = DBG_REASON_DBGRQ;
+		} else {
 			target->state = TARGET_RUNNING;
-		target->debug_reason = DBG_REASON_DBGRQ;
+			target->debug_reason = DBG_REASON_NOTHALTED;
+		}
 
 		if (get_field(dmstatus, DM_DMSTATUS_ALLHAVERESET)) {
 			/* Ack reset. */
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2452,7 +2452,10 @@ static int deassert_reset(struct target *target)
 				return ERROR_FAIL;
 			}
 		}
-		target->state = TARGET_HALTED;
+		if (target->reset_halt)
+			target->state = TARGET_HALTED;
+		else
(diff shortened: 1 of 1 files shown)


JanMatCodasip (approved):

src/target/riscv/riscv-013.c
```diff
@@ -2452,7 +2452,10 @@ static int deassert_reset(struct target *target)
 				return ERROR_FAIL;
 			}
 		}
-		target->state = TARGET_HALTED;
+		if (target->reset_halt)
+			target->state = TARGET_HALTED;
+		else
+			target->state = TARGET_RUNNING;
 		target->debug_reason = DBG_REASON_DBGRQ;
```
JanMatCodasip: I don't think any code in OpenOCD pays attention to `target->debug_reason` when `target->state == TARGET_RUNNING`. But for clarity & robustness, we better keep those two aligned:

```suggestion
		if (target->reset_halt) {
			target->state = TARGET_HALTED;
			target->debug_reason = DBG_REASON_DBGRQ;
		} else {
			target->state = TARGET_RUNNING;
			target->debug_reason = DBG_REASON_NOTHALTED;
		}
```

JanMatCodasip (changes requested):

GregSavin (approved):

JanMatCodasip (approved):

PR #752, opened 2022-10-21
target: Add TARGET_UNAVAILABLE state.

timsifive:
Change-Id: I8a062d59eea1e5b3c788281a75159592db024683

### DIFF
--- a/src/target/target.h
+++ b/src/target/target.h
@@ -56,6 +56,8 @@ struct gdb_fileio_info;
  * not sure how this is used with all the recent changes)
  * TARGET_DEBUG_RUNNING = 4: the target is running, but it is executing code on
  * behalf of the debugger (e.g. algorithm for flashing)
+ * TARGET_UNAVAILABLE = 5: The target is unavailable for some reason. It might
+ * be powered down, for instance.
  *
  * also see: target_state_name();
  */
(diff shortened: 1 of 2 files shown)


src/target/target.h
```diff
@@ -66,6 +66,9 @@ enum target_state {
 	TARGET_HALTED = 2,
 	TARGET_RESET = 3,
 	TARGET_DEBUG_RUNNING = 4,
+	/* The target is unavailable for some reason. It might be powered down,
+	 * for instance. */
+	TARGET_UNAVAILABLE = 5
```
JanMatCodasip: Please, move the comment to the top of the enum (for consistency with the existing comments).

In what circumstances would you like to use this new target state? What prompted this change? Is it a part of soluton for a concrete issue?

GregSavin (dismissed):

src/target/target.h
timsifive: Moved, and added a better commit message. If that doesn't answer your questions enough let me know and I'll expand on it there.

JanMatCodasip: Thank you.

From RISC-V point of view, this new state is perfectly fine. 

I'd only watch for existing OpenOCD code (outside of `riscv`) that uses `target->state` and that may not count with this newly added `TARGET_UNAVAILABLE` state. However, after quickly going through the existing code, the change looks safe.

JanMatCodasip (approved):

PR #753, opened 2022-10-25
Support trigger with multiple types

YenHaoChen:
This commit supports triggers with multiple types, i.e. type 15 and tries to use mcontrol and mcontrol6, i.e. type 2 and 6, on those triggers with multiple types.

Change-Id: I2f08512b2224ef13fefd8f51a5eda6cefd7dcb4f
Signed-off-by: YenHaoChen <howard25336284@gmail.com>

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -527,6 +527,7 @@ static int maybe_add_trigger_t2(struct target *target,
 	}
 
 	/* address/data match trigger */
+	tdata1 = set_field(tdata1, CSR_MCONTROL_TYPE(riscv_xlen(target)), 2);
 	tdata1 |= CSR_MCONTROL_DMODE(riscv_xlen(target));
 	tdata1 = set_field(tdata1, CSR_MCONTROL_ACTION,
 			CSR_MCONTROL_ACTION_DEBUG_MODE);
@@ -577,6 +578,7 @@ static int maybe_add_trigger_t6(struct target *target,
 	}
 
 	/* address/data match trigger */
+	tdata1 = set_field(tdata1, CSR_MCONTROL6_TYPE(riscv_xlen(target)), 6);
 	tdata1 |= CSR_MCONTROL6_DMODE(riscv_xlen(target));
 	tdata1 = set_field(tdata1, CSR_MCONTROL6_ACTION,
 			CSR_MCONTROL6_ACTION_DEBUG_MODE);
@@ -650,6 +652,12 @@ static int add_trigger(struct target *target, struct trigger *trigger)
 				result = maybe_add_trigger_t2(target, trigger, tdata1);
 				break;
 			case 6:
+				result = maybe_add_trigger_t6(target, trigger, tdata1);
+				break;
+			case 15:
+				result = maybe_add_trigger_t2(target, trigger, tdata1);
+				if (result == ERROR_OK)
+					break;
 				result = maybe_add_trigger_t6(target, trigger, tdata1);
 				break;
 			default:


src/target/riscv/riscv.c
```diff
@@ -527,6 +527,7 @@ static int maybe_add_trigger_t2(struct target *target,
 	}
 
 	/* address/data match trigger */
+	tdata1 = set_field(tdata1, CSR_MCONTROL_TYPE(riscv_xlen(target)), 2);
```
JanMatCodasip, line 530: This looks as a correct fix (we should always try to write the desired value to `tdata1.type` since it is not known here whether that field is read-only or not).

src/target/riscv/riscv.c
```diff
@@ -577,6 +578,7 @@ static int maybe_add_trigger_t6(struct target *target,
 	}
 
 	/* address/data match trigger */
+	tdata1 = set_field(tdata1, CSR_MCONTROL6_TYPE(riscv_xlen(target)), 6);
```
JanMatCodasip, line 581: This looks like a correct fix (same as above).

src/target/riscv/riscv.c
```diff
@@ -652,6 +654,12 @@ static int add_trigger(struct target *target, struct trigger *trigger)
 			case 6:
 				result = maybe_add_trigger_t6(target, trigger, tdata1);
 				break;
+			case 15:
+				result = maybe_add_trigger_t2(target, trigger, tdata1);
+				if (result == ERROR_OK)
+					break;
+				result = maybe_add_trigger_t6(target, trigger, tdata1);
+				break;
```
JanMatCodasip, line 662: 1) For `tdata1.type == 15`, RISC-V Debug spec says: `15: This trigger exists (...) but is not currently available`. I believe it is more correct for OpenOCD to just ignore such triggers (do not try to use them anyhow). @timsifive - What is your opinion, please?

2) If `tdata1.type` is writable, the supported trigger types should be discovered from `tinfo`. There is a parallel merge request that adds this proper trigger enumeration: https://github.com/riscv/riscv-openocd/pull/725/files @YenHaoChen  - Would it address your use case, too?

YenHaoChen: I thought tdata1.type=15 means the trigger is disabled due to reset or previously configured (writing 0) by the debugger. 

image

The reset handler writes 0 to the tdata1, which results in tdatat.type=15. Thus, initially, the tdata.type will be 15, and the debugger should try to write the tdata1 if it wants to use a trigger that supports multiple types. Additionally, writing 0 to tdata1 and making tdata1.type=15 are standard procedures for set-up triggers. Thus, I believe a debugger should not ignore a trigger with tdata1.type=15.

image

JanMatCodasip: > I thought tdata1.type=15 means the trigger is disabled 

You're right - thank you for correcting me.

Still, if the hardware supports multiple trigger types, then it is mandatory to have `tinfo` implemented which provides the information about which types are supported. The parallel merge request https://github.com/riscv/riscv-openocd/pull/725 already implements the use of `tinfo`. During trigger examination, `tinfo` is read & stored. That information is later used when placing triggers.

Would the change from https://github.com/riscv/riscv-openocd/pull/725 address your use case?

timsifive: @JanMatCodasip, the spec draft was updated 6 months ago to say this for type 15:
            This trigger is disabled. In this state, \RcsrTdataTwo and
            \RcsrTdataThree can be written with any value that is supported for
            any of the types this trigger implements. The remaining bits in this
            register are ignored.

Seems you're referring to an old draft. You can always see the latest stable draft at https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-stable.pdf

timsifive (commented):
This code seems like it'll conflict with #725. Let's get that merged and then figure out what this code should look like.

It's really close, I just have one final question over there.

YenHaoChen:
I close this PR due to the conflict with https://github.com/riscv/riscv-openocd/pull/725.

update: all tests passed with the patch https://github.com/riscv/riscv-openocd/pull/725.

src/target/riscv/riscv.c
JanMatCodasip: > Seems you're referring to an old draft.

You're right, I missed this trigger related update in the spec. Thank you for correcting me.

PR #754, opened 2022-10-26
Fix dm->current_hartid corruption on hartsellen discovery

Dolu1990:
target/riscv Fix dm->current_hartid corruption on hartsellen discovery

So basicaly, the hartsellen discovery code is changing the state of the dm without reflecting it into dm->current_hartid, leading the next dm013_select_hart to believe that hartsel is still zero (as initialized in https://github.com/riscv/riscv-openocd/blob/9eb07f258e7ae8c4a38b349481d27b0207c35100/src/target/riscv/riscv-013.c#L262), and so creating a missmatch between the hard and the soft (which appear for me at https://github.com/riscv/riscv-openocd/blob/9eb07f258e7ae8c4a38b349481d27b0207c35100/src/target/riscv/riscv-013.c#L1683)

That bug appeared from https://github.com/riscv/riscv-openocd/commit/a50b280558cc2908f43dba600461aca256bfbf10 i guess ? Unless i missed something in the spec ? and it isn't a bug but a feature ?

Signed-off-by: Charles Papon <charles.papon.90@gmail.com>

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -130,6 +130,7 @@ typedef enum {
 } yes_no_maybe_t;
 
 #define HART_INDEX_MULTIPLE	-1
+#define HART_INDEX_DIRTY	-2
 
 typedef struct {
 	struct list_head list;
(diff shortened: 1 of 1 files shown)


timsifive (commented):
I'm a bit nervous about this change. What happens in the rest of the code when dm->current_hartid is HART_INDEX_DIRTY?

I'd feel better if you have an assert in set_dmcontrol_hartsel() that hart_index passed there (which is dm->current_hartid in a couple of places) is not HART_INDEX_DIRTY.

Another possible solution would be just to write dmcontrol to match dm->current_hartid after hartsel has been discovered.

Dolu1990:
Now things should be more on the safe side. Also, i added a dm013_select_hart after the probing, to get out of that unusual state as fast as possible.

timsifive (approved):

PR #756, opened 2022-10-27
riscv/target: Replace is_halted() with get_hart_state()

timsifive:
Prep work for handling unavailable harts.

Change-Id: I9c00ed4cdad8edeaa5a13fbec7f88f40d8af9028

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -46,7 +46,6 @@ static int riscv013_step_current_hart(struct target *target);
 static int riscv013_on_halt(struct target *target);
 static int riscv013_on_step(struct target *target);
 static int riscv013_resume_prep(struct target *target);
-static bool riscv013_is_halted(struct target *target);
 static enum riscv_halt_reason riscv013_halt_reason(struct target *target);
 static int riscv013_write_debug_buffer(struct target *target, unsigned index,
 		riscv_insn_t d);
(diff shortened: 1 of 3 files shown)


src/target/riscv/riscv.h
```diff
@@ -65,6 +65,13 @@ enum riscv_isrmasking_mode {
 	RISCV_ISRMASK_STEPONLY,
 };
 
+enum riscv_state {
```
JanMatCodasip: Cosmetic thing: This could be called `enum riscv_hart_state`.

```suggestion
enum riscv_hart_state {
```

src/target/riscv/riscv-013.c
```diff
@@ -2279,6 +2280,51 @@ static int sample_memory(struct target *target,
...
+	if (get_field(dmstatus, DM_DMSTATUS_ALLHALTED)) {
+		*state = RISCV_STATE_HALTED;
+		return ERROR_OK;
+	}
+	if (get_field(dmstatus, DM_DMSTATUS_ALLRUNNING)) {
+		*state = RISCV_STATE_RUNNING;
+		return ERROR_OK;
+	}
+	LOG_TARGET_ERROR(target, "Couldn't determine state. dmstatus=0x%x", dmstatus);
+	assert(0);
```
JanMatCodasip: Please, remove this `assert(0);`. Asserts should be used for code invariants (not for runtime errors). 

OpenOCD should not crash even if `dmstatus` is unexpected.

JanMatCodasip (commented):
Checked visually - the changes look all right.

GregSavin (dismissed):

src/target/riscv/riscv-013.c
timsifive: Good catch.

src/target/riscv/riscv.h
timsifive: Will do.

timsifive:
@JanMatCodasip can you rereview or just approve after I addressed your feedback?

JanMatCodasip (approved):

JanMatCodasip:
> can you rereview or just approve after I addressed your feedback?

Thank you - the changes look fine.

PR #758, opened 2022-11-02
target/riscv: Deal with DMI busy in sample_memory_bus_v1()

timsifive:
Change-Id: I810a4c4b7f02cb40ea99b7a500dce23c1bbd9231

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2239,9 +2239,9 @@ static int sample_memory_bus_v1(struct target *target,
 			return result;
 		}
 
-		/* Discard the batch when we encounter a busy state. It's too
-		 * much hassle to try to recover partial data. We'll try again
-		 * with a larger delay. */
+		/* Discard the batch when we encounter a busy state on the DMI level.
+		 * It's too much hassle to try to recover partial data. We'll try again
+		 * with a larger DMI delay. */
 		unsigned int sbcs_read_op = riscv_batch_get_dmi_read_op(batch, sbcs_read_index);
 		if (sbcs_read_op == DTM_DMI_OP_BUSY) {
 			increase_dmi_busy_delay(target);
@@ -2250,6 +2250,8 @@ static int sample_memory_bus_v1(struct target *target,
 
 		uint32_t sbcs_read = riscv_batch_get_dmi_read_data(batch, sbcs_read_index);
 		if (get_field(sbcs_read, DM_SBCS_SBBUSYERROR)) {
+				/* Discard this batch when we encounter "busy error" state on the System Bus level.
+				 * We'll try next time with a larger System Bus read delay. */
 			info->bus_master_read_delay += info->bus_master_read_delay / 10 + 1;
 			dmi_write(target, DM_SBCS, sbcs_read | DM_SBCS_SBBUSYERROR | DM_SBCS_SBERROR);
 			riscv_batch_free(batch);
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2231,16 +2231,25 @@ static int sample_memory_bus_v1(struct target *target,
 			break;
 		}
 
-		size_t sbcs_key = riscv_batch_add_dmi_read(batch, DM_SBCS);
+		size_t sbcs_read_index = riscv_batch_add_dmi_read(batch, DM_SBCS);
 
 		int result = batch_run(target, batch);
-		if (result != ERROR_OK)
+		if (result != ERROR_OK) {
+			riscv_batch_free(batch);
 			return result;
+		}
 
-		uint32_t sbcs_read = riscv_batch_get_dmi_read_data(batch, sbcs_key);
+		/* Discard the batch when we encounter a busy state. It's too
+		 * much hassle to try to recover partial data. We'll try again
+		 * with a larger delay. */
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -2237,10 +2237,18 @@ static int sample_memory_bus_v1(struct target *target,
 		if (result != ERROR_OK)
 			return result;
```
JanMatCodasip, line 2239: (Unrelated to this change)

It seems there is `riscv_batch_free(batch);` missing before the `return`.

src/target/riscv/riscv-013.c
```diff
@@ -2237,10 +2237,18 @@ static int sample_memory_bus_v1(struct target *target,
...
 
+		/* Discard the batch when we encounter a busy state. It's too
+		 * much hassle to try to recover partial data. We'll try again
+		 * with a larger delay. */
+		unsigned int sbcs_read_op = riscv_batch_get_dmi_read_op(batch, sbcs_key);
```
JanMatCodasip: Cosmetic thing: `sbcs_key` could be renamed to `sbcs_read_index` (or similar), if you'd like.

src/target/riscv/riscv-013.c
```diff
@@ -2237,10 +2237,18 @@ static int sample_memory_bus_v1(struct target *target,
...
+			increase_dmi_busy_delay(target);
+			continue;
+		}
+
 		uint32_t sbcs_read = riscv_batch_get_dmi_read_data(batch, sbcs_key);
-		if (get_field(sbcs_read, DM_SBCS_SBBUSYERROR)) {
-			/* Discard this batch (too much hassle to try to recover partial
-			 * data) and try again with a larger delay. */
+		if (get_field(sbcs_read, DM_SBCS_SBBUSYERROR) ||
+				sbcs_read_op == DTM_DMI_OP_BUSY) {
```
JanMatCodasip: The second part of the condition - `sbcs_read_op == DTM_DMI_OP_BUSY` - will never be `true` because there is already an identical check above on line 2244.

Could you please double-check the code. (Maybe you intended some other condition here?)

src/target/riscv/riscv-013.c
timsifive: Fixed. Thanks.

src/target/riscv/riscv-013.c
timsifive: Yes, that's better.

src/target/riscv/riscv-013.c
timsifive: Good catch. That was left over from an earlier attempt where I tried to combine both busy checks. Then I made the new one on 2244 and didn't fix this.

GregSavin (dismissed):

src/target/riscv/riscv-013.c
```diff
@@ -2231,16 +2231,25 @@ static int sample_memory_bus_v1(struct target *target,
...
-		if (result != ERROR_OK)
+		if (result != ERROR_OK) {
+			riscv_batch_free(batch);
 			return result;
+		}
+
+		/* Discard the batch when we encounter a busy state. It's too
+		 * much hassle to try to recover partial data. We'll try again
+		 * with a larger delay. */
+		unsigned int sbcs_read_op = riscv_batch_get_dmi_read_op(batch, sbcs_read_index);
```
```suggestion
		/* Discard the batch when we encounter a busy state on the DMI level. 
		 * It's too much hassle to try to recover partial data. We'll try again
		 * with a larger DMI delay. */
		unsigned int sbcs_read_op = riscv_batch_get_dmi_read_op(batch, sbcs_read_index);
```

src/target/riscv/riscv-013.c
```diff
@@ -2231,16 +2231,25 @@ static int sample_memory_bus_v1(struct target *target,
...
+		if (sbcs_read_op == DTM_DMI_OP_BUSY) {
+			increase_dmi_busy_delay(target);
+			continue;
+		}
 
-		uint32_t sbcs_read = riscv_batch_get_dmi_read_data(batch, sbcs_key);
+		uint32_t sbcs_read = riscv_batch_get_dmi_read_data(batch, sbcs_read_index);
 		if (get_field(sbcs_read, DM_SBCS_SBBUSYERROR)) {
```
JanMatCodasip, line 2252:
```suggestion
		if (get_field(sbcs_read, DM_SBCS_SBBUSYERROR)) {
				/* Discard this batch when we encounter "busy error" state on the System Bus level.
				 * We'll try next time with a larger System Bus read delay. */
```

JanMatCodasip (dismissed):
The code looks all right - thank you.

One last thing - I'd recommend to update the comments to clearly distinguish the "DMI busy delay" vs. "System bus read delay".

src/target/riscv/riscv-013.c
```diff
@@ -2231,16 +2231,27 @@ static int sample_memory_bus_v1(struct target *target,
...
-			/* Discard this batch (too much hassle to try to recover partial
-			 * data) and try again with a larger delay. */
+				/* Discard this batch when we encounter "busy error" state on the System Bus level.
+				 * We'll try next time with a larger System Bus read delay. */
```
JanMatCodasip: Little thing: There is an extra tab.

```suggestion
			/* Discard this batch when we encounter "busy error" state on the System Bus level.
			 * We'll try next time with a larger System Bus read delay. */
```

JanMatCodasip (dismissed):

PR #759, opened 2022-11-03
target: Expose examine as a user command.

timsifive:
Change-Id: I7be9c75727cf78cedd438fec23374f3084c72be4

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -8712,6 +8712,14 @@ The other options will not work on all systems.
 @end itemize
 @end deffn
 
+@deffn {Command} {examine}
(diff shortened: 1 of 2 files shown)


JanMatCodasip (commented):
The examine operation is already exposed as user command `<target_name> arp_examine` - implemented in `jim_target_examine()`. So I believe there is no need to add any further examine command.

This is typically used together with `target create ... -defer-examine` - for targets which are not able to pass the examination early at `init` (for example due to being still powered down).

@timsifive - Would those existing commands be sufficient for your use case?

timsifive:
That works. Thank you.

PR #762, opened 2022-11-08
target/riscv: Use vlenb to check whether vector registers exist

timsifive:
E.g. the Zve* vector extensions have all the same registers as the full V extension, but leaves misa.V clear.

Change-Id: Ib08c3612c52bb3a6b074d9431e3396c8f2f0ff27

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -4766,7 +4766,7 @@ int riscv_init_registers(struct target *target)
 				case CSR_VL:
 				case CSR_VTYPE:
 				case CSR_VLENB:
-					r->exist = riscv_supports_extension(target, 'V');
+					r->exist = (info->vlenb > 0);
 					break;
 			}
 
@@ -4795,7 +4795,7 @@ int riscv_init_registers(struct target *target)
 
 		} else if (number >= GDB_REGNO_V0 && number <= GDB_REGNO_V31) {
 			r->caller_save = false;
-			r->exist = riscv_supports_extension(target, 'V') && info->vlenb;
+			r->exist = (info->vlenb > 0);
 			r->size = info->vlenb * 8;
 			sprintf(reg_name, "v%d", number - GDB_REGNO_V0);
 			r->group = "vector";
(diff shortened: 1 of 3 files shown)


GregSavin (dismissed):
Is misa.V == 0 congruent with the ISA specs for the Zve* extensions?  (In any case, this solution looks good, since the misa.V == 0 seems to be a reality)  [Edit: I now see in the ISA spec where it states that presence of misa.V == 1 implies some specific things that don't apply to Zve* extensions]

src/target/riscv/riscv.c
```diff
@@ -4795,7 +4795,7 @@ int riscv_init_registers(struct target *target)
 
 		} else if (number >= GDB_REGNO_V0 && number <= GDB_REGNO_V31) {
 			r->caller_save = false;
-			r->exist = riscv_supports_extension(target, 'V') && info->vlenb;
+			r->exist = info->vlenb && info->vlenb;
```
cgsfv: Only one of them is needed:
`r->exist = info->vlenb;`

src/target/riscv/riscv.c
```diff
@@ -4766,7 +4766,7 @@ int riscv_init_registers(struct target *target)
 				case CSR_VL:
 				case CSR_VTYPE:
 				case CSR_VLENB:
-					r->exist = riscv_supports_extension(target, 'V');
+					r->exist = info->vlenb;
```
JanMatCodasip: Detail: To make the code easier to read, I'd recommend:

```suggestion
					r->exist = (info->vlenb > 0);
```

(There's one more occurrence below.)

JanMatCodasip (commented):
Checked visually - the change looks all right.

src/target/riscv/riscv.c
timsifive: Oops, too quick with the search and replace.

src/target/riscv/riscv.c
timsifive: Changed. Thanks.

cgsfv (approved):
LGTM

JanMatCodasip (approved):

PR #766, opened 2022-11-11
Workaround for fp register access in case fp unit is disabled

en-sc:
On some boards there is a HW bug: if fp unit is disabled (fs in mstatus set to 0), accessing any fp register results in a hang (any abstract command timeouts, untill the board is rebooted).

Signed-off-by: Evgeniy Naydanov <evgeniy.naydanov@syntacore.com>
Change-Id: I0c0d1033889f15dcc326c4078bf9cbb5a8558565

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1304,6 +1304,10 @@ static int register_write_direct(struct target *target, unsigned number,
 {
 	LOG_TARGET_DEBUG(target, "%s <- 0x%" PRIx64, gdb_regno_name(number), value);
 
+	uint64_t mstatus;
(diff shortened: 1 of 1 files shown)


timsifive (commented):
This looks OK. It's a bit wasteful on targets that don't need it, but no more than it is for targets that use the program buffer to access FPRs.

Did you run the tests at https://github.com/riscv-software-src/riscv-tests/tree/master/debug?

en-sc:
Yes, I did ran the tests from `riscv-tests/debug` on `spike32` `spike32-2` `spike32-2-hwthread` `spike64` `spike64-2` `spike64-2-hwthread`  targets. They have passed.

timsifive (approved):

PR #767, opened 2022-11-11
Handle harts becoming unavailable while they're being debugged.

timsifive:
Tested using an FPGA target that supports the `cease` instruction. All regular tests still pass.

This commit history should be pretty clean, so if this change is too large to review all at once, just review the individual changes.


src/server/gdb_server.c
```diff
@@ -1736,7 +1759,7 @@ static int gdb_step_continue_packet(struct connection *connection,
 static int gdb_breakpoint_watchpoint_packet(struct connection *connection,
 		char const *packet, int packet_size)
 {
-	struct target *target = get_target_from_connection(connection);
+	struct target *target = get_available_target_from_connection(connection);
```
GregSavin, line 1762: As far as hardware breakpoints, is there a small chance that a previously active target might become unavailable, and then a hardware breakpoint might get set on a different hart than a user might've expected (if get_available_target_from_connection() returns one of the other peers in the SMP group rather than the previous active target that became newly-unavailable)?  Maybe that'd be an exceptionally rare scenario anyway, if it were to happen at all?

GregSavin (dismissed):

src/server/gdb_server.c
timsifive: That could happen, but it's OK because when asked to set a hardware breakpoint we always set that same breakpoint on every target in the SMP group. So it doesn't matter which of the targets in that SMP group is returned here as far as the user is concerned.

GregSavin: Good info, thanks.  Wasn't aware that HW breakpoints were fanned out to all group members; makes sense

src/target/riscv/riscv.c
```diff
@@ -2172,44 +2179,125 @@ static int riscv_checksum_memory(struct target *target,
...
+		switch (state) {
+			case RISCV_STATE_HALTED:
+				LOG_TARGET_DEBUG(target, "  triggered a halt; previous_target_state=%d",
+					previous_target_state);
+				target->state = TARGET_HALTED;
+				enum riscv_halt_reason halt_reason = riscv_halt_reason(target);
+				if (set_debug_reason(target, halt_reason) != ERROR_OK)
+					return ERROR_FAIL;
+
+				if (halt_reason == RISCV_HALT_BREAKPOINT) {
```
JanMatCodasip: This is not strictly related to this merge request, but could we rename `RISCV_HALT_BREAKPOINT` to `RISCV_HALT_EBREAK` (or `RISCV_HALT_SW_BREAK`, or similar)?

I think it would make this part of code more clear.

src/target/riscv/riscv-013.c
```diff
@@ -4304,9 +4304,26 @@ static int riscv013_halt_go(struct target *target)
 		target_list_t *entry;
```
JanMatCodasip, line 4304: There is possibly a mistake on lines 4283-5 (not touched in this merge request, but nearby):

I think the condition should check for "not ANYRUNNING" instead of "not ALLRUNNING":

```
		/* When no harts are running, there's no point in continuing this loop. */
		if (!get_field(dmstatus, DM_DMSTATUS_ANYRUNNING))
				break;
```

src/target/riscv/riscv-013.c
```diff
@@ -4304,9 +4304,26 @@ static int riscv013_halt_go(struct target *target)
...
+				if (dmi_read(target, &dmstatus2, DM_DMSTATUS) != ERROR_OK)
+					return ERROR_FAIL;
+			}
+			if (get_field(dmstatus2, DM_DMSTATUS_ALLHALTED)) {
+				t->state = TARGET_HALTED;
+				if (t->debug_reason == DBG_REASON_NOTHALTED)
+					t->debug_reason = DBG_REASON_DBGRQ;
+			} else if (get_field(dmstatus2, DM_DMSTATUS_ALLUNAVAIL)) {
+				t->state = TARGET_UNAVAILABLE;
+			}
```
JanMatCodasip: This part of code seems to me harder to understand. If you'd like, it can be re-strucutred as shown below (pseudocode). It also avoids the need to have the second dmstatus variable (`dmstatus2`). 

It may also be a helpful to add debug prints.

```
if (get_field(dmstatus, DM_DMSTATUS_ALLHALTED)) {
    list_for_each_entry(entry, &dm->target_list, list) {
        /* set target->state = TARGET_HALTED (+ debug_reason) */
    }
    
    LOG_TARGET_DEBUG(target, "All SMP targets are halted after halt request");
}
else if (get_field(dmstatus, DM_DMSTATUS_ALLUNAVAILABLE)) {
    list_for_each_entry(entry, &dm->target_list, list) {
        /* set target->state = TARGET_UNAVAILABLE */
    }
    
    LOG_TARGET_DEBUG(target, "All SMP targets are unavailable after halt request");    
}
else {
   /* Only some harts were halted. Need to check
    * status of each one individually. */
    
   /* Loop through the harts, select each one individually and determine its state 
    *  ...
    */
    
   /* Make a debug log entry for each: */ 
   LOG_TARGET_DEBUG(target, "Target is %s after halt request", is_unavail ? "unavailable" : "halted");    
}

```

src/target/riscv/riscv.c
```diff
@@ -2172,44 +2179,125 @@ static int riscv_checksum_memory(struct target *target,
...
+					int retval;
+					switch (riscv_semihosting(target, &retval)) {
```
JanMatCodasip, line 2253: Detail - a comment that may help someone:

```suggestion
					/* Detect if this EBREAK is a semihosting request. If so, handle it. */
					switch (riscv_semihosting(target, &retval)) {
```

src/target/riscv/riscv.c
```diff
@@ -2266,120 +2354,95 @@ int sample_memory(struct target *target)
...
-		} else if (out == RPH_ERROR) {
-			return ERROR_FAIL;
+		/* For targets that were discovered to be halted, call the
+		 * appropriate callback. */
+		foreach_smp_target(entry, targets)
+		{
+			struct target *t = entry->target;
+			riscv_info_t *info = riscv_info(t);
+			if (info->needs_callback)
+				target_call_event_callbacks(t, info->callback_event);
```
JanMatCodasip: For clarity, the `needs_callback` flag could be cleared at the same place where it is processed.

```suggestion
			if (info->needs_callback) {
				target_call_event_callbacks(t, info->callback_event);
				info->needs_callback = false;
			}				
```

src/target/riscv/riscv.c
```diff
@@ -2266,120 +2354,95 @@ int sample_memory(struct target *target)
...
+	struct target_list *entry;
+	foreach_smp_target(entry, targets) {
+		struct target *t = entry->target;
+		riscv_info_t *info = riscv_info(t);
 
-			case RPH_ERROR:
-				return ERROR_FAIL;
-			}
-		}
+		info->needs_callback = false;
```
JanMatCodasip: This assert would work with the change suggested at line 2434.

```suggestion
		/* Any pending callback for the target must have been processed already. */
		assert(!info->needs_callback);
```

src/target/riscv/riscv.c
JanMatCodasip: I wonder if there are better names for `needs_callback` and `callback_event`.

Maybe `halt_callback_pending` & `halt_callback_event` (or similar).

src/target/riscv/riscv-013.c
```diff
@@ -4304,9 +4304,26 @@ static int riscv013_halt_go(struct target *target)
...
 		}
 	}
 	/* The "else" case is handled in halt_go(). */
```
JanMatCodasip, line 4335: It would be good to handle both cases at one place - but this can be done sometime later.

JanMatCodasip (commented):
The changes overall appear fine. I came across only few items worth taking closer look at.

src/target/riscv/riscv.c

src/target/riscv/riscv-013.c
timsifive: Good catch.

src/target/riscv/riscv-013.c
timsifive: I dunno. The resulting code is longer and duplicates some logic (e.g. for setting debug_reason). I'll add some comments to this code to make it easier to read.

src/target/riscv/riscv.c
timsifive: Done.

src/target/riscv/riscv-013.c
JanMatCodasip: I won't insist here but I would still recommend to re-structure it since the original code is not easy to understand by  someone who is not the author. 

Making the code slightly longer and minor duplication (only the loop + setting debug reason) is fine when it improves overall readability.

JanMatCodasip: Here is another suggestion how to re-structure the code while still keeping it compact and without duplication. And it is closer to the original code.

```
    const bool all_halted = get_field(dmstatus, DM_DMSTATUS_ALLHALTED);   
    const bool all_unavail = get_field(dmstatus, DM_DMSTATUS_ALLUNAVAIL);
        /* const is fine here and improves readability - 
           it tells the reader the value won't change in the rest of the function */

    list_for_each_entry(entry, &dm->target_list, list) {

        bool this_hart_halted;
        if (all_halted) {
            this_hart_halted = true;
        }
        else if (all_unavail) {
           this_hart_halted = false;
        }
        else {
            /* Harts are in different states, need to check them individually */
            
            /* ... select the hart, check the state ... */
        }
        
        if (this_hart_halted) {
            /* this hart is halted, set the state and debug reason accordingly */
        }
        else {
            /* this hart is unavailable, set the state and debug reason accordingly */
        }
    }
```

src/target/riscv/riscv.c
timsifive: I like halted_needs_event_callback and halted_callback_event

src/target/riscv/riscv.c
timsifive: I don't like the assert. It's simple to clear the state before it might get set. It's complicated to track every possible error case and make sure the state is cleared if it was set and we didn't get to actually send the callback. On top of that we're unlikely to catch every possible of those error cases in a test, so it seems likely the code is incomplete and a user will encounter an assert that kills their debug session for a fairly innocuous reason.

src/target/riscv/riscv-013.c
timsifive: That feels more cumbersome than what I wrote. Take a look and tell me how you like this version.

timsifive:
I think I've addressed all feedback. (And I'm trying to keep the history clean. I think github can still show you the difference between this set of changes and the previous ones.)

src/target/riscv/riscv-013.c
```diff
@@ -4304,9 +4304,33 @@ static int riscv013_halt_go(struct target *target)
...
-			if (t->debug_reason == DBG_REASON_NOTHALTED)
-				t->debug_reason = DBG_REASON_DBGRQ;
+			uint32_t t_dmstatus;
+			if (get_field(dmstatus, DM_DMSTATUS_ALLHALTED) ||
+					get_field(dmstatus, DM_DMSTATUS_ALLUNAVAIL)) {
+				/* All harts are either halted or unavailable. No
+				 * need to read dmstatus for each hart. */
+				t_dmstatus = dmstatus;
+			} else {
+
```
JanMatCodasip: Cosmetic detail - this blank line can be removed.

src/target/riscv/riscv-013.c
JanMatCodasip: As a side note, it seems we have slightly different views on what code is more readable vs. more cumbersome :)

In any case, the current version looks much clearer to me than the previous one - I am fine with it. Thank you for the additional effort on making this part more readable.

src/target/riscv/riscv.c
JanMatCodasip: `info->needs_callback == false` is an invariant at this point in the code, and that is exactly what `assert()`s are for :)

In any case, please feel free to ignore this suggestion. It is not really important to have the assert here.

JanMatCodasip (dismissed):
Thank you for this pull request and for considering my previous suggestions.

The change looks fine - I have no more items to add.

src/target/riscv/riscv-013.c
timsifive: Removed.

src/target/riscv/riscv-013.c
timsifive: > As a side note, it seems we have slightly different views on what code is more readable vs. more cumbersome :)

That's why code reviews are good. :-)

timsifive:
Thanks. Can you approve one last time so I can merge with the latest fix?

JanMatCodasip (approved):

PR #769, opened 2022-11-17
Revive 0.11 debug spec support

timsifive:
This had bitrotted away. Tested against HiFive1. Several tests still don't pass, but I'm not convinced they ever did.

At least now you can use flash, set triggers, and single step again.

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -67,6 +67,8 @@
  * to the target. Afterwards use cache_get... to read results.
  */
 
+static int handle_halt(struct target *target, bool announce);
+
 #define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
 #define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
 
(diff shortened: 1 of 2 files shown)


JanMatCodasip (approved):
I have made quick visual check of the changes and all looks fine.

> Tested against HiFive1. Several tests still don't pass, but I'm not convinced they ever did.

Please, could you post here the list of tests that are currently known to fail on HiFive1. I'd like to run the test suite on HiFive (0.11) from time to time as a regression and would like to know the current state - to make sure I am not breaking things in 0.11 support when making 0.13 related changes. Thank you.

timsifive:
Based on 2 runs:
```
./gdbserver.py targets/SiFive/HiFive1.py
   DebugFunctionCall exception just once
   MemTestReadInvalid both exception
   MemorySampleMixed both exception
   MemorySampleSingle both exception
   RepeatReadTest both exception
   InstantHaltTest both failed
   Semihosting both failed
   SemihostingFileio both failed
```

Based on 2 runs:
```
./gdbserver.py targets/SiFive/HiFive1-flash.py
   DebugCompareSections failed
   InstantHaltTest failed twice
   Semihosting failed twice
   SemihostingFileio failed twice
   DebugExit exception once
   DebugFunctionCall exception
   DebugSymbols exception once
   DebugTurbostep exception once
   MemTestReadInvalid exception twice
   MemorySampleMixed exception twice
   MemorySampleSingle exception twice
   RepeatReadTest exception once
```

PR #770, opened 2022-11-21
target/riscv: Fix 0.13 examine target->status not being halted after halt_go

Dolu1990:
Hi,

So, i had a error while the "init" command execute : 
Error: 212 420 riscv-013.c:4826 riscv013_step_or_resume_current_hart(): [saxon.cpu0] Hart is not halted!

It was due to riscv013_halt_go which assume the target->status is updated outside itself (by the openocd core i think).
The issue is that riscv013_halt_go is also used in the examine function to halt the CPU.

So, the target->status wasn't updated to TARGET_HALTED, which produced the error later in the examine execution, durring the riscv013_step_or_resume_current_hart call.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1708,6 +1708,7 @@ static int examine(struct target *target)
 					info->index);
 			return ERROR_FAIL;
 		}
+		target->state = TARGET_HALTED;
 	}
 
 	/* Without knowing anything else we can at least mess with the


JanMatCodasip (commented):
Thank you for submitting a fix for this bug.

I'd like to recommend a different approach to the fix:

- Remove the check for `target->state != TARGET_HALTED` from `riscv013_step_or_resume_current_hart()`. That's because this is a low-level function that should just perform the resume the operation when asked. It should not try to check the overall target state.
- Instead, move the check one level up - to the two other places where this function is called from: 
  - riscv013_step_current_hart()
  - riscv013_resume_go()

timsifive:
I believe #767 might fix this already because it does a better job keeping target->state always matching what we did to the target. It should merge this week.

JanMatCodasip:
Good point. I no longer see this error after with the #767 change.

Dolu1990:
good ^^

PR #772, opened 2022-11-28
target/riscv: Set target->state in riscv013_halt_go()

### LINKED ISSUE #749: Error: [riscv.cpu0] Hart is not halted! in riscv-013 examine

timsifive:
Then also set it when we resume in examine(), which doesn't use the full abstractions because not all required data structures are filled out yet.

Hopefully fixes #749.

Change-Id: I0c6ab16da1f035ca2fbdb9f7be1462d44ddce3a0

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1765,8 +1765,11 @@ static int examine(struct target *target)
 		return ERROR_FAIL;
 	}
 
-	if (!halted)
+	if (!halted) {
 		riscv013_step_or_resume_current_hart(target, false);
+		target->state = TARGET_RUNNING;
+		target->debug_reason = DBG_REASON_NOTHALTED;
+	}
 
 	if (target->smp) {
 		bool haltgroup_supported;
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1765,8 +1765,10 @@ static int examine(struct target *target)
 		return ERROR_FAIL;
 	}
 
-	if (!halted)
+	if (!halted) {
 		riscv013_step_or_resume_current_hart(target, false);
+		target->state = TARGET_RUNNING;
```
JanMatCodasip, line 1770: I'd recommend to set here also `target->debug_reason` - to keep the "state" and "debug_reason" in sync. (This is just for consistency. I don't think omission of this would cause any issues.)

```suggestion
		target->state = TARGET_RUNNING;
		target->debug_reason = DBG_REASON_NOTHALTED;		
```

src/target/riscv/riscv-013.c
```diff
@@ -4331,8 +4333,17 @@ static int riscv013_halt_go(struct target *target)
 				t->state = TARGET_UNAVAILABLE;
 			}
 		}
+
```
JanMatCodasip, line 4337: Detail: This blank line can be removed.

```suggestion
```

JanMatCodasip (dismissed):

src/target/riscv/riscv-013.c
timsifive: Yeah, but I like whitespace in some places to separate different parts of the code.

JanMatCodasip (approved):

PR #773, opened 2022-11-29
riscv/run_algorithm : Add support for memory parameters

Dolu1990:
This PR add support for riscv_run_algorithm memory parameters. (read/write the main memory).
This is usefull for some openocd flash chip drivers, which load a little access program + the data into the main memory in order to read / write the flash efficiently.

Tested on hardware against : 
https://github.com/SpinalHDL/openocd_riscv/blob/058dfa50d625893bee9fecf8d604141911fac125/src/flash/nor/vexriscv_nor_spi.c#L345
https://github.com/SpinalHDL/openocd_riscv/blob/058dfa50d625893bee9fecf8d604141911fac125/src/flash/nor/vexriscv_nor_spi.c#L394
https://github.com/SpinalHDL/openocd_riscv/tree/riscv_spinal/src/flash/nor/vexriscv_algo

Note, the bandwidth is realy low, so there is ?likely? some overhead somewere in riscv_run_algorithm.

Signed-off-by: Dolu1990 <charles.papon.90@gmail.com>

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1946,14 +1946,20 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 {
 	RISCV_INFO(info);
 
-	if (num_mem_params > 0) {
-		LOG_ERROR("Memory parameters are not supported for RISC-V algorithms.");
-		return ERROR_FAIL;
-	}
-
 	if (target->state != TARGET_HALTED) {
 		LOG_WARNING("target not halted");
 		return ERROR_TARGET_NOT_HALTED;
+	}
+
+	/* Write memory parameters to the target memory */
+	for (int i = 0; i < num_mem_params; i++) {
+		if (mem_params[i].direction != PARAM_IN) {
+			int retval = target_write_buffer(target, mem_params[i].address, mem_params[i].size, mem_params[i].value);
+			if (retval != ERROR_OK) {
+				LOG_ERROR("Couldn't write input mem param into the memory.");
+				return retval;
+			}
+		}
 	}
 
 	/* Save registers */
@@ -2078,6 +2084,18 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 		if (r->type->set(r, buf) != ERROR_OK) {
 			LOG_ERROR("set(%s) failed", r->name);
 			return ERROR_FAIL;
+		}
+	}
+
+	/* Read memory parameters from the target memory */
+	for (int i = 0; i < num_mem_params; i++) {
+		if (mem_params[i].direction != PARAM_OUT) {
+			int retval = target_read_buffer(target, mem_params[i].address, mem_params[i].size,
+					mem_params[i].value);
+			if (retval != ERROR_OK) {
+				LOG_ERROR("Couldn't read output mem param from the memory.");
+				return retval;
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv.c
```diff
@@ -1953,7 +1953,8 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 
 	/* Write memory parameters to the target memory */
 	for (int i = 0; i < num_mem_params; i++) {
-		if (mem_params[i].direction != PARAM_IN) {
+		if (mem_params[i].direction == PARAM_OUT ||
+				mem_params[i].direction == PARAM_IN_OUT) {
```
timsifive: I don't understand how this change has an effect. param_direction only defines IN, OUT, and IN_OUT. So checking for "not IN" is the same as "OUT or IN_OUT", right? What am I missing?

Dolu1990: > I don't understand how this change has an effect

This is right, this change has kind of no behavioural effect.

> What am I missing?

I changed it to be more in line with the rest of the code base. Ex : 
https://github.com/riscv/riscv-openocd/blob/4fb152f733651a5461080f2ebe65eb234bf401b1/src/target/riscv/riscv.c#L2073

timsifive: Oh, I clicked the wrong link in github and all it showed me was this one cleanup, not the functional part of your code.

Dolu1990: Ahhhh XD, got it ^^

src/target/riscv/riscv.c
```diff
@@ -1946,16 +1946,23 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
...
 		return ERROR_TARGET_NOT_HALTED;
 	}
 
+	/* Write memory parameters to the target memory */
+	for (int i = 0; i < num_mem_params; i++) {
+		if (mem_params[i].direction == PARAM_OUT ||
+				mem_params[i].direction == PARAM_IN_OUT) {
+			int retval = target_write_buffer(target, mem_params[i].address, mem_params[i].size, mem_params[i].value);
+			if (retval != ERROR_OK) {
+				LOG_ERROR("Couldn't write input mem param into the memory.");
```
timsifive: Can you add address and size to this error message?

src/target/riscv/riscv.c
```diff
@@ -2081,6 +2088,19 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
...
 	}
 
+	/* Read memory parameters from the target memory */
+	for (int i = 0; i < num_mem_params; i++) {
+		if (mem_params[i].direction == PARAM_IN ||
+				mem_params[i].direction == PARAM_IN_OUT) {
+			int retval = target_read_buffer(target, mem_params[i].address, mem_params[i].size,
+					mem_params[i].value);
+			if (retval != ERROR_OK) {
+				LOG_ERROR("Couldn't read output mem param from the memory.");
```
timsifive: Same as above.

timsifive (changes requested):
Looks good, just a minor suggestion below.

src/target/riscv/riscv.c
Dolu1990: Sure, will do

src/target/riscv/riscv.c
Dolu1990: both done.

timsifive (approved):

PR #775, opened 2022-12-07
add 'init resethalt' command

Alan-19950616:
Change-Id: I6ae5e43fa816fd7e8bf427cc2c1402708e3ecadc

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1579,6 +1579,13 @@ static int examine(struct target *target)
 		riscv013_invalidate_cached_debug_buffer(target);
 	}
 
+	if (target->resethalt_during_init) {
+		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_SETRESETHALTREQ | DM_DMCONTROL_DMACTIVE);
+		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_NDMRESET | DM_DMCONTROL_DMACTIVE);
+		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_DMACTIVE);
+		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_CLRRESETHALTREQ | DM_DMCONTROL_DMACTIVE);
+	}
+
 	dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_HARTSELLO |
 			DM_DMCONTROL_HARTSELHI | DM_DMCONTROL_DMACTIVE |
 			DM_DMCONTROL_HASEL);
--- a/src/openocd.c
+++ b/src/openocd.c
@@ -119,9 +119,16 @@ COMMAND_HANDLER(handle_noinit_command)
 /* OpenOCD can't really handle failure of this command. Patches welcome! :-) */
 COMMAND_HANDLER(handle_init_command)
 {
+	bool resethalt = false;
 
-	if (CMD_ARGC != 0)
+	if (CMD_ARGC > 1)
 		return ERROR_COMMAND_SYNTAX_ERROR;
+
+	if (CMD_ARGC == 1) {
+		if(0 == strcmp("resethalt", CMD_ARGV[0])) {
+			resethalt = true;
+		}
+	}
 
 	int retval;
 	static int initialized;
@@ -130,7 +137,11 @@ COMMAND_HANDLER(handle_init_command)
 
 	initialized = 1;
 
-	retval = command_run_line(CMD_CTX, "target init");
+	if (resethalt) {
+		retval = command_run_line(CMD_CTX, "target init resethalt");
+	} else {
+		retval = command_run_line(CMD_CTX, "target init");
+	}
 	if (retval != ERROR_OK)
 		return ERROR_FAIL;
 
@@ -217,7 +228,7 @@ static const struct command_registration openocd_command_handlers[] = {
 			"Changes command mode from CONFIG to EXEC.  "
 			"Unless 'noinit' is called, this command is "
 			"called automatically at the end of startup.",
-		.usage = ""
+		.usage = "[resethalt]"
 	},
 	{
 		.name = "add_script_search_dir",
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -1616,7 +1616,7 @@ static int target_init_one(struct command_context *cmd_ctx,
 	return ERROR_OK;
 }
 
-static int target_init(struct command_context *cmd_ctx)
+static int target_init(struct command_context *cmd_ctx, bool resethalt)
 {
 	struct target *target;
 	int retval;
@@ -1625,6 +1625,7 @@ static int target_init(struct command_context *cmd_ctx)
 		retval = target_init_one(cmd_ctx, target);
 		if (retval != ERROR_OK)
 			return retval;
+		target->resethalt_during_init = resethalt;
 	}
 
 	if (!all_targets)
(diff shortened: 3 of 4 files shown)


sifiverobert:
Some suggestion ...

Instead of adding different 'init ...' flavors, options it would be better for 'init' to be smarter ...

During initialization sequence it is (usually ...) known if particular action brought desired effect. In case it did not, init may try different (stronger ...) action[s]. In order to provide down-compatibility it can be called 'init auto' or something. Or maybe configuration setting, which may improve how ordinary 'init' works?
It will IMO relieve a need to try all possible 'init' variants in case something is not working ...
Extra setting may be specified in command line, so there will be not necessary to edit existing configuration scripts.

fanghuaqi:
> Some suggestion ...
> 
> Instead of adding different 'init ...' flavors, options it would be better for 'init' to be smarter ...
> 
> During initialization sequence it is (usually ...) known if particular action brought desired effect. In case it did not, init may try different (stronger ...) action[s]. In order to provide down-compatibility it can be called 'init auto' or something. Or maybe configuration setting, which may improve how ordinary 'init' works? It will IMO relieve a need to try all possible 'init' variants in case something is not working ... Extra setting may be specified in command line, so there will be not necessary to edit existing configuration scripts.

I think this changes will not affect original `init` command, and it just add a subcomand `init resethalt` to cover some case that cpu is in hang up state due to bad program issue, so it can do init and reset and halt in the same time, we tried to use init command first, and then do reset halt, it can't help. So we add this type of command.

fanghuaqi:
@Alan-19950616 can provide more details use cases about this new subcommand under init

Alan-19950616:
> Some suggestion ...
> 
> Instead of adding different 'init ...' flavors, options it would be better for 'init' to be smarter ...
> 
> During initialization sequence it is (usually ...) known if particular action brought desired effect. In case it did not, init may try different (stronger ...) action[s]. In order to provide down-compatibility it can be called 'init auto' or something. Or maybe configuration setting, which may improve how ordinary 'init' works? It will IMO relieve a need to try all possible 'init' variants in case something is not working ... Extra setting may be specified in command line, so there will be not necessary to edit existing configuration scripts.



> Some suggestion ...
> 
> Instead of adding different 'init ...' flavors, options it would be better for 'init' to be smarter ...
> 
> During initialization sequence it is (usually ...) known if particular action brought desired effect. In case it did not, init may try different (stronger ...) action[s]. In order to provide down-compatibility it can be called 'init auto' or something. Or maybe configuration setting, which may improve how ordinary 'init' works? It will IMO relieve a need to try all possible 'init' variants in case something is not working ... Extra setting may be specified in command line, so there will be not necessary to edit existing configuration scripts.

First, the "init resethalt" has no effect on "init". 

Second, the current "init" isn't smart enough to solve the problem I'm having.

src/target/riscv/riscv-013.c
```diff
@@ -1579,6 +1579,13 @@ static int examine(struct target *target)
 		riscv013_invalidate_cached_debug_buffer(target);
 	}
 
+	if (target->resethalt_during_init) {
+		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_SETRESETHALTREQ | DM_DMCONTROL_DMACTIVE);
+		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_NDMRESET | DM_DMCONTROL_DMACTIVE);
+		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_DMACTIVE);
+		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_CLRRESETHALTREQ | DM_DMCONTROL_DMACTIVE);
+	}
```
JanMatCodasip, line 1587: Note that the halt-on-reset feature (`dmcontrol.setresethaltreq` and `dmcontrol.clrresethaltreq`) is an optional one - it need not be present in the given target.

JanMatCodasip (commented):
Hi Alan,

> (From https://github.com/riscv/riscv-openocd/issues/776:) Sometimes Openocd cannot halt the CPU due to a software or hardware problem that accesses the empty address.

What do you please mean by "accessing an empty address"? 

OpenOCD shall always be able to halt the CPU (regardless of what the software, running on the hart, is doing). If the CPU does not to halt when asked to do so, then I am afraid this is a bug in the hardware.


> Second, the current "init" isn't smart enough to solve the problem I'm having.

The current "init" command (examine operation) should work correctly on any hardware that behaves according to the RISC-V Debug Spec.

Assuming your hardware is misbehaving, it should be handled by hardware means (e.g. power-cycling the dev board, pressing reset button, restarting simulator of the hardware, fixing the hardware, ..., ...).

But if you've already discovered that you can recover from your specific HW issue by issuing some debug commands from OpenOCD, then you can do that via the existing commands:

- `<target_name> riscv dmi_write` and `<target_name> riscv dmi_read` commands to access the DM registers
- `<target_name> arp_examine` to re-examine the target when it is ready to pass the examination


TL;DR: I'd prefer to not modify the "init" command if this is a HW issue.

src/target/riscv/riscv-013.c
Alan-19950616: > Hi Alan,
> 
> > (From #776:) Sometimes Openocd cannot halt the CPU due to a software or hardware problem that accesses the empty address.
> 
> What do you please mean by "accessing an empty address"?
> 
> OpenOCD shall always be able to halt the CPU (regardless of what the software, running on the hart, is doing). If the CPU does not to halt when asked to do so, then I am afraid this is a bug in the hardware.
> 
> > Second, the current "init" isn't smart enough to solve the problem I'm having.
> 
> The current "init" command (examine operation) should work correctly on any hardware that behaves according to the RISC-V Debug Spec.
> 
> Assuming your hardware is misbehaving, it should be handled by hardware means (e.g. power-cycling the dev board, pressing reset button, restarting simulator of the hardware, fixing the hardware, ..., ...).
> 
> But if you've already discovered that you can recover from your specific HW issue by issuing some debug commands from OpenOCD, then you can do that via the existing commands:
> 
> * `<target_name> riscv dmi_write` and `<target_name> riscv dmi_read` commands to access the DM registers
> * `<target_name> arp_examine` to re-examine the target when it is ready to pass the examination
> 
> TL;DR: I'd prefer to not modify the "init" command if this is a HW issue.

Hi，JanMatCodasip, Thanks for your reply.

"accessing an empty address" it's means cpu accessed an address that should not have been accessed, at this time cpu will not respond to any operations unless reset and halt. 

use halt-on-reset feature Is to ensure that the cpu does not execute any instructions.

TommyMurphyTM1234: > "accessing an empty address" it's means cpu accessed an address that should not have been accessed, at this time cpu will not respond to any operations unless reset and halt.

I've experienced this myself in some CPU implementations, particularly soft IP ones in FPGA (and not just RISC-V). What happened was that an access to "unimplemented" memory regions was handled by a default (e.g. AMBA) slave (possibly implemented in the bus controller IP itself) which never "returned" and so left the CPU stuck in the memory access. I never really knew if this was a buggy hardware/IP implementation, but it certainly seems far from ideal at the very least!

Alan-19950616: > I've experienced this myself in some CPU implementations, particularly soft IP ones in FPGA (and not just RISC-V). What happened was that an access to "unimplemented" memory regions was handled by a default (e.g. AMBA) slave (possibly implemented in the bus controller IP itself) which never "returned" and so left the CPU stuck in the memory access. I never really knew if this was a buggy hardware/IP implementation, but it certainly seems far from ideal at the very least!

That's what I'm dealing with, and I agree with you, but it's a real problem.

PR #777, opened 2022-12-13
target/riscv: Add `riscv` `itrigger` and `etrigger` commands.

timsifive:
This lets the user set a itrigger and etrigger triggers, which don't fit in the normal breakpoint abstraction.

This implementation only allows control of a single trigger of each type. Hardware could support more than one, and that may be useful, but it would make the code/UI more complex and it feels like an unlikely use case.

Change-Id: I76c88636ee73d4bd298b2bd1435cb5d052e86c91

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10716,6 +10716,28 @@ Perform a 32-bit DMI read at address, returning the value.
 Perform a 32-bit DMI write of value at address.
 @end deffn
 
+@subsection RISC-V Trigger Commands
+
+The RISC-V Debug Specification defines several trigger types that don't map
+cleanly onto OpenOCD's notion of hardware breakpoints. These commands let you
+set those triggers directly. (It's also possible to do so by writing the
+appropriate CSRs.)
+
+@deffn {Command} {riscv itrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] [@option{nmi}] mie_bits
+Set an interrupt trigger (type 4) on the current target, which halts the target when it
+fires.  @option{m}, @option{s}, @option{u}, @option{vs}, and @option{vu} control
+which execution modes the trigger fires in.  If [@option{nmi}] is passed then
+the trigger will fire on non-maskable interrupts in those modes. @var{mie_bits}
+controls which interrupts the trigger fires on, using the same bit assignments
+as in the mie CSR (defined in the RISC-V Privileged Spec).
+
+For details on this trigger type, see the RISC-V Debug Specification.
+@end deffn
+
+@deffn {Command} {riscv itrigger clear}
+Clear the type 4 trigger that was set using @command{riscv itrigger set}.
+@end deffn
+
 @section ARC Architecture
 @cindex ARC
 
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10723,6 +10723,21 @@ cleanly onto OpenOCD's notion of hardware breakpoints. These commands let you
 set those triggers directly. (It's also possible to do so by writing the
 appropriate CSRs.)
 
+@deffn {Command} {riscv etrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] exception_codes
+Set an exception trigger (type 5) on the current target, which halts the target when it
+fires.  @option{m}, @option{s}, @option{u}, @option{vs}, and @option{vu} control
+which execution modes the trigger fires in. @var{exception_codes} is a bit
+field, where each bit corresponds to an exception code in mcause (defined in the
+RISC-V Privileged Spec). The etrigger will fire on the exceptions whose bits are
+set in @var{exception_codes}.
+
+For details on this trigger type, see the RISC-V Debug Specification.
+@end deffn
+
+@deffn {Command} {riscv etrigger clear}
+Clear the type 5 trigger that was set using @command{riscv etrigger set}.
+@end deffn
+
 @deffn {Command} {riscv itrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] [@option{nmi}] mie_bits
 Set an interrupt trigger (type 4) on the current target, which halts the target when it
 fires.  @option{m}, @option{s}, @option{u}, @option{vs}, and @option{vu} control
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -520,7 +520,7 @@ static int find_first_trigger_by_id(struct target *target, int unique_id)
 static int set_trigger(struct target *target, int idx, riscv_reg_t tdata1, riscv_reg_t tdata2,
 	riscv_reg_t tdata1_ignore_mask)
 {
-	riscv_reg_t tdata1_rb;
+	riscv_reg_t tdata1_rb, tdata2_rb;
 	if (riscv_set_register(target, GDB_REGNO_TSELECT, idx) != ERROR_OK)
 		return ERROR_FAIL;
 	if (riscv_set_register(target, GDB_REGNO_TDATA1, tdata1) != ERROR_OK)
@@ -528,16 +528,26 @@ static int set_trigger(struct target *target, int idx, riscv_reg_t tdata1, riscv
 	if (riscv_get_register(target, &tdata1_rb, GDB_REGNO_TDATA1) != ERROR_OK)
 		return ERROR_FAIL;
 	if ((tdata1 & ~tdata1_ignore_mask) != (tdata1_rb & ~tdata1_ignore_mask)) {
-		LOG_DEBUG("Trigger doesn't support what we need; After writing 0x%"
+		LOG_TARGET_DEBUG(target,
+			"Trigger %u doesn't support what we need; After writing 0x%"
 			PRIx64 " to tdata1 it contains 0x%" PRIx64
 			"; tdata1_ignore_mask=0x%" PRIx64,
-			tdata1, tdata1_rb, tdata1_ignore_mask);
+			idx, tdata1, tdata1_rb, tdata1_ignore_mask);
 		riscv_set_register(target, GDB_REGNO_TDATA1, 0);
 		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
 	}
-	LOG_DEBUG("tdata1=0x%" PRIx64, tdata1_rb);
 	if (riscv_set_register(target, GDB_REGNO_TDATA2, tdata2) != ERROR_OK)
 		return ERROR_FAIL;
+	if (riscv_get_register(target, &tdata2_rb, GDB_REGNO_TDATA2) != ERROR_OK)
+		return ERROR_FAIL;
+	if (tdata2 != tdata2_rb) {
+		LOG_TARGET_DEBUG(target,
+			"Trigger %u doesn't support what we need; wrote 0x%"
+			PRIx64 " to tdata2 but read back 0x%" PRIx64,
+			idx, tdata2, tdata2_rb);
+		riscv_set_register(target, GDB_REGNO_TDATA1, 0);
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	}
 	return ERROR_OK;
 }
 
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -476,7 +476,7 @@ static bool can_use_napot_match(struct trigger *trigger, riscv_reg_t *tdata2)
 	return false;
 }
 
-static int find_trigger(struct target *target, int type, bool chained, int *idx)
+static int find_trigger(struct target *target, int type, bool chained, unsigned int *idx)
 {
 	RISCV_INFO(r);
 
@@ -517,7 +517,7 @@ static int find_first_trigger_by_id(struct target *target, int unique_id)
 	return -1;
 }
 
-static int set_trigger(struct target *target, int idx, riscv_reg_t tdata1, riscv_reg_t tdata2,
+static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdata1, riscv_reg_t tdata2,
 	riscv_reg_t tdata1_ignore_mask)
 {
 	riscv_reg_t tdata1_rb, tdata2_rb;
@@ -553,7 +553,7 @@ static int set_trigger(struct target *target, int idx, riscv_reg_t tdata1, riscv
 
 static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
 {
-	int idx, ret;
+	int ret;
 	riscv_reg_t tdata1, tdata2;
 
 	RISCV_INFO(r);
@@ -568,6 +568,7 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
 	const uint32_t bpcontrol_bpmatch = 0xf << 7;
 	const uint32_t bpcontrol_bpaction = 0xff << 11;
 
+	unsigned int idx;
 	ret = find_trigger(target, CSR_TDATA1_TYPE_LEGACY, false, &idx);
 	if (ret != ERROR_OK)
 		return ret;
@@ -599,7 +600,7 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
 
 static int maybe_add_trigger_t2(struct target *target, struct trigger *trigger)
 {
-	int idx, ret;
+	int ret;
 	riscv_reg_t tdata1, tdata2;
 
 	RISCV_INFO(r);
@@ -622,6 +623,7 @@ static int maybe_add_trigger_t2(struct target *target, struct trigger *trigger)
 	if (!can_use_napot_match(trigger, &tdata2))
 		goto MATCH_GE_LT;
 
+	unsigned int idx;
 	ret = find_trigger(target, CSR_TDATA1_TYPE_MCONTROL, false, &idx);
 	if (ret != ERROR_OK)
 		return ret;
@@ -686,7 +688,7 @@ static int maybe_add_trigger_t4(struct target *target, bool vs, bool vu,
 				bool nmi, bool m, bool s, bool u, riscv_reg_t interrupts,
 				int unique_id)
 {
-	int idx, ret;
+	int ret;
 	riscv_reg_t tdata1, tdata2;
 
 	RISCV_INFO(r);
@@ -704,6 +706,7 @@ static int maybe_add_trigger_t4(struct target *target, bool vs, bool vu,
 
 	tdata2 = interrupts;
 
+	unsigned int idx;
 	ret = find_trigger(target, CSR_TDATA1_TYPE_ITRIGGER, false, &idx);
 	if (ret != ERROR_OK)
 		return ret;
@@ -718,7 +721,7 @@ static int maybe_add_trigger_t5(struct target *target, bool vs, bool vu,
 				bool m, bool s, bool u, riscv_reg_t exception_codes,
 				int unique_id)
 {
-	int idx, ret;
+	int ret;
 	riscv_reg_t tdata1, tdata2;
 
 	RISCV_INFO(r);
@@ -735,6 +738,7 @@ static int maybe_add_trigger_t5(struct target *target, bool vs, bool vu,
 
 	tdata2 = exception_codes;
 
+	unsigned int idx;
 	ret = find_trigger(target, CSR_TDATA1_TYPE_ETRIGGER, false, &idx);
 	if (ret != ERROR_OK)
 		return ret;
@@ -747,7 +751,7 @@ static int maybe_add_trigger_t5(struct target *target, bool vs, bool vu,
 
 static int maybe_add_trigger_t6(struct target *target, struct trigger *trigger)
 {
-	int idx, ret;
+	int ret;
 	riscv_reg_t tdata1, tdata2;
 
 	RISCV_INFO(r);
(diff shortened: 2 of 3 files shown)


src/target/riscv/riscv.c
```diff
@@ -506,6 +506,17 @@ static int find_trigger(struct target *target, int type, bool chained, int *idx)
...
+static int find_trigger_by_id(struct target *target, int unique_id)
+{
+	RISCV_INFO(r);
+
+	for (unsigned i = 0; i < r->trigger_count; i++) {
+		if (r->trigger_unique_id[i] == unique_id)
+			return i;
+	}
+	return -1;
+}
```
JanMatCodasip: Minor thing to note - In some cases, there will be multiple triggers with the with the same `unique_id`. For example here: https://github.com/riscv/riscv-openocd/pull/777/files#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR766

This function would find just the first of them. This is not an issue for itrigger but would be insufficient for other trigger types.

To avoid possible future mistakes, you may make the function specific for itriggers: `find_itrigger(target)`, if you'd like.

doc/openocd.texi
```diff
@@ -10716,6 +10716,28 @@ Perform a 32-bit DMI read at address, returning the value.
...
+The RISC-V Debug Specification defines several trigger types that don't map
+cleanly onto OpenOCD's notion of hardware breakpoints. These commands let you
+set those triggers directly. (It's also possible to do so by writing the
+appropriate CSRs.)
+
+@deffn {Command} {riscv itrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] [@option{nmi}] mie_bits
+Set a type 4 trigger on the current target. @option{m}, @option{s}, @option{u},
+@option{vs}, and @option{vu} control which execution modes the trigger fires in.
+If [@option{nmi}] is passed then the trigger will on fire non-maskable
+interrupts in those modes.  mie_bits controls which interrupts the trigger fires
```
JanMatCodasip: Minor things: 

* Should the "mie_bits" option be formatted like the others?
* You may also mention that the trigger action is to halt the hart (enter debug mode).

```suggestion
interrupts in those modes and cause the hart to halt.  @option{mie_bits} controls which interrupts the trigger fires
```

src/target/riscv/riscv.c
```diff
@@ -3080,6 +3123,67 @@ COMMAND_HANDLER(riscv_set_ebreaku)
...
+			else if (!strcmp(CMD_ARGV[i], "nmi"))
+				nmi = true;
+			else if (!strcmp(CMD_ARGV[i], "m"))
+				m = true;
+			else if (!strcmp(CMD_ARGV[i], "s"))
+				s = true;
+			else if (!strcmp(CMD_ARGV[i], "u"))
+				u = true;
+			else
+				COMMAND_PARSE_NUMBER(u64, CMD_ARGV[i], interrupts);
```
JanMatCodasip, line 3312: Do we wish to notify the user in case they were trying to set a useless trigger?

E.g.: 
* interrupts == 0 && nmi is not set
* none of the modes is set (m, s, u, vs, vu)

src/target/riscv/riscv.c
```diff
@@ -3080,6 +3123,67 @@ COMMAND_HANDLER(riscv_set_ebreaku)
...
+		if (result != ERROR_OK)
+			LOG_ERROR("Failed to set requested itrigger.");
+		return result;
+
+	} else if (!strcmp(CMD_ARGV[0], "clear")) {
+		if (CMD_ARGC != 1) {
+			LOG_ERROR("clear command takes no extra arguments.");
+			return ERROR_COMMAND_SYNTAX_ERROR;
+		}
+		return remove_trigger(target, -4);
```
JanMatCodasip: Minor point: Do we wish to check beforehand if the itrigger is set before trying to remove it?

This will give opportunity to print a better error message to the user, e.g. `Itrigger is not currently set, nothing to clear.`.

Now it would fail in `remove_trigger()` which would produce error message `Couldn't find the hardware resources used by hardware trigger.`

doc/openocd.texi
```diff
@@ -10716,6 +10716,28 @@ Perform a 32-bit DMI read at address, returning the value.
...
+on, using the same bit assignments as in the mie CSR (defined in the RISC-V
```
JanMatCodasip: Minor: Is there any suitable formatting that could be used for "mie" in this sentence?

src/target/riscv/riscv.c
JanMatCodasip: I'd recommend a defined constant instead of `-4` everywhere.

For example:

```suggestion
		return remove_trigger(target, ITRIGGER_UNIQUE_ID);
```

JanMatCodasip (dismissed):
Checked visually (I do not have a target at hand with itrigger support). It looks fine - I have just minor stylistic items.

src/target/riscv/riscv.c
timsifive: Good point. I don't want to limit it to itrigger(). I wrote this function because it will be helpful for etrigger and icount as well.

How about find_first_trigger_by_id()?

doc/openocd.texi
timsifive: Ah, I found the correct formatting for mie_bits.
I'm updating it to mention that firing means halting, but with slightly different wording than you suggested.

src/target/riscv/riscv.c
timsifive: I'll add it that. It's easy and will probably save someone some time.

src/target/riscv/riscv.c

doc/openocd.texi
timsifive: Maybe, but I can't easily find it. Browsing through the document, I see e.g.
```
@deffn {Command} {nrf5 info}
Decodes and shows information from FICR and UICR registers.
@end deffn
```
which doesn't use any special formatting.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
JanMatCodasip: That sounds fine (and very clear).

src/target/riscv/riscv.c
```diff
@@ -3151,6 +3194,81 @@ COMMAND_HANDLER(riscv_set_ebreaku)
...
+	}
+
+	struct target *target = get_current_target(CMD_CTX);
+	const int unique_id = -CSR_TDATA1_TYPE_ITRIGGER;
+
+	if (riscv_enumerate_triggers(target) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (!strcmp(CMD_ARGV[0], "set")) {
+		if (find_first_trigger_by_id(target, unique_id) >= 0) {
```
JanMatCodasip, line 3211: Minor stylistic thing - Here I'd recommend to have the unique ID for itrigger as a global constant. 

```
// somewhere near the top of riscv.c:
#define ITRIGGER_UNIQUE_ID (-CSR_TDATA1_TYPE_ITRIGGER)

// ...

// here in COMMAND_HANDLER(riscv_itrigger):
find_first_trigger_by_id(target, ITRIGGER_UNIQUE_ID)  // etc.
```

I understand that the unique ID is not used anywhere else outside this function (yet?). But I believe it will be quicker to the reader to understand the code when the uppercase constant is there, contrary to `unique_id` (which looks like an ordinary variable when reading the code).

src/target/riscv/riscv.c
```diff
@@ -3568,6 +3686,13 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.help = "Control dcsr.ebreaku. When off, U-mode ebreak instructions "
 			"don't trap to OpenOCD. Defaults to on."
 	},
+	{
+		.name = "itrigger",
+		.handler = riscv_itrigger,
+		.mode = COMMAND_ANY,
+		.usage = "set [vs] [vu] [nmi] [m] [s] [u] <mie bits>|clear",
+		.help = "Set or clear a single itrigger."
```
JanMatCodasip: Little favor to those who may not be familiar with the term `itrigger`:

```suggestion
		.help = "Set or clear a single itrigger (RISC-V Interrupt Trigger)."
```

src/target/riscv/riscv.c
JanMatCodasip: I'm sorry to open this one again. This is a minor thing, but I believe important for readability:

Here I'd still recommend to have the unique ID for itrigger as a global constant.

```
// somewhere near the top of riscv.c:
#define ITRIGGER_UNIQUE_ID (-CSR_TDATA1_TYPE_ITRIGGER)

// ...

// here in COMMAND_HANDLER(riscv_itrigger):
find_first_trigger_by_id(target, ITRIGGER_UNIQUE_ID)  // etc.
```

I understand that the unique ID is not used anywhere else outside this function (yet?). But I believe it will be quicker to the reader to understand the code when the uppercase constant is there, contrary to unique_id (which looks like an ordinary variable when reading the code).

src/target/riscv/riscv.c
JanMatCodasip: Another possibility: Leave the code as is but rename the variable

```
const int ITRIGGER_UNIQUE_ID = -CSR_TDATA1_TYPE_ITRIGGER;
```

src/target/riscv/riscv.c
JanMatCodasip: Another possibility: Leave the code as is, just rename the variable to make it clear it is a constant.

```
const int ITRIGGER_UNIQUE_ID = -CSR_TDATA1_TYPE_ITRIGGER;
```

JanMatCodasip:
In `riscv_enumerate_triggers()`, outstanding triggers from previous sessions are cleared:
https://github.com/riscv/riscv-openocd/blob/b337b0cfb4612c0474c109abeb511189bfd5559c/src/target/riscv/riscv.c#L4122

I believe this should be extended to clear itriggers, too.

src/target/riscv/riscv.c
timsifive: I went with the 2nd option.

src/target/riscv/riscv.c
timsifive: How about "Set or clear a single interrupt trigger."

doc/openocd.texi
```diff
@@ -10716,6 +10716,44 @@ Perform a 32-bit DMI read at address, returning the value.
...
+set in @var{exception_codes}.
+
+For details on this trigger type, see the RISC-V Debug Specification.
+@end deffn
+
+@deffn {Command} {riscv etrigger clear}
+Clear the type 5 trigger that was set using @command{riscv etrigger set}.
+@end deffn
+
+@section ARC Architecture
```
JanMatCodasip: This seems to be a copy-paste typo (an extra `@section ...`)

doc/openocd.texi
```diff
@@ -10716,6 +10716,44 @@ Perform a 32-bit DMI read at address, returning the value.
...
 
+@subsection RISC-V Trigger Commands
+
+The RISC-V Debug Specification defines several trigger types that don't map
+cleanly onto OpenOCD's notion of hardware breakpoints. These commands let you
+set those triggers directly. (It's also possible to do so by writing the
+appropriate CSRs.)
+
+@deffn {Command} {riscv etrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] exception_codes
+Set a type 5 trigger on the current target, which halts the target when it
```
JanMatCodasip: Here I'd prefer to put the full name of the trigger - for maximum clarity:

```suggestion
Set a RISC-V Exception Trigger (type 5 trigger) on the current target, which halts the target when it
```

doc/openocd.texi
```diff
@@ -10716,6 +10716,44 @@ Perform a 32-bit DMI read at address, returning the value.
...
+For details on this trigger type, see the RISC-V Debug Specification.
+@end deffn
+
+@deffn {Command} {riscv etrigger clear}
+Clear the type 5 trigger that was set using @command{riscv etrigger set}.
+@end deffn
+
+@section ARC Architecture
+@deffn {Command} {riscv itrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] [@option{nmi}] mie_bits
+Set a type 4 trigger on the current target, which halts the target when it
```
JanMatCodasip:
```suggestion
Set a RISC-V Interrupt trigger (type 4 trigger) on the current target, which halts the target when it
```

src/target/riscv/riscv.c
```diff
@@ -3590,6 +3817,20 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.help = "Control dcsr.ebreaku. When off, U-mode ebreak instructions "
 			"don't trap to OpenOCD. Defaults to on."
 	},
+	{
+		.name = "etrigger",
+		.handler = riscv_etrigger,
+		.mode = COMMAND_ANY,
+		.usage = "set [vs] [vu] [nmi] [m] [s] [u] <mie bits>|clear",
+		.help = "Set or clear a single interrupt trigger."
```
JanMatCodasip: There seem to be few copy-paste mistakes:

```suggestion
		.name = "etrigger",
		.handler = riscv_etrigger,
		.mode = COMMAND_ANY,
		.usage = "set [vs] [vu] [m] [s] [u] <exception_codes>|clear",
		.help = "Set or clear a single exception trigger."
```

src/target/riscv/riscv.c
```diff
@@ -3173,6 +3253,153 @@ COMMAND_HANDLER(riscv_set_ebreaku)
...
+
+	struct target *target = get_current_target(CMD_CTX);
+	const int ITRIGGER_UNIQUE_ID = -CSR_TDATA1_TYPE_ITRIGGER;
+
+	if (riscv_enumerate_triggers(target) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (!strcmp(CMD_ARGV[0], "set")) {
+		if (find_first_trigger_by_id(target, ITRIGGER_UNIQUE_ID) >= 0) {
+			LOG_ERROR("An itrigger is already set, and OpenOCD "
```
JanMatCodasip: Here it'd be good to use LOG_TARGET_ERROR:

```suggestion
			LOG_TARGET_ERROR(target, "An itrigger is already set, and OpenOCD "
```

There are more of these in `riscv_itrigger` and `riscv_etrigger`. I won't list all of them so that I don't clutter the review.

src/target/riscv/riscv.h
```diff
@@ -124,8 +124,10 @@ typedef struct {
 	/* record the tinfo of each trigger */
 	unsigned int trigger_tinfo[RISCV_MAX_TRIGGERS];
 
-	/* For each physical trigger, contains -1 if the hwbp is available, or the
-	 * unique_id of the breakpoint/watchpoint that is using it.
+	/* For each physical trigger contains:
+	 * -1: the hwbp is available
+	 * -4: The trigger is used by the itrigger command
```
JanMatCodasip, line 129: Now there is also the `-5` value for etriggers.

```suggestion
	 * -4: The trigger is used by the itrigger command
	 * -5: The trigger is used by the etrigger command	 
```

src/target/riscv/riscv.c
```diff
@@ -3590,6 +3817,20 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
+		.handler = riscv_etrigger,
+		.mode = COMMAND_ANY,
+		.usage = "set [vs] [vu] [nmi] [m] [s] [u] <mie bits>|clear",
+		.help = "Set or clear a single interrupt trigger."
+	},
+	{
+		.name = "itrigger",
+		.handler = riscv_itrigger,
+		.mode = COMMAND_ANY,
+		.usage = "set [vs] [vu] [nmi] [m] [s] [u] <mie bits>|clear",
```
JanMatCodasip: Detail: Here I'd add an underscore to make it very clear that `<mie_bits>` is a single argument:

```suggestion
		.usage = "set [vs] [vu] [nmi] [m] [s] [u] <mie_bits>|clear",
```

src/target/riscv/riscv.c
```diff
@@ -661,6 +672,69 @@ static int maybe_add_trigger_t2(struct target *target, struct trigger *trigger)
...
+	tdata1 = set_field(tdata1, CSR_ITRIGGER_M, m);
+	tdata1 = set_field(tdata1, CSR_ITRIGGER_S, s);
+	tdata1 = set_field(tdata1, CSR_ITRIGGER_U, u);
+
+	tdata2 = interrupts;
+
+	ret = find_trigger(target, CSR_TDATA1_TYPE_ITRIGGER, false, &idx);
+	if (ret != ERROR_OK)
+		return ret;
+	ret = set_trigger(target, idx, tdata1, tdata2, 0);
```
JanMatCodasip, line 713: The debug spec says (for both itrigger and etrigger): `A debugger must read back tdata2 after
writing it to confirm the requested functionality is actually supported.`

I believe the read-back & check of `tdata2` should be added into `set_trigger()`.

My understanding is that the tdata2 check should be performed in all cases - for all trigger types. (I hope I'm not missing any special corner case.)

src/target/riscv/riscv.c
```diff
@@ -3590,6 +3817,20 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.help = "Control dcsr.ebreaku. When off, U-mode ebreak instructions "
 			"don't trap to OpenOCD. Defaults to on."
 	},
+	{
+		.name = "etrigger",
+		.handler = riscv_etrigger,
+		.mode = COMMAND_ANY,
```
JanMatCodasip: Please use `COMMAND_EXEC` here:

```suggestion
		.mode = COMMAND_EXEC,
```

src/target/riscv/riscv.c
```diff
@@ -3590,6 +3817,20 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
+		.usage = "set [vs] [vu] [nmi] [m] [s] [u] <mie bits>|clear",
+		.help = "Set or clear a single interrupt trigger."
+	},
+	{
+		.name = "itrigger",
+		.handler = riscv_itrigger,
+		.mode = COMMAND_ANY,
```
JanMatCodasip: Same as above - `COMMAND_EXEC` should be used here:

```suggestion
		.mode = COMMAND_EXEC,
```

doc/openocd.texi
timsifive: Fixed. Thanks.

doc/openocd.texi
timsifive: Sounds good, but I'm removing "RISC-V". That seems redundant since it's already a `riscv` command in the RISC-V section of the document.

src/target/riscv/riscv.c
timsifive: Thank you.

src/target/riscv/riscv.c
timsifive: Done. Thanks.

src/target/riscv/riscv.h
timsifive: Done.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
timsifive: Good point.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
```diff
@@ -506,27 +506,48 @@ static int find_trigger(struct target *target, int type, bool chained, int *idx)
...
 		return ERROR_FAIL;
 	if (riscv_get_register(target, &tdata1_rb, GDB_REGNO_TDATA1) != ERROR_OK)
 		return ERROR_FAIL;
 	if ((tdata1 & ~tdata1_ignore_mask) != (tdata1_rb & ~tdata1_ignore_mask)) {
-		LOG_DEBUG("Trigger doesn't support what we need; After writing 0x%"
+		LOG_TARGET_DEBUG(target,
+			"Trigger doesn't support what we need; After writing 0x%"
 			PRIx64 " to tdata1 it contains 0x%" PRIx64
 			"; tdata1_ignore_mask=0x%" PRIx64,
 			tdata1, tdata1_rb, tdata1_ignore_mask);
```
JanMatCodasip: Minor thing: For easier troubleshooting, it will be good to have the trigger index in the debug message.

Also the `idx` parameter in `set_trigger()` (and in `find_trigger()`) can be made `unsigned`.

```suggestion
		LOG_TARGET_DEBUG(target,
			"Trigger #%u doesn't support what we need; After writing 0x%"
			PRIx64 " to tdata1 it contains 0x%" PRIx64
			"; tdata1_ignore_mask=0x%" PRIx64,
			idx, tdata1, tdata1_rb, tdata1_ignore_mask);
```

src/target/riscv/riscv.c
```diff
@@ -506,27 +506,48 @@ static int find_trigger(struct target *target, int type, bool chained, int *idx)
...
-	LOG_DEBUG("tdata1=0x%" PRIx64, tdata1_rb);
 	if (riscv_set_register(target, GDB_REGNO_TDATA2, tdata2) != ERROR_OK)
 		return ERROR_FAIL;
+	if (riscv_get_register(target, &tdata2_rb, GDB_REGNO_TDATA2) != ERROR_OK)
+		return ERROR_FAIL;
+	if (tdata2 != tdata2_rb) {
+		LOG_TARGET_DEBUG(target,
+			"Trigger doesn't support what we need; wrote 0x%"
+			PRIx64 " to tdata2 but read back 0x%" PRIx64,
+			tdata2, tdata2_rb);
```
```suggestion
		LOG_TARGET_DEBUG(target,
			"Trigger #%u doesn't support what we need; wrote 0x%"
			PRIx64 " to tdata2 but read back 0x%" PRIx64,
			idx, tdata2, tdata2_rb);
```

JanMatCodasip (dismissed):
I have just one last cosmetic suggestion. 

Otherwise the changes look fine to me (and useful), thank you.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

JanMatCodasip (approved):

PR #780, opened 2022-12-29
target/riscv: Remove `riscv test_sba_config_reg` command.

timsifive:
This command is supposed to be a start at a compliance test for system bus access. It doesn't pass against spike because it doesn't handle all cases where the interface might be busy. It's not documented. As far as I know nobody uses it.

So delete 400 lines of code instead of trying to fix it.

Change-Id: Ib94f2acb95a48f7c07d4f44206ff7373b03857f3


JanMatCodasip (approved):
Thank you for this cleanup. Happy to see this legacy and unmaintained code go away.

GregSavin (approved):
Looks good; it's always nice to take those rare opportunities to simplify when there's no downside.

PR #786, opened 2023-01-10
target/riscv: added support for missing VCSR register

aap-sc:
Change-Id: I0ce7b9e76c613400916c46fad0f19984ea4b482e


timsifive (approved):
This looks right. What testing have you done?

aap-sc:
I've tested locally with gdb and telnet (using spike) to make sure that we can actually read the csr . Also I've run the whole riscv-tests (external debugger part) just in case.

TommyMurphyTM1234:
> I've tested locally with gdb and telnet (using spike) to make sure that we can actually read the csr . Also I've run the whole riscv-tests (external debugger part) just in case.

Is Spike a good test? In the past Spike's implemention has masked bugs in OpenOCD. And doesn't Spike deal with registers directly than other targets?

aap-sc:
> Is Spike a good test? 

We don't have any other testsuite available to the public... I've also run some tests on FPGA in our lab with a set of our own .tcl-based tests, but we don't have those available in open source.

> In the past Spike's implemention has masked bugs in OpenOCD.

Yes. Spike is not ideal (though it's still better than nothing). However, for this change, I'm fairly certain that I won't introduce a regression. Since for this particular case it looks like the register was just skipped accidentally. There is nothing special about this register: it should work as all the remaining vector csrs.

aap-sc:
> And doesn't Spike deal with registers directly than other targets?

I'm not certain what you mean here.. Though if you are hinting at the fact that in order to read vector registers we should enable access by setting a bit to mstatus - there was a change a while ago that handles this situation (see dc49ed8ae25b6ac99f11864ce0ec83a2589b4ad8 - the commit is about FP registers, but it works for vector registers too)

timsifive (approved):

timsifive:
> Is Spike a good test? In the past Spike's implemention has masked bugs in OpenOCD. And doesn't Spike deal with registers directly than other targets?

Spike is the best test we have that people can just run. Spike supports register access with and without the program buffer. If you run all riscv-tests/debug then you get coverage of both those code paths in OpenOCD.

Issue #785, opened 2023-01-10
[question/feature] facilities to hide some CSRs from the user


aap-sc:
Hi!

Currently, OpenOCD exposes most of the CSRs defined by the riscv-privileged spec.

This results in some confusion or minor usability issues for the user:

1. Not every CSR is guarded by an extension. An example is `sedeleg` register. The register is deprecated (that is, the latest spec does not have it), however, the encoding still has its definition (for compatibility reasons, I guess). Since we write such registers to the gdb target .xml this leads to some confusion because gdb can query the state of non-existent registers (like when we do `info all-registers` command)

2. When we debug bare-metal applications often we don't need access to registers like CSR_MHPMCOUNTER* or manual access to trigger registers. We could even use IDE like Eclipse of VisualStudio. When the debugger is in control, these IDEs display all CSRs and attempt to retrieve their values. These activities slow down debugging and, in some cases, may cause instability.

On the other hand, OpenOCD has a concept of `hidden` registers - these are registers  not listed by `reg` command and which are not included in the target.xml sent to GDB. To avoid the above issues, I'd like to have the ability to explicitly mark CSRs as hidden ones. In fact, I have a patch that implements this.

However, I'm not sure if there are any objections regarding inclusion of such functionality in the upstream. Thus the question.

TommyMurphyTM1234:
There are a few older issues logged that might also be relevant here?

* https://github.com/riscv/riscv-openocd/issues/245
* https://github.com/riscv/riscv-openocd/issues/175
* https://github.com/riscv/riscv-openocd/issues/136
* https://github.com/riscv/riscv-openocd/issues/500

And some other possibly relevant content:

* https://stackoverflow.com/questions/72759791/why-does-gdb-do-not-show-all-risc-v-csrs-when-debugging-bare-metal-program-runni
* https://sourceware.org/pipermail/gdb-patches/2020-June/169563.html
* https://github.com/eclipse-embed-cdt/eclipse-plugins/issues/345
* https://review.openocd.org/c/openocd/+/6362

I'm just posting these links because, from past experience, changes in this context can lead to unintended consequences in all sorts of related tools.

aap-sc:
@TommyMurphyTM1234 , from the above list the most relevant is #175, I think.

I've decided to create a PR (#787), since changes are relatively small. The idea is to allow user/vendor to decide which CSRs are exposed during the target configuration stage.

TommyMurphyTM1234:
[...]

PR #787, opened 2023-01-10
target/riscv: hide_csrs configuration option

aap-sc:
This option allows users to mark certain CSRs as hidden so they could be expluded from *reg* output and target.xml

Change-Id: Iddf8456cd3901f572f8590329ebba5229974d24a

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10553,6 +10553,23 @@ $_TARGETNAME expose_custom 32=myregister
 @end example
 @end deffn
 
+@deffn {Config Command} {riscv hide_csrs} n[-m] [,n1[-m1]] [...]
+The RISC-V Specification defines many CSRs, and we may want to avoid showing
+each CSR to the user, as they may not be relevant to the task at hand. For
+example, we may choose not to show trigger or PMU registers for simple
+debugging scenarios. This command allows to mark individual registers or
+register ranges (inclusive) as "hidden". Such hidden registers won't be
+displayed in GDB or @code{reg} command output.
+
+@example
+
+# Hide range of RISC-V CSRs
+# CSR_TSELECT - 1952 and CSR_DATA1 - 1953
+$_TARGETNAME riscv hide_csrs 1952-1953
+
+@end example
+@end deffn
+
 @deffn {Command} {riscv memory_sample} bucket address|clear [size=4]
 Configure OpenOCD to frequently read size bytes at the given addresses.
 Execute the command with no arguments to see the current configuration. Use
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10564,7 +10564,7 @@ displayed in GDB or @code{reg} command output.
 @example
 
 # Hide range of RISC-V CSRs
-# CSR_TSELECT - 1952 and CSR_DATA1 - 1953
+# CSR_TSELECT - 1952 and CSR_TDATA1 - 1953
 $_TARGETNAME riscv hide_csrs 1952-1953
 
 @end example
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -433,6 +433,11 @@ static void riscv_deinit_target(struct target *target)
 	riscv_free_registers(target);
 
 	range_list_t *entry, *tmp;
+	list_for_each_entry_safe(entry, tmp, &info->hide_csr, list) {
+		free(entry->name);
+		free(entry);
+	}
+
 	list_for_each_entry_safe(entry, tmp, &info->expose_csr, list) {
 		free(entry->name);
 		free(entry);
@@ -2937,6 +2942,26 @@ COMMAND_HANDLER(riscv_set_expose_custom)
 	return ret;
 }
 
+COMMAND_HANDLER(riscv_hide_csrs)
+{
+	if (CMD_ARGC == 0) {
+		LOG_ERROR("Command expects parameters");
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(info);
+	int ret = ERROR_OK;
+
+	for (unsigned int i = 0; i < CMD_ARGC; i++) {
+		ret = parse_ranges(&info->hide_csr, CMD_ARGV[i], "csr", 0xfff);
+		if (ret != ERROR_OK)
+			break;
+	}
+
+	return ret;
+}
+
 COMMAND_HANDLER(riscv_authdata_read)
 {
 	unsigned int index = 0;
@@ -3676,6 +3701,16 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 			"etc. This must be executed before `init`."
 	},
 	{
+		.name = "hide_csrs",
+		.handler = riscv_hide_csrs,
+		.mode = COMMAND_CONFIG,
+		.usage = "{n0|n-m0}[,n1|n-m1]......",
+		.help = "Configure a list of inclusive ranges for CSRs to hide from gdb. "
+			"Hidden registers are still available, but are not listed in "
+			"gdb target description and `reg` command output."
+			"This must be executed before `init`."
+	},
+	{
 		.name = "authdata_read",
 		.handler = riscv_authdata_read,
 		.usage = "[index]",
@@ -3926,6 +3961,7 @@ void riscv_info_init(struct target *target, riscv_info_t *r)
 
 	INIT_LIST_HEAD(&r->expose_csr);
 	INIT_LIST_HEAD(&r->expose_custom);
+	INIT_LIST_HEAD(&r->hide_csr);
 
 	r->vsew64_supported = YNM_MAYBE;
 }
@@ -5213,6 +5249,14 @@ int riscv_init_registers(struct target *target)
 								csr_number, entry->name ? entry->name : reg_name);
 
 						r->exist = true;
+						break;
+					}
+			} else if (r->exist && !list_empty(&info->hide_csr)) {
+				range_list_t *entry;
+				list_for_each_entry(entry, &info->hide_csr, list)
+					if ((entry->low <= csr_number) && (csr_number <= entry->high)) {
+						LOG_TARGET_DEBUG(target, "Hiding CSR %d (name=%s)", csr_number, r->name);
+						r->hidden = true;
 						break;
 					}
 			}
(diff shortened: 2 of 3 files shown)


aap-sc:
I describe the motivation for this PR here: #785

src/target/riscv/riscv.h
```diff
@@ -248,6 +248,10 @@ typedef struct {
 	 * from range 0xc000 ... 0xffff. */
 	struct list_head expose_custom;
 
+	/* The list of registers to mark as "hidden". Hidden registers are available
+	 * but do not appear in gdb targets description or reg command output. */
+	struct list_head hidden_csr;
```
timsifive:
```suggestion
	struct list_head hide_csr;
```
To keep naming consistent with `expose_csr`.

src/target/riscv/riscv.c
```diff
@@ -3675,6 +3700,16 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 			"expose. custom0 is accessed as abstract register number 0xc000, "
 			"etc. This must be executed before `init`."
 	},
+	{
+		.name = "hide_csrs",
+		.handler = riscv_set_hidden_csrs,
+		.mode = COMMAND_CONFIG,
+		.usage = "n0[-m0|=name0][,n1[-m1|=name1]]...",
+		.help = "Configure a list of inclusive ranges for CSRs to mark them as "
+			"hidden. Hidden registers are still available, but are not listed in "
```
```suggestion
		.help = "Configure a list of inclusive ranges for CSRs to hide from gdb. "
			"Hidden registers are still available, but are not listed in "
```

src/target/riscv/riscv.c
```diff
@@ -5209,6 +5245,14 @@ int riscv_init_registers(struct target *target)
 						r->exist = true;
 						break;
 					}
+			} else if (r->exist && !list_empty(&info->hidden_csr)) {
+				range_list_t *entry;
+				list_for_each_entry(entry, &info->hidden_csr, list)
+					if ((entry->low <= csr_number) && (csr_number <= entry->high)) {
+						LOG_DEBUG("Hiding CSR %d (name=%s)", csr_number, r->name);
```
```suggestion
						LOG_TARGET_DEBUG(target, "Hiding CSR %d (name=%s)", csr_number, r->name);
```

timsifive (changes requested):
Looks alright. Can you document this new command in doc/openocd.texi?

TommyMurphyTM1234:
> This option allows users to mark certain CSRs as hidden so they could be expluded from _reg_ output and target.xml
> 
> Change-Id: Iddf8456cd3901f572f8590329ebba5229974d24a

If it's going to be added then why not generalise it to allow ANY register to be hidden - integer, floating point, CSR - using any of the canonical name, the assembler mnemonic name, the ABI name(s?), any other relevant names (e.g. CSR alias names, deprecated/demoted CSR names)?

src/target/riscv/riscv.c
```diff
@@ -3675,6 +3700,16 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 			"expose. custom0 is accessed as abstract register number 0xc000, "
```
TommyMurphyTM1234, line 3700: Why does it have to be executed before `init`?
Might the GDB user want to do `monitor ...` to change the register list?

Will the commands work via the OpenOCD telnet interface?

See here for my comment about identifying registers by name:

* https://github.com/riscv/riscv-openocd/pull/787#issuecomment-1380934928

aap-sc: > Why does it have to be executed before init?

Because list of available registers is built/initialized during target examination procedure.

> Might the GDB user want to do monitor ... to change the register list?
> Will the commands work via the OpenOCD telnet interface?

This won't work. There are a couple of reasons for that. 
1. As is - the command  can be run during the configuration stage only. That is before gdbserver client can connect.  
2. AFAIK the remote target configuration is requested by gdb only once upon connection. Maybe there are ways to re-request it, but I'm not aware of them currently (+ see the next item).
3. The way registers are initialized in OpenOCD complicates things. These are initialized during the examine procedure. And making OpenOCD to re-examine target may imply way too many changes.

aap-sc:
Regarding this one (from @TommyMurphyTM1234 ):

> If it's going to be added then why not generalise it to allow ANY register to be hidden - integer, floating point, CSR - using any of the canonical name, the assembler mnemonic name, the ABI name(s?), any other relevant names (e.g. CSR alias names, deprecated/demoted CSR names)?

1. Regarding other classes of registers like GPR, FPR and VPR. I'll investigate this, however GDB may have some special handling for these kinds of registers and this may require some additional testing. I'm fairly certain that the code works for CSR's since I experimented with it quite a lot - not so sure for others.

2. Regarding aliases and canonical/ABI names... I don't believe we have canonical naming scheme present in the current codebase of OpenOCD and GDB. It was removed a couple of years ago from the respected codebases. As for aliases - I'll double check, but I don't believe we have any.

aap-sc:
@timsifive, regarding this one:
> Can you document this new command in doc/openocd.texi?

Sure (I've completely forgotten about it).

timsifive:
> If it's going to be added then why not generalise it to allow ANY register to be hidden - integer, floating point, CSR - using any of the canonical name, the assembler mnemonic name, the ABI name(s?), any other relevant names (e.g. CSR alias names, deprecated/demoted CSR names)?

That sounds like a lot of work for registers that nobody might ever want to hide. As it is the CSR work is already done so easy to hook into for this feature. At a minimum the parser would get a lot more complex to support what you mentioned above.

TommyMurphyTM1234:
> That sounds like a lot of work for registers that nobody might ever want to hide. As it is the CSR work is already done so easy to hook into for this feature. At a minimum the parser would get a lot more complex to support what you mentioned above.

OK - fair enough. :-)

src/target/riscv/riscv.c
```diff
@@ -3675,6 +3700,16 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
 			"etc. This must be executed before `init`."
 	},
+	{
+		.name = "hide_csrs",
+		.handler = riscv_set_hidden_csrs,
+		.mode = COMMAND_CONFIG,
+		.usage = "n0[-m0|=name0][,n1[-m1|=name1]]...",
```
JanMatCodasip: I recommend the following syntax for the list of hidden CSRs: `{n0|n0-m0|name0}[,n1|n1-m1|name1]...`

Or even a simpler variant, without the name, would suffice: `{n0|n-m0}[,n1|n-m1]...`

The current syntax (implemented in `parse_ranges()`) makes sense for `expose_csrs` command but not for `hide_csrs`. In `hide_csrs`, there is no point in setting a custom name for a certain CSR number (`some_num=some_name`).

src/target/riscv/riscv.c
```diff
@@ -3675,6 +3700,16 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 			"expose. custom0 is accessed as abstract register number 0xc000, "
 			"etc. This must be executed before `init`."
 	},
+	{
+		.name = "hide_csrs",
+		.handler = riscv_set_hidden_csrs,
```
JanMatCodasip: For consistency, it would be good to call this `riscv_hide_csrs`.

```suggestion
		.handler = riscv_hide_csrs,
```

JanMatCodasip (changes requested):

src/target/riscv/riscv.h
aap-sc: addressed

src/target/riscv/riscv.c
aap-sc: Thanks for suggestion, addressed

src/target/riscv/riscv.c

src/target/riscv/riscv.c
aap-sc: addressed (I should mention that I haven't touched the actual parsing function, though - hopefully it's not a big deal)

src/target/riscv/riscv.c
aap-sc: addressed, thanks

aap-sc:
@timsifive 

> Looks alright. Can you document this new command in doc/openocd.texi?

I've added a short section.

@TommyMurphyTM1234 and @timsifive 

> That sounds like a lot of work for registers that nobody might ever want to hide

yep. I would prefer not to touch that code. It's messy even in GDB.

timsifive (dismissed):
Looks good.

doc/openocd.texi
```diff
@@ -10553,6 +10553,23 @@ $_TARGETNAME expose_custom 32=myregister
...
+each CSR to the user, as they may not be relevant to the task at hand. For
+example, we may choose not to show trigger or PMU registers for simple
+debugging scenarios. This command allows to mark individual registers or
+register ranges (inclusive) as "hidden". Such hidden registers won't be
+displayed in GDB or @code{reg} command output.
+
+@example
+
+# Hide range of RISC-V CSRs
+# CSR_TSELECT - 1952 and CSR_DATA1 - 1953
```
JanMatCodasip: This seems to be a typo: DATA1 --> TDATA1

```suggestion
# CSR_TSELECT - 1952 and CSR_TDATA1 - 1953
```

src/target/riscv/riscv.c
```diff
@@ -3675,6 +3700,16 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
 			"etc. This must be executed before `init`."
 	},
+	{
+		.name = "hide_csrs",
+		.handler = riscv_hide_csrs,
+		.mode = COMMAND_CONFIG,
+		.usage = "{n0|n-m0}[,n1|n-m1]......",
+		.help = "Configure a list of inclusive ranges for CSRs to hide from gdb. "
+			"Hidden registers are still available, but are not listed in "
+			"gdb target description and `reg` command output."
```
JanMatCodasip: There is a very minor typo - missing space at the end:

```suggestion
			"gdb target description and `reg` command output. "
```

JanMatCodasip (dismissed):
This looks fine. I have noticed just two small typos.

src/target/riscv/riscv.c
aap-sc: thanks! Addressed.

doc/openocd.texi
aap-sc: addressed

timsifive (approved):
It's even better now. :-)

JanMatCodasip (approved):

PR #794, opened 2023-02-01
Fix opcode for the "fence" instruction

JanMatCodasip:
OpenOCD currently uses improper "fence" instruction: "FENCE" opcode with empty predecessor and successor sets.

Such instruction has no effect and is reserved for future use as a HINT instruction (RISC-V Unprivileged ISA spec V20191213, section 2.9).

This patch fixes it by using the proper "fence rw,rw" instruction.

Change-Id: Ia2a66059009153efef27279410850ddfd73dae38
Signed-off-by: Jan Matyas <jan.matyas@codasip.com>

### DIFF
--- a/src/target/riscv/opcodes.h
+++ b/src/target/riscv/opcodes.h
@@ -294,10 +294,11 @@ static uint32_t srli(unsigned int dest, unsigned int src, uint8_t shamt)
 	return inst_rs2(shamt) | inst_rs1(src) | inst_rd(dest) | MATCH_SRLI;
 }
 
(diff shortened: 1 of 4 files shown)


JanMatCodasip:
I have tested this by `riscv-tests/debug`. The only failing item was `EtriggerTest` on spike32_2 and spike64_2 which I assume is an unrelated issue. Other tests have passed.

timsifive (approved):
Looks good, thanks.

I regression tested this against HiFive Unleashed.

PR #796, opened 2023-02-08
Improve a debug log in freertos_update_threads()

Du-Chao:
To make the log more accurate and comprehensible. In case the offset is non-zero.

Signed-off-by: Chao Du <duchao@eswincomputing.com>

### DIFF
--- a/src/rtos/FreeRTOS.c
+++ b/src/rtos/FreeRTOS.c
@@ -590,7 +590,7 @@ static int freertos_update_threads(struct rtos *rtos)
 			return retval;
 		}
 		LOG_DEBUG("FreeRTOS: Read thread count for list %u at 0x%" PRIx64 ", value %" PRIu64,
-										i, list_of_lists[i], list_thread_count);
+								i, list_of_lists[i] + freertos->list_uxNumberOfItems_offset, list_thread_count);
 
 		if (list_thread_count == 0)
 			continue;


timsifive (commented):
This code is not RISC-V-specific, so this change really should be submitted upstream, as described in https://openocd.org/doc/doxygen/html/patchguide.html

There are already too many differences between the two code bases, so I really want to avoid more of those.

Du-Chao:
> This code is not RISC-V-specific, so this change really should be submitted upstream, as described in https://openocd.org/doc/doxygen/html/patchguide.html
> 
> There are already too many differences between the two code bases, so I really want to avoid more of those.

Thanks Tim for the comment.

I checked the upstream before I created this PR. This part is different between two code bases.
In riscv-openocd the _list_thread_count_ value is read by _freertos_read_struct_value_ which is introduced by riscv-openocd. The address is calculated by a base address and a offset.

This PR just make the debug log matching with the read value operation. Like other references.

timsifive:
There is a lot of difference, but I think this function (and especially these few lines of code) are essentially unchanged. I think https://sourceforge.net/p/openocd/code/ci/master/tree/src/rtos/FreeRTOS.c#l291 is identical to the lines you're changing here. Am I missing some detail?

timsifive:
The detail I'm missing you already pointed out to me. Sorry for not reading more carefully.

timsifive (approved):

PR #797, opened 2023-02-08
If XLEN=64 and vsew=64 fails, fall back to vsew=32.

timsifive:
This should make vector accesses work on 64-bit harts that implement Zve32*. There doesn't appear to be any way to easily determine what vsew values are allowed, so try and notice the failure.

### DIFF
--- a/src/target/riscv/riscv.h
+++ b/src/target/riscv/riscv.h
@@ -259,6 +259,8 @@ typedef struct {
 
 	/* Track when we were last asked to do something substantial. */
 	int64_t last_activity;
+
+	yes_no_maybe_t vsew64_supported;
 } riscv_info_t;
 
 COMMAND_HELPER(riscv_print_info_line, const char *section, const char *key,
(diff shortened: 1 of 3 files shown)


timsifive:
Independently confirmed that this works on hardware where this is the case.

src/target/riscv/riscv.h
```diff
@@ -253,6 +259,8 @@ typedef struct {
 
 	/* Track when we were last asked to do something substantial. */
 	int64_t last_activity;
+
+	yes_no_maybe_t vsew64_supported;
```
GregSavin, line 263: So this new data member doesn't need an initialization in riscv_info_init because the entire structure it's a part of gets allocated using calloc(), and YNM_MAYBE is the zero representation and what we want it initialized to, and that's what we get.  From a code readability standpoint, I wonder if it's worth initializing vsew64_supported in riscv_info_init, so there's no room for (human) misunderstanding the intent regarding the initial value.  (If OpenOCD community coding standards aren't against that sort of thing).

GregSavin (dismissed):
Looks good to me.

src/target/riscv/riscv.h
timsifive: You're right about the logic, and it does seem better to explicitly initialize anyway. Done.

GregSavin (approved):

PR #798, opened 2023-02-09
CSR_MCOUNTEREN should not exist if U-mode is not supported

aap-sc:
Change-Id: I1a2420fb88bd3ee37f6a539992e8dc119fdd6e0e


timsifive:
The privspec says:
> In systems without U-mode, the mcounteren register should not exist.

That sounds like it might exist, but is not recommended.

I'm OK with this change. If we encounter an implementation that needs access to mcounteren even without having U-mode, we can provide a mechanism to expose that register anyway.

timsifive (approved):

PR #799, opened 2023-02-14
Add `riscv icount` command.

timsifive:
Also refactor shared code for clearing itrigger/etrigger/icount.

Change-Id: Iac2e756332c89d2ed43435391e3c097abc825255

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10718,8 +10718,9 @@ Perform a 32-bit DMI write of value at address.
 
 @subsection RISC-V Trigger Commands
 
-The RISC-V Debug Specification defines several trigger types that don't map
-cleanly onto OpenOCD's notion of hardware breakpoints. These commands let you
+The RISC-V Debug Specification defines several optional trigger types that don't
+map cleanly onto OpenOCD's notion of hardware breakpoints. For the types that
+the target supports, these commands let you
 set those triggers directly. (It's also possible to do so by writing the
 appropriate CSRs.)
 
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10738,6 +10738,23 @@ For details on this trigger type, see the RISC-V Debug Specification.
 Clear the type 5 trigger that was set using @command{riscv etrigger set}.
 @end deffn
 
+@deffn {Command} {riscv icount set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] [@option{pending}] count
+Set an instruction count
+trigger (type 3) on the current target, which halts the target when it fires.
+@option{m}, @option{s}, @option{u}, @option{vs}, and @option{vu} control which
+execution modes the trigger fires in. If [@option{pending}] is passed then the
+pending bit is set, which is unlikely to be useful unless you're debugging the
+hardware implementation of this trigger.
+@var{count} sets the number of instructions to execute before the trigger is
+taken.
+
+For details on this trigger type, see the RISC-V Debug Specification.
+@end deffn
+
+@deffn {Command} {riscv icount clear}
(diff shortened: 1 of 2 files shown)


doc/openocd.texi
```diff
@@ -10738,6 +10738,23 @@ For details on this trigger type, see the RISC-V Debug Specification.
 Clear the type 5 trigger that was set using @command{riscv etrigger set}.
 @end deffn
 
+@deffn {Command} {riscv icount set} [@option{m}] [@option{s}] [@option{u}]
+[@option{vs}] [@option{vu}] [@option{pending}] count Set an instruction count
```
JanMatCodasip: There seems to be a newline typo:

```suggestion
@deffn {Command} {riscv icount set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] [@option{pending}] count 
Set an instruction count [....]
```

doc/openocd.texi
```diff
@@ -10738,6 +10738,23 @@ For details on this trigger type, see the RISC-V Debug Specification.
 Clear the type 5 trigger that was set using @command{riscv etrigger set}.
 @end deffn
```
JanMatCodasip, line 10741: I wonder if we should note in the doc that the triggers are an optional feature - making it clearer to the user that  the commands will not generally work on any RISC-V target.

For instance:
`These commands let you set those triggers directly.` ---> `The triggers are an optional feature. The following commands allow to set certain trigger types, provided they are supported by the given RISC-V target.`
(or similar wording)

src/target/riscv/riscv.c
```diff
@@ -3797,6 +3897,13 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.usage = "set [vs] [vu] [m] [s] [u] <exception_codes>|clear",
 		.help = "Set or clear a single exception trigger."
 	},
+	{
+		.name = "icount",
```
JanMatCodasip, line 3901: For clarity, I'd propose a more descriptive name:  `icount_trigger`

JanMatCodasip (commented):
I've checked this visually and it looks all right - thank you.

I have just three minor suggestions.

doc/openocd.texi
timsifive: Fixed.

doc/openocd.texi
timsifive: Good point. Added something like that.

src/target/riscv/riscv.c
timsifive: How strongly do you feel about that? `icount_trigger` is such a lot to type, and the other trigger types don't have _trigger appended because the names already contain `trigger`. (The debug spec naming is inconsistent in this regard.)

GregSavin (approved):
Looks good to me.

src/target/riscv/riscv.c
JanMatCodasip: I definitely do not feel strongly about this one.

Generally, I tend to prefer clarity over short names. If I imagine I don't know this command and need to study an OpenOCD's .cfg file written by someone else where it is used, I'd like to see there `riscv icount_trigger set <...>` rather than the shorter form (`riscv icount set <...>`). The former immediately tells me to what is going on and where to look for details (that is, the "Triggers" chapter in the RISC-V Debug Spec).

JanMatCodasip (approved):

src/target/riscv/riscv.c
timsifive: If you don't feel strongly, I'll just say that the naming is this way to be consistent with the debug spec and the existing itrigger and etrigger commands.

If someone out there wants to weigh in on Jan's side, I'll happily change the name.

Issue #788, opened 2023-01-13
[bug] looks like the current implementation of breakpoints/watchpoints is broken (for some HW implementations)


aap-sc:
In fact, this is what we observe on the current OpenOCD TOT with our (Syntacore) designs.

I've investigated the observed failures and I believe I understand the problem. Please, see below.

A while ago there was a change in how OpenOCD processes configuration of the triggers #725. More specifically, the region of interest is this commit: https://github.com/riscv/riscv-openocd/pull/725/commits/5a5c5e02ae6aa3c455cb7fe7881fa83e0ef4ba36 file riscv.c:627 and below.

Previously, we had the following code:

```
	for (i = 0; i < r->trigger_count; i++) {
		if (r->trigger_unique_id[i] != -1)
			continue;

		riscv_set_register(target, GDB_REGNO_TSELECT, i);

		uint64_t tdata1;
		int result = riscv_get_register(target, &tdata1, GDB_REGNO_TDATA1);
		if (result != ERROR_OK)
			return result;
		int type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));

		result = ERROR_OK;
		switch (type) {
			case 1:
				result = maybe_add_trigger_t1(target, trigger, tdata1);
				break;
			case 2:
				result = maybe_add_trigger_t2(target, trigger, tdata1);
				break;
			case 6:
				result = maybe_add_trigger_t6(target, trigger, tdata1);
				break;
			default:
				LOG_DEBUG("trigger %d has unknown type %d", i, type);
				continue;
		}
               // continued
```

That is we iterated over each available trigger `for (i = 0; i < r->trigger_count; i++) {` in an attempt to find one suitable for the requested configuration.

The loop was removed and now code looks as follows:

```
 do {                                                                                                         
   ret = maybe_add_trigger_t1(target, trigger);                                                               
   if (ret == ERROR_OK)                                                                                       
     break;                                                                                                   
   ret = maybe_add_trigger_t2(target, trigger);                                                               
   if (ret == ERROR_OK)                                                                                       
     break;                                                                                                   
   ret = maybe_add_trigger_t6(target, trigger);                                                               
   if (ret == ERROR_OK)                                                                                       
     break;                                                                                                   
  } while (0);
```

[...]

PR #800, opened 2023-02-15
Try all triggers in maybe_add_trigger_t*

en-sc:
It is possible for triggers of the same type to support different match field values, so it is needed to try all the triggers, not just the first one.

Fixes issue #788.

Signed-off-by: Evgeniy Naydanov evgeniy.naydanov@syntacore.com
Change-Id: I4c9fbc98bae7259377456d9ad8e770232724a592

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -28,6 +28,7 @@
 
 #define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
 #define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
+#define field_value(mask, val) set_field((riscv_reg_t) 0, mask, val)
 
 /*** JTAG registers. ***/
 
@@ -468,36 +469,37 @@ static void trigger_from_breakpoint(struct trigger *trigger,
 	trigger->unique_id = breakpoint->unique_id;
 }
 
-static bool can_use_napot_match(struct trigger *trigger, riscv_reg_t *tdata2)
+static bool can_use_napot_match(struct trigger *trigger)
 {
 	riscv_reg_t addr = trigger->address;
 	riscv_reg_t size = trigger->length;
 	bool sizePowerOf2 = (size & (size - 1)) == 0;
 	bool addrAligned = (addr & (size - 1)) == 0;
-	if (size > 1 && sizePowerOf2 && addrAligned) {
-		if (tdata2)
-			*tdata2 = addr | ((size - 1) >> 1);
-		return true;
-	}
-	return false;
+	return size > 1 && sizePowerOf2 && addrAligned;
 }
 
-static int find_trigger(struct target *target, int type, bool chained, unsigned int *idx)
+/* Find the next free trigger of the given type, without talking to the target. */
+static int find_next_free_trigger(struct target *target, int type, bool chained,
+		unsigned int *idx)
 {
+	assert(idx);
 	RISCV_INFO(r);
 
 	unsigned int num_found = 0;
 	unsigned int num_required = chained ? 2 : 1;
 
-	for (unsigned i = 0; i < r->trigger_count; i++) {
+	for (unsigned i = *idx; i < r->trigger_count; i++) {
 		if (r->trigger_unique_id[i] == -1) {
 			if (r->trigger_tinfo[i] & (1 << type)) {
 				num_found++;
-				bool done = (num_required == num_found);
-				if (done) {
+				if (num_required == num_found) {
 					/* Found num_required consecutive free triggers - success, done. */
-					if (idx)
-						*idx = i - (num_required - 1);
+					*idx = i - (num_required - 1);
+					LOG_TARGET_DEBUG(target,
+							"%d trigger(s) of type %d found on index %u, "
+							"chained == %s",
+							num_required, type, *idx,
+							chained ? "true" : "false");
 					return ERROR_OK;
 				}
 				/* Found a trigger but need more consecutive ones */
@@ -574,8 +576,8 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
 	const uint32_t bpcontrol_bpmatch = 0xf << 7;
 	const uint32_t bpcontrol_bpaction = 0xff << 11;
 
-	unsigned int idx;
-	ret = find_trigger(target, CSR_TDATA1_TYPE_LEGACY, false, &idx);
+	unsigned int idx = 0;
+	ret = find_next_free_trigger(target, CSR_TDATA1_TYPE_LEGACY, false, &idx);
 	if (ret != ERROR_OK)
 		return ret;
 
@@ -604,90 +606,227 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
 	return ERROR_OK;
 }
 
-static int maybe_add_trigger_t2(struct target *target, struct trigger *trigger)
-{
-	int ret;
-	riscv_reg_t tdata1, tdata2;
+struct trigger_request_info {
+	riscv_reg_t tdata1;
+	riscv_reg_t tdata2;
+	riscv_reg_t tdata1_ignore_mask;
+};
 
+static int try_setup_single_match_trigger(struct target *target,
+		struct trigger *trigger, struct trigger_request_info trig_info)
+{
+	int trigger_type =
+		get_field(trig_info.tdata1, CSR_MCONTROL_TYPE(riscv_xlen(target)));
+	int ret = ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
 	RISCV_INFO(r);
 
-	tdata1 = 0;
-	tdata1 = set_field(tdata1, CSR_MCONTROL_TYPE(riscv_xlen(target)), 2);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_DMODE(riscv_xlen(target)), 1);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_ACTION, CSR_MCONTROL_ACTION_DEBUG_MODE);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_M, 1);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_S, !!(r->misa & (1 << ('S' - 'A'))));
-	tdata1 = set_field(tdata1, CSR_MCONTROL_U, !!(r->misa & (1 << ('U' - 'A'))));
-	tdata1 = set_field(tdata1, CSR_MCONTROL_EXECUTE, trigger->execute);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_LOAD, trigger->read);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_STORE, trigger->write);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_SIZELO, CSR_MCONTROL_SIZELO_ANY & 3);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_SIZEHI, (CSR_MCONTROL_SIZELO_ANY >> 2) & 3);
-
-	if (trigger->execute || trigger->length == 1)
-		goto MATCH_EQUAL;
-	if (!can_use_napot_match(trigger, &tdata2))
-		goto MATCH_GE_LT;
-
-	unsigned int idx;
-	ret = find_trigger(target, CSR_TDATA1_TYPE_MCONTROL, false, &idx);
-	if (ret != ERROR_OK)
-		return ret;
-	tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_NAPOT);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_DISABLED);
-	ret = set_trigger(target, idx, tdata1, tdata2, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
-	if (ret != ERROR_OK) {
-		if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
-			goto MATCH_GE_LT; /* Fallback to chained MATCH using GT and LE */
-		return ret;
+	/* Find the first trigger, supporting required tdata1 value */
+	for (unsigned int idx = 0;
+			find_next_free_trigger(target, trigger_type, false, &idx) == ERROR_OK;
+			++idx) {
+		ret = set_trigger(target, idx, trig_info.tdata1, trig_info.tdata2,
+				trig_info.tdata1_ignore_mask);
+		if (ret == ERROR_OK) {
+			r->trigger_unique_id[idx] = trigger->unique_id;
+			return ERROR_OK;
+		}
+		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+			return ret;
 	}
-	r->trigger_unique_id[idx] = trigger->unique_id;
-	return ERROR_OK;
+	return ret;
+}
 
-MATCH_GE_LT:
-	ret = find_trigger(target, CSR_TDATA1_TYPE_MCONTROL, true, &idx);
-	if (ret != ERROR_OK)
-		return ret;
-	tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_GE);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_ENABLED);
-	tdata2 = trigger->address;
-	ret = set_trigger(target, idx, tdata1, tdata2, 0);
-	if (ret != ERROR_OK) {
-		if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
-			goto MATCH_EQUAL; /* Fallback to EQUAL MATCH */
-		return ret;
-	}
-	tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_LT);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_DISABLED);
-	tdata2 = trigger->address + trigger->length;
-	ret = set_trigger(target, idx + 1, tdata1, tdata2, 0);
-	if (ret != ERROR_OK) {
-		/* Undo the setting of the previous trigger*/
-		set_trigger(target, idx, 0, 0, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
-		if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
-			goto MATCH_EQUAL; /* Fallback to EQUAL MATCH */
-		return ret;
-	}
-	r->trigger_unique_id[idx] = trigger->unique_id;
-	r->trigger_unique_id[idx + 1] = trigger->unique_id;
-	return ERROR_OK;
+static int try_setup_chained_match_triggers(struct target *target,
+		struct trigger *trigger, struct trigger_request_info t1,
+		struct trigger_request_info t2)
+{
+	int trigger_type =
+		get_field(t1.tdata1, CSR_MCONTROL_TYPE(riscv_xlen(target)));
+	int ret = ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	RISCV_INFO(r);
 
-MATCH_EQUAL:
-	ret = find_trigger(target, CSR_TDATA1_TYPE_MCONTROL, false, &idx);
-	if (ret != ERROR_OK)
-		return ret;
-	tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_EQUAL);
-	if (trigger->length == 1) {
-		tdata1 = set_field(tdata1, CSR_MCONTROL_SIZELO, CSR_MCONTROL_SIZELO_8BIT & 3);
-		tdata1 = set_field(tdata1, CSR_MCONTROL_SIZEHI, (CSR_MCONTROL_SIZELO_8BIT >> 2) & 3);
+	/* Find the first 2 consecutive triggers, supporting required tdata1 values */
+	for (unsigned int idx = 0;
+			find_next_free_trigger(target, trigger_type, true, &idx) == ERROR_OK;
+			++idx) {
+		ret = set_trigger(target, idx, t1.tdata1, t1.tdata2,
+				t1.tdata1_ignore_mask);
+		if (ret != ERROR_OK)
+			continue;
+		ret = set_trigger(target, idx + 1, t2.tdata1, t2.tdata2,
+				t2.tdata1_ignore_mask);
+		if (ret == ERROR_OK) {
+			r->trigger_unique_id[idx] = trigger->unique_id;
+			r->trigger_unique_id[idx + 1] = trigger->unique_id;
+			return ERROR_OK;
+		}
+		/* Undo the setting of the previous trigger */
+		int ret_undo = set_trigger(target, idx, 0, 0, 0);
+		if (ret_undo != ERROR_OK)
+			return ret_undo;
+
+		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+			return ret;
 	}
-	tdata1 = set_field(tdata1, CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_DISABLED);
-	tdata2 = trigger->address;
-	ret = set_trigger(target, idx, tdata1, tdata2, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
-	if (ret != ERROR_OK)
-		return ret;
-	r->trigger_unique_id[idx] = trigger->unique_id;
-	return ERROR_OK;
+	return ret;
+}
+
+struct match_triggers_tdata1_fields {
+	riscv_reg_t common;
+	struct {
+		riscv_reg_t any;
+		riscv_reg_t s8bit;
+	} size;
+	struct {
+		riscv_reg_t enable;
+		riscv_reg_t disable;
+	} chain;
+	struct {
+		riscv_reg_t napot;
+		riscv_reg_t lt;
+		riscv_reg_t ge;
+		riscv_reg_t eq;
+	} match;
+	riscv_reg_t tdata1_ignore_mask;
+};
+
+static struct match_triggers_tdata1_fields fill_match_triggers_tdata1_fields_t2(
+		struct target *target, struct trigger *trigger)
+{
+	RISCV_INFO(r);
+
+	struct match_triggers_tdata1_fields result = {
+		.common =
+			field_value(CSR_MCONTROL_TYPE(riscv_xlen(target)), CSR_TDATA1_TYPE_MCONTROL) |
+			field_value(CSR_MCONTROL_DMODE(riscv_xlen(target)), 1) |
+			field_value(CSR_MCONTROL_ACTION, CSR_MCONTROL_ACTION_DEBUG_MODE) |
+			field_value(CSR_MCONTROL_M, 1) |
+			field_value(CSR_MCONTROL_S, !!(r->misa & BIT('S' - 'A'))) |
+			field_value(CSR_MCONTROL_U, !!(r->misa & BIT('U' - 'A'))) |
+			field_value(CSR_MCONTROL_EXECUTE, trigger->execute) |
+			field_value(CSR_MCONTROL_LOAD, trigger->read) |
+			field_value(CSR_MCONTROL_STORE, trigger->write),
+		.size = {
+			.any =
+				field_value(CSR_MCONTROL_SIZELO, CSR_MCONTROL_SIZELO_ANY & 3) |
+				field_value(CSR_MCONTROL_SIZEHI, (CSR_MCONTROL_SIZELO_ANY >> 2) & 3),
+			.s8bit =
+				field_value(CSR_MCONTROL_SIZELO, CSR_MCONTROL_SIZELO_8BIT & 3) |
+				field_value(CSR_MCONTROL_SIZEHI, (CSR_MCONTROL_SIZELO_8BIT >> 2) & 3)
+		},
+		.chain = {
+			.enable = field_value(CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_ENABLED),
+			.disable = field_value(CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_DISABLED)
+		},
+		.match = {
+			.napot = field_value(CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_NAPOT),
+			.lt = field_value(CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_LT),
+			.ge = field_value(CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_GE),
+			.eq = field_value(CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_EQUAL)
+		},
+		.tdata1_ignore_mask = CSR_MCONTROL_MASKMAX(riscv_xlen(target))
+	};
+	return result;
+}
+
+static struct match_triggers_tdata1_fields fill_match_triggers_tdata1_fields_t6(
+		struct target *target, struct trigger *trigger)
+{
+	RISCV_INFO(r);
+
+	struct match_triggers_tdata1_fields result = {
+		.common =
+			field_value(CSR_MCONTROL6_TYPE(riscv_xlen(target)), CSR_TDATA1_TYPE_MCONTROL6) |
+			field_value(CSR_MCONTROL6_DMODE(riscv_xlen(target)), 1) |
+			field_value(CSR_MCONTROL6_ACTION, CSR_MCONTROL_ACTION_DEBUG_MODE) |
+			field_value(CSR_MCONTROL6_M, 1) |
+			field_value(CSR_MCONTROL6_S, !!(r->misa & BIT('S' - 'A'))) |
+			field_value(CSR_MCONTROL6_U, !!(r->misa & BIT('U' - 'A'))) |
+			field_value(CSR_MCONTROL6_EXECUTE, trigger->execute) |
+			field_value(CSR_MCONTROL6_LOAD, trigger->read) |
+			field_value(CSR_MCONTROL6_STORE, trigger->write),
+		.size = {
+			.any = field_value(CSR_MCONTROL6_SIZE, CSR_MCONTROL6_SIZE_ANY),
+			.s8bit = field_value(CSR_MCONTROL6_SIZE, CSR_MCONTROL6_SIZE_8BIT)
+		},
+		.chain = {
+			.enable = field_value(CSR_MCONTROL6_CHAIN, CSR_MCONTROL6_CHAIN_ENABLED),
+			.disable = field_value(CSR_MCONTROL6_CHAIN, CSR_MCONTROL6_CHAIN_DISABLED)
+		},
+		.match = {
+			.napot = field_value(CSR_MCONTROL6_MATCH, CSR_MCONTROL6_MATCH_NAPOT),
+			.lt = field_value(CSR_MCONTROL6_MATCH, CSR_MCONTROL6_MATCH_LT),
+			.ge = field_value(CSR_MCONTROL6_MATCH, CSR_MCONTROL6_MATCH_GE),
+			.eq = field_value(CSR_MCONTROL6_MATCH, CSR_MCONTROL6_MATCH_EQUAL)
+		},
+		.tdata1_ignore_mask = 0
+	};
+	return result;
+}
+
+static int maybe_add_trigger_t2_t6(struct target *target,
+		struct trigger *trigger, struct match_triggers_tdata1_fields fields)
+{
+	int ret = ERROR_OK;
+
+	if (!trigger->execute && trigger->length > 1) {
+		/* Setting a load/store trigger ("watchpoint") on a range of addresses */
+
+		if (can_use_napot_match(trigger)) {
+			LOG_TARGET_DEBUG(target, "trying to setup NAPOT match trigger");
+			struct trigger_request_info napot = {
+				.tdata1 = fields.common | fields.size.any |
+					fields.chain.disable | fields.match.napot,
+				.tdata2 = trigger->address | ((trigger->length - 1) >> 1),
+				.tdata1_ignore_mask = fields.tdata1_ignore_mask
+			};
+			ret = try_setup_single_match_trigger(target, trigger, napot);
+			if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+				return ret;
+		}
+		LOG_TARGET_DEBUG(target, "trying to setup GE+LT chained match trigger pair");
+		struct trigger_request_info ge_1 = {
+			.tdata1 = fields.common | fields.size.any | fields.chain.enable |
+				fields.match.ge,
+			.tdata2 = trigger->address,
+			.tdata1_ignore_mask = fields.tdata1_ignore_mask
+		};
+		struct trigger_request_info lt_2 = {
+			.tdata1 = fields.common | fields.size.any | fields.chain.disable |
+				fields.match.lt,
+			.tdata2 = trigger->address + trigger->length,
+			.tdata1_ignore_mask = fields.tdata1_ignore_mask
+		};
+		ret = try_setup_chained_match_triggers(target, trigger, ge_1, lt_2);
+		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+			return ret;
+
+		LOG_TARGET_DEBUG(target, "trying to setup LT+GE chained match trigger pair");
+		struct trigger_request_info lt_1 = {
+			.tdata1 = fields.common | fields.size.any | fields.chain.enable |
+				fields.match.lt,
+			.tdata2 = trigger->address,
+			.tdata1_ignore_mask = fields.tdata1_ignore_mask
+		};
+		struct trigger_request_info ge_2 = {
+			.tdata1 = fields.common | fields.size.any | fields.chain.disable |
+				fields.match.ge,
+			.tdata2 = trigger->address + trigger->length,
+			.tdata1_ignore_mask = fields.tdata1_ignore_mask
+		};
+		ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
+		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+			return ret;
+	}
+	LOG_TARGET_DEBUG(target, "trying to setup equality match trigger");
+	struct trigger_request_info eq = {
+		.tdata1 = fields.common |
+			(trigger->length == 1 ? fields.size.s8bit : fields.size.any) |
+			fields.chain.disable | fields.match.eq,
+		.tdata2 = trigger->address,
+		.tdata1_ignore_mask = fields.tdata1_ignore_mask
+	};
+	return try_setup_single_match_trigger(target, trigger, eq);
 }
 
 static int maybe_add_trigger_t3(struct target *target, bool vs, bool vu,
@@ -711,8 +850,8 @@ static int maybe_add_trigger_t3(struct target *target, bool vs, bool vu,
 	tdata1 = set_field(tdata1, CSR_ICOUNT_U, u);
 	tdata1 = set_field(tdata1, CSR_ICOUNT_COUNT, count);
 
-	unsigned int idx;
-	ret = find_trigger(target, CSR_TDATA1_TYPE_ICOUNT, false, &idx);
+	unsigned int idx = 0;
+	ret = find_next_free_trigger(target, CSR_TDATA1_TYPE_ICOUNT, false, &idx);
 	if (ret != ERROR_OK)
 		return ret;
 	ret = set_trigger(target, idx, tdata1, 0, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
@@ -744,8 +883,8 @@ static int maybe_add_trigger_t4(struct target *target, bool vs, bool vu,
 
 	tdata2 = interrupts;
 
-	unsigned int idx;
-	ret = find_trigger(target, CSR_TDATA1_TYPE_ITRIGGER, false, &idx);
+	unsigned int idx = 0;
+	ret = find_next_free_trigger(target, CSR_TDATA1_TYPE_ITRIGGER, false, &idx);
 	if (ret != ERROR_OK)
 		return ret;
 	ret = set_trigger(target, idx, tdata1, tdata2, 0);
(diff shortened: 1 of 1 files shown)


aap-sc:
@en-sc , looks like formatting is messed up.

src/target/riscv/riscv.c
```diff
@@ -506,6 +506,14 @@ static int find_trigger(struct target *target, int type, bool chained, unsigned
...
 }
 
+static int find_trigger(struct target *target, unsigned int type, bool chained,
+						unsigned int *idx)
+{
+	if (idx)
+		*idx = 0;
+	return find_next_trigger(target, type, chained, idx);
+}
+
```
timsifive: This feels like unnecessary complexity. find_trigger() is always called with a non-zero idx pointer. It is better to modify all the call sites to force the initial value to 0 than adding this tiny function.

src/target/riscv/riscv.c
```diff
@@ -598,90 +606,174 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
-	riscv_reg_t tdata1, tdata2;
+struct trigger_request_info {
+	riscv_reg_t tdata1;
+	riscv_reg_t tdata2;
+	riscv_reg_t tdata1_ignore_mask;
+};
 
+static int try_trigger_request_match(struct target *target,
+									 struct trigger *trigger, int trigger_type,
+									 struct trigger_request_info trig_info)
```
timsifive: Excessive indent here and elsewhere.

src/target/riscv/riscv.c
```diff
@@ -598,90 +606,174 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
-	tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_NAPOT);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_DISABLED);
-	ret = set_trigger(target, idx, tdata1, tdata2, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
-	if (ret != ERROR_OK) {
-		if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
-			goto MATCH_GE_LT; /* Fallback to chained MATCH using GT and LE */
-		return ret;
+	for (unsigned int idx = 0;
+		 find_next_trigger(target, trigger_type, false, &idx) == ERROR_OK &&
+		 ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
```
timsifive: I much prefer a simpler loop, where you do these calls inside the loop and break/continue as necessary.

timsifive (changes requested):
Can you rename the PR to make it explicit that you're only changing maybe_add_trigger_t2 and maybe_add_trigger_t2?

Haven't read it all in detail yet. I'll do another pass later.

timsifive:
What have you done to test these changes?

src/target/riscv/riscv.c
JanMatCodasip: Hint: The rest of the OpenOCD code typically uses 2 tabs when breaking statements or function declarations to more lines.

Example: https://github.com/riscv/riscv-openocd/blob/riscv/src/target/target.c#L637

src/target/riscv/riscv.c
```diff
@@ -492,6 +488,10 @@ static int find_trigger(struct target *target, int type, bool chained, unsigned
 					/* Found num_required consecutive free triggers - success, done. */
 					if (idx)
 						*idx = i - (num_required - 1);
+					LOG_TARGET_DEBUG(
+						target,
+						"Trigger of type %d found on index %u, chained == %d",
+						type, i - (num_required - 1), chained);
```
JanMatCodasip: Minor thing: Here you can print `chained == true|false` instead of `1|0` to make it clearer what it really means.

```suggestion
					LOG_TARGET_DEBUG(
						target,
						"Trigger of type %d found on index %u, chained == %s",
						type, i - (num_required - 1), chained ? "true" : "false");
```

src/target/riscv/riscv.c
```diff
@@ -598,90 +606,174 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+							  !!(r->misa & (1 << ('U' - 'A'))));
+	common.tdata1 =
+		set_field(common.tdata1, CSR_MCONTROL_EXECUTE, trigger->execute);
+	common.tdata1 = set_field(common.tdata1, CSR_MCONTROL_LOAD, trigger->read);
+	common.tdata1 =
+		set_field(common.tdata1, CSR_MCONTROL_STORE, trigger->write);
+	common.tdata1 = set_field(common.tdata1, CSR_MCONTROL_SIZELO,
+							  CSR_MCONTROL_SIZELO_ANY & 3);
+	common.tdata1 = set_field(common.tdata1, CSR_MCONTROL_SIZEHI,
+							  (CSR_MCONTROL_SIZELO_ANY >> 2) & 3);
```
JanMatCodasip: Please, reformat the code so that each assignment is on a single line - to make it easy to read.

If the line lengths would exceed the maximum, consider using just a simple variable `tdata1` and create the `common` structure later, like this:

```
riscv_reg_t tdata1 = 0;
tdata1 = set_field(tdata1, /* ... */ , /* ... */);
tdata1 = set_field(tdata1, /* ... */ , /* ... */);
/* ... */
struct trigger_request_info common = {
		.tdata1 = tdata1, 
		.tdata2 = trigger->address, 
		.tdata1_ignore_mask = 0};
```

src/target/riscv/riscv.c
```diff
@@ -598,90 +606,174 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+	eq.tdata1 =
+		set_field(eq.tdata1, CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_DISABLED);
 	if (trigger->length == 1) {
-		tdata1 = set_field(tdata1, CSR_MCONTROL_SIZELO, CSR_MCONTROL_SIZELO_8BIT & 3);
-		tdata1 = set_field(tdata1, CSR_MCONTROL_SIZEHI, (CSR_MCONTROL_SIZELO_8BIT >> 2) & 3);
+		eq.tdata1 = set_field(eq.tdata1, CSR_MCONTROL_SIZELO,
+							  CSR_MCONTROL_SIZELO_8BIT & 3);
+		eq.tdata1 = set_field(eq.tdata1, CSR_MCONTROL_SIZEHI,
+							  (CSR_MCONTROL_SIZELO_8BIT >> 2) & 3);
 	}
```
JanMatCodasip: Same as above - please find a way to put the `set_field` assignments onto single lines to make it readable.

src/target/riscv/riscv.c
```diff
@@ -598,90 +606,174 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
-	ret = find_trigger(target, CSR_TDATA1_TYPE_MCONTROL, false, &idx);
-	if (ret != ERROR_OK)
-		return ret;
-	tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_EQUAL);
+static int maybe_add_trigger_t2(struct target *target, struct trigger *trigger)
+{
+	RISCV_INFO(r);
+
+	struct trigger_request_info common = {
+		.tdata1 = 0, .tdata2 = trigger->address, .tdata1_ignore_mask = 0};
```
JanMatCodasip: Please, format the code like this - each assignment on a separate line:

```suggestion
	struct trigger_request_info common = {
		.tdata1 = 0,
		.tdata2 = trigger->address,
		.tdata1_ignore_mask = 0};
```

src/target/riscv/riscv.c
```diff
@@ -598,90 +606,174 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
-{
-	int ret;
-	riscv_reg_t tdata1, tdata2;
+struct trigger_request_info {
+	riscv_reg_t tdata1;
+	riscv_reg_t tdata2;
+	riscv_reg_t tdata1_ignore_mask;
+};
 
+static int try_trigger_request_match(struct target *target,
```
JanMatCodasip: I'd recommend a more descriptive name, e.g. `try_setup_single_match_trigger`.

src/target/riscv/riscv.c
```diff
@@ -598,90 +606,174 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+			continue;
+		ret = set_trigger(target, idx + 1, trig2_info.tdata1, trig2_info.tdata2,
+						  trig2_info.tdata1_ignore_mask);
+		if (ret == ERROR_OK) {
+			r->trigger_unique_id[idx] = trigger->unique_id;
+			r->trigger_unique_id[idx + 1] = trigger->unique_id;
+			return ERROR_OK;
+		} else
+			set_trigger(target, idx, 0, 0,
+						0); /* Undo the setting of the previous trigger*/
```
JanMatCodasip: You can reformat it like this:

```suggestion
			/* Undo the setting of the previous trigger */
			set_trigger(target, idx, 0, 0, 0); 
```

src/target/riscv/riscv.c
```diff
@@ -598,90 +606,174 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
-	tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_GE);
-	tdata1 = set_field(tdata1, CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_ENABLED);
-	tdata2 = trigger->address;
-	ret = set_trigger(target, idx, tdata1, tdata2, 0);
-	if (ret != ERROR_OK) {
-		if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
-			goto MATCH_EQUAL; /* Fallback to EQUAL MATCH */
-		return ret;
+static int
+try_chained_trigger_request_match(struct target *target,
```
JanMatCodasip: I'd recommend to rename it for clarity - similary to the above. 

E.g.: `try_setup_chained_match_triggers`

src/target/riscv/riscv.c
```diff
@@ -598,90 +606,174 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+			return ret;
 	}
-	r->trigger_unique_id[idx] = trigger->unique_id;
-	r->trigger_unique_id[idx + 1] = trigger->unique_id;
-	return ERROR_OK;
+	LOG_TARGET_DEBUG(target, "trying t request EQUAL match");
+	return try_trigger_request_match(target, trigger, trigger_type,
+									 trig_eq_info);
+}
```
JanMatCodasip: I recommend to remove this function completely and put its logic directly to `maybe_add_trigger_t2` and `maybe_add_trigger_t6`. The reason is that now all the possible trigger combinations are always created and passed to this function, even if they do not make sense for the given scenario.

For instance, if we have an execute=1 trigger, all the `napot_trigger_info`, `trig1_ge_info` ... structures are created, passed to this function but are irrelevant - and that decision is made only at the very end of the chain. This makes it harder to whoever reads the code to understand it properly.

Once this function is deleted, the `maybe_add_trigger_tX()` function will look like this (pseudocode):

```c
int maybe_add_trigger_tX( /*...*/ )
{
    /* Prepare the common trigger structure */
    /* ... */

    /* Prepare the exact equality match trigger */
    /* trig_eq_info */

    if (execute==1 || size == 1) {
        /* HW watchpoint of size=1 or HW breakpoint. Use simply an exact-match trigger. */
        return try_setup_single_match_trigger( /*...*/ );
    }

    /* Otherwise this is a watchpoint of size > 1. 
     * Try multiple trigger variants to cover this case. */

    /* Find out if we can use NAPOT match */
    if (can_use_napot_match( /* ... */ )) {
        /* Prepare napot_trigger_info */
        /* ... */

        ret = try_setup_single_match_trigger( /* ... */ );
        if (ret == ERROR_OK)
            return ERROR_OK;  /* Done - success */
    }

    /* NAPOT not possible or not supported by the target */

    /* Try a pair of chained triggers GE + LT */
    /* Prepare trig1_ge_info */
    /* ... */
    /* Prepare trig2_lt_info */
    /* ... */
    ret = try_setup_chained_match_triggers( /*...*/ );
    if (ret == ERROR_OK)
        return ERROR_OK;

    /* Try a pair of chained triggers LT + GE */
    /* Prepare trig1_lt_info */
    /* ... */
    /* Prepare trig2_ge_info */
    /* ... */
    ret = try_setup_chained_match_triggers( /*...*/ );
    if (ret == ERROR_OK)
        return ERROR_OK; 

    /* Last fallback: Try at least an exact match against the beginning of the range */
    return try_setup_single_match_trigger( trigger_eq_info );
}
```

JanMatCodasip (changes requested):
@en-sc Thank you for submitting this change. I like the idea that multiple trigger variants are tried to cover watchpoints of size > 1 until we find something that the target actually supports.

My comments are related to:
- 1) structure of the code (I'd recommend to remove function `try_trigger_request_match_with_fallbacks` and move the logic elsewhere - see the respective comment)
- 2) naming of functions
- 3) formatting of the code (line breaks, indents, etc.)

For now, I've checked the code only visually. Later when the first review round is addressed, I'd like to test the changes locally.

src/target/riscv/riscv.c
en-sc: Thank you for the suggestion. As I see it there are two approaches to rewriting the loop:

```
for (unsigned int idx = 0; ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE; ++idx) {
    if (find_next_trigger(target, trigger_type, false, &idx) != ERROR_OK)
        break;
    /* loop body */
```

There are several reasons why I do not like this approach:

1. The initialization and mutation of `idx` are in the loop's header, but it is unclear on first look that the loop is finite (loop condition does not check the index).

2. The mutations of `idx` are further separated, so one can assume that the loop checks all `idx` values in order, which is untrue.

The other approach:

```

int ret = ERROR_OK;
for (unsigned int idx = 0; find_next_trigger(target, trigger_type, false, &idx) != ERROR_OK; ++idx) {
    /* loop body */
    if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
        break;
}

```

I find that this is the better method since I only see one small drawback: it is not clear at first glance that the loop is here to find the first trigger that can be used, not to simply iterate over triggers of the same type. It is also somewhat better than the initial version since the initialization of `ret` is more straightforward.

Moreover, in both approaches I fail to see in which way the loop is actually simplified. Minor changes to the condition trade off major changes to the body.

However, if you insist, I'm ready to do as you suggested

src/target/riscv/riscv.c
en-sc: Thank you for the suggestion.

 I would like to preserve the function in question. In my opinion, it is preferable to minimize code repetition, especially if the code has non-trivial control flow. Moving control flow into a separate function simplifies debugging, modification, and makes it clear what exactly is the difference between _t2 and _t6 variants.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

en-sc:
> What have you done to test these changes?

I've run tests from `riscv-tests/debug` on spike, as well as testing on the setup mentioned in #788.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: Addressed. In some places, the 80-column limit is exceeded. Hope it is not crucial.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
timsifive: Your "other approach" looks good to me.

timsifive:
> I've run tests from riscv-tests/debug on spike, as well as testing on the setup mentioned in https://github.com/riscv/riscv-openocd/issues/788.

Sounds good, thanks.

src/target/riscv/riscv.c
```diff
@@ -468,36 +468,37 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
 				if (done) {
 					/* Found num_required consecutive free triggers - success, done. */
-					if (idx)
-						*idx = i - (num_required - 1);
+					*idx = i - (num_required - 1);
+					LOG_TARGET_DEBUG(target,
+							"Trigger of type %d found on index %u, "
+							"chained == %s",
+							type, i - (num_required - 1),
+							chained ? "true" : "false");
```
```suggestion
					LOG_TARGET_DEBUG(target,
							"%d trigger(s) of type %d found on index %u, "
							"chained == %s",
							num_required, type, i - (num_required - 1),
							chained ? "true" : "false");
```

src/target/riscv/riscv.c
```diff
@@ -604,90 +605,146 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+			return ERROR_OK;
+		} else
+			/* Undo the setting of the previous trigger */
+			set_trigger(target, idx, 0, 0, 0);
+	}
+	return ret;
+}
+
+static int setup_match_trigger(struct target *target,
+		struct trigger *trigger, int trigger_type,
```
timsifive: You shouldn't need trigger_type as an argument here (and in try_setup_single_match_trigger and try_setup_chained_match_triggers), because it's already encoded in tdata1 of every trigger_request_info.

src/target/riscv/riscv.c
```diff
@@ -468,36 +468,37 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
-		if (tdata2)
-			*tdata2 = addr | ((size - 1) >> 1);
-		return true;
-	}
-	return false;
+	return size > 1 && sizePowerOf2 && addrAligned;
 }
 
-static int find_trigger(struct target *target, int type, bool chained, unsigned int *idx)
+static int find_next_trigger(struct target *target, int type, bool chained,
```
```suggestion
/* Find the next unused trigger of the given type, without talking to the target. */
static int find_next_trigger(struct target *target, int type, bool chained,
```

src/target/riscv/riscv.c
```diff
@@ -604,90 +605,146 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+
+static int setup_match_trigger(struct target *target,
+		struct trigger *trigger, int trigger_type,
+		struct trigger_request_info t_napot, struct trigger_request_info t1_ge,
+		struct trigger_request_info t2_lt, struct trigger_request_info t1_lt,
+		struct trigger_request_info t2_ge, struct trigger_request_info t_eq)
+{
+	int ret = ERROR_OK;
+
+	if (!(trigger->execute || trigger->length == 1)) {
```
```suggestion
	if (!trigger->execute && trigger->length != 1) {
```

src/target/riscv/riscv.c
timsifive: An alternate option is to make a struct that contains all the relevant bitmasks for MCONTROL and MCONTROL6, then pass that struct to a single function that performs all the logic.

src/target/riscv/riscv.c
```diff
@@ -604,90 +605,146 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+		if (ret != ERROR_OK)
+			continue;
+		ret = set_trigger(target, idx + 1, t2.tdata1, t2.tdata2, t2.imask);
+		if (ret == ERROR_OK) {
+			r->trigger_unique_id[idx] = trigger->unique_id;
+			r->trigger_unique_id[idx + 1] = trigger->unique_id;
+			return ERROR_OK;
+		} else
+			/* Undo the setting of the previous trigger */
+			set_trigger(target, idx, 0, 0, 0);
```
JanMatCodasip: Detail: If the `checkpatch` script won't complain about the braces `{` `}` here, I'd add them.

```suggestion
		} else {
			/* Undo the setting of the previous trigger */
			set_trigger(target, idx, 0, 0, 0);
		}			
```

src/target/riscv/riscv.c
JanMatCodasip: I also like the "other approach".

> I only see one small drawback: it is not clear at first glance that the loop is here to find the first trigger that can be used, not to simply iterate over triggers of the same type.

You can add a comment above the loop, describing this: `/* Find the first trigger of given type that supports what we need */` (or similar wording).

src/target/riscv/riscv.c
JanMatCodasip: If the `checkpatch` script doesn't complain, than it is fine.

src/target/riscv/riscv.c
JanMatCodasip: This can even be called `find_next_unused_trigger()` or `find_next_free_trigger()`, if you'd like.

src/target/riscv/riscv.c
JanMatCodasip: > I would like to preserve the function in question. In my opinion, it is preferable to minimize code repetition, especially if the code has non-trivial control flow. Moving control flow into a separate function simplifies debugging, modification, and makes it clear what exactly is the difference between _t2 and _t6 variants.

Thank you for your reply. I agree with your arguments but I'd like that to be achieved by a different code structure. My updated proposal, that takes into account your points, would be:

- Have a single `maybe_add_trigger_t2_t6()` function that has the structure shown above - this would contain the control flow and would cover triggers of both type=2 and type=6
- Have a helper function(s) that fill the data structures (`napot_trigger_info`, `trig1_ge_info` ...) depending on the trigger type but do not contain any control flow. This would be called from `maybe_add_trigger_t2_t6()`.

This will avoid code repetition, separate control flow and data, and make the code easier to grasp by showing the control flow first ("what happens and in what sequence") and filling the data structures will be secondary ("how that happens internally").

src/target/riscv/riscv.c
JanMatCodasip: (Or even more idiomatic: `trigger->length > 1`)

Anyway, I'd recommend to move the control flow code (sequence of actions attempted) into the upper-level function, as discussed [in another comment](https://github.com/riscv/riscv-openocd/pull/800#discussion_r1109459134), not to bury it into the last function in the chain.

JanMatCodasip (changes requested):

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: There are no complaints from `checkpath`

src/target/riscv/riscv.c
en-sc: Addressed. I've reconfigured the code in such a way, that all the modifications to `trigger_request_info`, necessary for `maybe_add_trigger_t2_t6` are represented in `match_triggers_field_info`.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: Addressed. There is actually no need for `else`, because inside the `if` there is a `return`

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,349 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+}
+
+static int maybe_add_trigger_t2_t6(struct target *target,
+		struct trigger *trigger, struct match_triggers_field_info set)
+{
+	int ret = ERROR_OK;
+
+	if (!trigger->execute && trigger->length > 1) {
+		if (can_use_napot_match(trigger)) {
+			LOG_TARGET_DEBUG(target, "trying t request NAPOT match");
```
JanMatCodasip: If you'd like:

```suggestion
	if (!trigger->execute && trigger->length > 1) {
		/* Setting a load/store trigger ("watchpoint") on a range of addresses */

		if (can_use_napot_match(trigger)) {
			LOG_TARGET_DEBUG(target, "trying t request NAPOT match");
```

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,349 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+		},
+		.tdata1_ignore_mask = 0
+	};
+	return result;
+}
+
+static int maybe_add_trigger_t2_t6(struct target *target,
+		struct trigger *trigger, struct match_triggers_field_info set)
+{
+	int ret = ERROR_OK;
```
JanMatCodasip: Could you please simplify it further by passing just the trigger type:

```suggestion
static int maybe_add_trigger_t2_t6(struct target *target,
		struct trigger *trigger, unsigned int trigger_type)
{
	assert(trigger_type == 2 || trigger_type == 6);
	int ret = ERROR_OK;
```

The `trigger_type` argument can be an unsigned int or an enum.

en-sc:
Did you have a chance to take a look at the last version, @timsifive, @JanMatCodasip?

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,349 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+		riscv_reg_t napot;
+		riscv_reg_t lt;
+		riscv_reg_t ge;
+		riscv_reg_t eq;
+	} match;
+	riscv_reg_t tdata1_ignore_mask;
+};
+
+static riscv_reg_t
+set_common_t2(struct target *target, struct trigger *trigger)
```
JanMatCodasip: This can be called `fill_common_tdata1_t2` (or similar).

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,349 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+	tdata1 = set_field(tdata1, CSR_MCONTROL_S, !!(r->misa & BIT('S' - 'A')));
+	tdata1 = set_field(tdata1, CSR_MCONTROL_U, !!(r->misa & BIT('U' - 'A')));
 	tdata1 = set_field(tdata1, CSR_MCONTROL_EXECUTE, trigger->execute);
 	tdata1 = set_field(tdata1, CSR_MCONTROL_LOAD, trigger->read);
 	tdata1 = set_field(tdata1, CSR_MCONTROL_STORE, trigger->write);
+	return tdata1;
+}
+
+static riscv_reg_t
+set_common_t6(struct target *target, struct trigger *trigger)
```
JanMatCodasip: This can be called `fill_common_tdata1_t6` (or similar).

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,349 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+	tdata1 = set_field(tdata1, CSR_MCONTROL6_CHAIN, CSR_MCONTROL6_CHAIN_DISABLED);
+	return tdata1;
+}
+
+static riscv_reg_t set_match_napot_t2(void)
+{
+	riscv_reg_t tdata1 = 0;
+	tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_NAPOT);
+	return tdata1;
+}
```
JanMatCodasip: All these simple 3-line functions can be inlined directly into the proposed `fill_trigger_request_info_*` functions. It would simply get reduced to a single line, for example: `tdata1 = set_field(tdata1, CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_NAPOT);`

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,349 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+
+	if (!trigger->execute && trigger->length > 1) {
+		if (can_use_napot_match(trigger)) {
+			LOG_TARGET_DEBUG(target, "trying t request NAPOT match");
+			struct trigger_request_info napot = {
+				.tdata1 = set.common | set.size.any | set.chain.disable |
+					set.match.napot,
+				.tdata2 = trigger->address | ((trigger->length - 1) >> 1),
+				.tdata1_ignore_mask = set.tdata1_ignore_mask
+			};
```
JanMatCodasip: When `trigger_type` is passed to `maybe_add_trigger_t2_t6()`, as suggested in another comment, the creation of structures like `napot`, `ge_1`, `lt_1` ... can be simplified this way:

```suggestion
			struct trigger_request_info napot = fill_trigger_request_info_napot(
					trigger_type, false, trigger->address | ((trigger->length - 1) >> 1));
```

Function `fill_trigger_request_info_napot()` (and `*_gt()`, `*_lt()`, `*_eq()`) can look like this (pseudocode):

```
struct trigger_request_info fill_trigger_request_info_napot(unsigned int trigger_type, struct trigger *trigger, bool is_chained, riscv_reg_t tdata2)
{
	riscv_reg_t tdata1;
	riscv_reg_t tdata1_ignore_mask;	
	if (trigger_type == 2) {
		tdata1 = fill_common_tdata1_t2(trigger);
		/* ... */
	} else {
		tdata1 = fill_common_tdata1_t6(trigger);
		/* ... */	
	}
	struct trigger_request_info result = {
		.tdata1 = tdata1, .tdata2 = tdata2, .tdata1_ignore_mask = tdata1_ignore_mask
	};
	return result;
}
```

The relatively complex (nested) `struct match_triggers_field_info` will then not be needed.

JanMatCodasip (changes requested):
Thank you for addressing the previous comments. I have few additional suggestions how to reduce the number of code lines and functions, and also the complex data type `struct match_triggers_field_info`.

Other than that, the code looks fine.

src/target/riscv/riscv.c
JanMatCodasip: Thank you for the changes. I have submitted one last round of suggestions how to simplify the code even further.

src/target/riscv/riscv.c
en-sc: Thank you for the suggestion. I would like to keep the current approach. Now there is no `if`s dependent on trigger type since it is known on the call site `maybe_add_trigger_t2_t6`. If I were to follow your suggestion, it would create many `if`s with the same condition, so, in my opinion, it would complicate control flow and the code would become more error-prone.

Now:
```
static int maybe_add_trigger_t2_t6(struct target *target,
        struct trigger *trigger, struct match_triggers_field_info set)
{
   /*...*/
    if (!trigger->execute && trigger->length > 1) {
         /* Prepare corresponding structure */
         ret = try_setup_single_match_trigger(target, trigger, napot);
            return ret;
        /* Prepare corresponding structures */
        ret = try_setup_chained_match_triggers(target, trigger, ge_1, lt_2);
            return ret;
        /* Prepare corresponding structures */
        ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
            return ret;
    }
    /* Prepare corresponding structure */
    return try_setup_single_match_trigger(target, trigger, eq);
}
```

Your suggestion:
```
struct trigger_request_info fill_trigger_request_info_napot(unsigned int trigger_type, ...)
{
	if (trigger_type == 2) {
		/* ... */
	} else {
		/* ... */	
	}
       /*...*/
}

struct trigger_request_info fill_trigger_request_info_ge(unsigned int trigger_type, ...)
{
	if (trigger_type == 2) {
		/* ... */
	} else {
		/* ... */	
	}
       /*...*/
}

struct trigger_request_info fill_trigger_request_info_lt(unsigned int trigger_type, ...)
{
	if (trigger_type == 2) {
		/* ... */
	} else {
		/* ... */	
	}
       /*...*/
}

struct trigger_request_info fill_trigger_request_info_eq(unsigned int trigger_type, ...)
{
	if (trigger_type == 2) {
		/* ... */
	} else {
		/* ... */	
	}
       /*...*/
}

static int maybe_add_trigger_t2_t6(struct target *target, struct trigger *trigger, unsigned int trigger_type)
{
   /*...*/
    if (!trigger->execute && trigger->length > 1) {
           /* ...*/
            ret = try_setup_single_match_trigger(target, trigger, fill_trigger_request_info_napot(trigger_type, ...));
           /*...*/
        }
        /*...*/
        ret = try_setup_chained_match_triggers(target, trigger, fill_trigger_request_info_ge(trigger_type, /*chained = */ true, ...), fill_trigger_request_info_lt(trigger_type, /*chained = */ false, ...));
        /*...*/
        ret = try_setup_chained_match_triggers(target, trigger, fill_trigger_request_info_lt(trigger_type, /*chained = */ true, ...), fill_trigger_request_info_ge(trigger_type, /*chained = */ false, ...));
        /*...*/
    return try_setup_single_match_trigger(target, trigger, fill_trigger_request_info_eq(trigger_type, ...));
}
```

src/target/riscv/riscv.c

src/target/riscv/riscv.c
JanMatCodasip: Hi @en-sc, thank you for replying.

I am afraid the code comparison above is not fair. That's because the code titled `Now:` does not show the complex structure of `struct match_triggers_field_info` data type nor the code needed to fill that structure.

Having the `if (trigger_type == 2) {...} else {...}` conditions inside the `fill_trigger_request_info_*` functions looks all right to me, I do not have any concern about complexity. In fact such functions will be straightforward to read, debug and would even be easily unit-testable. 

Also the last piece of code above can be made nicer by getting rid of the nested calls to `fill_trigger_request_info_*`. Instead, the results can be stored in variables:

```
        // Original:
        ret = try_setup_chained_match_triggers(target, trigger, fill_trigger_request_info_ge(trigger_type, /*chained = */ true, ...), fill_trigger_request_info_lt(trigger_type, /*chained = */ false, ...));
        
        // Suggested:
        struct trigger_request_info ge_1 = fill_trigger_request_info_ge(trigger_type, /*chained = */ true, ...);
        struct trigger_request_info lt_2 = fill_trigger_request_info_lt(trigger_type, /*chained = */ false, ...);
        ret = try_setup_chained_match_triggers(target, trigger, ge_1, lt_2);        
```

Please, if you don't have strong feeling about keeping `struct match_triggers_field_info` in the code, it would be my recommendation to simplify the code as suggested in this thread.

src/target/riscv/riscv.c
```diff
@@ -468,36 +468,38 @@ static void trigger_from_breakpoint(struct trigger *trigger,
...
 	unsigned int num_found = 0;
 	unsigned int num_required = chained ? 2 : 1;
 
-	for (unsigned i = 0; i < r->trigger_count; i++) {
+	for (unsigned i = *idx; i < r->trigger_count; i++) {
 		if (r->trigger_unique_id[i] == -1) {
 			if (r->trigger_tinfo[i] & (1 << type)) {
 				num_found++;
 				bool done = (num_required == num_found);
 				if (done) {
```
```suggestion
				if (num_required == num_found) {
```

src/target/riscv/riscv.c
timsifive: Agreed. I appreciate that you're trying to avoid duplicating code and using symbolic names, but it feels unnecessarily complex to make so many tiny functions.

src/target/riscv/riscv.c
timsifive: I think I like where @en-sc is going with this. The complicated data structure only contains constants. It could be flattened but that wouldn't really change much.
If all these set_match_ge_t2() were inlined then I think it would be fairly compact and simple as well.

timsifive (changes requested):

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
```diff
@@ -604,90 +605,277 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+			.disable = chain_disable
+		},
+		.match = {
+			.napot = match_napot,
+			.lt = match_lt,
+			.ge = match_ge,
+			.eq = match_eq
+		},
+		.tdata1_ignore_mask = CSR_MCONTROL_MASKMAX(riscv_xlen(target))
+	};
```
timsifive: I'd just inline everything here. The extra variables add a lot of lines, and don't improve understanding. So:
```
struct match_triggers_tdata1_info result = {
   .size = {
      .any = ((CSR_MCONTROL_SIZELO_ANY & 3) << CSR_MCONTROL_SIZELO_OFFSET) |
         (((CSR_MCONTROL_SIZELO_ANY >> 2) & 3) << CSR_MCONTROL_SIZEHI_OFFSET,
      },
      .chain = {
         .enable = set_field(0, CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_DISABLED)
      }
   }
```
etc.

There's no need to use set_field() for everything when other use of the macros is equally clear.

If you really want you create a `struct match_triggers_tdata1_info` and 0 it. Then use set_field macros directly on the struct fields instead of having all the unnecessary extra variables.

timsifive (changes requested):
This looks close to me.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
JanMatCodasip: Even though I'd prefer to pass around just the trigger type (and not a structure that tells how to set up a particular trigger), the current code looks all right. It is definitely great improvement over the previous state and also over the first revision of this merge request.

Please feel free to move on, I'm giving it a green light :)

And thanks for the effort invested into this so far!

src/target/riscv/riscv.c
```diff
@@ -28,6 +28,14 @@
...
 #define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
 #define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
+#define set_2_fields(reg, m1, v1, ...) set_field(set_field(reg, m1, v1), __VA_ARGS__)
+#define set_3_fields(reg, m1, v1, ...) set_2_fields(set_field(reg, m1, v1), __VA_ARGS__)
+#define set_4_fields(reg, m1, v1, ...) set_3_fields(set_field(reg, m1, v1), __VA_ARGS__)
+#define set_5_fields(reg, m1, v1, ...) set_4_fields(set_field(reg, m1, v1), __VA_ARGS__)
+#define set_6_fields(reg, m1, v1, ...) set_5_fields(set_field(reg, m1, v1), __VA_ARGS__)
+#define set_7_fields(reg, m1, v1, ...) set_6_fields(set_field(reg, m1, v1), __VA_ARGS__)
+#define set_8_fields(reg, m1, v1, ...) set_7_fields(set_field(reg, m1, v1), __VA_ARGS__)
+#define set_9_fields(reg, m1, v1, ...) set_8_fields(set_field(reg, m1, v1), __VA_ARGS__)
```
timsifive: How about:
```
#define field_value(mask, val) set_field(0, mask, val)
```

Then use that everywhere with | in between, instead of the set_N_fields macros. In the places where you use them you still don't care about the initial value, and it seems awkward to count the number of arguments and counting all the commas in a macro with 19 arguments.

It's fine to acknowledge that these macros are setting bit fields. We don't have to hide the fact that these can be combined with bit-wise OR inside a macro, because that will never change.

timsifive (changes requested):
I'm really sorry to push back yet again, but this is introducing a whole new (creative) way of specifying constants.

src/target/riscv/riscv.c

timsifive (dismissed):

timsifive:
This looks good to me. @JanMatCodasip, can you approve your requested changes (or explain what's still missing)?

JanMatCodasip:
> can you approve your requested changes (or explain what's still missing)?

Overall, the changes look fine, I don't expect to post any more review finding. Please let me do one last overall review pass tomorrow before the final approval.

src/target/riscv/riscv.c
JanMatCodasip: Comment no longer relevant - marking as resolved.

src/target/riscv/riscv.c
```diff
@@ -28,6 +28,7 @@
 
 #define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
 #define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
+#define field_value(mask, val) set_field(0, mask, val)
```
JanMatCodasip: Here I'd recommend `0u` (unsigned). This is for safety - to avoid potentially undefined behavior when someone uses `<<` or `>>` operators on the result.

`#define field_value(mask, val) set_field(0u, mask, val)`

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,227 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+			.ge = field_value(CSR_MCONTROL6_MATCH, CSR_MCONTROL6_MATCH_GE),
+			.eq = field_value(CSR_MCONTROL6_MATCH, CSR_MCONTROL6_MATCH_EQUAL)
+		},
+		.tdata1_ignore_mask = 0
+	};
+	return result;
+}
+
+static int maybe_add_trigger_t2_t6(struct target *target,
+		struct trigger *trigger, struct match_triggers_tdata1_info set)
```
```suggestion
		struct trigger *trigger, struct match_trigger_tdata1_masks masks)
```

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,227 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+{
+	int ret = ERROR_OK;
+
+	if (!trigger->execute && trigger->length > 1) {
+		/* Setting a load/store trigger ("watchpoint") on a range of addresses */
+
+		if (can_use_napot_match(trigger)) {
+			LOG_TARGET_DEBUG(target, "trying t request NAPOT match");
```
```suggestion
			LOG_TARGET_DEBUG(target, "trying to setup NAPOT match trigger");
```

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,227 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+				.tdata1 = set.common | set.size.any | set.chain.disable |
+					set.match.napot,
+				.tdata2 = trigger->address | ((trigger->length - 1) >> 1),
+				.tdata1_ignore_mask = set.tdata1_ignore_mask
+			};
+			ret = try_setup_single_match_trigger(target, trigger, napot);
+			if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+				return ret;
+		}
+		LOG_TARGET_DEBUG(target, "trying t request GE/LT match");
```
```suggestion
		LOG_TARGET_DEBUG(target, "trying to setup GE+LT chained trigger pair");
```

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,227 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+			.tdata1 = set.common | set.size.any | set.chain.disable |
+				set.match.lt,
+			.tdata2 = trigger->address + trigger->length,
+			.tdata1_ignore_mask = set.tdata1_ignore_mask
+		};
+		ret = try_setup_chained_match_triggers(target, trigger, ge_1, lt_2);
+		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+			return ret;
+
+		LOG_TARGET_DEBUG(target, "trying t request LT/GE match");
```
```suggestion
		LOG_TARGET_DEBUG(target, "trying to setup LT+GE chained trigger pair");
```

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,227 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+			.tdata1 = set.common | set.size.any | set.chain.disable |
+				set.match.ge,
+			.tdata2 = trigger->address + trigger->length,
+			.tdata1_ignore_mask = set.tdata1_ignore_mask
+		};
+		ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
+		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+			return ret;
+	}
+	LOG_TARGET_DEBUG(target, "trying t request EQUAL match");
```
```suggestion
	LOG_TARGET_DEBUG(target, "trying to setup equality match trigger");
```

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,227 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
-	tdata2 = trigger->address;
-	ret = set_trigger(target, idx, tdata1, tdata2, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
-	if (ret != ERROR_OK)
-		return ret;
-	r->trigger_unique_id[idx] = trigger->unique_id;
-	return ERROR_OK;
+	return ret;
+}
+
+struct match_triggers_tdata1_info {
```
JanMatCodasip: For clarity, I recommend to use the word `masks` (bit masks) instead of just `info`.

```suggestion
struct match_trigger_tdata1_masks {
```

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,227 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+	struct {
+		riscv_reg_t napot;
+		riscv_reg_t lt;
+		riscv_reg_t ge;
+		riscv_reg_t eq;
+	} match;
+	riscv_reg_t tdata1_ignore_mask;
+};
+
+static struct match_triggers_tdata1_info fill_match_triggers_field_info_t2(
```
```suggestion
static struct match_triggers_tdata1_info fill_match_trigger_masks_t2(
```

src/target/riscv/riscv.c
```diff
@@ -604,90 +606,227 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
...
+			.lt = field_value(CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_LT),
+			.ge = field_value(CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_GE),
+			.eq = field_value(CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_EQUAL)
+		},
+		.tdata1_ignore_mask = CSR_MCONTROL_MASKMAX(riscv_xlen(target))
+	};
+	return result;
+}
+
+static struct match_triggers_tdata1_info fill_match_triggers_field_info_t6(
```
```suggestion
static struct match_triggers_tdata1_info fill_match_triggers_masks_t6(
```

JanMatCodasip (changes requested):
The code looks all right. 

I have just very last round of review suggestions. Those are just minor items - mostly naming and debug prints.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: Thank you for the suggestion! I do agree that the name of this parameter is quite cryptic, however, in my opinion, the meaning behind it is quite clear when you consider its use. I would refrain from renaming it to `masks`, since as I understand the fields inside are not masks, but rather offsetted field values.

src/target/riscv/riscv.c
en-sc: (Commented above)

src/target/riscv/riscv.c
en-sc: (Commented above)

src/target/riscv/riscv.c
en-sc: (Commented above)

src/target/riscv/riscv.c
timsifive: Agree that these are not masks. How about `_constants`? `_definition`?

en-sc: I have another suggestion: `match_triggers_tdata1_descriptor`?

JanMatCodasip: You're right, they are not `masks`, thanks for pointing it out.

In that case I'd recommend to call it `fields` or leave it as is.

JanMatCodasip: Thank you.

JanMatCodasip (approved):

PR #801, opened 2023-02-17
Set the current_thread when no FreeRTOS task was created.

Du-Chao:
The 'current thread' was not set when no FreeRTOS task created. Which could lead a wrongly invoking of freertos_get_thread_reg_list.

Change-Id: I0e0f8327080ef698d7ed4aae5ac2a630d532ddeb

### DIFF
--- a/src/rtos/FreeRTOS.c
+++ b/src/rtos/FreeRTOS.c
@@ -501,6 +501,7 @@ static int freertos_update_threads(struct rtos *rtos)
 			LOG_ERROR("Error allocating memory for %" PRIu64 " threads", thread_list_size);
 			return ERROR_FAIL;
 		}
+		rtos->current_thread = 1;
 		rtos->thread_details->threadid = 1;
 		rtos->thread_details->exists = true;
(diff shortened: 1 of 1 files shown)


timsifive (approved):
Did you run the tests in https://github.com/riscv-software-src/riscv-tests/tree/master/debug?

Du-Chao:
> Did you run the tests in https://github.com/riscv-software-src/riscv-tests/tree/master/debug?

Not yet. I will check how to run the tests and then update.

Du-Chao:
The result of 'make all' under https://github.com/riscv-software-src/riscv-tests/tree/master/debug :

Total: 426
pass: 347
not_applicable: 79

6 FreeRtosTest: 2 pass, 4 not_applicable.

But I see some openocd Error logs even in the pass log, not sure if it is normal.

timsifive:
Those results are good. Thanks.

There are some OpenOCD error messages. I'm pretty sure you'll at least get one complaining about IDCODE. But if the test passes that's all we need.

PR #802, opened 2023-02-17
Smoke test OpenOCD against spike.

timsifive:
Change-Id: I17321d70e2b54086e8f3fbb01744746633d7a119

### DIFF
--- a/.github/workflows/spike-openocd-tests.yml
+++ b/.github/workflows/spike-openocd-tests.yml
@@ -1,0 +1,76 @@
+# Build spike and run a couple of debug tests.
+
+name: Test OpenOCD against 2 spike configurations
+
+on: pull_request
+
+# There is some commented out code below that would be useful in adding this
+# workflow to other repos. Ideally we can come up with something that would
+# leave this file almost identical between repos, so they can all easily run
+# this test suite.
+
+jobs:
+  test:
+    name: Test debug (Ubuntu)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Dependencies
+        run: |
+          sudo apt install build-essential device-tree-compiler
+
+#      - name: Download OpenOCD
+#        run: |
+#          git clone --recurse-submodules https://github.com/riscv/riscv-openocd.git
+#          cd riscv-openocd
+#          git checkout 43ea20dfbb6c815004a51106a3b2009d7f6c4940
+
+      - name: Build OpenOCD
+        run: |
+          #cd riscv-openocd
+          ./bootstrap
+          ./configure
+          make -j"$(nproc 2> /dev/null || sysctl -n hw.ncpu)"
+          ls -l src/openocd
+
+      - name: Download Toolchain
+        run: wget --progress=dot:giga https://github.com/xpack-dev-tools/riscv-none-elf-gcc-xpack/releases/download/v12.2.0-1/xpack-riscv-none-elf-gcc-12.2.0-1-linux-x64.tar.gz
+
+      - name: Install Toolchain
+        run: tar zxf xpack-riscv-none-elf-gcc-12.2.0-1-linux-x64.tar.gz
+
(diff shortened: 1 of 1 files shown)


JanMatCodasip (commented):
The new workflow looks fine. Thank you for putting your time into this.

I like how compact it is - just one workflow file that is easy to read (compared to what I attempted earlier).

I have created a branch on top of yours at https://github.com/riscv/riscv-openocd/pull/803 - to prototype few additional ideas (most notably, cache for the dependencies). Please feel free to check it out.

How stable are `riscv-tests/debug` these days? Are there any tests that are known to fail sporadically?

.github/workflows/spike-openocd-tests.yml
```diff
@@ -0,0 +1,76 @@
+# Build spike and run a couple of debug tests.
+
+name: Test OpenOCD against 2 spike configurations
+
+on: pull_request
+
+# There is some commented out code below that would be useful in adding this
+# workflow to other repos. Ideally we can come up with something that would
+# leave this file almost identical between repos, so they can all easily run
+# this test suite.
```
JanMatCodasip, line 10: I thought that other forks of riscv-openocd could use the workflow file as is - without changes. What would be the commented pieces of code for?

GregSavin (approved):
I'm not (yet) familiar with the continuous integration stuff on github, but just the general concept here looks awesome.

.github/workflows/spike-openocd-tests.yml
timsifive: To deal with source showing up in different directories, and also to optionally check out a specific git commit instead of just the latest. Compare this PR with https://github.com/riscv-software-src/riscv-isa-sim/blob/master/.github/workflows/debug-smoke.yml

timsifive:
> How stable are riscv-tests/debug these days? Are there any tests that are known to fail sporadically?

As far as I know it's rock solid now.

timsifive:
@JanMatCodasip I took a quick look at your branch. Looks like you've got several promising improvements, but let's get this in use and improve it later instead of trying to make it perfect first.

.github/workflows/spike-openocd-tests.yml
JanMatCodasip: Ok, I see, so this is not about forks of riscv-openocd but about different repos altogether (riscv-isa-sim).

Related question: 
How would you like to manage the versions of the dependencies? 
- The [debug-smoke.yml](https://github.com/riscv-software-src/riscv-isa-sim/blob/master/.github/workflows/debug-smoke.yml) workflow in spike repo (linked above) uses fixed revisions of riscv-tests and openocd. 
- On the other hand, this OpenOCD workflow takes master branches of the dependencies. I understand that you may want to leave it like that on purpose, but then it's necessary to accept the risk of tests suddenly getting broken just because of changes in those external repos.

Regarding the code reuse in other repos: 
Since the workflow file is very short and easy to maintain as is, I would not worry about trying to share pieces of it across different repos. It seems not worth the effort.

JanMatCodasip:
> let's get this in use and improve it later instead of trying to make it perfect first

I agree. The current initial version looks fine. I'll submit my improvements later in a separate merge request.

JanMatCodasip (approved):

.github/workflows/spike-openocd-tests.yml
timsifive: My workflow has always attempted to ensure that mainline spike, OpenOCD and riscv-tests always pass. Generally the process is:
1. Add feature to spike.
2. Add feature to OpenOCD.
3. Add test to riscv-tests

I would rather know that we're always testing the latest stuff and sometimes deal with a problem resulting from that, than not be testing the latest stuff (and frequently forgetting/having to update the hashes in the workflow).

JanMatCodasip: Ok, understood, thank you for explanation.

PR #804, opened 2023-02-23
Properly populate the FreeRTOS type sizes and offsets.

Du-Chao:
1. According to FreeRTOS/portable/GCC/RISC-V/portmacro.h, align _TickType_t_ with UBASE_TYPE.
2. Make the alignment of _ListItem_t_ clear.

### DIFF
--- a/src/rtos/FreeRTOS.c
+++ b/src/rtos/FreeRTOS.c
@@ -59,6 +59,8 @@ struct FreeRTOS {
 	unsigned ubasetype_size;
 	/* sizeof(void *) */
 	unsigned pointer_size;
+	/* sizeof(TickType_t) */
+	unsigned ticktype_size;
 	unsigned list_width;
 	unsigned list_item_width;
 	unsigned list_elem_next_offset;
@@ -305,13 +307,12 @@ static unsigned populate_offset_size(struct FreeRTOS *freertos,
 				align = freertos->pointer_size;
 				break;
 			case TYPE_TICKTYPE:
-				/* Could be either 16 or 32 bits, depending on configUSE_16_BIT_TICKS. */
-				info[i].size = 4;
-				align = 4;
+				info[i].size = freertos->ticktype_size;
+				align = freertos->ticktype_size;
 				break;
 			case TYPE_LIST_ITEM:
 				info[i].size = freertos->list_item_width;
-				align = MAX(freertos->ubasetype_size, freertos->pointer_size);
+				align = MAX(freertos->ticktype_size, freertos->pointer_size);
 				break;
 			case TYPE_CHAR_ARRAY:
 				/* size is set by the caller. */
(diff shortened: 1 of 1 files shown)


src/rtos/FreeRTOS.c
```diff
@@ -352,6 +353,12 @@ static void freertos_compute_offsets(struct rtos *rtos)
 
 	freertos->pointer_size = DIV_ROUND_UP(target_address_bits(rtos->target), 8);
 	freertos->ubasetype_size = DIV_ROUND_UP(target_data_bits(rtos->target), 8);
+	/*
+	 * According to FreeRTOS/portable/GCC/RISC-V/portmacro.h,
+	 * the TickType_t is defined as UBASE_TYPE,
+	 * Not controlled by configUSE_16_BIT_TICKS.
+	 */
+	freertos->ticktype_size = freertos->ubasetype_size;
```
timsifive, line 361: If ticktype_size is always set to ubasetype_size, what's the benefit of having ticktype_size defined in the struct?

src/rtos/FreeRTOS.c
```diff
@@ -352,6 +353,12 @@ static void freertos_compute_offsets(struct rtos *rtos)
 
 	freertos->pointer_size = DIV_ROUND_UP(target_address_bits(rtos->target), 8);
 	freertos->ubasetype_size = DIV_ROUND_UP(target_data_bits(rtos->target), 8);
+	/*
+	 * According to FreeRTOS/portable/GCC/RISC-V/portmacro.h,
```
timsifive, line 357: What about other ports? This code should work on non-RISC-V targets as well.

src/rtos/FreeRTOS.c
Du-Chao: The benefit is to make the calculation more clear and comprehensible.

src/rtos/FreeRTOS.c
Du-Chao: I thought that _riscv-openocd_ is only for RISC-V port, so this PR only considered the RISC-V case.

Du-Chao:
Thanks Tim for the comment.

I will close this PR and then try to prepare another patch which is suitable for all targets.

src/rtos/FreeRTOS.c
TommyMurphyTM1234: > I thought that _riscv-openocd_ is only for RISC-V port, so this PR only considered the RISC-V case.

RISC-V specific changes, in the `src/target/riscv` folder and above (the latter potentially affecting other/all target architectures) in this repo, are eventually upstreamed to the master OpenOCD project.

Du-Chao: > 

I see. Thanks for the explanation.
BTW, I created another PR #806 which may benefit all targets.

Du-Chao: > > I thought that _riscv-openocd_ is only for RISC-V port, so this PR only considered the RISC-V case.
> 
> RISC-V specific changes, in the `src/target/riscv` folder and above (the latter potentially affecting other/all target architectures) in this repo, are eventually upstreamed to the master OpenOCD project.

I see. Thanks for the explanation.
BTW, I created another PR https://github.com/riscv/riscv-openocd/pull/806 which may benefit all targets.

PR #806, opened 2023-03-01
Calculate the FreeRTOS type sizes and offsets more adaptively.

Du-Chao:
The definition of TickType_t varies between different targets. And it is also related to configUSE_16_BIT_TICKS option. Thus introduce a new command to make sure we are using a correct ticktype size.

### DIFF
--- a/src/rtos/FreeRTOS.c
+++ b/src/rtos/FreeRTOS.c
@@ -59,6 +59,8 @@ struct FreeRTOS {
 	unsigned ubasetype_size;
 	/* sizeof(void *) */
 	unsigned pointer_size;
+	/* sizeof(TickType_t) */
+	unsigned ticktype_size;
 	unsigned list_width;
 	unsigned list_item_width;
 	unsigned list_elem_next_offset;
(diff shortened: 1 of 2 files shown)


src/rtos/FreeRTOS.c
```diff
@@ -127,6 +129,45 @@ static int nds32_stacking(struct rtos *rtos, const struct rtos_register_stacking
...
+			break;
+		default:
+			LOG_ERROR("Invalid ticktype size. Should be 2, 4 or 8.");
+			return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	return ERROR_OK;
+}
+
+static const struct command_registration freetros_commands[] = {
```
timsifive:
```suggestion
static const struct command_registration freertos_commands[] = {
```

src/rtos/FreeRTOS.c
```diff
@@ -127,6 +129,45 @@ static int nds32_stacking(struct rtos *rtos, const struct rtos_register_stacking
...
+	{
+		.name = "freertos_ticktype_size",
+		.handler = handle_freertos_ticktype_size,
+		.mode = COMMAND_ANY,
+		.usage = "<2|4|8>",
+		.help = "Pass the size (in bytes) of TickType_t to OpenOCD. To make sure the "
+		"calculation of offsets and sizes is correct. Defaults to 4."
```
timsifive: Please document this command in `doc/openocd.texi` as well.

src/rtos/FreeRTOS.c
```diff
@@ -127,6 +129,45 @@ static int nds32_stacking(struct rtos *rtos, const struct rtos_register_stacking
 	return ERROR_OK;
 }
 
+/* take 4 bytes (32 bits) as the default size,
+ * which is suitable for most 32-bit targets and
+ * configuration of configUSE_16_BIT_TICKS = 0. */
+static unsigned freertos_ticktype_size = 4;
```
timsifive:
```suggestion
static unsigned int freertos_ticktype_size = 4;
```
OpenOCD style requires this. I realize that this file uses plain `unsigned` all over the place. Some day I'll have to fix that, but for now I'm trying to at least have new code conform to the new style.

src/rtos/FreeRTOS.c
```diff
@@ -127,6 +129,45 @@ static int nds32_stacking(struct rtos *rtos, const struct rtos_register_stacking
...
+COMMAND_HANDLER(handle_freertos_ticktype_size)
+{
+	if (CMD_ARGC != 1) {
+		LOG_ERROR("Command takes exactly 1 parameter");
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	unsigned size;
```
timsifive:
```suggestion
	unsigned int size;
```

timsifive (changes requested):
I like adding the command.

Du-Chao:
Thanks Tim for the review and comment.

I will commit a new patch later. (for the doc part, I need to check where is the right place to insert this new command.)

timsifive (approved):

PR #809, opened 2023-03-06
Multiple improvements to Spike smoke-test workflow

JanMatCodasip:
Change-Id: I5b8e9200c5d8cbaa3116bac565e009089bb6b36b

### DIFF
--- a/.github/workflows/spike-openocd-tests.yml
+++ b/.github/workflows/spike-openocd-tests.yml
@@ -1,6 +1,13 @@
-# Build spike and run a couple of debug tests.
+# Build Spike and run a couple of debug tests.
 
 name: Test OpenOCD against 2 spike configurations
+
+env:
+  SPIKE_REPO: https://github.com/riscv-software-src/riscv-isa-sim.git
+  SPIKE_REV: master
+  RISCV_TESTS_REPO: https://github.com/riscv-software-src/riscv-tests.git
+  RISCV_TESTS_REV: master
+  TOOLCHAIN_URL: https://github.com/xpack-dev-tools/riscv-none-elf-gcc-xpack/releases/download/v12.2.0-1/xpack-riscv-none-elf-gcc-12.2.0-1-linux-x64.tar.gz
 
 on: pull_request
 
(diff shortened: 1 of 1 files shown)


JanMatCodasip:
The following updates were made:

- use explicit runner name (ubuntu-22.04 instead of ubuntu-latest)
- use checkout action v3 (not v2) - silences a Github warning
- cache dependencies to speed-up the process (Spike + toolchain) 
- reordered actions so that cached items are handled first
- moved dependencies to /opt/riscv for easier caching
- archive logs from the tests (downloadable artifact)
- more descriptive names for some steps
- changed 'apt' to 'apt-get' to supress a warning

JanMatCodasip:
The cache of Spike + RISC-V toolchain reduces the runtime by approx. 40 % (18m 11s --> 11m 13s).

timsifive:
> The following updates were made:

Can you put that list in the change description itself?

.github/workflows/spike-openocd-tests.yml
```diff
@@ -12,65 +19,99 @@ on: pull_request
 jobs:
   test:
     name: Test debug (Ubuntu)
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
timsifive: I prefer latest. If we run against latest and it breaks, we can quickly fix it by changing back to the previous version. If the fixed version stops working because github removes it (as just happened for ubuntu-18.04) then the only way to quick fix is to move to a newer version and hope that it works. If it doesn't we have to make the test work with newer ubuntu right now.

timsifive (dismissed):
Looks good. It's great that the spike build can be cached.
Thanks for these improvements.

.github/workflows/spike-openocd-tests.yml
JanMatCodasip: I usually prefer to be explicit about requirements/dependencies, that's why I proposed `ubuntu-22.04` (and `ubuntu-latest` is at the moment just an alias to `ubuntu-22.04`). But I do not feel stronly about this one  - I reverted it to `ubuntu-latest`.

JanMatCodasip:
> > The following updates were made

> Can you put that list in the change description itself?

Good point. I have squashed the commits and updated the commit message to include the list of changes.

.github/workflows/spike-openocd-tests.yml
```diff
@@ -14,63 +21,97 @@ jobs:
...
+          echo "RISC_V_TESTS_COMMIT=$RISC_V_TESTS_COMMIT" >> $GITHUB_ENV
 
-      - name: Download Toolchain
-        run: wget --progress=dot:giga https://github.com/xpack-dev-tools/riscv-none-elf-gcc-xpack/releases/download/v12.2.0-1/xpack-riscv-none-elf-gcc-12.2.0-1-linux-x64.tar.gz
+      - name: Get dependencies from cache
+        id: cache-deps
+        uses: actions/cache@v3
+        with:
+          path: /opt/riscv
+          key: "spike-${{env.SPIKE_COMMIT}}-toolchain-${{env.TOOLCHAIN_URL}}"
```
timsifive: Can you turn this into 2 caches: one for spike, and one for the toolchain?

The toolchain is quite large but only changes when we update this file. spike changes all the time. It seems better not to store a new copy of the toolchain just because spike changes.

timsifive (changes requested):

.github/workflows/spike-openocd-tests.yml
JanMatCodasip: That's a good suggestion - I will do that. (Just a note: It may take me a week or so to get back to this.)

timsifive (approved):
Looks good. Thanks!

PR #812, opened 2023-03-12
target/riscv: fix the bug of using S2 register in read_memory_progbuf

XuHangHub:
We should avoid using x16~x31 register in program buffer because there are no such general purpose registers in RVE(Embedded) extension. For targets that support rvE, when the parameter increment=0 and count>1 of the read_memory_progbuf function, openocd will cause an error due to the use of the s2 register.
For example:
{Command} {riscv repeat_read} count address [size=4]

Change-Id: I8b74dcc15cd00a400f2f1354c577a82132394435

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3197,7 +3197,7 @@ static int read_memory_progbuf_inner(struct target *target, target_addr_t addres
 		return result;
 
 	if (increment == 0 &&
-			register_write_direct(target, GDB_REGNO_S2, 0) != ERROR_OK)
+			register_write_direct(target, GDB_REGNO_A0, 0) != ERROR_OK)
 		return ERROR_FAIL;
 
 	uint32_t command = access_register_command(target, GDB_REGNO_S1,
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -3197,7 +3197,7 @@ static int read_memory_progbuf_inner(struct target *target, target_addr_t addres
 		return result;
 
 	if (increment == 0 &&
-			register_write_direct(target, GDB_REGNO_S2, 0) != ERROR_OK)
+			register_write_direct(target, GDB_REGNO_A0, 0) != ERROR_OK)
```
JanMatCodasip, line 3200: Just a detail concerning formatting: Could this expression be on a single line? Would it pass through `checkpatch`?

JanMatCodasip (approved):
Checked visually - this looks fine. 

@XuHangHub - Thank you for the fix.

@timsifive - IIUC, this functionality is tested by `RepeatReadTest` but I'd assuem it is not run on RV32E regularly. Would you like to include that configuration to the CI?

src/target/riscv/riscv-013.c
XuHangHub: @JanMatCodasip Yes, it passed the checkpath, there are many similar practices in the source code, probably to avoid lines of code that are longer than 72-80 column.

timsifive (approved):
Looks good. Thanks.

timsifive:
@JanMatCodasip It would be cool to test against RV32E, but it would take a bunch of work. I've never run these tests against RV32E, and there are a number of tests that access all GPRs without any check whether that means 16 or 32 of them.

JanMatCodasip:
> It would be cool to test against RV32E, but it would take a bunch of work. I've never run these tests against RV32E, and there are a number of tests that access all GPRs without any check whether that means 16 or 32 of them.

Understood. Thank you for explanation.

PR #815, opened 2023-03-16
target/riscv: Expose S?aia CSRs if they're on the target.

timsifive:
Untested, because I don't have a target that implements this.


GregSavin (approved):
Looks good to me (took a quick look at the spec)

JanMatCodasip (approved):
Quick-checked visually - looks all right.

PR #816, opened 2023-03-17
Merge up to commit '1293ddd65713d6551775b67169387622ada477c1' from upstream

### LINKED ISSUE #814: OPENOCD command: read_memory $address 64 $count, displaying 32-bit values with count=1
Hi, 
   Whenever I try to use read_memory 64, I only get 32 bit output (data1) on my terminal for both abstract and sysbus access. This issue is not present while reading 64 bit using mdd.

image

timsifive:
This includes https://sourceforge.net/p/openocd/mailman/message/37710818/, which should fix #814.

No regressions against spike, HiFive Unleashed, and HiFive1.


timsifive:
checkpatch fails because a bunch of the upstream code doesn't pass it. My plan is to ignore the code style check and merge this anyway. But not until someone reviews it.

GregSavin (dismissed):
Lots of license boilerplate cleanup.

TommyMurphyTM1234:
> checkpatch fails because a bunch of the upstream code doesn't pass it. My plan is to ignore the code style check and merge this anyway. But not until someone reviews it.

Hi @timsifive - thanks for updating the upstream merge. Apologies if it's a dumb question, but what exactly needs to be reviewed? I presume not all ~ 200 commits that are part of this PR, but, rather, something more specific?

JanMatCodasip (commented):
I'd be happy to take a look but will be able to get to this only in few days (estimated: Fri/Mon). I'll leave this up to you whether you'd be willing to wait a moment prior to merging.

timsifive:
> Hi @timsifive - thanks for updating the upstream merge. Apologies if it's a dumb question, but what exactly needs to be reviewed? I presume not all ~ 200 commits that are part of this PR, but, rather, something more specific?

@TommyMurphyTM1234 This is not a dumb question. I honestly don't really know. My experience has been that when people review code, even these crazy merges, sometimes they find bugs. So I stick with the process. :-)

@JanMatCodasip I can certainly wait a few days for your review.

TommyMurphyTM1234:
Hi @timsifive - since my first question above wasn't dumb after all, I'll try again. :-D

> checkpatch fails because a bunch of the upstream code doesn't pass it. 

Is this because upstream does not actually enforce checkpatch passing before accepting patches and (often?) accepts patches even if they fail the check?

* https://github.com/openocd-org/openocd/blob/b6b4f9d46a48aadc1de6bb5152ff4913661c9059/HACKING#L239

timsifive:
>> checkpatch fails because a bunch of the upstream code doesn't pass it.

> Is this because upstream does not actually enforce checkpatch passing before accepting patches and (often?) accepts patches even if they fail the check?

I have never dug into why upstream code does not satisfy checkpatch. All I know is that we're getting errors that would have been obvious to anyone using the tool, in code that was not modified in this branch. This is not a new thing.

erhankur:
Several checkpatch rules were added just before 0.12.0-rc1 tag. Now errors coming because the new checkpatch runs against the old codes.  e.g. I see a lot of xtensa related errors that were fine during the review stage.

TommyMurphyTM1234:
> Several checkpatch rules were added just before 0.12.0-rc1 tag. Now errors coming because the new checkpatch runs against the old codes. e.g. I see a lot of xtensa related errors that were fine during the review stage.

Thanks @erhankur - that's interesting and could explain some (or all?) of the discrepancies alright.

erhankur:
Instead of disabling the checkpatch for the next PRs, I'd suggest getting the commit count in the branch and running git diff against just PRs commit. Something like; 
```
cnt=$(git rev-list --count HEAD ^origin/riscv)
git diff HEAD~${cnt} \
 filterdiff \
 -x "a/src/jtag/drivers/libjaylink/*" \
 ....
```

timsifive:
> Instead of disabling the checkpatch for the next PRs, I'd suggest getting the commit count in the branch and running git diff against just PRs commit. Something like;

That is better than the current fixed count of 20, but does not actually solve the problem of checkpatch failing because of existing code. (I just tried the commands manually, and checkpatch still complained about all kinds of old code.)

A few representative errors:
```
WARNING:EMBEDDED_FILENAME: It's generally not useful to have the filename in the file
#772: FILE: COPYING:16:
+All contributions to OpenOCD are subject to this COPYING file.
```
```
WARNING:TYPO_SPELLING: 'addres' may be misspelled - perhaps 'address'?
#8413: FILE: contrib/loaders/flash/rsl10/rom_launcher.S:24:
+    // variables are already set, addres to jump is in r3
                                   ^^^^^^
```
```
CHECK:SPACING: spaces preferred around that '/' (ctx:VxV)
#39024: FILE: src/flash/nor/em357.c:712:
+       LOG_INFO("flash size = %d KiB", num_pages*page_size/1024);
```
```
CHECK:UNNECESSARY_PARENTHESES: Unnecessary parentheses around 'last >= bank->num_sectors'
#64189: FILE: src/flash/nor/stmsmi.c:319:
+       if ((last < first) || (last >= bank->num_sectors)) {
```
And so on...

There are 1694 of them.

JanMatCodasip:
@timsifive <del>Could you please check `riscv/encodings.h` and `gdb_regs.h` files. It seems that this merge is trying to bring in older versions from upstream, compared to what is in the riscv-openocd repository.</del>

Edit: I take this back, this was my oversight. Both these files were changed in the trunk (`riscv` branch) only after this branch was forked.

JanMatCodasip:
@timsifive This merge request also seems to be reverting this change https://github.com/riscv/riscv-openocd/pull/815

TommyMurphyTM1234:
> That is better than the current fixed count of 20, but does not actually solve the problem of checkpatch failing because of existing code. (I just tried the commands manually, and checkpatch still complained about all kinds of old code.)

@timsifive and others - do you think it's of any use to flag the issue on the OpenOCD development mailing list or via an OpenOCD SourceForge issue and ask if there might be any willingness to align historical patches or the codebase as a whole with the latest checkpatch requirements? If there is then I am happy to raise the issue there. And maybe to help upstream if necessary.

JanMatCodasip:
I have quick-reviewed this merge visually and it looks all right.

I noticed just a detail: In `spi.c`, the entry "micron mt25qu01" should be right before "micron mt25ql02" (not after it). This is really a little thing, but it'd be good to clean it up so that it no longer shows up in the diffs against upstream.

JanMatCodasip (dismissed):

timsifive:
Thanks, @JanMatCodasip. I moved the flash device entry.

timsifive:
> @timsifive and others - do you think it's of any use to flag the issue on the OpenOCD development mailing list or via an OpenOCD SourceForge issue and ask if there might be any willingness to align historical patches or the codebase as a whole with the latest checkpatch requirements? If there is then I am happy to raise the issue there. And maybe to help upstream if necessary.

This is probably an improvement, and some people there seem to really enjoy this kind of work. (Look at e.g. replacing `unsigned` with `unsigned int` everywhere.)

JanMatCodasip (approved):
I have quick-reviewed this by comparing the branch:
- against the upstream base, and
- again the riscv-openocd trunk (`riscv` branch)

The merge looks all right.

TommyMurphyTM1234:
> > @timsifive and others - do you think it's of any use to flag the issue on the OpenOCD development mailing list or via an OpenOCD SourceForge issue and ask if there might be any willingness to align historical patches or the codebase as a whole with the latest checkpatch requirements? If there is then I am happy to raise the issue there. And maybe to help upstream if necessary.
> 
> This is probably an improvement, and some people there seem to really enjoy this kind of work. (Look at e.g. replacing `unsigned` with `unsigned int` everywhere.)

Hi @timsifive - sorry, I don't understand your reply. Are you expressing positivity or negativitity towards what I suggested? If the former then I will raise the issue upstream if you/others think it might be of any use in making future downstream <-> upstream merges easier. But if you/others think it's of no real use then I won't bother.

timsifive:
I was a bit on the fence about it, because I thought it would only affect things when merging down. But now it's affecting every other PR (until we get enough changes merged that this is far in history again). So yes, this would be helpful.

TommyMurphyTM1234:
> I was a bit on the fence about it, because I thought it would only affect things when merging down. But now it's affecting every other PR (until we get enough changes merged that this is far in history again). So yes, this would be helpful.

Thanks @timsifive - ok, I will raise the issue upstream via a SourceForge issue or on the openocd-devel mailing list to see what others there think. I wouldn't expect any quick action given that obvious rebuttals are (a) it could cause instability by changing lots of historical patches and already committed code and (b) why are you doing development in a fork and not upstream in the master project. But let's see what happens....

TommyMurphyTM1234:
> > I was a bit on the fence about it, because I thought it would only affect things when merging down. But now it's affecting every other PR (until we get enough changes merged that this is far in history again). So yes, this would be helpful.
> 
> Thanks @timsifive - ok, I will raise the issue upstream via a SourceForge issue or on the openocd-devel mailing list to see what others there think. I wouldn't expect any quick action given that obvious rebuttals are (a) it could cause instability by changing lots of historical patches and already committed code and (b) why are you doing development in a fork and not upstream in the master project. But let's see what happens....

I've posted here about this so let's see what people have to say about it.

* https://sourceforge.net/p/openocd/mailman/message/37801122/

TommyMurphyTM1234:
> > > I was a bit on the fence about it, because I thought it would only affect things when merging down. But now it's affecting every other PR (until we get enough changes merged that this is far in history again). So yes, this would be helpful.
> > 
> > 
> > Thanks @timsifive - ok, I will raise the issue upstream via a SourceForge issue or on the openocd-devel mailing list to see what others there think. I wouldn't expect any quick action given that obvious rebuttals are (a) it could cause instability by changing lots of historical patches and already committed code and (b) why are you doing development in a fork and not upstream in the master project. But let's see what happens....
> 
> I've posted here about this so let's see what people have to say about it.
> 
> * https://sourceforge.net/p/openocd/mailman/message/37801122/

Hi @timsifive

As mentioned by @AntonioBorneo here:

* https://sourceforge.net/p/openocd/mailman/message/37801185/

would adding (via some script/automation?) `Checkpatch-ignore: ...` (not sure if a wildcard can be used for "all checks"?) to the commit comments for patches pulled from upstream be a workable option or has that been tried before and found to be problematic?

* https://github.com/openocd-org/openocd/blob/99ec5760961d264599a9c9fb1a4d5d6042bc3ba8/HACKING#L275

PR #818, opened 2023-03-21
Try to upload Windows binary as artifact.

timsifive:
Change-Id: Iee52373a367eed61e98e828198921c9f840d9a81

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -90,3 +90,9 @@ jobs:
           echo "RELEASE_NAME=$RELEASE_NAME" >> $GITHUB_ENV
           echo "IS_PRE_RELEASE=$IS_PRE_RELEASE" >> $GITHUB_ENV
           echo "ARTIFACT_PATH=$PWD/$ARTIFACT" >> $GITHUB_ENV
+          echo "ARTIFACT_NAME=openocd-windows-${OPENOCD_TAG}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/snapshot.yml
```diff
@@ -90,3 +90,9 @@ jobs:
           echo "RELEASE_NAME=$RELEASE_NAME" >> $GITHUB_ENV
           echo "IS_PRE_RELEASE=$IS_PRE_RELEASE" >> $GITHUB_ENV
           echo "ARTIFACT_PATH=$PWD/$ARTIFACT" >> $GITHUB_ENV
+          echo "ARTIFACT_NAME=openocd-windows-${OPENOCD_TAG}" >> $GITHUB_ENV
+      - name: Publish OpenOCD packaged for windows
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ env.ARTIFACT_NAME }}
+          path: ${{ env.ARTIFACT_PATH }}
```
JanMatCodasip, line 98: @timsifive - If you'd like, the ARTIFACT_NAME improvement can be submitted also to the upstream so that there are less differences to maintain.

JanMatCodasip (approved):

.github/workflows/snapshot.yml
timsifive: Yes, although I'm not super concerned with this file since it's changed very rarely.

PR #819, opened 2023-03-22
target/riscv: support log memory access128

zqb-all:
log_memory_access only support size 1/2/4/8, let write_memory_bus_v1 call log_memory_access multiple times to fix size assert fail.

Change-Id: I6b22c97f81fac26703b66d3dbd8b6d41aaea4875

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2570,6 +2570,18 @@ static int execute_fence(struct target *target)
 	return ERROR_OK;
 }
 
+static void log_memory_access128(target_addr_t address, uint64_t value_h,
+		uint64_t value_l, bool read)
+{
+	if (debug_level < LOG_LVL_DEBUG)
+		return;
+
+	char fmt[80];
+	sprintf(fmt, "M[0x%" TARGET_PRIxADDR "] %ss 0x%%016" PRIx64 "%%016" PRIx64,
+			address, read ? "read" : "write");
+	LOG_DEBUG(fmt, value_h, value_l);
+}
+
 static void log_memory_access(target_addr_t address, uint64_t value,
 		unsigned size_bytes, bool read)
 {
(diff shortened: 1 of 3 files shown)


JanMatCodasip (changes requested):
Hi @zqb-all, thank you for submittng this change.

My recommendation would be to implement this differently: extend `log_memory_access()` function to accept 128-bit values instead of calling it multiple times with smaller values. That's because one 128-bit memory access is a different thing than four 32-bit accesses - and the log messages should clearly reflect that.

Note: I understand it can be tricky to extend `log_memory_access()` to accept 128-bit inputs since support of 128-bit integer data types is not required by the C standard ([1](https://stackoverflow.com/questions/34588650/uint128-t-does-not-name-a-type/34588976#34588976), [2](https://en.cppreference.com/w/cpp/header/cstdint)). So you can create a second function `log_memory_access128()` that would accept an array as an argument.

Also, how have you tested these changes?

zqb-all:
> My recommendation would be to implement this differently: extend `log_memory_access()` function to accept 128-bit values instead of calling it multiple times with smaller values. That's because one 128-bit memory access is a different thing than four 32-bit accesses - and the log messages should clearly reflect that.

Thanks，I will modify patch in this way

> Also, how have you tested these changes?

I have a riscv board with a cpu that supports sbaccess128，debug it with openocd and do some memory writes can trigger this assert

zqb-all:
log show like this:
```
Debug: 28502 73591 riscv-013.c:2582 log_memory_access128(): M[0x40020f70] writes 0x0000001300000013b7ed56f580820141
Debug: 28503 73591 riscv-013.c:2582 log_memory_access128(): M[0x40020f80] writes 0x879700a7e86303f007930800e4221141
Debug: 28504 73592 riscv-013.c:2582 log_memory_access128(): M[0x40020f90] writes 0x80820141642276a7d00b0a2787930001
```

src/target/riscv/riscv-013.c
```diff
@@ -2570,6 +2570,18 @@ static int execute_fence(struct target *target)
 	return ERROR_OK;
 }
 
+static void log_memory_access128(target_addr_t address, uint64_t value_h,
+		uint64_t value_l, bool read)
```
JanMatCodasip: Please, while making changes here, change also the name of the `read` argument to something else. That's because `read` is a system function (`man 2 read`):

```suggestion
		uint64_t value_l, bool is_read)
```

src/target/riscv/riscv-013.c
```diff
@@ -2570,6 +2570,18 @@ static int execute_fence(struct target *target)
...
+		return;
+
+	char fmt[80];
+	sprintf(fmt, "M[0x%" TARGET_PRIxADDR "] %ss 0x%%016" PRIx64 "%%016" PRIx64,
+			address, read ? "read" : "write");
+	LOG_DEBUG(fmt, value_h, value_l);
+}
+
 static void log_memory_access(target_addr_t address, uint64_t value,
 		unsigned size_bytes, bool read)
```

```suggestion
static void log_memory_access(target_addr_t address, uint64_t value,
		unsigned size_bytes, bool is_read)
```

src/target/riscv/riscv-013.c
```diff
@@ -3764,35 +3776,51 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
-			riscv_batch_add_dmi_write(batch, DM_SBDATA0, value, false);
+				value32 |= ((uint32_t) p[1]) << 8;
+
+			value128_l = value128_l | ((uint64_t)value32);
+			riscv_batch_add_dmi_write(batch, DM_SBDATA0, value32, false);
+
+			if (size == 16)
+				log_memory_access128(address + i * size, value128_h, value128_l, false);
+			else
+				log_memory_access(address + i * size, value128_l, size, false);
```
JanMatCodasip: Thank you for this change.

For improved readability, I'd recommend to make four `uint32_t` variables `sbdata0`, `sbdata1`, `sbdata2` and `sbdata3`, initialize them to zero and then assign them in the `if`s.

The last part would then look this way (or similar):

```suggestion
			if (size == 16) {
				uint64_t value_h = ((uint64_t)sbaccess3 << 32) | sbaccess2;
				uint64_t value_l = ((uint64_t)sbaccess1 << 32) | sbaccess0;
				log_memory_access128(address + i * size, value_h, value_l, false);
			} else {
				uint64_t value = ((uint64_t)sbaccess1 << 32) | sbaccess0;			
				log_memory_access(address + i * size, value128_l, size, false);
			}
```

JanMatCodasip:
I assume that [function `read_memory_bus_v1()`](https://github.com/riscv/riscv-openocd/blob/78ea7cf8bab3529138204eb7dbbd404a35f6dbb8/src/target/riscv/riscv-013.c#L2784) would need the same update, is that correct?

If you do not wish to do it in this merge request, that is OK. In that case please create an issue so that this "TODO" item is recorded. The changes made so far are already nice improvement over the current state.

JanMatCodasip (changes requested):

src/target/riscv/riscv-013.c
zqb-all: There are other places in this file that also use bool read, should I change all of them to is_read in this patch

src/target/riscv/riscv-013.c
zqb-all: done

src/target/riscv/riscv-013.c
zqb-all: I submitted a separate patch to rename variables

zqb-all:
> I assume that [function `read_memory_bus_v1()`](https://github.com/riscv/riscv-openocd/blob/78ea7cf8bab3529138204eb7dbbd404a35f6dbb8/src/target/riscv/riscv-013.c#L2784) would need the same update, is that correct?

Yes, thank you for pointing that out

> If you do not wish to do it in this merge request, that is OK. In that case please create an issue so that this "TODO" item is recorded. The changes made so far are already nice improvement over the current state.

After this patch is merged, I will submit another patch to update read_memory_bus_v1

JanMatCodasip (approved):
@zqb-all Thank you for the changes (and I'm sorry for my not so prompt reply).

It looks all right - approved.

timsifive (approved):
I'm not a fan of renaming a bunch of trigger variables in a commit that's about logging 128-bit memory accesses.

But the code looks OK.

TommyMurphyTM1234:
> I'm not a fan of renaming a bunch of trigger variables in a commit that's about logging 128-bit memory accesses.
> 
> But the code looks OK.

Would it make more sense to split it into two discrete PRs?

zqb-all:
> Would it make more sense to split it into two discrete PRs?

Thanks, next time there is a similar change, I will split it into a separate PR

zqb-all:
> I assume that [function `read_memory_bus_v1()`](https://github.com/riscv/riscv-openocd/blob/78ea7cf8bab3529138204eb7dbbd404a35f6dbb8/src/target/riscv/riscv-013.c#L2784) would need the same update, is that correct?
> 
> If you do not wish to do it in this merge request, that is OK. In that case please create an issue so that this "TODO" item is recorded. The changes made so far are already nice improvement over the current state.

https://github.com/riscv/riscv-openocd/pull/833

PR #823, opened 2023-03-23
target/riscv: leaf PTE check PTE_W missing

panciyan:
When permission bits R, W, and X in PTE all three are zero, the PTE is a pointer to the next level of the page table; otherwise, it is a leaf PTE. Here PTE_W is missed.

Change-Id: I82a4cc4e64280f0fcad75b20e51b617520aff29b


timsifive (approved):
Good catch.

What testing did you do for this?

panciyan:
It is based on the riscv-privileged spec, I'd bring up the Linux opensbi, boot and kernel and lookup the virtual address which could be translated to physical address.

panciyan:
> Good catch.
> 
> What testing did you do for this?

I had tested the enable mmu walk to change virtual address to physical address by tcl command, it works normally.

PR #824, opened 2023-03-24
target/riscv: AIA regs, check for H not V

timsifive:
Change-Id: Iac37b79dc737fd64a21dce83b3ef36f1a8aae118


GregSavin (approved):
That one gets me too.  It's such an easy mistake to make "V stands for virtualization extension right?", nope, vector :).

PR #825, opened 2023-03-24
target/riscv: Set hypervisor bits.

timsifive:
No other attempt is made at doing anything hypervisor-specific. Are other things necessary?

Change-Id: Ib65f114888840cf0878f9bfe028c9a42b436aa3e


src/target/riscv/riscv.c
```diff
@@ -732,16 +732,20 @@ static struct match_triggers_tdata1_fields fill_match_triggers_tdata1_fields_t2(
 static struct match_triggers_tdata1_fields fill_match_triggers_tdata1_fields_t6(
 		struct target *target, struct trigger *trigger)
 {
-	RISCV_INFO(r);
+	bool misa_s = riscv_supports_extension(target, 'S');
+	bool misa_u = riscv_supports_extension(target, 'U');
+	bool misa_v = riscv_supports_extension(target, 'V');
```
GregSavin: Is this intended to be checking for the hypervisor extension?

timsifive: Right. The virtualization extension.

Good catch. Thanks.

GregSavin (dismissed):
Looks good to me, though I have no experience yet with bring-up (or debug) of a RISC-V hypervisor environment.  I guess the OpenOCD level of debugging is exactly what somebody would use to get that sort of thing wiggling.

timsifive:
Now why did github not rerun the actions so I get a Windows binary?

timsifive:
I pushed a no-op change, and now they're running...

JanMatCodasip (commented):
I have little practical experience with H extension but the code change looks all right to me.

timsifive:
I'm ignoring `checkpatch` because it's complaining about upstream changes, not changes in this branch.

GregSavin (approved):

PR #828, opened 2023-03-28
Preserve artifact of Linux build, too.

timsifive:
That can be handy to e-mail to people who don't have a build setup.

Change-Id: I7b7e2d6f0033edf71f53211168f74b80ada64b97

### DIFF
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -22,7 +22,6 @@ jobs:
       - run: file src/openocd | grep 32-bit
       - run: src/openocd --version
 
-
   # 64-bit, gcc
   build64:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


GregSavin (dismissed):

GregSavin (dismissed):

.github/workflows/linux-build.yml
```diff
@@ -22,7 +22,6 @@ jobs:
       - run: file src/openocd | grep 32-bit
       - run: src/openocd --version
 
-
   # 64-bit, gcc
   build64:
     runs-on: ubuntu-latest
```
JanMatCodasip, line 27: The 32-bit build (which is not archived) runs on Ubuntu 20.04 whereas the 64-bit build (published as an artifact) runs on the latest one (22.04). Is that intentional, please?

.github/workflows/linux-build.yml
```diff
@@ -32,12 +31,28 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
```
JanMatCodasip: I recommend to use `v3` of the checkout action to suppress a github warning.

```suggestion
        uses: actions/checkout@v3
```

.github/workflows/linux-build.yml
```diff
@@ -32,12 +31,28 @@ jobs:
...
+          src/openocd --version
+      - name: Package
+        run: |
+          make install
+          tar zcvf ${{ env.NAME }}.tgz -C /tmp ${{ env.NAME }}
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ env.NAME }}
+          path: ${{ env.NAME }}.tgz
```
JanMatCodasip: This results in a .zip package that contains a .tgz package that finally contains the built openocd.

For that reason, please consider removing the "Package" step.

```suggestion
      - name: Upload Artifact
        uses: actions/upload-artifact@v3
        with:
          name: ${{ env.NAME }}
          path: /tmp/${{ env.NAME }}
```

.github/workflows/linux-build.yml
timsifive: I needed to provide a binary to someone on Ubuntu 20.04, so I switched it. I figure it doesn't actually matter which one we use.

.github/workflows/linux-build.yml
timsifive: https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts says "billing is calculated based on the size of the uploaded artifact and not the size of the zip file." I don't actually knows how pays, presumably the RISC-V Foundation. But whoever it is, we save them money by compressing the data before uploading it. I agree it's annoying that this means you have a tgz that is zipped.

.github/workflows/linux-build.yml
JanMatCodasip: Actually, what I meant was that the workflow currently publishes as an artifact the 64-bit openocd binary built on ubuntu-latest (22.04). Not the 32-bit binary built on ubuntu-20.04.

Is that what you intended?

.github/workflows/linux-build.yml
JanMatCodasip: That page also says: "GitHub Actions usage is free for standard GitHub-hosted runners in public repositories", which I believe applies to this public riscv-openocd repository.

In any case, the artifact size reduction w.r.t. potential billing is a good argument (even though it probably does not make any difference in this repo at the moment). If you'd like to keep the tgz compression step in the workflow, please add there just a short comment to explain the reason.

.github/workflows/linux-build.yml
timsifive: Yes. I figure the 64-bit one is useful. The 32-bit one is really just to guard against compile errors. I can't imagine anyone actually wanting that binary.

JanMatCodasip (approved):

PR #829, opened 2023-03-29
doc/openocd.texi: fix expose_csrs example

XuHangHub:
The first part of the command should be only "riscv expose_csrs", the $_TARGETNAME value is not necessarily "riscv", in fact it is generally riscv.cpux.


timsifive (approved):
Thank you for catching that.

PR #830, opened 2023-04-03
target/riscv: set some csr size to 32

zqb-all:
Change-Id: I4703b7b8ad492b14dc8d188ebb8f645c568fd515


zqb-all:
Without this patch, my openocd will try to write dcsr with size 64 and fail, then show log: "Disableing abstract command write to CSRs"

timsifive (approved):
Looks good. Thanks. I suspect you're the first person really using an implementation where abstract commands are used for CSRs.

PR #833, opened 2023-04-05
target/riscv: support log memory access128 for read

zqb-all:
Change-Id: I9235150fa00c03a1d75d0b44a7500758daa56e2b

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2541,8 +2541,8 @@ static void log_memory_access128(target_addr_t address, uint64_t value_h,
 	LOG_DEBUG(fmt, value_h, value_l);
 }
 
-static void log_memory_access(target_addr_t address, uint64_t value,
-		unsigned size_bytes, bool is_read)
+static void log_memory_access64(target_addr_t address, uint64_t value,
+		unsigned int size_bytes, bool is_read)
 {
 	if (debug_level < LOG_LVL_DEBUG)
 		return;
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -2831,33 +2839,50 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
...
+					if (next_read_j == 0) {
+						if (size == 16) {
+							uint64_t value_h = ((uint64_t)sbvalue[3] << 32) | sbvalue[2];
+							uint64_t value_l = ((uint64_t)sbvalue[1] << 32) | sbvalue[0];
+							log_memory_access128(next_read, value_h, value_l, true);
+						} else {
+							uint64_t value = ((uint64_t)sbvalue[1] << 32) | sbvalue[0];
+							log_memory_access(next_read, value, size, true);
+						}
+						memset(sbvalue, 0, size);
```
zqb-all, line 2820: I'm not sure whether to keep this memset, can anyone give me some advice

src/target/riscv/riscv-013.c
```diff
@@ -2831,33 +2839,50 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
...
+					buf_set_u32(buffer + next_read - address, 0, 8 * MIN(size, 4), sbvalue[next_read_j]);
+					if (next_read_j == 0) {
+						if (size == 16) {
+							uint64_t value_h = ((uint64_t)sbvalue[3] << 32) | sbvalue[2];
+							uint64_t value_l = ((uint64_t)sbvalue[1] << 32) | sbvalue[0];
+							log_memory_access128(next_read, value_h, value_l, true);
+						} else {
+							uint64_t value = ((uint64_t)sbvalue[1] << 32) | sbvalue[0];
+							log_memory_access(next_read, value, size, true);
+						}
```
timsifive: Can you refactor this into a function that takes sbvalue and size, so the code can be shared with the same code below?

src/target/riscv/riscv-013.c
timsifive: Keep it. If it turns out to slow things down (I doubt it), then we can put this whole block inside a check so it only runs when debug is enabled.

timsifive (changes requested):

src/target/riscv/riscv-013.c
zqb-all: sure, just submitted the new patch

timsifive (approved):

PR #834, opened 2023-04-07
Enable the RISC-V semihosting control command

lidie626:
After modification, the following commands will take effect: monitor riscv semihosting enable/disable
monitor riscv semihosting_fileio enable/disable

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -9533,6 +9533,10 @@ requests by using a special SVC instruction that is trapped at the
 Supervisor Call vector by OpenOCD.
 @end deffn
 
+@deffn {Command} {riscv semihosting} [@option{enable}|@option{disable}]
+@cindex RISCV semihosting
+@end deffn
+
 @deffn {Command} {arm semihosting_redirect} (@option{disable} | @option{tcp} <port>
 [@option{debug}|@option{stdio}|@option{all})
 @cindex ARM semihosting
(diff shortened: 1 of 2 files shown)


TommyMurphyTM1234:
I take it that the rationale for this change is to allow `riscv` to be used as an alternative to `arm` when invoking the relevant semihosting commands? 

I'm not sure that the documenation changes are correct as they seem to add `riscv semihosting` headings/sections with no associated text?

lidie626:
> I take it that the rationale for this change is to allow `riscv` to be used as an alternative to `arm` when invoking the relevant semihosting commands?
> 
> I'm not sure that the documenation changes are correct as they seem to add `riscv semihosting` headings/sections with no associated text?

RISC-V already has support for semihosting, including the semihosting library. I think the architecture-related field of the relevant commands could possibly include 'riscv'.

timsifive:
The command is under `arm semihosting` because it implements a protocol that is basically the same as ARM semihosting. What problem are you trying to solve?

lidie626:
> The command is under `arm semihosting` because it implements a protocol that is basically the same as ARM semihosting. What problem are you trying to solve?
Thanks for your reply. Considering that RISC-V does not have its own semihosting protocol, this patch is not needed for now.

JanMatCodasip:
> Thanks for your reply. Considering that RISC-V does not have its own semihosting protocol, this patch is not needed for now.

@lidie626 - Can this merge request be closed, then?

Thank you.

PR #840, opened 2023-04-24
fix bp handling during resume

aap-sc:
Depending client parameters OpenOCD resume command can do step+resume to avoid triggering a pending breakpoint

Change-Id: Ib7ae544e1a1f13843584f4c1c87db17851642b89

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1733,23 +1733,23 @@ static int resume_prep(struct target *target, int current,
 		target_addr_t address, int handle_breakpoints, int debug_execution)
 {
 	RISCV_INFO(r);
-	LOG_TARGET_DEBUG(target, "target->state=%d", target->state);
+
+	LOG_TARGET_DEBUG(target, "target->state=%d, current=%s, address=0x%"
+		TARGET_PRIxADDR ", handle_breakpoints=%s, debug_exec=%s",
+		target->state,
+		current ? "true" : "false",
+		address,
+		handle_breakpoints ? "true" : "false",
+		debug_execution ? "true" : "false");
 
 	if (!current && riscv_set_register(target, GDB_REGNO_PC, address) != ERROR_OK)
 		return ERROR_FAIL;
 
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv.c
```diff
@@ -1733,23 +1733,23 @@ static int resume_prep(struct target *target, int current,
...
-			return ERROR_FAIL;
-
-		if (old_or_new_riscv_step(target, true, 0, false) != ERROR_OK)
-			return ERROR_FAIL;
-
-		if (enable_triggers(target, trigger_state) != ERROR_OK)
+	if (handle_breakpoints) {
+		// To be able to run off a trigger, we perform a step operation and then
+		// resume. If handle_breakpoints is true then step temporarily disables
+		// pending breakpoints so we can safely perform the step.
```
timsifive: Can you rewrite this with `/* */` comments? I'm not sure why checkpatch doesn't complain about it anymore, but I've had upstream requests denied for using `//` comments, so I'd like to stick with `/* */` for now.

timsifive (changes requested):
Looks like a good simplification. How did you test this?

src/target/riscv/riscv.c
aap-sc: > Looks like a good simplification. How did you test this?

1. For the directed test - I've tested it locally using a directed test with a telnet-based connection (it can't be tested with gdb since gdb manages breakpoints by itself, that is `handle_breakpoints` set to false). I used both hw and sw breakpoints for the test.
2. In addition, I've launched the whole riscv-tests/debug  testsuite.  I used `48491dadafcd59442c5bf22603fb0a7f1c589cb9` revision (it's from around december 27).

aap-sc: > Can you rewrite this with /* */ comments? 

Oh, sure.

timsifive (approved):

JanMatCodasip (approved):
The change looks all right, and addresses one of my points from https://github.com/riscv/riscv-openocd/issues/836

Thank you.

PR #841, opened 2023-04-24
target/riscv: respect error code from dm013_select_target in select_prepped_harts

aap-sc:
Change-Id: I3099589521538590e366d60629e49cfc74e2d0c6


timsifive (approved):
Thank you.

PR #842, opened 2023-04-24
target:riscv: optimize register accesses

en-sc:
These commits reduce the number of accesses to target's registers by better utilizing register cache.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -58,6 +58,7 @@ static void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a,
 static void riscv013_fill_dmi_read_u64(struct target *target, char *buf, int a);
 static int riscv013_dmi_write_u64_bits(struct target *target);
 static void riscv013_fill_dmi_nop_u64(struct target *target, char *buf);
+static unsigned int register_size(struct target *target, enum gdb_regno number);
 static int register_read_direct(struct target *target, riscv_reg_t *value,
 		enum gdb_regno number);
 static int register_write_direct(struct target *target, enum gdb_regno number,
@@ -877,8 +878,8 @@ static uint32_t access_register_command(struct target *target, uint32_t number,
 	return command;
 }
 
-static int register_read_abstract(struct target *target, riscv_reg_t *value,
-		enum gdb_regno number, unsigned int size)
+static int register_read_abstract_with_size(struct target *target,
+		riscv_reg_t *value, enum gdb_regno number, unsigned int size)
 {
 	RISCV013_INFO(info);
 
@@ -915,10 +916,19 @@ static int register_read_abstract(struct target *target, riscv_reg_t *value,
 	return ERROR_OK;
 }
 
+static int register_read_abstract(struct target *target, riscv_reg_t *value,
+		enum gdb_regno number)
+{
+	const unsigned int size = register_size(target, number);
+
+	return register_read_abstract_with_size(target, value, number, size);
+}
+
 static int register_write_abstract(struct target *target, enum gdb_regno number,
-		riscv_reg_t value, unsigned int size)
+		riscv_reg_t value)
 {
 	RISCV013_INFO(info);
+	const unsigned int size = register_size(target, number);
 
 	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31 &&
 			!info->abstract_write_fpr_supported)
@@ -1294,22 +1304,43 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
 }
 
 /**
- * Immediately write the new value to the requested register. This mechanism
- * bypasses any caches.
+ * This function is used to read a 64-bit value from a register by executing a
+ * program.
+ * The program stores a register to address located in S0.
+ * The caller should save S0.
  */
-static int register_write_direct(struct target *target, enum gdb_regno number,
-		riscv_reg_t value)
+static int internal_register_read64_progbuf_scratch(struct target *target,
+		struct riscv_program *program, riscv_reg_t *value)
 {
-	LOG_TARGET_DEBUG(target, "%s <- 0x%" PRIx64, gdb_regno_name(number), value);
+	scratch_mem_t scratch;
 
-	riscv_reg_t mstatus;
-	if (prep_for_register_access(target, &mstatus, number) != ERROR_OK)
+	if (scratch_reserve(target, &scratch, program, 8) != ERROR_OK)
 		return ERROR_FAIL;
 
-	int result = register_write_abstract(target, number, value,
-			register_size(target, number));
-	if (result == ERROR_OK || !has_sufficient_progbuf(target, 2))
-		return result;
+	if (register_write_abstract(target, GDB_REGNO_S0, scratch.hart_address)
+			!= ERROR_OK) {
+		scratch_release(target, &scratch);
+		return ERROR_FAIL;
+	}
+	if (riscv_program_exec(program, target) != ERROR_OK) {
+		scratch_release(target, &scratch);
+		return ERROR_FAIL;
+	}
+
+	int result = scratch_read64(target, &scratch, value);
+
+	scratch_release(target, &scratch);
+	return result;
+}
+
+/**
+ * This function reads a register by writing a program to program buffer and
+ * executing it.
+ */
+static int register_read_progbuf(struct target *target, uint64_t *value,
+		enum gdb_regno number)
+{
+	assert(target->state == TARGET_HALTED);
 
 	struct riscv_program program;
 	riscv_program_init(&program, target);
@@ -1317,170 +1348,183 @@ static int register_write_direct(struct target *target, enum gdb_regno number,
 	if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
 		return ERROR_FAIL;
 
-	scratch_mem_t scratch;
-	bool use_scratch = false;
-	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31 &&
-			riscv_supports_extension(target, 'D') &&
-			riscv_xlen(target) < 64) {
-		/* There are no instructions to move all the bits from a register, so
-		 * we need to use some scratch RAM. */
-		use_scratch = true;
-		if (riscv_program_insert(&program, fld(number - GDB_REGNO_FPR0, S0, 0))
-				!= ERROR_OK)
-			return ERROR_FAIL;
+	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31) {
+		const unsigned int freg = number - GDB_REGNO_FPR0;
 
-		if (scratch_reserve(target, &scratch, &program, 8) != ERROR_OK)
-			return ERROR_FAIL;
-
-		if (register_write_direct(target, GDB_REGNO_S0, scratch.hart_address)
-				!= ERROR_OK) {
-			scratch_release(target, &scratch);
-			return ERROR_FAIL;
+		if (riscv_supports_extension(target, 'D') && riscv_xlen(target) < 64) {
+			/* There are no instructions to move all the bits from a
+			 * register, so we need to use some scratch RAM.
+			 */
+			if (riscv_program_insert(&program, fsd(freg, S0, 0)) != ERROR_OK)
+				return ERROR_FAIL;
+			return internal_register_read64_progbuf_scratch(target, &program,
+					value);
 		}
-
-		if (scratch_write64(target, &scratch, value) != ERROR_OK) {
-			scratch_release(target, &scratch);
+		if (riscv_program_insert(&program,
+					riscv_supports_extension(target, 'D') ?
+					fmv_x_d(S0, freg) : fmv_x_w(S0, freg)) != ERROR_OK)
 			return ERROR_FAIL;
-		}
-
+	} else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095) {
+		if (riscv_program_csrr(&program, S0, number) != ERROR_OK)
+			return ERROR_FAIL;
 	} else {
-		if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31) {
-			if (riscv_supports_extension(target, 'D')) {
-				if (riscv_program_insert(&program,
-							fmv_d_x(number - GDB_REGNO_FPR0, S0)) != ERROR_OK)
-					return ERROR_FAIL;
-			} else {
-				if (riscv_program_insert(&program,
-							fmv_w_x(number - GDB_REGNO_FPR0, S0)) != ERROR_OK)
-					return ERROR_FAIL;
-			}
-		} else if (number == GDB_REGNO_VTYPE) {
-			if (riscv_save_register(target, GDB_REGNO_S1) != ERROR_OK)
-				return ERROR_FAIL;
-			if (riscv_program_insert(&program, csrr(S1, CSR_VL)) != ERROR_OK)
-				return ERROR_FAIL;
-			if (riscv_program_insert(&program, vsetvl(ZERO, S1, S0)) != ERROR_OK)
-				return ERROR_FAIL;
-		} else if (number == GDB_REGNO_VL) {
-			/* "The XLEN-bit-wide read-only vl CSR can only be updated by the
-			 * vsetvli and vsetvl instructions, and the fault-only-rst vector
-			 * load instruction variants." */
-			if (riscv_save_register(target, GDB_REGNO_S1) != ERROR_OK)
-				return ERROR_FAIL;
-			if (riscv_program_insert(&program, csrr(S1, CSR_VTYPE)) != ERROR_OK)
-				return ERROR_FAIL;
-			if (riscv_program_insert(&program, vsetvl(ZERO, S0, S1)) != ERROR_OK)
-				return ERROR_FAIL;
-		} else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095) {
-			if (riscv_program_csrw(&program, S0, number) != ERROR_OK)
-				return ERROR_FAIL;
-		} else {
-			LOG_ERROR("Unsupported register (enum gdb_regno)(%d)", number);
-			return ERROR_FAIL;
-		}
-		if (register_write_direct(target, GDB_REGNO_S0, value) != ERROR_OK)
-			return ERROR_FAIL;
+		LOG_ERROR("Unsupported register: %s", gdb_regno_name(number));
+		return ERROR_FAIL;
 	}
 
-	int exec_out = riscv_program_exec(&program, target);
-	/* Don't message on error. Probably the register doesn't exist. */
-	if (exec_out == ERROR_OK && target->reg_cache) {
-		struct reg *reg = &target->reg_cache->reg_list[number];
-		buf_set_u64(reg->value, 0, reg->size, value);
-	}
+	if (riscv_program_exec(&program, target) != ERROR_OK)
+		return ERROR_FAIL;
 
-	if (use_scratch)
+	return register_read_abstract(target, value, GDB_REGNO_S0) != ERROR_OK;
+}
+
+/**
+ * This function is used to write a 64-bit value to a register by executing a
+ * program.
+ * The program loads a value from address located in S0 to a register.
+ * The caller should save S0.
+ */
+static int internal_register_write64_progbuf_scratch(struct target *target,
+		struct riscv_program *program, riscv_reg_t value)
+{
+	scratch_mem_t scratch;
+
+	if (scratch_reserve(target, &scratch, program, 8) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (register_write_abstract(target, GDB_REGNO_S0, scratch.hart_address)
+			!= ERROR_OK) {
 		scratch_release(target, &scratch);
+		return ERROR_FAIL;
+	}
+	if (scratch_write64(target, &scratch, value) != ERROR_OK) {
+		scratch_release(target, &scratch);
+		return ERROR_FAIL;
+	}
+	int result = riscv_program_exec(program, target);
+
+	scratch_release(target, &scratch);
+	return result;
+}
+
+/**
+ * This function writes a register by writing a program to program buffer and
+ * executing it.
+ */
+static int register_write_progbuf(struct target *target, enum gdb_regno number,
+		riscv_reg_t value)
+{
+	assert(target->state == TARGET_HALTED);
+
+	struct riscv_program program;
+
+	riscv_program_init(&program, target);
+
+	if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31 &&
+			riscv_supports_extension(target, 'D') && riscv_xlen(target) < 64) {
+		/* There are no instructions to move all the bits from a register,
+		 * so we need to use some scratch RAM.
+		 */
+		const unsigned int freg = number - GDB_REGNO_FPR0;
+
+		if (riscv_program_insert(&program, fld(freg, S0, 0)) != ERROR_OK)
+			return ERROR_FAIL;
+		return internal_register_write64_progbuf_scratch(target, &program,
+				value);
+	}
+
+	if (register_write_abstract(target, GDB_REGNO_S0, value) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31) {
+		const unsigned int freg = number - GDB_REGNO_FPR0;
+
+		if (riscv_program_insert(&program,
+					riscv_supports_extension(target, 'D') ?
+					fmv_d_x(freg, S0) : fmv_w_x(freg, S0)) != ERROR_OK)
+			return ERROR_FAIL;
+	} else if (number == GDB_REGNO_VTYPE) {
+		if (riscv_save_register(target, GDB_REGNO_S1) != ERROR_OK)
+			return ERROR_FAIL;
+		if (riscv_program_insert(&program, csrr(S1, CSR_VL)) != ERROR_OK)
+			return ERROR_FAIL;
+		if (riscv_program_insert(&program, vsetvl(ZERO, S1, S0)) != ERROR_OK)
+			return ERROR_FAIL;
+	} else if (number == GDB_REGNO_VL) {
+		/* "The XLEN-bit-wide read-only vl CSR can only be updated by the
+		 * vsetvli and vsetvl instructions, and the fault-only-rst vector
+		 * load instruction variants."
+		 */
+		if (riscv_save_register(target, GDB_REGNO_S1) != ERROR_OK)
+			return ERROR_FAIL;
+		if (riscv_program_insert(&program, csrr(S1, CSR_VTYPE)) != ERROR_OK)
+			return ERROR_FAIL;
+		if (riscv_program_insert(&program, vsetvl(ZERO, S0, S1)) != ERROR_OK)
+			return ERROR_FAIL;
+	} else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095) {
+		if (riscv_program_csrw(&program, S0, number) != ERROR_OK)
+			return ERROR_FAIL;
+	} else {
+		LOG_ERROR("Unsupported register (enum gdb_regno)(%d)", number);
+		return ERROR_FAIL;
+	}
+	return riscv_program_exec(&program, target);
+}
+
+/**
+ * Immediately write the new value to the requested register. This mechanism
+ * bypasses any caches.
+ */
+static int register_write_direct(struct target *target, enum gdb_regno number,
+		riscv_reg_t value)
+{
+	LOG_TARGET_DEBUG(target, "Writing 0x%" PRIx64 " to %s", value,
+			gdb_regno_name(number));
+
+	riscv_reg_t mstatus;
+	if (prep_for_register_access(target, &mstatus, number) != ERROR_OK)
+		return ERROR_FAIL;
+
+	int result = register_write_abstract(target, number, value);
+
+	if (result != ERROR_OK && target->state == TARGET_HALTED)
+		result = register_write_progbuf(target, number, value);
 
 	if (cleanup_after_register_access(target, mstatus, number) != ERROR_OK)
 		return ERROR_FAIL;
 
-	return exec_out;
+	if (result == ERROR_OK)
+		LOG_TARGET_DEBUG(target, "%s <- 0x%" PRIx64, gdb_regno_name(number),
+				value);
+
+	return result;
 }
 
 /** Actually read registers from the target right now. */
 static int register_read_direct(struct target *target, riscv_reg_t *value,
 		enum gdb_regno number)
 {
-	if (dm013_select_target(target) != ERROR_OK)
-		return ERROR_FAIL;
+	LOG_TARGET_DEBUG(target, "Reading %s", gdb_regno_name(number));
 
 	riscv_reg_t mstatus;
+
 	if (prep_for_register_access(target, &mstatus, number) != ERROR_OK)
 		return ERROR_FAIL;
 
-	int result = register_read_abstract(target, value, number,
-			register_size(target, number));
+	int result = register_read_abstract(target, value, number);
 
-	if (result != ERROR_OK &&
-			has_sufficient_progbuf(target, 2) &&
-			number > GDB_REGNO_XPR31) {
-		struct riscv_program program;
-		riscv_program_init(&program, target);
-
-		scratch_mem_t scratch;
-		bool use_scratch = false;
-
-		if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
-			return ERROR_FAIL;
-
-		/* Write program to move data into s0. */
-
-		if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31) {
-			if (riscv_supports_extension(target, 'D')
-					&& riscv_xlen(target) < 64) {
-				/* There are no instructions to move all the bits from a
-				 * register, so we need to use some scratch RAM. */
-				if (riscv_program_insert(&program,
-							fsd(number - GDB_REGNO_FPR0, S0, 0)) != ERROR_OK)
-					return ERROR_FAIL;
-				if (scratch_reserve(target, &scratch, &program, 8) != ERROR_OK)
-					return ERROR_FAIL;
-				use_scratch = true;
-
-				if (register_write_direct(target, GDB_REGNO_S0,
-							scratch.hart_address) != ERROR_OK) {
-					scratch_release(target, &scratch);
-					return ERROR_FAIL;
-				}
-			} else if (riscv_supports_extension(target, 'D')) {
-				if (riscv_program_insert(&program, fmv_x_d(S0, number - GDB_REGNO_FPR0)) !=
-						ERROR_OK)
-					return ERROR_FAIL;
-			} else {
-				if (riscv_program_insert(&program, fmv_x_w(S0, number - GDB_REGNO_FPR0)) !=
-						ERROR_OK)
-					return ERROR_FAIL;
-			}
-		} else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095) {
-			if (riscv_program_csrr(&program, S0, number) != ERROR_OK)
-				return ERROR_FAIL;
-		} else {
-			LOG_ERROR("Unsupported register: %s", gdb_regno_name(number));
-			return ERROR_FAIL;
-		}
-
-		/* Execute program. */
-		result = riscv_program_exec(&program, target);
-		/* Don't message on error. Probably the register doesn't exist. */
-
-		if (use_scratch) {
-			result = scratch_read64(target, &scratch, value);
-			scratch_release(target, &scratch);
-			if (result != ERROR_OK)
-				return result;
-		} else {
-			/* Read S0 */
-			if (register_read_direct(target, value, GDB_REGNO_S0) != ERROR_OK)
-				return ERROR_FAIL;
-		}
-	}
+	if (result != ERROR_OK && target->state == TARGET_HALTED)
+		result = register_read_progbuf(target, value, number);
 
 	if (cleanup_after_register_access(target, mstatus, number) != ERROR_OK)
 		return ERROR_FAIL;
 
 	if (result == ERROR_OK)
-		LOG_TARGET_DEBUG(target, "%s = 0x%" PRIx64, gdb_regno_name(number), *value);
+		LOG_TARGET_DEBUG(target, "%s = 0x%" PRIx64, gdb_regno_name(number),
+				*value);
 
 	return result;
 }
@@ -1710,13 +1754,29 @@ static int examine(struct target *target)
 					info->index);
 			return ERROR_FAIL;
 		}
+		target->state = TARGET_HALTED;
+		target->debug_reason = DBG_REASON_DBGRQ;
+	}
+	/* FIXME: This is needed since register_read_direct relies on target->state
+	 * to work correctly, so, if target->state does not represent current state
+	 * of target, e.g. if a target is halted, but target->state is
+	 * TARGET_UNKNOWN, it can fail early, (e.g. accessing registers via program
+	 * buffer can not be done atomically on a running hart becuse mstatus can't
+	 * be prepared for a register access and then restored)
+	 * See https://github.com/riscv/riscv-openocd/pull/842#discussion_r1179414089
+	 */
+	const enum target_state saved_tgt_state = target->state;
+	const enum target_debug_reason saved_dbg_reason = target->debug_reason;
+	if (target->state != TARGET_HALTED) {
+		target->state = TARGET_HALTED;
+		target->debug_reason = DBG_REASON_DBGRQ;
 	}
 
 	/* Without knowing anything else we can at least mess with the
 	 * program buffer. */
 	r->debug_buffer_size = info->progbufsize;
 
-	int result = register_read_abstract(target, NULL, GDB_REGNO_S0, 64);
+	int result = register_read_abstract_with_size(target, NULL, GDB_REGNO_S0, 64);
 	if (result == ERROR_OK)
 		r->xlen = 64;
 	else
@@ -1725,11 +1785,11 @@ static int examine(struct target *target)
 	/* Save s0 and s1. The register cache hasn't be initialized yet so we
 	 * need to take care of this manually. */
 	uint64_t s0, s1;
-	if (register_read_direct(target, &s0, GDB_REGNO_S0) != ERROR_OK) {
+	if (register_read_abstract(target, &s0, GDB_REGNO_S0) != ERROR_OK) {
 		LOG_TARGET_ERROR(target, "Fatal: Failed to read s0.");
 		return ERROR_FAIL;
 	}
-	if (register_read_direct(target, &s1, GDB_REGNO_S1) != ERROR_OK) {
+	if (register_read_abstract(target, &s1, GDB_REGNO_S1) != ERROR_OK) {
 		LOG_TARGET_ERROR(target, "Fatal: Failed to read s1.");
 		return ERROR_FAIL;
 	}
@@ -1780,6 +1840,9 @@ static int examine(struct target *target)
 		LOG_TARGET_ERROR(target, "Fatal: Failed to write back s1.");
 		return ERROR_FAIL;
 	}
+
+	target->state = saved_tgt_state;
+	target->debug_reason = saved_dbg_reason;
 
 	if (!halted) {
 		riscv013_step_or_resume_current_hart(target, false);
@@ -2019,14 +2082,15 @@ static int riscv013_get_register_buf(struct target *target,
 
 	riscv_reg_t mstatus, vtype, vl;
 	unsigned int debug_vl, debug_vsew;
+
 	if (prep_for_vector_access(target, &mstatus, &vtype, &vl,
 				&debug_vl, &debug_vsew) != ERROR_OK)
 		return ERROR_FAIL;
 
-	unsigned vnum = regno - GDB_REGNO_V0;
+	unsigned int vnum = regno - GDB_REGNO_V0;
 
 	int result = ERROR_OK;
-	for (unsigned i = 0; i < debug_vl; i++) {
+	for (unsigned int i = 0; i < debug_vl; i++) {
 		/* Can't reuse the same program because riscv_program_exec() adds
 		 * ebreak to the end every time. */
 		struct riscv_program program;
@@ -2073,17 +2137,18 @@ static int riscv013_set_register_buf(struct target *target,
 
 	riscv_reg_t mstatus, vtype, vl;
 	unsigned int debug_vl, debug_vsew;
+
 	if (prep_for_vector_access(target, &mstatus, &vtype, &vl,
 				&debug_vl, &debug_vsew) != ERROR_OK)
 		return ERROR_FAIL;
 
-	unsigned vnum = regno - GDB_REGNO_V0;
+	unsigned int vnum = regno - GDB_REGNO_V0;
 
 	struct riscv_program program;
 	riscv_program_init(&program, target);
 	riscv_program_insert(&program, vslide1down_vx(vnum, vnum, S0, true));
 	int result = ERROR_OK;
-	for (unsigned i = 0; i < debug_vl; i++) {
+	for (unsigned int i = 0; i < debug_vl; i++) {
 		if (register_write_direct(target, GDB_REGNO_S0,
 					buf_get_u64(value, debug_vsew * i, debug_vsew)) != ERROR_OK)
 			return ERROR_FAIL;
@@ -4164,8 +4229,7 @@ struct target_type riscv013_target = {
 static int riscv013_get_register(struct target *target,
 		riscv_reg_t *value, enum gdb_regno rid)
 {
-	LOG_DEBUG("[%s] reading register %s", target_name(target),
-			gdb_regno_name(rid));
+	LOG_TARGET_DEBUG(target, "reading register %s",	gdb_regno_name(rid));
 
 	if (dm013_select_target(target) != ERROR_OK)
 		return ERROR_FAIL;
@@ -4178,7 +4242,8 @@ static int riscv013_get_register(struct target *target,
 	} else if (rid == GDB_REGNO_PRIV) {
 		uint64_t dcsr;
 		/* TODO: move this into riscv.c. */
-		result = register_read_direct(target, &dcsr, GDB_REGNO_DCSR);
+		if (register_read_direct(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
+			return ERROR_FAIL;
 		*value = set_field(0, VIRT_PRIV_V, get_field(dcsr, CSR_DCSR_V));
 		*value = set_field(*value, VIRT_PRIV_PRV, get_field(dcsr, CSR_DCSR_PRV));
 	} else {
@@ -4193,27 +4258,22 @@ static int riscv013_get_register(struct target *target,
 static int riscv013_set_register(struct target *target, enum gdb_regno rid,
 		riscv_reg_t value)
 {
-	if (dm013_select_target(target) != ERROR_OK)
-		return ERROR_FAIL;
 	LOG_TARGET_DEBUG(target, "writing 0x%" PRIx64 " to register %s",
 			value, gdb_regno_name(rid));
+
+	if (dm013_select_target(target) != ERROR_OK)
+		return ERROR_FAIL;
 
 	if (rid <= GDB_REGNO_XPR31) {
 		return register_write_direct(target, rid, value);
 	} else if (rid == GDB_REGNO_PC) {
 		LOG_TARGET_DEBUG(target, "writing PC to DPC: 0x%" PRIx64, value);
-		register_write_direct(target, GDB_REGNO_DPC, value);
-		riscv_reg_t actual_value;
-		register_read_direct(target, &actual_value, GDB_REGNO_DPC);
-		LOG_TARGET_DEBUG(target, "  actual DPC written: 0x%016" PRIx64, actual_value);
-		if (value != actual_value) {
-			LOG_ERROR("Written PC (0x%" PRIx64 ") does not match read back "
-					"value (0x%" PRIx64 ")", value, actual_value);
-			return ERROR_FAIL;
-		}
+		return register_write_direct(target, GDB_REGNO_DPC, value);
 	} else if (rid == GDB_REGNO_PRIV) {
 		riscv_reg_t dcsr;
-		register_read_direct(target, &dcsr, GDB_REGNO_DCSR);
+
+		if (register_read_direct(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
+			return ERROR_FAIL;
 		dcsr = set_field(dcsr, CSR_DCSR_PRV, get_field(value, VIRT_PRIV_PRV));
 		dcsr = set_field(dcsr, CSR_DCSR_V, get_field(value, VIRT_PRIV_V));
 		return register_write_direct(target, GDB_REGNO_DCSR, dcsr);
@@ -4544,7 +4604,7 @@ static int riscv013_on_step_or_resume(struct target *target, bool step)
 
 	/* We want to twiddle some bits in the debug CSR so debugging works. */
 	riscv_reg_t dcsr;
-	int result = register_read_direct(target, &dcsr, GDB_REGNO_DCSR);
+	int result = riscv_get_register(target, &dcsr, GDB_REGNO_DCSR);
 	if (result != ERROR_OK)
 		return result;
 	dcsr = set_field(dcsr, CSR_DCSR_STEP, step);
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -36,8 +36,9 @@ static void riscv013_clear_abstract_error(struct target *target);
 
 /* Implementations of the functions in struct riscv_info. */
 static int riscv013_get_register(struct target *target,
-		riscv_reg_t *value, int rid);
-static int riscv013_set_register(struct target *target, int regid, uint64_t value);
+		riscv_reg_t *value, enum gdb_regno rid);
+static int riscv013_set_register(struct target *target, enum gdb_regno regid,
+		riscv_reg_t value);
 static int dm013_select_hart(struct target *target, int hart_index);
 static int riscv013_halt_prep(struct target *target);
 static int riscv013_halt_go(struct target *target);
@@ -57,9 +58,10 @@ static void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a,
 static void riscv013_fill_dmi_read_u64(struct target *target, char *buf, int a);
 static int riscv013_dmi_write_u64_bits(struct target *target);
 static void riscv013_fill_dmi_nop_u64(struct target *target, char *buf);
-static int register_read_direct(struct target *target, uint64_t *value, uint32_t number);
-static int register_write_direct(struct target *target, unsigned number,
-		uint64_t value);
+static int register_read_direct(struct target *target, riscv_reg_t *value,
+		enum gdb_regno number);
+static int register_write_direct(struct target *target, enum gdb_regno number,
+		riscv_reg_t value);
 static int read_memory(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, uint8_t *buffer, uint32_t increment);
 static int write_memory(struct target *target, target_addr_t address,
@@ -875,8 +877,8 @@ static uint32_t access_register_command(struct target *target, uint32_t number,
 	return command;
 }
 
-static int register_read_abstract(struct target *target, uint64_t *value,
-		uint32_t number, unsigned size)
+static int register_read_abstract(struct target *target, riscv_reg_t *value,
+		enum gdb_regno number, unsigned int size)
 {
 	RISCV013_INFO(info);
 
@@ -913,8 +915,8 @@ static int register_read_abstract(struct target *target, uint64_t *value,
 	return ERROR_OK;
 }
 
-static int register_write_abstract(struct target *target, uint32_t number,
-		uint64_t value, unsigned size)
+static int register_write_abstract(struct target *target, enum gdb_regno number,
+		riscv_reg_t value, unsigned int size)
 {
 	RISCV013_INFO(info);
 
@@ -1043,7 +1045,7 @@ static int examine_progbuf(struct target *target)
 	return ERROR_OK;
 }
 
-static int is_fpu_reg(uint32_t gdb_regno)
+static int is_fpu_reg(enum gdb_regno gdb_regno)
 {
 	return (gdb_regno >= GDB_REGNO_FPR0 && gdb_regno <= GDB_REGNO_FPR31) ||
 		(gdb_regno == GDB_REGNO_CSR0 + CSR_FFLAGS) ||
@@ -1051,7 +1053,7 @@ static int is_fpu_reg(uint32_t gdb_regno)
 		(gdb_regno == GDB_REGNO_CSR0 + CSR_FCSR);
 }
 
-static int is_vector_reg(uint32_t gdb_regno)
+static int is_vector_reg(enum gdb_regno gdb_regno)
 {
 	return (gdb_regno >= GDB_REGNO_V0 && gdb_regno <= GDB_REGNO_V31) ||
 		gdb_regno == GDB_REGNO_VSTART ||
@@ -1063,8 +1065,8 @@ static int is_vector_reg(uint32_t gdb_regno)
 		gdb_regno == GDB_REGNO_VLENB;
 }
 
-static int prep_for_register_access(struct target *target, uint64_t *mstatus,
-		int regno)
+static int prep_for_register_access(struct target *target, riscv_reg_t *mstatus,
+		enum gdb_regno regno)
 {
 	if (is_fpu_reg(regno) || is_vector_reg(regno)) {
 		if (register_read_direct(target, mstatus, GDB_REGNO_MSTATUS) != ERROR_OK)
@@ -1085,7 +1087,7 @@ static int prep_for_register_access(struct target *target, uint64_t *mstatus,
 }
 
 static int cleanup_after_register_access(struct target *target,
-		uint64_t mstatus, int regno)
+		riscv_reg_t mstatus, enum gdb_regno regno)
 {
 	if ((is_fpu_reg(regno) && (mstatus & MSTATUS_FS) == 0) ||
 			(is_vector_reg(regno) && (mstatus & MSTATUS_VS) == 0))
@@ -1254,7 +1256,7 @@ static int scratch_write64(struct target *target, scratch_mem_t *scratch,
 }
 
 /** Return register size in bits. */
-static unsigned register_size(struct target *target, unsigned number)
+static unsigned int register_size(struct target *target, enum gdb_regno number)
 {
 	/* If reg_cache hasn't been initialized yet, make a guess. We need this for
 	 * when this function is called during examine(). */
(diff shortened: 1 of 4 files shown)


timsifive (commented):
I love this. Unnecessary register accesses annoy me every time I look at a log and I just haven't gotten around to fixing them.

I haven't done a proper review yet.

Can you add short comments to each register function that highlights the difference between all of them?

sifiverobert:
I am not sure if using 'gdb_regno' in low-level debug access functions is appropriate.
These indices are rather defined by ISA (and DM ...) specification and NOT by GDB.
I know these GDB_REGNO_xxx were there before, but when we re-factor the code, we should make it more generic and not more (GDB) specific.

en-sc:
> I am not sure if using 'gdb_regno' in low-level debug access functions is appropriate. These indices are rather defined by ISA (and DM ...) specification and NOT by GDB. I know these GDB_REGNO_xxx were there before, but when we re-factor the code, we should make it more generic and not more (GDB) specific.

I also think this is a valid point of concern. However, the transition from `int` to `gdb_regno`, in my opinion, is justified for a handful of reasons: 

* There are quite a few places in which such a variable is used in the following way:
```
if (<var> >= GDB_REGNO_<XXX> && <var> <= GDB_REGNO_<YYY>)

```
In such uses it is silently expected that `<var>` is part of some enumeration.

* In my opinion, it is better to use one enumeration consistently through the codebase, rather than supporting two enumerations, one representing gdb's point of view and the other representing HW's point, together with the tools to convert between them.

* Furthermore, if it becomes necessary to create such HW regno enumeration, it would still be easier if the use of the currently present enumeration was consistent.

So, in my opinion, the creation of such an enumeration and conversion should be done in a separate commit and is somewhat out of scope of this PR.

en-sc:
> Can you add short comments to each register function that highlights the difference between all of them?

Addressed.

src/target/riscv/riscv-013.c
```diff
@@ -1063,34 +1065,47 @@ static int is_vector_reg(uint32_t gdb_regno)
...
+	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
+		return ERROR_OK;
+
+	LOG_TARGET_DEBUG(target, "Preparing mstatus to access %s",
+			gdb_regno_name(regno));
+	if (target->state != TARGET_HALTED) {
+		LOG_TARGET_ERROR(target, "Unable to access %s: target not halted",
+				gdb_regno_name(regno));
+		return ERROR_FAIL;
 	}
```
timsifive: That doesn't seem right. If the target is running registers might still be accessible.

If the target is running, however, change mstatus, access a register, and restore mstatus, probably has nasty side effects. So we should report failure to the user, but only if mstatus actually needs to be modified.

src/target/riscv/riscv-013.c
```diff
@@ -1063,34 +1065,47 @@ static int is_vector_reg(uint32_t gdb_regno)
...
-			(is_vector_reg(regno) && (mstatus & MSTATUS_VS) == 0))
-		if (register_write_direct(target, GDB_REGNO_MSTATUS, mstatus) != ERROR_OK)
-			return ERROR_FAIL;
+	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
+		return ERROR_OK;
+
+	LOG_TARGET_DEBUG(target, "Restoring mstatus to 0x%" PRIx64, mstatus);
+	if (riscv_write_register(target, GDB_REGNO_MSTATUS, mstatus))
+		return ERROR_FAIL;
+	LOG_TARGET_DEBUG(target, "mstatus is restored");
```
timsifive: Don't really need this, since there are plenty of debug messages already showing that the write occurred.

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1287,210 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
 	return info->progbufsize + r->impebreak >= size;
 }
 
-/**
- * Immediately write the new value to the requested register. This mechanism
- * bypasses any caches.
- */
-static int register_write_direct(struct target *target, unsigned number,
-		uint64_t value)
+static int internal_register_read64_progbuf_scratch(struct target *target,
```
timsifive, line 1312: Nice refactor.

Can you add a brief comment for this function, mentioning what the riscv_program is supposed to do?

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1287,210 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
+	if (riscv_program_exec(program, target) != ERROR_OK)
+		return ERROR_FAIL;
+
+	int result = scratch_read64(target, &scratch, value);
+
+	scratch_release(target, &scratch);
+	return result;
+}
+
+static int register_read_progbuf(struct target *target, uint64_t *value,
```
timsifive, line 1340: Add a short comment.

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1287,210 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
+		LOG_ERROR("Unsupported register: %s", gdb_regno_name(number));
+		return ERROR_FAIL;
 	}
 
-	int exec_out = riscv_program_exec(&program, target);
-	/* Don't message on error. Probably the register doesn't exist. */
-	if (exec_out == ERROR_OK && target->reg_cache) {
-		struct reg *reg = &target->reg_cache->reg_list[number];
-		buf_set_u64(reg->value, 0, reg->size, value);
-	}
```
timsifive, line 1369: This seems like a big change in behavior. Is it handled elsewhere, or does it need to be part of the change description?

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1287,210 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
+	if (riscv_program_exec(&program, target) != ERROR_OK)
+		return ERROR_FAIL;
 
-	if (use_scratch)
-		scratch_release(target, &scratch);
+	return register_read_direct(target, value, GDB_REGNO_S0) != ERROR_OK;
+}
 
-	if (cleanup_after_register_access(target, mstatus, number) != ERROR_OK)
+static int internal_register_write64_progbuf_scratch(struct target *target,
```
timsifive, line 1387: Comment, please.

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1287,210 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
+	int result = riscv_program_exec(program, target);
 
-	return exec_out;
+	scratch_release(target, &scratch);
+	return result;
 }
 
-/** Actually read registers from the target right now. */
-static int register_read_direct(struct target *target, uint64_t *value, uint32_t number)
+static int register_write_progbuf(struct target *target, enum gdb_regno number,
```
timsifive, line 1414: Comment.

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1287,210 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
+		uint64_t value)
```
```suggestion
		riscv_reg_t value)
```

src/target/riscv/riscv-013.c
```diff
@@ -1688,6 +1720,23 @@ static int examine(struct target *target)
...
+		target->state = TARGET_HALTED;
+		target->debug_reason = DBG_REASON_DBGRQ;
+	}
+
+	/* FIXME: This is needed since register_read_direct relies on target->state
+	 * to work correctly, so, if target->state does not represent current state
+	 * of target, e.g. if a target is halted, but target->state is
+	 * TARGET_UNKNOWN, it can fail early, (e.g. accessing registers via program
+	 * buffer can not be done atomicaly on a running hart nither msattus can not
+	 * be prepared for a register access and then restored)
```
```suggestion
	 * TARGET_UNKNOWN, it can fail early, (e.g. accessing registers via program
	 * buffer can not be done atomically on a running hart because mstatus can not
	 * be prepared for a register access and then restored)
```

src/target/riscv/riscv-013.c
```diff
@@ -1688,6 +1720,23 @@ static int examine(struct target *target)
...
+	 */
+	const enum target_state saved_tgt_state = target->state;
+	const enum target_debug_reason saved_dbg_reason = target->debug_reason;
```
timsifive, line 1769: What happens if you don't restore the target state?

I think you can replace all this state changing with simply updating target->state and debug_reason after the call to riscv_get_hart_state() a few lines above.

src/target/riscv/riscv.c
```diff
@@ -1481,6 +1487,25 @@ int riscv_flush_registers(struct target *target)
...
+		return ERROR_OK;
+
+	LOG_TARGET_DEBUG(target, "");
+
+	if (internal_flush_registers(target, 0, target->reg_cache->num_regs)
+			!= ERROR_OK)
+		return ERROR_FAIL;
+	/* Writing registers in the first flush_loop can leave some registers dirty
+	 * (e.g. S0, S1), so we need to flush GPRs again
+	 */
```
timsifive: Can we avoid that by flushing the registers in reverse order? Might still want the extra pass to double-check, but I think that should solve the problem and save a few writes here.

src/target/riscv/riscv.c
```diff
@@ -4375,97 +4400,155 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
 	}
 }
 
-/**
- * This function is called when the debug user wants to change the value of a
- * register. The new value may be cached, and may not be written until the hart
- * is resumed. */
-int riscv_set_register(struct target *target, enum gdb_regno regid, riscv_reg_t value)
+static int riscv_set_or_write_register(struct target *target,
```
timsifive, line 4642: Comment? What is the difference between set and write?

src/target/riscv/riscv.c
```diff
@@ -4375,97 +4400,155 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
...
+				"Target not halted, writing to target: %s <- 0x%" PRIx64,
+				reg->name, value);
+		return r->set_register(target, regid, value);
+	}
+
+	const bool need_to_write = !reg->valid || reg->dirty ||
+		value != buf_get_u64(reg->value, 0, reg->size);
+	const bool cacheble = gdb_regno_cacheable(regid, need_to_write);
+
+	if (!cacheble || (write_request & need_to_write)) {
```
```suggestion
	if (!cacheble || (write_request && need_to_write)) {
```

src/target/riscv/riscv.c
```diff
@@ -4375,97 +4400,155 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
 	}
 }
 
-/**
- * This function is called when the debug user wants to change the value of a
- * register. The new value may be cached, and may not be written until the hart
- * is resumed. */
-int riscv_set_register(struct target *target, enum gdb_regno regid, riscv_reg_t value)
+static int riscv_set_or_write_register(struct target *target,
+		enum gdb_regno regid, riscv_reg_t value, bool write_request)
```
timsifive: What does write_request mean? Looks like it's more like must_write, or write_through?

timsifive (changes requested):
Can you make sure to run `make all` in riscv-tests/debug? That has coverage for at least some of the scratch RAM cases.

This is a large change, and I haven't entirely grokked it. Can you give an example or two of cases where registers are now cached that they wouldn't be previously?

src/target/riscv/riscv-013.c
```diff
@@ -1063,34 +1065,47 @@ static int is_vector_reg(uint32_t gdb_regno)
...
-				return ERROR_FAIL;
-		} else if (is_vector_reg(regno) && (*mstatus & MSTATUS_VS) == 0) {
-			if (register_write_direct(target, GDB_REGNO_MSTATUS,
-						set_field(*mstatus, MSTATUS_VS, 1)) != ERROR_OK)
-				return ERROR_FAIL;
-		}
-	} else {
-		*mstatus = 0;
+	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
+		return ERROR_OK;
```
JanMatCodasip: Detail: Just a little favor to whoever will read the code:
```suggestion
	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
		/* No special preparation needed */
		return ERROR_OK;
```

src/target/riscv/riscv-013.c
```diff
@@ -1063,34 +1065,47 @@ static int is_vector_reg(uint32_t gdb_regno)
...
 static int cleanup_after_register_access(struct target *target,
-		uint64_t mstatus, int regno)
+		riscv_reg_t mstatus, enum gdb_regno regno)
 {
-	if ((is_fpu_reg(regno) && (mstatus & MSTATUS_FS) == 0) ||
-			(is_vector_reg(regno) && (mstatus & MSTATUS_VS) == 0))
-		if (register_write_direct(target, GDB_REGNO_MSTATUS, mstatus) != ERROR_OK)
-			return ERROR_FAIL;
+	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
+		return ERROR_OK;
```
```suggestion
	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
		/* Mstatus was not changed for this register access. No need to restore it. */
		return ERROR_OK;
```

src/target/riscv/riscv-013.c
```diff
@@ -1063,34 +1065,47 @@ static int is_vector_reg(uint32_t gdb_regno)
...
+		return ERROR_FAIL;
+
+	LOG_TARGET_DEBUG(target, "Prepared to access %s (mstatus=0x%" PRIx64 ")",
+			gdb_regno_name(regno), new_mstatus);
 	return ERROR_OK;
 }
 
 static int cleanup_after_register_access(struct target *target,
-		uint64_t mstatus, int regno)
+		riscv_reg_t mstatus, enum gdb_regno regno)
```
JanMatCodasip, line 1118: Detail: This can be called `orig_mstatus`, if you'd like

```suggestion
		riscv_reg_t orig_mstatus, enum gdb_regno regno)
```

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Addressed.

src/target/riscv/riscv-013.c
en-sc: Addressed.

src/target/riscv/riscv-013.c
en-sc: Addressed.

src/target/riscv/riscv-013.c
en-sc: Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv.c
en-sc: Thanks! The solution you suggested is much nicer. Addressed.

src/target/riscv/riscv.c
en-sc: Addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: Renamed to `write_through` and commented on.

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1286,224 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
-	if (exec_out == ERROR_OK && target->reg_cache) {
-		struct reg *reg = &target->reg_cache->reg_list[number];
-		buf_set_u64(reg->value, 0, reg->size, value);
-	}
+	if (riscv_program_exec(&program, target) != ERROR_OK)
+		return ERROR_FAIL;
 
-	if (use_scratch)
-		scratch_release(target, &scratch);
+	return register_read_direct(target, value, GDB_REGNO_S0) != ERROR_OK;
```
JanMatCodasip: Please, use directly `register_read_abstract` here.

Otherwise this may look like to the reader (or to static analysis tools) like a mutual recursion between `register_read_direct` and `register_read_progbuf`.

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1286,224 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
+		 */
+		const unsigned int freg = number - GDB_REGNO_FPR0;
+
+		if (riscv_program_insert(&program, fld(freg, S0, 0)) != ERROR_OK)
+			return ERROR_FAIL;
+		return internal_register_write64_progbuf_scratch(target, &program,
+				value);
+	}
+
+	if (register_write_direct(target, GDB_REGNO_S0, value) != ERROR_OK)
```
JanMatCodasip: Similarly to the above: It'd be better to use `register_write_abstract()` here.

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1286,224 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
+static int internal_register_write64_progbuf_scratch(struct target *target,
+		struct riscv_program *program, riscv_reg_t value)
+{
+	scratch_mem_t scratch;
+
+	if (scratch_reserve(target, &scratch, program, 8) != ERROR_OK)
 		return ERROR_FAIL;
 
-	return exec_out;
+	if (register_write_direct(target, GDB_REGNO_S0, scratch.hart_address)
```
JanMatCodasip: Please, use register_write_abstract().

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1286,224 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
-	uint64_t mstatus;
-	if (prep_for_register_access(target, &mstatus, number) != ERROR_OK)
+	if (scratch_reserve(target, &scratch, program, 8) != ERROR_OK)
 		return ERROR_FAIL;
 
-	int result = register_write_abstract(target, number, value,
-			register_size(target, number));
-	if (result == ERROR_OK || !has_sufficient_progbuf(target, 2))
-		return result;
+	if (register_write_direct(target, GDB_REGNO_S0, scratch.hart_address)
```
JanMatCodasip: Please, use `register_write_abstract()`.

Can you please check all the progbuf-related functions, and replace register_*_direct() by register_*_abstract() where possible (that is, when the register is a GPR).

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1286,224 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
 	return info->progbufsize + r->impebreak >= size;
 }
 
-/**
- * Immediately write the new value to the requested register. This mechanism
- * bypasses any caches.
+/** This function recives a program that stores a register at a memory address
+ * stored at S0, allocates the memory needed, executes it and reads the value
+ * of the register.
  */
```
JanMatCodasip: Detail: Please start the comment text on the second line, not directly after `/**`.

Reasons:
- Consistency with other comments
- Consistency with the recommended doxygen style (https://www.doxygen.nl/manual/docblocks.html)

(There are multiple occurrences in this merge request - I won't comment on each.)

```suggestion
/** 
 * This function recives a program that stores a register at a memory address
 * stored at S0, allocates the memory needed, executes it and reads the value
 * of the register.
 */
```

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1286,224 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
-	if (result == ERROR_OK || !has_sufficient_progbuf(target, 2))
-		return result;
+	if (register_write_direct(target, GDB_REGNO_S0, scratch.hart_address)
+			!= ERROR_OK) {
+		scratch_release(target, &scratch);
+		return ERROR_FAIL;
+	}
+
+	if (riscv_program_exec(program, target) != ERROR_OK)
+		return ERROR_FAIL;
```
JanMatCodasip, line 1327: Is `scratch_release()` missing here before the `return`?

src/target/riscv/riscv-013.c
JanMatCodasip: Thanks for the comment. I'd suggest slightly different wording:

```suggestion
/** 
 * This function allocates 64 bits of scratch memory, writes its address to s0 and executes program buffer.
 * Then it reads the value of the stratch memory.
 * The scratch memory is deallocated at the end.
 *
 * Use case: Reading of D-extension registers (64-bit) on RV32, which needs to be done through scratch memory.
 *
 * The caller should save s0.
 * The caller should prepare a suitable program that fills the scratch memory location.
 */
```

src/target/riscv/riscv-013.c
```diff
@@ -1272,193 +1286,224 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
-	if (use_scratch)
-		scratch_release(target, &scratch);
+	return register_read_direct(target, value, GDB_REGNO_S0) != ERROR_OK;
+}
 
-	if (cleanup_after_register_access(target, mstatus, number) != ERROR_OK)
+/** This function recives a program that writes a value stored at a memory
+ * address from S0 to a register, allocates the memory needed, writes the value
+ * to the memory and executes the program.
+ */
```
JanMatCodasip: Similarly to the above, I'd suggest different wording:

```suggestion
/** 
 * This function allocates 64 bits of scratch memory, fills it with the provided `value`,
 * fills the stratch address into register s0 and executes program buffer.
 * The scratch memory is deallocated at the end.
 *
 * Use case: Writing of D-extension registers (64-bit) on RV32, which needs to be done through scratch memory.
 *
 * The caller should save s0.
 * The caller should prepare a suitable program that stores the value from the memory to the register.
 */
```

src/target/riscv/riscv-013.c
```diff
@@ -1688,6 +1733,23 @@ static int examine(struct target *target)
...
+	 * TARGET_UNKNOWN, it can fail early, (e.g. accessing registers via program
+	 * buffer can not be done atomically on a running hart becuse mstatus can't
+	 * be prepared for a register access and then restored)
+	 */
+	const enum target_state saved_tgt_state = target->state;
+	const enum target_debug_reason saved_dbg_reason = target->debug_reason;
+
+	if (target->state != TARGET_HALTED) {
+		target->state = TARGET_HALTED;
+		target->debug_reason = DBG_REASON_DBGRQ;
```
JanMatCodasip: IIUC, at this point the target is guaranteed to be halted. So is it possible to simply change the `target->state` directly (as shonw below)?

I don't think it is necessary to store the previous state of the target.

```suggestion
	}

	/* Record that the target is now halted. This is needed for register 
	   access functions that work with progbuf to operate. */
	target->state = TARGET_HALTED;
	target->debug_reason = DBG_REASON_DBGRQ;
```

src/target/riscv/riscv-013.c
```diff
@@ -1759,6 +1821,9 @@ static int examine(struct target *target)
 		return ERROR_FAIL;
 	}
 
+	target->state = saved_tgt_state;
+	target->debug_reason = saved_dbg_reason;
+
 	if (!halted) {
 		riscv013_step_or_resume_current_hart(target, false);
 		target->state = TARGET_RUNNING;
```
JanMatCodasip, line 1849: I'd recommend something along these lines:

```
	if (!halted) {
		/* The hart was running before the examine. So resume it now that 
		   examination is complete. */
	}
	else {
		/* The hart was already halted before the examine. Remain halted.
		   Let the periodic poll function interrogate the hart and 
		   determine the exact halt cause. */
		target->state = TARGET_UNKNOWN;
		target->state = DBG_REASON_UNDEFINED;
	}
```

Would this work?

src/target/riscv/riscv.c
```diff
@@ -4886,15 +4974,19 @@ static int register_set(struct reg *reg, uint8_t *buf)
...
 			return ERROR_FAIL;
+
+		memcpy(reg->value, buf, DIV_ROUND_UP(reg->size, 8));
+		reg->valid = gdb_regno_cacheable(reg->number, /* is write? */ true);
 	} else {
-		uint64_t value = buf_get_u64(buf, 0, reg->size);
+		const riscv_reg_t value = buf_get_u64(buf, 0, reg->size);
 		if (riscv_set_register(target, reg->number, value) != ERROR_OK)
 			return ERROR_FAIL;
 	}
```
JanMatCodasip, line 5244: In `riscv_set_register`, the cache is handled internally. However, `r->set_register_buf` does not contain that logic.

Could you please add it, so that these two functions are symmetric?

src/target/riscv/riscv.h
```diff
@@ -370,8 +373,10 @@ int riscv_current_hartid(const struct target *target);
  * consecutive and start with mhartid=0. */
 unsigned int riscv_count_harts(struct target *target);
 
-/** Set register, updating the cache. */
+/** Set register, if possible, only updating the cache.*/
 int riscv_set_register(struct target *target, enum gdb_regno i, riscv_reg_t v);
+/** Write register to target, updating the cache. */
+int riscv_write_register(struct target *target, enum gdb_regno i, riscv_reg_t v);
```
JanMatCodasip: I tried to make the comments even more clear:

```suggestion
/** Set the register value. For cacheable registers, only the cache is updated (write-back mode). */
int riscv_set_register(struct target *target, enum gdb_regno i, riscv_reg_t v);
/** Set the register value and immediately write it to the target (write-through mode). 
    Update the cache (for cacheable registers only). */
int riscv_write_register(struct target *target, enum gdb_regno i, riscv_reg_t v);
```

src/target/riscv/riscv.c
```diff
@@ -4376,96 +4388,160 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
...
 
+/**
+ * This function is used to save the value of a register in cache. The register
+ * is marked as dirty, and writeback is delayed for as long as possible.
+ */
 int riscv_save_register(struct target *target, enum gdb_regno regid)
 {
 	RISCV_INFO(r);
 	riscv_reg_t value;
 	if (!target->reg_cache) {
```
JanMatCodasip, line 4780: While touching this function, it would be good to include this safety check:

```suggestion
	/* Only cacheable registers can be saved. */
	assert(gdb_regno_cacheable(regid, false);

	RISCV_INFO(r);
	riscv_reg_t value;
	if (!target->reg_cache) {
```

src/target/riscv/riscv.c
```diff
@@ -4376,96 +4388,160 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
...
-	return result;
+	LOG_TARGET_DEBUG(target, "Read %s: 0x%" PRIx64, reg->name, *value);
+	return ERROR_OK;
 }
 
+/**
+ * This function is used to save the value of a register in cache. The register
+ * is marked as dirty, and writeback is delayed for as long as possible.
+ */
 int riscv_save_register(struct target *target, enum gdb_regno regid)
```
JanMatCodasip, line 4771: The `gdb_regno` arguments are sometimes called `regid`, `rid` and at other times `regno`.

If you have a moment, this would be a good opportunity to unify the parameter names (preferrably `regno`).

JanMatCodasip (changes requested):
These changes look really fine :+1:

Thank you for taking the time to improve the code.

I have only smaller suggestions.

src/target/riscv/riscv.c
timsifive: Let's put that off for another PR. This PR is already cluttered with cleanups. Adding more just makes it harder to review.

src/target/riscv/riscv.c
```diff
@@ -4376,96 +4388,160 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
 }
 
 /**
- * This function is called when the debug user wants to change the value of a
- * register. The new value may be cached, and may not be written until the hart
- * is resumed. */
-int riscv_set_register(struct target *target, enum gdb_regno regid, riscv_reg_t value)
+ * This function is used internally by functions that change register values.
+ * If `write_trough` is true, it is ensured that the value of the target's
```
```suggestion
 * If `write_through` is true, it is ensured that the value of the target's
```

src/target/riscv/riscv.c
timsifive: I think that happens below in the reg->valid check. Probably the assert is OK since whether registers can be cached is static. And I do like being more explicit about it.

src/target/riscv/riscv.c
JanMatCodasip: Agreed, let's do this in a separate merge request.

src/target/riscv/riscv.c
JanMatCodasip: Good point, I missed that `if (!reg->valid)` check below. 

Still, let's please be explicit and catch that potential programmer error as soon as possible.

The `if (!reg->valid)` check can be replaced by an assertion, too (I've posted a separate comment).

src/target/riscv/riscv.c
```diff
@@ -4376,96 +4388,160 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
...
+
+	LOG_TARGET_DEBUG(target, "Saving %s", reg->name);
 	if (riscv_get_register(target, &value, regid) != ERROR_OK)
 		return ERROR_FAIL;
 
-	if (!reg->valid)
+	if (!reg->valid) {
+		LOG_TARGET_DEBUG(target, "%s can't be cached", reg->name);
 		return ERROR_FAIL;
+	}
```
JanMatCodasip: As noted above, please replace this by an assert. It is an invariant.

```suggestion
	/* The register is cacheable, so the cache entry must be valid now. */
	assert(reg->is_valid);

```

src/target/riscv/riscv.c
JanMatCodasip, line 4780: For safety, this check can be added, too:

```suggestion
	/* Doesn't make sense to populate register cache on non-halted targets. */
	assert(target->state == TARGET_HALTED);

	if (!target->reg_cache) {
```

en-sc:
> This is a large change, and I haven't entirely grokked it. Can you give an example or two of cases where registers are now cached that they wouldn't be previously?

* This is an annotated log of vector register write prior to the change (this is a second write in a row with `mstatus` set to `0x8000000a00000600` before accessing the registers):
```
Debug: 2400 469 riscv.c:4861 register_set(): [riscv.cpu] write 0xfffffffffffffffc0000000000000004 to v1 (valid=0)
...
//Reading mstatus from hart as this is a vector register access
...
Debug: 2412 470 riscv-013.c:1472 register_read_direct(): [riscv.cpu] mstatus = 0x8000000a00000600
...
//Reading mstatus from hart to prepare to access vtype
...
Debug: 2422 471 riscv-013.c:1472 register_read_direct(): [riscv.cpu] mstatus = 0x8000000a00000600
...
//Reading vtype
...
Debug: 2432 471 riscv-013.c:1472 register_read_direct(): [riscv.cpu] vtype = 0x18
...
//Reading mstatus from hart to prepare to read vl
...
Debug: 2442 472 riscv-013.c:1472 register_read_direct(): [riscv.cpu] mstatus = 0x8000000a00000600
...
//Reading vl
...
Debug: 2452 472 riscv-013.c:1472 register_read_direct(): [riscv.cpu] vl = 0x2
Debug: 2453 472 riscv-013.c:1293 register_write_direct(): [riscv.cpu] vtype <- 0x18
...
//Reading mstatus from hart to prepare to read vtype
...
Debug: 2463 473 riscv-013.c:1472 register_read_direct(): [riscv.cpu] mstatus = 0x8000000a00000600
...
//Wrote 0x18 to vtype
...
Debug: 2505 476 riscv-013.c:1293 register_write_direct(): [riscv.cpu] vl <- 0x2
...
//Reading mstatus from hart to prepare to write vl
...
Debug: 2515 477 riscv-013.c:1472 register_read_direct(): [riscv.cpu] mstatus = 0x8000000a00000600
...
//Wrote  0x2 to vl
...
//Writing the vector register
...
Debug: 2599 481 riscv-013.c:1293 register_write_direct(): [riscv.cpu] vtype <- 0x18
...
//Reading mstatus from hart to prepare to write vtype
...
Debug: 2609 481 riscv-013.c:1472 register_read_direct(): [riscv.cpu] mstatus = 0x8000000a00000600
...
//vtype restored 
...
Debug: 2651 483 riscv-013.c:1293 register_write_direct(): [riscv.cpu] vl <- 0x2
...
//Reading mstatus from hart to prepare to write vl
...
Debug: 2661 485 riscv-013.c:1472 register_read_direct(): [riscv.cpu] mstatus = 0x8000000a00000600
...
//vl restored
...
User : 2703 488 options.c:63 configuration_output_handler(): 0:v1 (/128): 0xfffffffffffffffc0000000000000004
```
* And this is the same log after the change:
``` 
Debug: 2235 539 riscv.c:4951 register_set(): [riscv.cpu] write 0xfffffffffffffffc0000000000000004 to v1 (valid=0)
...
Debug: 2238 539 riscv-013.c:1085 prep_for_register_access(): [riscv.cpu] Preparing mstatus to access vl
Debug: 2239 539 riscv.c:4498 riscv_get_register(): [riscv.cpu] Read mstatus: 0x8000000a00000600 (cached)
Debug: 2240 539 riscv.c:4498 riscv_get_register(): [riscv.cpu] Read vtype: 0x18 (cached)
Debug: 2241 539 riscv.c:4498 riscv_get_register(): [riscv.cpu] Read vl: 0x2 (cached)
Debug: 2242 539 riscv.c:4443 riscv_set_or_write_register(): [riscv.cpu] Wrote 0x18 to vtype (valid=true, dirty=false)
Debug: 2243 539 riscv.c:4443 riscv_set_or_write_register(): [riscv.cpu] Wrote 0x2 to vl (valid=true, dirty=false)
...
//Reading the vector register
...
Debug: 2287 540 riscv.c:4443 riscv_set_or_write_register(): [riscv.cpu] Wrote 0x18 to vtype (valid=true, dirty=false)
Debug: 2288 540 riscv.c:4443 riscv_set_or_write_register(): [riscv.cpu] Wrote 0x2 to vl (valid=true, dirty=false)
Debug: 2289 540 riscv-013.c:1118 cleanup_after_register_access(): [riscv.cpu] Restoring mstatus to 0x8000000a00000600
Debug: 2290 540 riscv.c:4443 riscv_set_or_write_register(): [riscv.cpu] Wrote 0x8000000a00000600 to mstatus (valid=true, dirty=false)
User : 2291 540 options.c:63 configuration_output_handler(): 0:v1 (/128): 0xfffffffffffffffc0000000000000004
```

### To sum up, after the change:
* Writes to a register of the same value as the one present in the cache do not write to hart, in particular to a WARL register (MSTATUS, VL and VTYPE in the example above). This allows reading MSTATUS only ones, when no change is needed.

* Using the `riscv_get_register` and the new `riscv_write_register` functions allows MSTATUS to be read at most twice and written once during preparation for vector register access (read to get the value, write to set it, read to get the updated value, since it is a WARL register).

* Similarly, the number of accesses to VL and VTYPE is reduced.

src/target/riscv/riscv.c
TommyMurphyTM1234: > Writes to a register of the same value as the one present in the cache do not write to hart

Is is definitely and always the case that there are no registers that cause side effects when the value already stored is written again and, for which, ignoring such writes will result in incorrect behaviour?

TommyMurphyTM1234:
> Writes to a register of the same value as the one present in the cache do not write to hart

Is is definitely and always the case that there are no registers that cause side effects when the value already stored is written again and, for which, ignoring such writes will result in incorrect behaviour?

en-sc:
> Can you make sure to run `make all` in riscv-tests/debug? That has coverage for at least some of the scratch RAM cases.

I've run all the tests from `make all`. Tests on the spike-multy platform hang and does not finish. I've gotten the same behavior without my changes. This could be an issue with spike/gdb version. Can you please provide me with info on which versions these tests were passing before?

en-sc:
> > Writes to a register of the same value as the one present in the cache do not write to hart
> 
> Is is definitely and always the case that there are no registers that cause side effects when the value already stored is written again and, for which, ignoring such writes will result in incorrect behaviour?

This question interests me as well. My understanding is that there isn't one, but I'm not sure.

src/target/riscv/riscv.c
en-sc: I would like for accessing vector and non-vector registers to be symmetrical. But there is an issue:
`riscv_set_register` handles the cache internally, not `r->set_register`. Moreover, `r->set_register_buf` is defined in `riscv-013.c` where currently there is no cache handling at all, so I'm not sure it would be better.
One could move it to a separate function, like `riscv_set_register_buf`, but I'm not sure if it is necessary - it would be a very short function called only here.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Addressed.

src/target/riscv/riscv-013.c
en-sc: Thanks for the suggestion! Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Thanks for the catch! Addressed.

src/target/riscv/riscv-013.c
en-sc: IMHO, this is an excessively long comment for such a small and local function. However, my wording was really poor. I have rewritten the comment.

src/target/riscv/riscv-013.c
en-sc: Addressed, similarly to the above.

src/target/riscv/riscv.h

src/target/riscv/riscv.c
en-sc: Addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: Addressed.

src/target/riscv/riscv-013.c
en-sc: Are you referring to not setting the `reg_cahce` to value? If I understand correctly it is handled by:
```
/* TODO: Hack to deal with gdb that thinks these registers still exist. */
...
```
in `register_set/get`.
I'm not sure if this hack is still needed, but I don't know which GDB versions should be supported.

src/target/riscv/riscv-013.c
en-sc: I'm not sure that the debug reason would be set correctly in this case, since logic for setting it in `riscv_poll_hart` is quite complex. We (at Syntacore) are looking to refactor `examine/poll/halt/resume` and address this in future commits.

src/target/riscv/riscv-013.c
en-sc: As far as I understand, this would work, but I'm not quite sure. My solution does not change the behavior, so it most definitely works. Please, see [this comment](https://github.com/riscv/riscv-openocd/pull/842#discussion_r1182675115).

src/target/riscv/riscv-013.c
en-sc: Addressed [below](https://github.com/riscv/riscv-openocd/pull/842#discussion_r1182698676)

src/target/riscv/riscv-013.c
timsifive: I was referring to the comment "Don't message on error. Probably the register doesn't exist."

But looking at the code, you just removed the comment (which wasn't super helpful) and didn't change the messaging behavior. So I think this is fine.

src/target/riscv/riscv-013.c
timsifive: Almost all that complexity is to make sure the world looks as gdb needs it to. I don't think gdb cares about the debug reason when first connecting.

I just don't like this save/restore, which only exists to preserve existing behavior, which itself is already hack. It's hacks upon hacks.

If you want to set the halt reason correctly, you can probably just use:
```
				enum riscv_halt_reason halt_reason = riscv_halt_reason(target);
				if (set_debug_reason(target, halt_reason) != ERROR_OK)
					return ERROR_FAIL;
```

src/target/riscv/riscv-013.c
JanMatCodasip: Looks all right - thanks.

src/target/riscv/riscv.h
```diff
@@ -370,8 +373,14 @@ int riscv_current_hartid(const struct target *target);
  * consecutive and start with mhartid=0. */
 unsigned int riscv_count_harts(struct target *target);
 
-/** Set register, updating the cache. */
+/** Set the register value. For cacheable registers, only the cache is updated
+ *  (write-back mode).
+ */
```
JanMatCodasip: Detail - style of the doxygen comments:

```suggestion
/** 
 * Set the register value. For cacheable registers, only the cache is updated
 * (write-back mode).
 */
```

src/target/riscv/riscv.h
```diff
@@ -370,8 +373,14 @@ int riscv_current_hartid(const struct target *target);
...
 int riscv_set_register(struct target *target, enum gdb_regno i, riscv_reg_t v);
+/** Set the register value and immediately write it to the target
+ *  (write-through mode).
+ */
```
JanMatCodasip: (same as above)

```suggestion
/** 
 * Set the register value and immediately write it to the target
 * (write-through mode).
 */
```

src/target/riscv/riscv.c
JanMatCodasip: Ok, feel free to leave the current code as is, then. This can be refactored later.

src/target/riscv/riscv-013.c
```diff
@@ -913,10 +918,21 @@ static int register_read_abstract(struct target *target, uint64_t *value,
...
 
-static int register_write_abstract(struct target *target, uint32_t number,
-		uint64_t value, unsigned size)
+static unsigned int register_size(struct target *target, enum gdb_regno number);
+
+static int register_read_abstract(struct target *target, riscv_reg_t *value,
+		enum gdb_regno number)
+{
+	const unsigned int size = register_size(target, number);
+
```
JanMatCodasip, line 923: Detail: This blank line can be removed.

```suggestion
```

src/target/riscv/riscv-013.c
JanMatCodasip: For clarity, let's continue the discussion about the target's state under the below comment.

src/target/riscv/riscv-013.c
JanMatCodasip: I also recommend to avoid the save/restore. 

Instead, please set the target state directly based on whether we resume (or keep the hart halted) at the end of the examine.

I like Tim's suggestion: for harts that remain halted, get the debug reason via the code snippet above. It is possible that GDB does not actually need the reason upon first connect, but for consistency (and as a defensive programming measure), I recommend to do collect the reason.

(My earlier proposal with TARGET_UNKNOWN would probably work, but the suggestion above is better.)

src/target/riscv/riscv-013.c
aap-sc: @en-sc , @timsifive  I think that if access to a register requires some bits in mstatus register to be set we should always return an error if the target is not halted. The reason is that we don't have transactional semantics here. Meaning, that even if we've seen that FS is set it does not mean that it will still be set by the time  we actually  perform  the access. (and this can lead to nasty results as described here: https://github.com/riscv/riscv-openocd/pull/766)

src/target/riscv/riscv-013.c
en-sc: After removing this line and running `checkpatch`:
```
WARNING: Missing a blank line after declarations
#91: FILE: src/target/riscv/riscv-013.c:927:
+       const unsigned int size = register_size(target, number);
+       return register_read_abstract_with_size(target, value, number, size);
```
So, if it's not crucial, I'd like to leave it as-is.

src/target/riscv/riscv.h

src/target/riscv/riscv.h

src/target/riscv/riscv-013.c
en-sc: Addressed.

src/target/riscv/riscv-013.c
JanMatCodasip: Ah, I didn't know checkpatch requires blank line. 

So yes, please feel free to leave the code as is.

src/target/riscv/riscv-013.c
timsifive: @aap-sc I agree, and I believe that is what the new code does. See https://github.com/riscv/riscv-openocd/pull/842/files#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63R1101

src/target/riscv/riscv-013.c
timsifive: This looks good. Thanks.

src/target/riscv/riscv.c
```diff
@@ -1481,8 +1493,7 @@ int riscv_flush_registers(struct target *target)
 	return ERROR_OK;
 }
 
-/* Convert: RISC-V hart's halt reason --> OpenOCD's generic debug reason */
```
timsifive, line 1523: Can you make this comment correct instead of deleting it? Something like:
Set OpenOCD's generic debug reason from the RISC-V halt reason.

timsifive:
Thanks for showing the difference in vector accesses. That's a very nice improvement.

> I've run all the tests from make all. Tests on the spike-multy platform hang and does not finish. I've gotten the same behavior without my changes. This could be an issue with spike/gdb version. Can you please provide me with info on which versions these tests were passing before?

Last I really looked, spike-multi passed for me but it was hugely slower than it used to be. I'm not sure what happened, and I've been ignoring the problem. They're really just to test connecting OpenOCD RV32 and RV64 targets. I don't think your code would have been affected by that.

src/target/riscv/riscv-013.c
JanMatCodasip: I've checked the current code visually and it also looks good to me. Thank you.

src/target/riscv/riscv.c
```diff
@@ -4376,96 +4387,163 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
...
+/**
+ * This function is used to save the value of a register in cache. The register
+ * is marked as dirty, and writeback is delayed for as long as possible.
+ */
 int riscv_save_register(struct target *target, enum gdb_regno regid)
 {
+	assert(target->state == TARGET_HALTED &&
+			"Doesn't make sense to populate register cache on non-halted targets.");
+	assert(gdb_regno_cacheable(regid, /* is write? */ false) &&
+			"Only cacheable registers can be saved.");
```
JanMatCodasip, line 4776: I am not aware that this "hack" to [include the textual message into the assert](https://stackoverflow.com/a/3692961) would be used anywhere else in OpenOCD.

For that reason, and to make the future upstreaming more smooth, I'd recommend to not introduce this new pattern.

I think it is more than sufficient to have the texutal explanation in an ordinary comment.

```suggestion
	/* Doesn't make sense to populate register cache on non-halted targets. */
	assert(target->state == TARGET_HALTED);
	/* Only cacheable registers can be saved. */
	assert(gdb_regno_cacheable(regid, /* is write? */ false));
```

timsifive: FWIW, I'm fine with this. If upstream complains then we can easily translate them back. This style seems better to me because when the user sees the assert at least there's some indication on what went wrong.

JanMatCodasip: I'm not happy to see this creative abuse of `&&` inside the asserts, but granted, the error message that gets displayed is added value. And this pattern doesn't seem to be totally uncommon (there e.g. several uses in the jimtcl library, used by OpenOCD). So please feel free to leave it there, and we will see later whether there will be any push-back from the upstream.

As a related note: Most of OpenOCD code apparently follows a different style: `LOG_ERROR("BUG: ...")` is used in places where common `assert`s would seem to be more fitting.

src/target/riscv/riscv-013.c
en-sc: If I understand correctly, the issue is similar to the following:
1. target is running.
2. `mstatus` is read by OpenOCD, `mstatus.fs != 0`
3. the code being executed on the target sets `mstatus.fs = 0`
4. OpenOCD writes to a FPU register

This would create a situation similar to #766 (not exactly the same, since there the target was halted).
I see this as a valid justification for rejecting access to a register on a running target, that could potentially need a modification to `mstatus` , even when this modification is perceived unnecessary, based on the current value of `mstatus`.

However, some target could implement abstract command accesses to FPU registers on running target. There would be no way to leverage this feature by an OpenOCD user.

IMHO, the proper way to tackle this would be:
1. To fail in `prep_for_register_access` if target is not halted prior to reading `mstatus`
2. To execute an extra abstract access prior to prep_for_register_access on running target -- This would allow access to an FPU register on a running target if mstatus does not need to be changed, while also preserving the WA for  #766. I would like to address this in another commit since this is a major change in behavior that needs to be tested on hardware.

src/target/riscv/riscv.c
en-sc: Addressed.

timsifive: You say it's addressed but I still see the comment here. What's going on?

src/target/riscv/riscv-013.c
timsifive: Your solution of trying and then updating mstatus if necessary (assuming the target is currently running) is a good one. I also agree that it's large enough to warrant a separate PR.

However, from this PR please remove the is-running check. Otherwise this PR causes a regression where accessing FPRs while running no longer works when it used to work.

src/target/riscv/riscv.c
en-sc: If I understood correctly, the suggestion was to rephrase the comment. I did so in [defenition](https://github.com/riscv/riscv-openocd/pull/842/files#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR1496-R1498) and [declaration](https://github.com/riscv/riscv-openocd/pull/842/files#diff-6b0d494208e1b1a563ca5a3ea6c4ee820028e589f3b8d69613868d9c8e169e13R398-R400). Should I remove it from the definition?

src/target/riscv/riscv-013.c

src/target/riscv/riscv.c
timsifive: Yes. That was sloppy reading on my part. Sorry.

timsifive (dismissed):
Thanks for sticking with all the review...

en-sc:
Please, do not merge it yet: there is a bug. Will fix soon.
Update: Bug is fixed. Please, see [this](https://github.com/riscv/riscv-openocd/pull/842#discussion_r1197639330) for details.

src/target/riscv/riscv-013.c
```diff
@@ -875,8 +877,11 @@ static uint32_t access_register_command(struct target *target, uint32_t number,
 	return command;
 }
 
-static int register_read_abstract(struct target *target, uint64_t *value,
-		uint32_t number, unsigned size)
+/**
+ * This function is used explicitly to determine xlen during examine.
+ */
```
JanMatCodasip: I'd recommend to drop this comment. That's because this function is used also for all abstract reads (indirectly from `register_read_abstract`), not just exclusively during examine.

```suggestion
```

src/target/riscv/riscv-013.c
```diff
@@ -913,10 +918,21 @@ static int register_read_abstract(struct target *target, uint64_t *value,
 	return ERROR_OK;
 }
 
-static int register_write_abstract(struct target *target, uint32_t number,
-		uint64_t value, unsigned size)
+static unsigned int register_size(struct target *target, enum gdb_regno number);
```
JanMatCodasip: Please, move this forward declaration to the top of the file - to the same place where the other forward declarations are.

src/target/riscv/riscv-013.c
```diff
@@ -1063,35 +1079,52 @@ static int is_vector_reg(uint32_t gdb_regno)
...
-				return ERROR_FAIL;
-		} else if (is_vector_reg(regno) && (*mstatus & MSTATUS_VS) == 0) {
-			if (register_write_direct(target, GDB_REGNO_MSTATUS,
-						set_field(*mstatus, MSTATUS_VS, 1)) != ERROR_OK)
-				return ERROR_FAIL;
-		}
-	} else {
-		*mstatus = 0;
-	}
+	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
```
JanMatCodasip, line 1083: Detail: For additional NULL safety, you can add an assert here:

```suggestion
	assert(orig_mstatus);

	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
```

src/target/riscv/riscv-013.c
```diff
@@ -1273,192 +1306,228 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
+		if (riscv_program_csrw(&program, S0, number) != ERROR_OK)
+			return ERROR_FAIL;
+	} else {
+		LOG_ERROR("Unsupported register (enum gdb_regno)(%d)", number);
+		return ERROR_FAIL;
+	}
+	return riscv_program_exec(&program, target);
+}
+
+
```
JanMatCodasip: Detail: You can remove the extra blank line here. 

```suggestion
```

src/target/riscv/riscv-013.c
```diff
@@ -1273,192 +1306,228 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
+	riscv_program_init(&program, target);
+
+	if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31 &&
+			riscv_supports_extension(target, 'D') && riscv_xlen(target) < 64) {
+		/* There are no instructions to move all the bits from a register,
+		 * so we need to use some scratch RAM.
+		 */
```
JanMatCodasip, line 1430: Detail: No need to have the closing `*/` on a new line.

```suggestion
		/* There are no instructions to move all the bits from a register,
		 * so we need to use some scratch RAM. */
```

src/target/riscv/riscv-013.c
```diff
@@ -1273,192 +1306,228 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
+			return ERROR_FAIL;
+		if (riscv_program_insert(&program, csrr(S1, CSR_VL)) != ERROR_OK)
+			return ERROR_FAIL;
+		if (riscv_program_insert(&program, vsetvl(ZERO, S1, S0)) != ERROR_OK)
+			return ERROR_FAIL;
+	} else if (number == GDB_REGNO_VL) {
+		/* "The XLEN-bit-wide read-only vl CSR can only be updated by the
+		 * vsetvli and vsetvl instructions, and the fault-only-rst vector
+		 * load instruction variants."
+		 */
```
JanMatCodasip, line 1460: (same as above)

```suggestion
		/* "The XLEN-bit-wide read-only vl CSR can only be updated by the
		 * vsetvli and vsetvl instructions, and the fault-only-rst vector
		 * load instruction variants." */
```

src/target/riscv/riscv-013.c
```diff
@@ -1273,192 +1306,228 @@ static bool has_sufficient_progbuf(struct target *target, unsigned size)
...
-			if (riscv_program_insert(&program, csrr(S1, CSR_VTYPE)) != ERROR_OK)
-				return ERROR_FAIL;
-			if (riscv_program_insert(&program, vsetvl(ZERO, S0, S1)) != ERROR_OK)
-				return ERROR_FAIL;
-		} else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095) {
-			if (riscv_program_csrw(&program, S0, number) != ERROR_OK)
+		if (riscv_supports_extension(target, 'D') && riscv_xlen(target) < 64) {
+			/* There are no instructions to move all the bits from a
+			 * register, so we need to use some scratch RAM.
+			 */
```
JanMatCodasip, line 1357: (same as above)

```suggestion
			/* There are no instructions to move all the bits from a
			 * register, so we need to use some scratch RAM. */
```

src/target/riscv/riscv.c
```diff
@@ -4376,96 +4390,163 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
...
+
 	RISCV_INFO(r);
 	riscv_reg_t value;
 	if (!target->reg_cache) {
 		assert(!target_was_examined(target));
+		/* To create register cache it is needed to examine the target first,
+		 * therefore during examine, any needed register should be saved
+		 * manually
+		 */
```
JanMatCodasip: Very minor reword:

```suggestion
		/* To create register cache, it is needed to examine the target first.
		 * Therefore, during examine, any changed register needs to be saved
		 * and restored manually. */
```

src/target/riscv/riscv-013.c
```diff
@@ -1688,13 +1757,29 @@ static int examine(struct target *target)
...
+	/* FIXME: This is needed since register_read_direct relies on target->state
+	 * to work correctly, so, if target->state does not represent current state
+	 * of target, e.g. if a target is halted, but target->state is
+	 * TARGET_UNKNOWN, it can fail early, (e.g. accessing registers via program
+	 * buffer can not be done atomically on a running hart becuse mstatus can't
+	 * be prepared for a register access and then restored)
+	 * See https://github.com/riscv/riscv-openocd/pull/842#discussion_r1179414089
+	 */
+	const enum target_state saved_tgt_state = target->state;
+	const enum target_debug_reason saved_dbg_reason = target->debug_reason;
```
JanMatCodasip, line 1769: In an earlier version of this merge request that I have reviewed, this hack to save & restore of the target state had been already removed and done differently (as [discussed here](https://github.com/riscv/riscv-openocd/pull/842/files#r1183104333)). I was happy with that version.

Now the save & restore is back. Can you please double-check if that is an intentional change. Or did the older code get back just by mistake?

src/target/riscv/riscv.c
```diff
@@ -1456,23 +1456,35 @@ static int old_or_new_riscv_poll(struct target *target)
...
 		return ERROR_OK;
 
-	LOG_DEBUG("[%s]", target_name(target));
+	LOG_TARGET_DEBUG(target, "");
 
-	for (uint32_t number = 0; number < target->reg_cache->num_regs; number++) {
-		struct reg *reg = &target->reg_cache->reg_list[number];
+	/* Since writing a register can make some GPRs dirty (e.g. S0, S1),
+	 * registers should be flushed in reverse order, so GPRs are flushed last.
+	 */
```
JanMatCodasip: Since this is an important piece of information, I've tried to make it even more clear:

```suggestion
	/* Writing non-GPR registers may require progbuf execution, and some GPRs may 
	 * become dirty in the process (e.g. S0, S1). For that reason, flush registers in 
	 * reverse order, so that GPRs are flushed last. */
```

JanMatCodasip (changes requested):
@en-sc Thank you for addressing the review feedback so far.

I have visually re-checked the merge request once again and it looks fine. In summary, I have only:
- Few minor suggestions about comments & code style.
- One doubt about the save & restore of target state in `examine()`. I think this part had been reworked in a better way already, but now this save & restore is back. Why? :)

src/target/riscv/riscv-013.c
en-sc: When I try to delete this newline there is a warning from `checkpatch`:
```
git diff --patch github/en-sc/optimize-access | filterdiff -x ... | ~/openocd/tools/scripts/checkpatch.pl --no-signoff -
WARNING: Block comments use a trailing */ on a separate line
#18: FILE: src/target/riscv/riscv-013.c:1431:
+                * so we need to use some scratch RAM.*/

total: 0 errors, 1 warnings, 20 lines checked
...
```
If you insist, I'm open to accepting your suggestion, but shouldn't then `.github/workflows/checkpatch.yml` and/or `tools/scripts/checkpatch.pl` be adjusted?

src/target/riscv/riscv-013.c
en-sc: Please, see [this comment](https://github.com/riscv/riscv-openocd/pull/842#discussion_r1197606649).

src/target/riscv/riscv-013.c
en-sc: Please, see [this comment](https://github.com/riscv/riscv-openocd/pull/842#discussion_r1197606649).

src/target/riscv/riscv-013.c
en-sc: This is an intentional revert. Yesterday I've found a bug in prior implementation:
if `dcsr.cause == CSR_DCSR_CAUSE_TRIGGER`, calling `riscv013_halt_reason` will result in a call to `riscv_enumerate_triggers`, which is defined in `riscv.c` (universal for 0.11 and 0.13), so it uses `reg_cache` to determine size of `tselect`, and, since `reg_cache` is not initialized yet, gets `-1`:
```
Debug: 136 5 riscv.c:4494 riscv_get_register(): [riscv.cpu] No cache, reading tselect from target
Debug: 137 5 riscv-013.c:4178 riscv013_get_register(): [riscv.cpu] reading register tselect
Debug: 138 5 riscv-013.c:1529 register_read_direct(): [riscv.cpu] Reading tselect
Error: 139 5 riscv-013.c:868 access_register_command(): [riscv.cpu] -1-bit register tselect not supported.
```
I was going to explain this today, but didn't finish it in time for your review)

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv-013.c
en-sc: The change was reverted. Please see [this](https://github.com/riscv/riscv-openocd/pull/842#discussion_r1197639330)

src/target/riscv/riscv-013.c
en-sc: Your [suggestion](https://github.com/riscv/riscv-openocd/pull/842#discussion_r1180436311) will probably work, but there are no tests, either in `riscv-tests/debug` nor in our (at Syntacore) internal testing, which would explicitly target examine, so, I would like to stick with current solution, since it definitely won't change the behavior and introduce new bugs.

JanMatCodasip: @en-sc Thank you for the explanation, understood.

In that case, I am OK with keeping the current save & restore solution temporarily. This merge request is already large enough. But the save & restore should be reworked in a future merge request (and it does not have to be you making that change :)).

Could you please [open an issue](https://github.com/riscv/riscv-openocd/issues) and describe what needs to be done, so that this idea is not lost.

Thank you.

src/target/riscv/riscv-013.c
JanMatCodasip: Ah, I didn't know `checkpatch` requires this style.

What if you remove the star `*` at the beginning of the line like shown below? Will `checkpatch` still complain?

```suggestion
		/* There are no instructions to move all the bits from a register,
		   so we need to use some scratch RAM. */
```

src/target/riscv/riscv-013.c
en-sc: Done. Link to the issue: #852. If it is ok, I would like for it to be assigned to me. Who should I ask to do so?

src/target/riscv/riscv-013.c
en-sc: Yes, It will
```
> git diff github/en-sc/optimize-access
...
                /* There are no instructions to move all the bits from a register,
-                * so we need to use some scratch RAM.
-                */
+                  so we need to use some scratch RAM. */
...
>  git diff --patch github/en-sc/optimize-access | filterdiff -x ... | ~/openocd/tools/scripts/checkpatch.pl --no-signoff -
WARNING: Block comments use a trailing */ on a separate line
#18: FILE: src/target/riscv/riscv-013.c:1429:
+                  so we need to use some scratch RAM. */
...
```

src/target/riscv/riscv-013.c
JanMatCodasip: > Done. Link to the issue: https://github.com/riscv/riscv-openocd/issues/852.

Thank you.

> I would like for it to be assigned to me. Who should I ask to do so?

I was able to assign the issue to you.

src/target/riscv/riscv-013.c
JanMatCodasip: Ok, thanks for checking it. So let's leave the comments as they are now.

src/target/riscv/riscv-013.c
JanMatCodasip: As discussed above, the comments need to stay as they are due to checkpatch. No change needed.

src/target/riscv/riscv-013.c
JanMatCodasip: As discussed above, the comments need to stay as they are due to checkpatch. No change needed.

src/target/riscv/riscv.c
```diff
@@ -4376,96 +4391,163 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
...
+
+	const bool need_to_write = !reg->valid || reg->dirty ||
+		value != buf_get_u64(reg->value, 0, reg->size);
+	const bool cacheble = gdb_regno_cacheable(regid, need_to_write);
+
+	if (!cacheble || (write_through && need_to_write)) {
+		LOG_TARGET_DEBUG(target,
+				"Writing to target: %s <- 0x%" PRIx64 " (valid=%s, dirty=%s)",
+				reg->name, value, reg->valid ? "true" : "false",
+				reg->dirty ? "true" : "false");
```
JanMatCodasip: For easier troubleshooting, here you can also print whether the register is cacheable or not.

```suggestion
		LOG_TARGET_DEBUG(target,
				"Writing to target: %s <- 0x%" PRIx64 " (cacheable=%s, valid=%s, dirty=%s)",
				reg->name, value, cacheable ? "true" : "false", reg->valid ? "true" : "false",
				reg->dirty ? "true" : "false");
```

src/target/riscv/riscv.c
```diff
@@ -4376,96 +4391,163 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
...
 	}
 
-	LOG_DEBUG("[%s] wrote 0x%" PRIx64 " to %s valid=%d",
-			  target_name(target), value, reg->name, reg->valid);
+	buf_set_u64(reg->value, 0, reg->size, value);
+	reg->valid = cacheble;
+
+	LOG_TARGET_DEBUG(target, "Wrote 0x%" PRIx64 " to %s (valid=%s, dirty=%s)",
+			value, reg->name, reg->valid ? "true" : "false",
+			reg->dirty ? "true" : "false");
```

The information about cacheability can be printed here, too.

src/target/riscv/riscv.c
```diff
@@ -1481,8 +1494,10 @@ int riscv_flush_registers(struct target *target)
 	return ERROR_OK;
```
JanMatCodasip, line 1533: For easier orientation in the debug log (to be able to easily find when the cache flush started and when it finished), you can add a debug print also to the end of the function:

```suggestion
	LOG_TARGET_DEBUG(target, "Flush of register cache completed");
	return ERROR_OK;
```

JanMatCodasip (dismissed):
I've done one last visual check of the changes and it looks good to me. Thanks.

src/target/riscv/riscv.c
en-sc: Addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

JanMatCodasip (dismissed):

JanMatCodasip:
@en-sc Also please make sure your branch is rebased on top of current `riscv` trunk. That's because a merge from the upstream (openocd.org) has recently taken place.

en-sc:
> @en-sc Also please make sure your branch is rebased on top of current `riscv` trunk. That's because a merge from the upstream (openocd.org) has recently taken place.

Done.

en-sc:
[Last change](https://github.com/riscv/riscv-openocd/compare/de8da1506611ce8b3f8c42021ecad09d32f818c3..5a29a7399fb0aeccab4d03946ff5d982a32c0747) was a bugfix. Current version passes tests from `riscv-tests/debug`.

timsifive (approved):

timsifive (approved):

en-sc:
@timsifive, can this PR be merged today, or do I need to do something?

timsifive:
I'm giving @JanMatCodasip a final chance at commenting, but otherwise I was going to merge it tomorrow.

timsifive:
:fireworks:

PR #843, opened 2023-04-24
target/riscv: Support hypervisor address translation

timsifive:
Address translation is different in hypervisor mode. Add support for that.

The main difference is that address translation has two stages, and in the first stage page table addresses are affected by the second stage.


src/target/riscv/riscv.c
```diff
@@ -1864,6 +1903,31 @@ static int riscv_target_resume(struct target *target, int current, target_addr_t
...
+		*effective_mode = get_field(mstatus, MSTATUS_MPP);
+	else
+		*effective_mode = get_field(priv, VIRT_PRIV_PRV);
+
+	LOG_TARGET_DEBUG(target, "Effective mode=%d; v=%d", *effective_mode, *v_mode);
+
+	return ERROR_OK;
+}
+
 static int riscv_mmu(struct target *target, int *enabled)
```
GregSavin, line 1931: ah, so this routine was already in place for non-hypervisor privileged architecture, and now for the hypervisor case it checks whether either of the 2 translation stages that might apply are active; interesting

src/target/riscv/riscv.c
```diff
@@ -2003,31 +2101,167 @@ static int riscv_address_translate(struct target *target,
...
-	if (riscv_mmu(target, &enabled) == ERROR_OK) {
-		if (!enabled)
+	if (riscv_mmu(target, &enabled) != ERROR_OK)
+		return ERROR_FAIL;
+	if (!enabled)
+		return ERROR_FAIL;
+
+	unsigned int xlen = riscv_xlen(target);
+
+	riscv_reg_t priv;
```
GregSavin: Can this new section be factored out into its own function?

GregSavin (approved):
Looks good (and it was nice to have a practical reason to read through the Hypervisor section of the privileged arch spec)

PR #844, opened 2023-05-01
Merge 228fe7 from upstream

timsifive:
Merge commit '228fe7300c7df7aa05ba2c0bc19edde6d0156401' from upstream


GregSavin (approved):
The changes in hwthread.c look interesting.  So if that "SMP node change" warning ever happens, I wonder if that's basically saying to debugger user "GDB thinks a hw thread is current but that thread is no longer available; you'll have to ask GDB to switch its focus to a different thread if you want useful context before the next resume"?

JanMatCodasip:
@timsifive - I would like to review this merge briefly, and will manage it by the end of the week. Would that be an OK time frame, could you please wait a bit before the merge?

JanMatCodasip (approved):
I've checked the changes visually and it looks all right.

I noticed only several little things that I have submitted separately:
- https://github.com/riscv/riscv-openocd/pull/848
- https://github.com/riscv/riscv-openocd/pull/850
- https://review.openocd.org/c/openocd/+/7687

Issue #845, opened 2023-05-08
probable issue with a loop inside riscv/program.c


bieganski:
https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/program.c#L75

does that loop always has no effect?

if there is some magic and the answer is "no", then I think it should be marked with a comment.

TommyMurphyTM1234:
> does that loop always has no effect?

Why do you think that this is the case?
Isn't the loop executed `riscv_debug_buffer_size(p->target)` times?

* https://github.com/riscv/riscv-openocd/blob/80d529cad3706a9f8ad23ac8f73ae7434d655dc7/src/target/riscv/riscv.c#L4739
* https://github.com/riscv/riscv-openocd/blob/80d529cad3706a9f8ad23ac8f73ae7434d655dc7/src/target/riscv/riscv-013.c#L1695
* https://github.com/riscv/riscv-openocd/blob/80d529cad3706a9f8ad23ac8f73ae7434d655dc7/src/target/riscv/riscv-013.c#L2396
* https://github.com/riscv/riscv-openocd/blob/80d529cad3706a9f8ad23ac8f73ae7434d655dc7/src/target/riscv/riscv-013.c#L1626

TommyMurphyTM1234:
Certainly looks like the loop has no effect given the `if` statement inside the loop!

* https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/program.c#L76

```
75	for (size_t i = 0; i < riscv_debug_buffer_size(p->target); ++i)
76		if (i >= riscv_debug_buffer_size(p->target))
77			p->debug_buffer[i] = riscv_read_debug_buffer(t, i);
```
Looks like this commit changed (broke?) the `if` condition?

* https://github.com/riscv/riscv-openocd/blob/85bfab36ad549aa53e5054bb5b850d11f8f80386/src/target/riscv/program.c#L86

Before this it was as follows:

* https://github.com/riscv/riscv-openocd/blob/d94b38279a10f7e9d012a225be8d78cd8e5f1e39/src/target/riscv/program.c#L87

TommyMurphyTM1234:
I presume that this issue can now be closed as a result of resolution via PR #846?

en-sc:
@TommyMurphyTM1234, thanks for noticing and linking the fix!

PR #846, opened 2023-05-10
target/riscv: Remove non-functional code in riscv_program_exec().

timsifive:
Addresses #845.

Change-Id: If4eee383f92946669a84f92e52a3ac3600707525


JanMatCodasip (approved):
This change is all right, the code is safe to be removed:
- the loop has been broken already (as correctly pointed out in #845 - thanks :+1:); 
- and there is no code that would make use of `p->debug_buffer` after calling `riscv_program_exec()`, so reading back the program buffer contents after execution is useless anyhow.

GregSavin (approved):

PR #847, opened 2023-05-12
target/riscv: Comment that data1 might change.

timsifive:
In case in the future I have the same idea of optimizing progbuf writes again.


JanMatCodasip (commented):
I'm sorry but I believe this optimization is not legal w.r.t. the debug spec, and could make the memory write to stop woring on some hardware that is perfectly compliant.

The reason is that the debug spec does not guarantee `data1` retains its value after the execution of this abstract command, so we can't rely on it:

image

timsifive:
Thanks for pointing that out. Changing this to a comment about why the optimization can't be done.

JanMatCodasip (approved):

PR #850, opened 2023-05-15
Minor cleanup in target.c

JanMatCodasip:
Small cleanup in target.c to get rid of few upstream differences:

- removed a pointless check for `reg->exists` - already checked few lines above
- unify one log message with what's in upstream

Change-Id: I3fd761157382670611fa90de84e2dfc90192f473


JanMatCodasip:
Spotted during the review of #844

timsifive (approved):

PR #851, opened 2023-05-16
target/riscv: Handle harts powering down and coming back up.

timsifive:


src/target/riscv/riscv-013.c
```diff
@@ -587,6 +587,7 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 			break;
 		} else {
 			LOG_ERROR("failed %s at 0x%x, status=%d", op_name, address, status);
+			dtmcontrol_scan(target, DTM_DTMCS_DMIRESET);
 			return ERROR_FAIL;
```
JanMatCodasip: Here you can just reset the DMI and break from the loop. The error message and function return is already handled below.

```suggestion
			dtmcontrol_scan(target, DTM_DTMCS_DMIRESET);
			break;
```

src/target/riscv/riscv-013.c
```diff
@@ -595,6 +596,7 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 
 	if (status != DMI_STATUS_SUCCESS) {
 		LOG_ERROR("Failed %s at 0x%x; status=%d", op_name, address, status);
+		dtmcontrol_scan(target, DTM_DTMCS_DMIRESET);
 		return ERROR_FAIL;
 	}
```
JanMatCodasip: No need to call `dtmcontrol_scan(target, DTM_DTMCS_DMIRESET)` here again. It has already been handled above.

The word "DMI" can be included in the error message to make it clearer to the user what type of error it actually is.

```suggestion
	if (status != DMI_STATUS_SUCCESS) {
		LOG_ERROR("Failed DMI %s at 0x%x; status=%d", op_name, address, status);
		return ERROR_FAIL;
	}
```

src/target/riscv/riscv-013.c
```diff
@@ -619,6 +621,7 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 					LOG_ERROR("Failed %s (NOP) at 0x%x; status=%d", op_name, address,
```
```suggestion
					LOG_ERROR("Failed DMI %s (NOP) at 0x%x; status=%d", op_name, address,
```

(There's also one more occurrence above.)

src/target/target.c
```diff
@@ -3073,8 +3076,7 @@ static int handle_target(void *priv)
 			target_reset_examined(target);
 			retval = target_examine_one(target);
 			if (retval != ERROR_OK) {
-				LOG_TARGET_DEBUG(target, "Examination failed, GDB will be halted. "
-					"Polling again in %dms",
+				LOG_TARGET_DEBUG(target, "Examination failed. Polling again in %dms",
```
JanMatCodasip, line 3073: Detail: 
```suggestion
				LOG_TARGET_DEBUG(target, "Examination failed. Polling again in %d ms",
```

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/target.c
timsifive: In the OpenOCD code that I did not write, I see 3 cases of "%d ms" and 3 cases of "%dms". I'm inclined to leave this as-is, and let upstream worry about it when they get around to nitpicking whitespace in debug messages.

JanMatCodasip: Agreed :)

src/target/riscv/riscv-013.c
```diff
@@ -1741,6 +1744,18 @@ static int examine(struct target *target)
 		r->mtopi_readable = false;
 	}
 
+	riscv_reg_t dcsr;
```
GregSavin: Does this code get reached if power down/up happens after the first call to "init" command happens?

src/target/riscv/riscv-013.c
```diff
@@ -1744,6 +1744,18 @@ static int examine(struct target *target)
...
+		return ERROR_FAIL;
+	dcsr = set_field(dcsr, CSR_DCSR_STEP, 0);
+	dcsr = set_field(dcsr, CSR_DCSR_EBREAKM, riscv_ebreakm);
+	dcsr = set_field(dcsr, CSR_DCSR_EBREAKS, riscv_ebreaks);
+	dcsr = set_field(dcsr, CSR_DCSR_EBREAKU, riscv_ebreaku);
+	dcsr = set_field(dcsr, CSR_DCSR_EBREAKVS, riscv_ebreaku);
+	dcsr = set_field(dcsr, CSR_DCSR_EBREAKVU, riscv_ebreaku);
+	if (register_write_direct(target, GDB_REGNO_DCSR, dcsr) != ERROR_OK)
+		return ERROR_FAIL;
+
```
JanMatCodasip: The same operation is performed in `riscv013_on_step_or_resume()`, so the code can be factored out to a single helper function, called multiple times.

The setting of `dcsr.ebreak*` bits should be also done in `deassert_reset`. I have recorded it as an issue: https://github.com/riscv/riscv-openocd/issues/853 .

timsifive: I've refactored. Let's leave the deassert fix for another PR.

src/target/riscv/riscv-013.c
timsifive: Yes, but the mechanism is fragile. What happens is you get a failure in poll(), and then in handle_target() that failure is seen and it calls target_reset_examined().

I'll try to make this mechanism more robust.

src/target/riscv/riscv-013.c
JanMatCodasip: Sure, the setting of `dcsr.ebreak* ` in `deassert_reset` is a topic for a separate PR. I've mentioned it just for completeness :) Thanks.

GregSavin (dismissed):

JanMatCodasip (dismissed):
Checked visually - looks all right.

timsifive:
Force pushed to rebase, avoiding conflicts.

timsifive:
I'd like to merge this because it makes it better. I'm working on a more robust solution (which will also help with the reset case), but it's getting complex quickly when there are multiple harts and they might halt for reasons besides that we requested them to halt.

GregSavin (approved):

PR #854, opened 2023-05-29
target/riscv: fix register access on running target

en-sc:
Register access on running target should fail if mstatus needs to be modified.

Change-Id: Iec8e8d514ef2f5ca42606a5534cce55aaaa99180

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1087,13 +1087,8 @@ static int prep_for_register_access(struct target *target,
 	LOG_TARGET_DEBUG(target, "Preparing mstatus to access %s",
 			gdb_regno_name(regno));
 
-	/* FIXME: On a running target, there is no way to make sure mstatus won't
-	 * change between reading and writing it, so
-	 * if target->state != TARGET_HALTED an error should be returned here.
-	 * However, this would not allow access to FPU registers on the running
-	 * target.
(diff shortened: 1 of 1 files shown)


en-sc:
For a detailed description of the issue, please refer to this [discussion](https://github.com/riscv/riscv-openocd/pull/842#discussion_r1178500114).

src/target/riscv/riscv-013.c
```diff
@@ -1087,13 +1087,8 @@ static int prep_for_register_access(struct target *target,
...
 
-	/* FIXME: On a running target, there is no way to make sure mstatus won't
-	 * change between reading and writing it, so
-	 * if target->state != TARGET_HALTED an error should be returned here.
-	 * However, this would not allow access to FPU registers on the running
-	 * target.
-	 * See https://github.com/riscv/riscv-openocd/pull/842#discussion_r1178500114
-	 */
+	assert(target->state == TARGET_HALTED &&
+			"The target should be halted to modify and then restore mstatus");
```
```suggestion
			"The target must be halted to modify and then restore mstatus");
```

timsifive (changes requested):
This looks good to me. Abstract register accesses are still attempted when running, but we won't try to modify mstatus while running, which would lead to confusing behavior.

src/target/riscv/riscv-013.c

timsifive (approved):

PR #856, opened 2023-06-01
target/riscv: Select hart in update_dcsr()

timsifive:
Otherwise we may end up modifying DCSR of a different hart.

Fixes a bug introduced in f0898155d1e83c0f8691396eb01103cd3beb470c.

Change-Id: I39bde21a1444623ed150f2b3d504b9318b9d6191

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1558,6 +1558,9 @@ static int wait_for_authbusy(struct target *target, uint32_t *dmstatus)
 
 static int update_dcsr(struct target *target, bool step)
 {
+	if (dm013_select_target(target) != ERROR_OK)
+		return ERROR_FAIL;
+
 	riscv_reg_t dcsr;
 	/* We want to twiddle some bits in the debug CSR so debugging works. */
 	int result = register_read_direct(target, &dcsr, GDB_REGNO_DCSR);


GregSavin (dismissed):

JanMatCodasip (commented):
I have checked this only visually. Several notes:

1
I think this is a wider issue, which I'd like to address in the future: Virtually all functions accept the `target` as the argument but it is not very clear
- a) which of them will select the corresponding hart 
- b) which of them simply assume that the caller (or some grand-caller) has selected the target already, and it can be directly used

For the functions in the b) category, my preference would be to add an assertion  that checks the number of the last selected hart. I assume that would have caught this bug as well. 

2
The commit message says `... in update_dcsr()` but the change is actually made in `register_read_direct()`. 

3
`register_write_direct()` does not contain the call to `dm013_select_target` but `register_read_direct()` now has it. This asymmetry does not look right. It seems to me that adding the select_target call to `update_dcsr()` would make more sense.

timsifive:
1. Agreed, but I don't think it's that big a problem. Just adding the occasional call to dm013_select_target() fixes it, and that seems better than an assert().
3. I'll move the change back to update_dcsr() then.

JanMatCodasip (approved):
Re-checked visually and looks all right. Thanks.

GregSavin (approved):

Issue #853, opened 2023-05-24
Update deassert_reset() in riscv-013 to set dcsr.ebreak{m,s,u,vs,vu} bits


JanMatCodasip:
The function `deassert_reset()` in riscv-013.c should be updated to set the `dcsr.ebreak*` bits as configured by `riscv set_ebreakm`, `set_ebreaks` and `set_ebreaku` commands.

This is especially important when the TCL `reset run` command is issued. If the dcsr.ebreak bits are not set correctly after `reset run`, syscalls and SW breaks may not work, for example.

PR #857, opened 2023-06-06
When dcsr.ebreak* might be cleared, halt the target and set it again

timsifive:
I don't have a great target to test this on. What I've done:
1. Hack the code to force halt_set_dcsr_ebreak() to be called at various times even when it's not necessary. Ran all the usual tests that way.
2. Tested this against a target which returns a single DMI failure when a hart is powered up, while having the call to set_dcsr_ebreak() from examine() commented out. After examine() completed, halt_set_dcsr_ebreak() was called from tick() as expected.

This also helps with #853, although it's not really optimal.

### DIFF
--- a/src/jtag/drivers/xds110.c
+++ b/src/jtag/drivers/xds110.c
@@ -579,7 +579,7 @@ static bool usb_get_response(uint32_t *total_bytes_read, uint32_t timeout)
 
 static bool usb_send_command(uint16_t size)
 {
-	int written;
+	int written = 0;
 	bool success = true;
 
 	/* Check the packet length */
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -206,6 +206,10 @@ typedef struct {
 	/* This target was selected using hasel. */
 	bool selected;
 
+	/* When false, we need to set dcsr.ebreak*, halting the target if that's
+	 * necessary. */
+	bool dcsr_ebreak_is_set;
+
 	/* This hart was placed into a halt group in examine(). */
 	bool haltgroup_supported;
 } riscv013_info_t;
@@ -1573,6 +1577,7 @@ static int set_dcsr_ebreak(struct target *target, bool step)
 	if (dm013_select_target(target) != ERROR_OK)
 		return ERROR_FAIL;
 
+	RISCV013_INFO(info);
 	riscv_reg_t original_dcsr, dcsr;
 	/* We want to twiddle some bits in the debug CSR so debugging works. */
 	if (riscv_get_register(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
@@ -1587,6 +1592,7 @@ static int set_dcsr_ebreak(struct target *target, bool step)
 	if (dcsr != original_dcsr &&
 			riscv_set_register(target, GDB_REGNO_DCSR, dcsr) != ERROR_OK)
 		return ERROR_FAIL;
+	info->dcsr_ebreak_is_set = true;
 	return ERROR_OK;
 }
 
@@ -2440,6 +2446,7 @@ static int riscv013_get_hart_state(struct target *target, enum riscv_hart_state 
 		return ERROR_FAIL;
 	if (get_field(dmstatus, DM_DMSTATUS_ANYHAVERESET)) {
 		LOG_TARGET_INFO(target, "Hart unexpectedly reset!");
+		info->dcsr_ebreak_is_set = false;
 		/* TODO: Can we make this more obvious to eg. a gdb user? */
 		uint32_t dmcontrol = DM_DMCONTROL_DMACTIVE |
 			DM_DMCONTROL_ACKHAVERESET;
@@ -2471,6 +2478,14 @@ static int riscv013_get_hart_state(struct target *target, enum riscv_hart_state 
 	}
 	LOG_TARGET_ERROR(target, "Couldn't determine state. dmstatus=0x%x", dmstatus);
 	return ERROR_FAIL;
+}
+
+static int handle_became_unavailable(struct target *target,
+		enum riscv_hart_state previous_riscv_state)
+{
+	RISCV013_INFO(info);
+	info->dcsr_ebreak_is_set = false;
+	return ERROR_OK;
 }
 
 static int init_target(struct command_context *cmd_ctx,
@@ -2508,6 +2523,9 @@ static int init_target(struct command_context *cmd_ctx,
 	generic_info->hart_count = &riscv013_hart_count;
 	generic_info->data_bits = &riscv013_data_bits;
 	generic_info->print_info = &riscv013_print_info;
+
+	generic_info->handle_became_unavailable = &handle_became_unavailable;
+
 	if (!generic_info->version_specific) {
 		generic_info->version_specific = calloc(1, sizeof(riscv013_info_t));
 		if (!generic_info->version_specific)
@@ -2625,6 +2643,7 @@ static int deassert_reset(struct target *target)
 		target->state = TARGET_RUNNING;
 		target->debug_reason = DBG_REASON_NOTHALTED;
 	}
+	info->dcsr_ebreak_is_set = false;
 
 	/* Ack reset and clear DM_DMCONTROL_HALTREQ if previously set */
 	control = 0;
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -44,7 +44,6 @@ static int riscv013_halt_prep(struct target *target);
 static int riscv013_halt_go(struct target *target);
 static int riscv013_resume_go(struct target *target);
 static int riscv013_step_current_hart(struct target *target);
-static int riscv013_on_halt(struct target *target);
 static int riscv013_on_step(struct target *target);
 static int riscv013_resume_prep(struct target *target);
 static enum riscv_halt_reason riscv013_halt_reason(struct target *target);
@@ -2476,7 +2475,6 @@ static int init_target(struct command_context *cmd_ctx,
 	generic_info->get_hart_state = &riscv013_get_hart_state;
 	generic_info->resume_go = &riscv013_resume_go;
 	generic_info->step_current_hart = &riscv013_step_current_hart;
-	generic_info->on_halt = &riscv013_on_halt;
 	generic_info->resume_prep = &riscv013_resume_prep;
 	generic_info->halt_prep = &riscv013_halt_prep;
 	generic_info->halt_go = &riscv013_halt_go;
@@ -4500,11 +4498,6 @@ static int riscv013_resume_prep(struct target *target)
 static int riscv013_on_step(struct target *target)
 {
 	return riscv013_on_step_or_resume(target, true);
-}
-
-static int riscv013_on_halt(struct target *target)
-{
-	return ERROR_OK;
 }
 
 static enum riscv_halt_reason riscv013_halt_reason(struct target *target)
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1596,6 +1596,69 @@ static int set_dcsr_ebreak(struct target *target, bool step)
 	return ERROR_OK;
 }
 
+static int halt_set_dcsr_ebreak(struct target *target)
+{
+	RISCV_INFO(r);
+	RISCV013_INFO(info);
+	LOG_TARGET_DEBUG(target, "Halt to set DCSR.ebreak*");
+
+	/* Remove this hart from the halt group.  This won't work on all targets
+	 * because the debug spec allows halt groups to be hard-coded, but I
+	 * haven't actually encountered those in the wild yet.
+	 *
+	 * There is a possible race condition when another hart halts, and
+	 * this one is expected to also halt because it's supposed to be in the
+	 * same halt group. Or when this hart is halted when that happens.
+	 *
+	 * A better solution might be to leave the halt groups alone, and track
+	 * why we're halting when a halt occurs. When there are halt groups,
+	 * that leads to extra halting if not all harts need to set dcsr.ebreak
+	 * at the same time.  It also makes for more complicated code.
+	 *
+	 * The perfect solution would be Quick Access, but I'm not aware of any
+	 * hardware that implements it.
+	 *
+	 * We don't need a perfect solution, because we only get here when a
+	 * hart spontaneously resets, or when it powers down and back up again.
+	 * Those are both relatively rare. (At least I hope so. Maybe some
+	 * design just powers each hart down for 90ms out of every 100ms)
+	 */
+
+
+	if (info->haltgroup_supported) {
+		bool supported;
+		if (set_group(target, &supported, 0, HALT_GROUP) != ERROR_OK)
+			return ERROR_FAIL;
+		if (!supported)
+			LOG_TARGET_ERROR(target, "Couldn't place hart in halt group 0. "
+						 "Some harts may be unexpectedly halted.");
+	}
+
+	int result = ERROR_OK;
+
+	r->prepped = true;
+	if (riscv013_halt_go(target) != ERROR_OK ||
+			set_dcsr_ebreak(target, false) != ERROR_OK ||
+			riscv013_step_or_resume_current_hart(target, false) != ERROR_OK) {
+		result = ERROR_FAIL;
+	} else {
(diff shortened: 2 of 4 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -67,6 +66,12 @@ static int read_memory(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, uint8_t *buffer, uint32_t increment);
 static int write_memory(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, const uint8_t *buffer);
+typedef enum {
+	HALTGROUP,
+	RESUMEGROUP
+} grouptype_t;
```
JanMatCodasip: Style detail - there can be blank lines around this typedef:

```suggestion

typedef enum {
	HALTGROUP,
	RESUMEGROUP
} grouptype_t;

```

src/target/riscv/riscv-013.c
```diff
@@ -1556,21 +1569,77 @@ static int wait_for_authbusy(struct target *target, uint32_t *dmstatus)
...
+	LOG_TARGET_DEBUG(target, "Halt to set DCSR.ebreak*");
+
+	/* Remove this hart from the halt group.  This won't work on all targets
+	 * because the debug spec allows halt groups to be hard-coded, but I
+	 * haven't actually encountered those in the wild yet.
+	 *
+	 * There are is a possible race condition when another hart halts, and
+	 * this one is expected to also halt because it's supposed to be in the
+	 * same halt group. Or when this hart is halted when that happens.
+	 *
```
JanMatCodasip: Typo:

```suggestion
	 * There is a possible race condition when another hart halts, and
	 * this one is expected to also halt because it's supposed to be in the
	 * same halt group. Or when this hart is halted when that happens.
	 *
```

src/target/riscv/riscv.h
```diff
@@ -193,7 +193,19 @@ struct riscv_info {
...
 	int (*step_current_hart)(struct target *target);
-	int (*on_halt)(struct target *target);
+
+	/* These get called from riscv_poll_hart(), which is a house of cards
+	 * together with openocd_poll(), so be careful not to upset things too
+	 * much. */
+	int (*became_halted)(struct target *target);
+	int (*became_running)(struct target *target);
+	int (*became_unavailable)(struct target *target);
+	int (*became_nonexistent)(struct target *target);
```
JanMatCodasip: I wonder whether handling of "became_nonexistent" is needed. The RISC-V debug spec says: `"Harts are nonexistent if they will never be part of this hardware platform, no matter how long a user waits"`.

So in a compliant hardware, nonexistent harts should remain nonexistent forever, and should never change state.

Did you mean to handle this "transition to nonexistent state" just in case; just for additional safety?

JanMatCodasip (commented):
I haven't yet reviewed this in as much detail as I wanted, but I am sending at least few early comments.

src/target/riscv/riscv-013.c
timsifive: I agree with empty line before, but I like not having it after because it closely puts the enum with the function, which is the only reason the enum exists.

src/target/riscv/riscv-013.c
timsifive: I'll fix this.

src/target/riscv/riscv.h
timsifive: You are right in that technically this should never occur. I'll add a comment and a warning when this does happen.

timsifive: Actually, I think it's better just to remove this. There's already an error message when a hart is nonexistent.

src/target/riscv/riscv-013.c
```diff
@@ -1556,24 +1570,77 @@ static int wait_for_authbusy(struct target *target, uint32_t *dmstatus)
...
+}
+
+static int halt_set_dcsr_ebreak(struct target *target)
+{
+	RISCV_INFO(r);
+	RISCV013_INFO(info);
+	LOG_TARGET_DEBUG(target, "Halt to set DCSR.ebreak*");
+
+	/* Remove this hart from the halt group.  This won't work on all targets
+	 * because the debug spec allows halt groups to be hard-coded, but I
```
GregSavin, line 1606: The detailed comments are helpful.  It'd be nice if Spike could be modified to have a stress/fuzz mode where it can optionally simulate hart power cycles and resets in a wacky aggressive and/or random way (e.g. your example of powering down for 90ms out of every 100ms) but that seems like it would be an intensive project in itself.

GregSavin (dismissed):

src/target/riscv/riscv-013.c
timsifive: Adding spike support for this kind of stuff is on my list. Right now it doesn't support harts becoming unavailable in any way.

JanMatCodasip: > It'd be nice if Spike could be modified to have a stress/fuzz

Actually, this "stress mode" is a more general topic - there are multiple areas in the debug interface that could be fuzzed this way. For instance: DMI busy condition, abstract command busy status, system bus busy status, perhaps even more.

src/target/riscv/riscv-013.c
```diff
@@ -68,6 +67,13 @@ static int read_memory(struct target *target, target_addr_t address,
 static int write_memory(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, const uint8_t *buffer);
 
+typedef enum {
+	HALTGROUP,
+	RESUMEGROUP
```
JanMatCodasip: Detail - if you'd like:

```suggestion
	HALT_GROUP,
	RESUME_GROUP
```

src/target/riscv/riscv-013.c
```diff
@@ -1606,7 +1670,8 @@ static int set_group(struct target *target, bool *supported, unsigned group, gro
 	uint32_t read_val;
 	if (dmi_read(target, &read_val, DM_DMCS2) != ERROR_OK)
 		return ERROR_FAIL;
-	*supported = get_field(read_val, DM_DMCS2_GROUP) == group;
+	if (supported)
+		*supported = get_field(read_val, DM_DMCS2_GROUP) == group;
```

```suggestion
		*supported = (get_field(read_val, DM_DMCS2_GROUP) == group);
```

src/target/riscv/riscv.h
```diff
@@ -193,7 +193,18 @@ struct riscv_info {
...
 	int (*resume_go)(struct target *target);
 	int (*step_current_hart)(struct target *target);
-	int (*on_halt)(struct target *target);
+
+	/* These get called from riscv_poll_hart(), which is a house of cards
+	 * together with openocd_poll(), so be careful not to upset things too
+	 * much. */
+	int (*became_halted)(struct target *target);
+	int (*became_running)(struct target *target);
+	int (*became_unavailable)(struct target *target);
```
JanMatCodasip: Could we call these `handle_became_*`? To make it very clear from the name that these are handlers/callback, and not for instance just bool flags.

```suggestion
	int (*handle_became_halted)(struct target *target);
	int (*handle_became_running)(struct target *target);
	int (*handle_became_unavailable)(struct target *target);
```

JanMatCodasip (dismissed):
I have just few minor suggestions related to the code style. Other than that, I have checked this visually and it looks all right.

src/target/riscv/riscv-013.c
timsifive: DMI busy already has a stressor by riscv-tests using `riscv reset_delays`. For the other ones I agree.

Still, the first thing to implement is to actually make spike able to have harts become unavailable and then available again so at least we can get some coverage of this new code.

timsifive:
@SiFiveGregS @JanMatCodasip can I get a re-review so I can merge this?

src/target/riscv/riscv-013.c
```diff
@@ -1556,24 +1570,77 @@ static int wait_for_authbusy(struct target *target, uint32_t *dmstatus)
...
 	dcsr = set_field(dcsr, CSR_DCSR_STEP, step);
 	dcsr = set_field(dcsr, CSR_DCSR_EBREAKM, riscv_ebreakm);
-	dcsr = set_field(dcsr, CSR_DCSR_EBREAKS, riscv_ebreaks);
-	dcsr = set_field(dcsr, CSR_DCSR_EBREAKU, riscv_ebreaku);
-	dcsr = set_field(dcsr, CSR_DCSR_EBREAKVS, riscv_ebreaku);
-	dcsr = set_field(dcsr, CSR_DCSR_EBREAKVU, riscv_ebreaku);
-	if (riscv_set_register(target, GDB_REGNO_DCSR, dcsr) != ERROR_OK)
+	dcsr = set_field(dcsr, CSR_DCSR_EBREAKS, riscv_ebreaks && riscv_supports_extension(target, 'S'));
+	dcsr = set_field(dcsr, CSR_DCSR_EBREAKU, riscv_ebreaku && riscv_supports_extension(target, 'U'));
+	dcsr = set_field(dcsr, CSR_DCSR_EBREAKVS, riscv_ebreaku && riscv_supports_extension(target, 'V'));
```
GregSavin: Should the 'V' extension references here be 'H' instead?

src/target/riscv/riscv-013.c
```diff
@@ -199,6 +205,13 @@ typedef struct {
 
 	/* This target was selected using hasel. */
 	bool selected;
+
+	/* When false, we need to set dcsr.ebreak*, halting the target if that's
+	 * necessary. */
+	bool dcsr_ebreak_set;
```
JanMatCodasip: Could we please rename this to `is_dcsr_ebreak_set`? To make it very clear that `true` means the work is done. (Not that `true` means a pending action that is yet to be done.)

```suggestion
	bool is_dcsr_ebreak_set;
```

or 

```suggestion
	bool dcsr_ebreak_is_set;
```

src/target/riscv/riscv-013.c
```diff
@@ -1556,24 +1570,77 @@ static int wait_for_authbusy(struct target *target, uint32_t *dmstatus)
...
+	if (set_dcsr_ebreak(target, false) != ERROR_OK)
+		return ERROR_FAIL;
+	if (riscv013_step_or_resume_current_hart(target, false) != ERROR_OK)
+		return ERROR_FAIL;
+	target->state = TARGET_RUNNING;
+	target->debug_reason = DBG_REASON_NOTHALTED;
+
+	/* Add it back to the halt group. */
+	if (info->haltgroup_supported && set_group(target, NULL, target->smp, HALT_GROUP) != ERROR_OK)
+		return ERROR_FAIL;
```
JanMatCodasip: I believe the halt group membership should still be restored even if any of the above operation fails. We should at the very least try.

src/target/riscv/riscv-013.c
```diff
@@ -1556,24 +1570,77 @@ static int wait_for_authbusy(struct target *target, uint32_t *dmstatus)
...
+	 * hardware that implements it.
+	 *
+	 * We don't need a perfect solution, because we only get here when a
+	 * hart spontaneously resets, or when it powers down and back up again.
+	 * Those are both relatively rare. (At least I hope so. Maybe some
+	 * design just powers each hart down for 90ms out of every 100ms)
+	 */
+
+	if (info->haltgroup_supported && set_group(target, NULL, 0, HALT_GROUP) != ERROR_OK)
+		return ERROR_FAIL;
```
JanMatCodasip: There is no error print inside `set_group()` if it fails, and I understand it is like that on purpose because in examine() we use `set_group()` to test if the feature is supported -- and don't wish to print errors at that point.

So I'd recommend to put the error prints directly here into `halt_set_dcsr_ebreak()`.

There is one more occurrence of this same case at the bottom of this function.

src/target/riscv/riscv.c
```diff
@@ -2926,6 +2931,13 @@ int riscv_openocd_poll(struct target *target)
 		}
 	}
 
+	foreach_smp_target(entry, targets) {
+		struct target *t = entry->target;
+		struct riscv_info *info = riscv_info(t);
+		if (info->tick && info->tick(t) != ERROR_OK)
+			return ERROR_FAIL;
+	}
+
```
JanMatCodasip, line 2947: What case is this loop supposed to cover? My understanding is that we iterate over the individual harts already above (line 2871) and the dcsr.ebreak* manipulation should be handled inside riscv_poll_hart() (called on line 2883).

Is this loop to cover the `unavailable --> running` transition? If so, then it appears cleaner to put the logic into the newly added `handle_running` callback (and to pass also the previous state to the handler, as discussed below).

src/target/riscv/riscv.c
```diff
@@ -2771,12 +2772,16 @@ static int riscv_poll_hart(struct target *target, enum riscv_next_action *next_a
 				LOG_TARGET_DEBUG(target, "  triggered running");
 				target->state = TARGET_RUNNING;
 				target->debug_reason = DBG_REASON_NOTHALTED;
+				if (r->handle_became_running && r->handle_became_running(target) != ERROR_OK)
```
JanMatCodasip: These callbacks to handle the changed state look useful. I recommend to pass the `previous_riscv_state` to that callback, too. 

That's because the actions to take will likely be different also based on the original state. For instance, transition `halted --> running` is very different from `unavailable --> running`.

JanMatCodasip (changes requested):

src/target/riscv/riscv-013.c
timsifive: Yes. Thank you. I must've cut and pasted this before you originally gave me that feedback.

src/target/riscv/riscv-013.c
timsifive: Yes. That's more clear. Thanks.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
timsifive: set_group() only fails if dmi_write() or dmi_read() fails, and those do print errors (I'm fairly sure).

src/target/riscv/riscv.c
timsifive: The reason for this loop is to call tick() for every hart. What happens in tick() is opaque to this layer. The reason it's outside the previous loop is that at this point the state of every hart has settled, so any side effects happening in tick() won't affect the delicate poll() code.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
JanMatCodasip: Ok, understood, thank you for the explanation. 

If you like, you can put this as a short comment on top of this "foreach".

src/target/riscv/riscv-013.c
JanMatCodasip: Actually, there are two failure modes for set_group():

- 1) ERROR_FAIL returned (due to DMI read/write error)
- 2) supported=false

I had the second failure mode in mind. So in case the removal from the halt group is not possible (for instance due to hardcoded values in the hardware - unlikely but possible), we would not be notified about that. That's what I recommend to cover by an error print - the added print may save some troubleshooting in the future.

src/target/riscv/riscv-013.c
JanMatCodasip: @timsifive Can you please check if you pushed your changes? I can see here just the older revision.

src/target/riscv/riscv-013.c
timsifive: Alright. This seems like a pretty extreme corner case.

src/target/riscv/riscv.c
timsifive: Done.

GregSavin (dismissed):

GregSavin (approved):

src/jtag/drivers/xds110.c
```diff
@@ -579,7 +579,7 @@ static bool usb_get_response(uint32_t *total_bytes_read, uint32_t timeout)
 
 static bool usb_send_command(uint16_t size)
 {
-	int written;
+	int written = 0;
```
JanMatCodasip, line 582: I assume this is to silence a compiler warning (or something similar), am I correct?

JanMatCodasip (approved):

src/jtag/drivers/xds110.c
timsifive: Yes. I must be using a newer compiler or something, that now knows to look inside usb_write() and can tell that there is a return there before *written is ever assigned.

PR #858, opened 2023-06-06
Add register cache flushing and invalidation to protobuf execution.

MarekVCodasip:
Previously, progbuf execution did not flush or invalidate the register cache which could lead to incorrect behavior. This patch fixes it as well as refactors few sore points in the code related to it.

Change-Id: I353b931ca70a1828d4a9cc512aead00441730875

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1511,7 +1511,7 @@ int riscv_flush_registers(struct target *target)
 	if (!target->reg_cache)
 		return ERROR_OK;
 
-	LOG_TARGET_DEBUG(target, "");
+	LOG_TARGET_DEBUG(target, "Flushing register cache");
 
 	/* Writing non-GPR registers may require progbuf execution, and some GPRs
 	 * may become dirty in the process (e.g. S0, S1). For that reason, flush
@@ -4078,10 +4078,17 @@ COMMAND_HANDLER(riscv_exec_progbuf)
 			return ERROR_FAIL;
 	}
 
-	if (riscv_program_exec(&prog, target) == ERROR_OK)
-		LOG_TARGET_DEBUG(target, "exec_progbuf: Program buffer execution successful.");
-	else
+	if (riscv_flush_registers(target) != ERROR_OK)
+		return ERROR_FAIL;
+	int error = riscv_program_exec(&prog, target);
+	riscv_invalidate_register_cache(target);
+
+	if (error != ERROR_OK) {
 		LOG_TARGET_ERROR(target, "exec_progbuf: Program buffer execution failed.");
+		return ERROR_FAIL;
+	}
+
+	LOG_TARGET_DEBUG(target, "exec_progbuf: Program buffer execution successful.");
 
 	return ERROR_OK;
 }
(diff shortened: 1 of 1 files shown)


src/target/register.c
```diff
@@ -93,9 +93,8 @@ void register_unlink_cache(struct reg_cache **cache_p, const struct reg_cache *c
 /** Marks the contents of the register cache as invalid (and clean). */
 void register_cache_invalidate(struct reg_cache *cache)
 {
-	struct reg *reg = cache->reg_list;
-
-	for (unsigned int n = cache->num_regs; n != 0; n--, reg++) {
+	for (size_t n = 0; n < cache->num_regs; n++) {
+		struct reg *reg = &cache->reg_list[n];
```
timsifive: Can you leave this as-is? I'd rather not make unnecessary target-independent changes in this branch, which has already drifted significantly from upstream.

src/target/riscv/riscv.c
```diff
@@ -4040,8 +4040,7 @@ COMMAND_HANDLER(handle_info)
 	return 0;
 }
 
-COMMAND_HANDLER(riscv_exec_progbuf)
-{
+COMMAND_HANDLER(riscv_exec_progbuf) {
```
timsifive: At a glance, I don't see any COMMAND_HANDLER() examples where { is on the same line.

src/target/riscv/riscv.c
```diff
@@ -4078,10 +4077,16 @@ COMMAND_HANDLER(riscv_exec_progbuf)
 			return ERROR_FAIL;
 	}
 
-	if (riscv_program_exec(&prog, target) == ERROR_OK)
-		LOG_TARGET_DEBUG(target, "exec_progbuf: Program buffer execution successful.");
-	else
+	riscv_flush_registers(target);
```
timsifive: Check whether this succeeds.

src/target/riscv/riscv.c
```diff
@@ -4565,12 +4570,8 @@ static void riscv_invalidate_register_cache(struct target *target)
 	if (!target->reg_cache)
 		return;
 
-	LOG_DEBUG("[%d]", target->coreid);
+	LOG_TARGET_DEBUG(target, "Invalidating register cache of hart: [%d]", target->coreid);
```
timsifive: The target gets us better info than the coreid number.
```suggestion
	LOG_TARGET_DEBUG(target, "Invalidating register cache");
```

src/target/riscv/riscv.c
```diff
@@ -4565,12 +4570,8 @@ static void riscv_invalidate_register_cache(struct target *target)
...
 	register_cache_invalidate(target->reg_cache);
-	for (size_t i = 0; i < target->reg_cache->num_regs; ++i) {
-		struct reg *reg = &target->reg_cache->reg_list[i];
-		reg->valid = false;
-	}
```
timsifive, line 4573: This looks right, but I worry there is some subtle reason this code exists. What testing have you done? I think `make all` in riscv-tests/debug should be good enough. (Some of the `multispike` tests may fail. That's OK.)

timsifive (changes requested):
Looks like a good improvement.

src/target/register.c
MarekVCodasip: Yes, you are right, in the meantime I published an MR to the upstream about this.

src/target/riscv/riscv.c
MarekVCodasip: I tested it manually on spike via telnet console. I will tell you once I have ran the tests.

MarekVCodasip: Ok, so I can the tests, the only test which does not pass is the IcountTest (for all configurations)

Also I took a look into the original commit which introduced setting reg->valid to false and it was part of a larger commit implementing debug specification 0.13. What we see here are the leftovers after 6 years of refactoring.

timsifive: That sounds like your spike doesn't implement icount triggers, which is fine for this change.

I agree that this is probably just left-over code.

timsifive (approved):

JanMatCodasip (approved):
We've already reviewed this internally with Marek.

PR #862, opened 2023-06-09
rtos/FreeRTOS: pxCurrentTCB should be used for judgment.

Du-Chao:
The current TCB is stored in pxCurrentTCB, which is somehow RISC-V-specific, should not be overwritten from upstream (#816).

### DIFF
--- a/src/rtos/FreeRTOS.c
+++ b/src/rtos/FreeRTOS.c
@@ -268,7 +268,7 @@ struct symbols {
 };
 
 static const struct symbols freertos_symbol_list[] = {
-	{ "px_current_tcb", false },
+	{ "pxCurrentTCB", false },
 	{ "pxReadyTasksLists", false },
 	{ "xDelayedTaskList1", false },
 	{ "xDelayedTaskList2", false },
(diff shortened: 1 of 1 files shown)


timsifive:
Thank you for fixing this.

1. Please fix the code style as pointed out by checkpatch. I know you didn't invent this style and are just adjusting the existing code, but checkpatch enforces the style OpenOCD requires for when this is eventually upstreamed. It's an annoying quirk that some of the existing OpenOCD code does not currently conform.
2. There are a couple of FreeRTOS smoke tests in https://github.com/riscv-software-src/riscv-tests/tree/master/debug. Can you suggest a simple test that would have caught this problem?

Du-Chao:
Thanks Tim.

1. I have fixed the code style check with a new patch.
2. I will check the riscv-tests debug infrastructure and try to add some test cases.

timsifive (approved):

timsifive:
I'm still looking forward to a way to test this some day.

Du-Chao:
> I'm still looking forward to a way to test this some day.

Hi Tim,

I have optimized the FreeRtosTest case with https://github.com/riscv-software-src/riscv-tests/pull/478.
More assertions are introduced to make sure GDB could always get a right thread list.

In the same time, I made another OpenOCD patch (#865), without which the new testcase may fail.

Please review both.

PR #865, opened 2023-06-15
rtos/FreeRTOS: solve some conflicting usage of thread id.

Du-Chao:
1.
There are some RISCV-specific usage of thread_id, which has conflict with upstream. Some adaptions are made in this patch, to make sure OpenOCD is sending a clear thread list to gdb.

2.
Use freertos_read_struct_value for xSchedulerRunning.

Change-Id: I001a88a0c6d8eac98a389c0217b4897f28124840

### DIFF
--- a/src/rtos/FreeRTOS.c
+++ b/src/rtos/FreeRTOS.c
@@ -717,7 +717,7 @@ static int freertos_update_threads(struct rtos *rtos)
 			strcpy(rtos->thread_details[tasks_found].thread_name_str, tmp_str);
 			rtos->thread_details[tasks_found].exists = true;
 
-			if (value->tcb == px_current_tcb && rtos->current_thread != 1) {
+			if (value->tcb == px_current_tcb && rtos->current_thread != FREERTOS_CURRENT_EXECUTION_ID) {
 				char running_str[] = "State: Running";
 				rtos->current_thread = value->threadid;
 				rtos->thread_details[tasks_found].extra_info_str = malloc(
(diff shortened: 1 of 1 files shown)


src/rtos/FreeRTOS.c
```diff
@@ -511,15 +511,17 @@ static int freertos_update_threads(struct rtos *rtos)
...
-	uint32_t scheduler_running;
-	retval = target_read_u32(rtos->target,
-			rtos->symbols[FREERTOS_VAL_X_SCHEDULER_RUNNING].address,
-			&scheduler_running);
+	uint64_t scheduler_running;
+	retval = freertos_read_struct_value(rtos->target,
+										rtos->symbols[FREERTOS_VAL_X_SCHEDULER_RUNNING].address,
+										0,
+										freertos->ubasetype_size,
+										&scheduler_running);
```
timsifive: Can you indent this less?

src/rtos/FreeRTOS.c
```diff
@@ -711,7 +716,7 @@ static int freertos_update_threads(struct rtos *rtos)
 			strcpy(rtos->thread_details[tasks_found].thread_name_str, tmp_str);
 			rtos->thread_details[tasks_found].exists = true;
 
-			if (value->tcb == px_current_tcb) {
+			if (value->tcb == px_current_tcb && rtos->current_thread != 1) {
```
timsifive: Can you make a constant for FREERTOS_CURRENT_EXECUTION or something instead of using 1, so it's more clear what's going on? Then also use that around line 540: `rtos->thread_details->threadid = 1;`

timsifive (changes requested):
This all looks right, but I don't really remember this code enough to really know. I really appreciate that you wrote a test for this.

Just a few style comments.

timsifive (approved):

PR #866, opened 2023-06-15
target/riscv: Early exit magic sequence checks in riscv_semihosting

timsifive:
When halted we don't need to read all 3 instructions before deciding the sequence doesn't match.

Change-Id: I9f8345960ce27e859265af901a368166a70b9fde


GregSavin (approved):
Short circuit evaluation for detecting the "this is a normal breakpoint and not a semihosting call" case.   Looks good.  I guess maybe for slower RTL level simulator-based backends, or JTAG topologies driven at slow TCK rates, that could make a noticeable human-scale performance difference in some cases.

PR #871, opened 2023-06-22
target/riscv: refactor read_memory_progbuf()

en-sc:
There were a couple of problems with previous implementation:

* Misalligned read would return ERROR_OK and print all zeroes.

* CMDERR_BUSY for abstract access was improperly handled:

According to the spec, no assumptions can be made about DM_DATA* contents in such a case, but these were considered valid values from memory.

* A fallback to one element read was implemented when DMI_STATUS_BUSY occurred during batch reads, even though this can be accounted for.

Change-Id: I09174c61c951b2bb97a529b7f0aa5afaa995179b

### DIFF
--- a/src/target/riscv/program.c
+++ b/src/target/riscv/program.c
@@ -119,6 +119,23 @@ int riscv_program_lbr(struct riscv_program *p, enum gdb_regno d, enum gdb_regno 
 	return riscv_program_insert(p, lb(d, b, offset));
 }
 
+int riscv_program_load(struct riscv_program *p, enum gdb_regno d, enum gdb_regno b, int offset,
+		unsigned int size)
+{
+	switch (size) {
+	case 1:
+		return riscv_program_lbr(p, d, b, offset);
+	case 2:
+		return riscv_program_lhr(p, d, b, offset);
+	case 4:
+		return riscv_program_lwr(p, d, b, offset);
+	case 8:
+		return riscv_program_ldr(p, d, b, offset);
+	}
+	assert(false && "Unsupported size");
+	return ERROR_FAIL;
+}
+
 int riscv_program_csrrsi(struct riscv_program *p, enum gdb_regno d, unsigned int z, enum gdb_regno csr)
 {
 	assert(csr >= GDB_REGNO_CSR0 && csr <= GDB_REGNO_CSR4095);
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3396,310 +3396,514 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
 }
 
 /**
- * Read the requested memory, taking care to execute every read exactly once,
- * even if cmderr=busy is encountered.
+ * This function is used to start the memory-reading pipeline.
+ * The pipeline looks like this:
+ * memory -> s1 -> dm_data[0:1] -> debugger
+ * Prior to calling it, the program buffer should contain the appropriate
+ * program.
+ * This function sets DM_ABSTRACTAUTO_AUTOEXECDATA to trigger second stage of the
+ * pipeline (s1 -> dm_data[0:1]) whenever dm_data is read.
  */
-static int read_memory_progbuf_inner(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, uint8_t *buffer, uint32_t increment)
+static int read_memory_progbuf_inner_startup(struct target *target,
+		target_addr_t address, uint32_t increment, uint32_t index)
+{
+	RISCV013_INFO(info);
+	/* s0 holds the next address to read from.
+	 * s1 holds the next data value read.
+	 * a0 is a counter in case increment is 0.
+	 */
+	if (register_write_direct(target, GDB_REGNO_S0, address + index * increment)
+			!= ERROR_OK)
+		return ERROR_FAIL;
+
+	if (/*is_repeated_read*/ increment == 0 &&
+			register_write_direct(target, GDB_REGNO_A0, index) != ERROR_OK)
+		return ERROR_FAIL;
+
+	/* AC_ACCESS_REGISTER_POSTEXEC is used to trigger first stage of the
+	 * pipeline (memory -> s1) whenever this command is executed.
+	 */
+	const uint32_t startup_command = access_register_command(target,
+			GDB_REGNO_S1, riscv_xlen(target),
+			AC_ACCESS_REGISTER_TRANSFER | AC_ACCESS_REGISTER_POSTEXEC);
+	if (execute_abstract_command(target, startup_command) != ERROR_OK)
+		return ERROR_FAIL;
+
+	/* First read has just triggered. Result is in s1.
+	 * dm_data registers contain the previous value of s1 (garbage).
+	 */
+	if (dmi_write(target, DM_ABSTRACTAUTO,
+				set_field(0, DM_ABSTRACTAUTO_AUTOEXECDATA, 1)) != ERROR_OK)
+		return ERROR_FAIL;
+
+	/* Read garbage from dm_data0, which triggers another execution of the
+	 * program. Now dm_data contains the first good result (from s1),
+	 * and s1 the next memory value.
+	 */
+	if (dmi_read_exec(target, NULL, DM_DATA0) != ERROR_OK)
+		goto clear_abstractauto_and_fail;
+
+	uint32_t abstractcs;
+
+	if (wait_for_idle(target, &abstractcs) != ERROR_OK)
+		goto clear_abstractauto_and_fail;
+
+	info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
+	switch (info->cmderr) {
+	case CMDERR_NONE:
+		return ERROR_OK;
+	case CMDERR_BUSY:
+		LOG_TARGET_ERROR(target, "Unexpected busy error. This is probably a hardware bug.");
+		/* fall through */
+	default:
+		LOG_TARGET_DEBUG(target, "error when reading memory, cmderr=0x%" PRIx32,
+				info->cmderr);
+		riscv013_clear_abstract_error(target);
+		goto clear_abstractauto_and_fail;
+	}
+clear_abstractauto_and_fail:
+	dmi_write(target, DM_ABSTRACTAUTO, 0);
+	return ERROR_FAIL;
+}
+
+struct memory_access_info {
+	uint8_t *buffer_address;
+	target_addr_t target_address;
+	uint32_t element_size;
+	uint32_t increment;
+};
+
+/**
+ * This function attempts to restore the pipeline after a busy on abstract
+ * access.
+ * Target's state is as follows:
+ * s0 contains address + index_on_target * increment
+ * s1 contains mem[address + (index_on_target - 1) * increment]
+ * dm_data[0:1] contains mem[address + (index_on_target - 2) * increment]
+ */
+static int read_memory_progbuf_inner_on_ac_busy(struct target *target,
+		uint32_t start_index, uint32_t elements_to_read, uint32_t *elements_read,
+		struct memory_access_info access)
+{
+	increase_ac_busy_delay(target);
+	riscv013_clear_abstract_error(target);
+
+	if (dmi_write(target, DM_ABSTRACTAUTO, 0) != ERROR_OK)
+		return ERROR_FAIL;
+
+	/* See how far we got by reading s0/a0 */
+	uint32_t index_on_target;
+
+	if (/*is_repeated_read*/ access.increment == 0) {
+		/* s0 is constant, a0 is incremented by one each execution */
+		riscv_reg_t counter;
+
+		if (register_read_direct(target, &counter, GDB_REGNO_A0) != ERROR_OK)
+			return ERROR_FAIL;
+		index_on_target = counter;
+	} else {
+		target_addr_t address_on_target;
+
+		if (register_read_direct(target, &address_on_target, GDB_REGNO_S0) != ERROR_OK)
+			return ERROR_FAIL;
+		index_on_target = (address_on_target - access.target_address) /
+			access.increment;
+	}
+
+	/* According to the spec, if an abstract command fails, one can't make any
+	 * assumptions about dm_data registers, so all the values in the pipeline
+	 * are clobbered now and need to be reread.
+	 */
+	const uint32_t min_index_on_target = start_index + 2;
+	if (index_on_target < min_index_on_target) {
+		LOG_TARGET_ERROR(target, "Arithmetic does not work correctly on the target");
+		return ERROR_FAIL;
+	} else if (index_on_target == min_index_on_target) {
+		LOG_TARGET_DEBUG(target, "No forward progress");
+	}
+	const uint32_t next_index = (index_on_target - 2);
+	*elements_read = next_index - start_index;
+	LOG_TARGET_WARNING(target, "Re-reading memory from addresses 0x%"
+			TARGET_PRIxADDR " and 0x%" TARGET_PRIxADDR ".",
+			access.target_address + access.increment * next_index,
+			access.target_address + access.increment * (next_index + 1));
+	return read_memory_progbuf_inner_startup(target, access.target_address,
+			access.increment, next_index);
+}
+
+/**
+ * This function attempts to restore the pipeline after a dmi busy.
+ */
+static int read_memory_progbuf_inner_on_dmi_busy(struct target *target,
+		uint32_t start_index, uint32_t next_start_index,
+		struct memory_access_info access)
+{
+	LOG_TARGET_DEBUG(target, "DMI_STATUS_BUSY encountered in batch. Memory read [%"
+			PRIu32 ", %" PRIu32 ")", start_index, next_start_index);
+	if (start_index == next_start_index)
+		LOG_TARGET_DEBUG(target, "No forward progress");
+
+	if (dmi_write(target, DM_ABSTRACTAUTO, 0) != ERROR_OK)
+		return ERROR_FAIL;
+	return read_memory_progbuf_inner_startup(target, access.target_address,
+			access.increment, next_start_index);
+}
+
+/**
+ * This function extracts the data from the batch.
+ */
+static int read_memory_progbuf_inner_extract_batch_data(struct target *target,
+		const struct riscv_batch *batch,
+		uint32_t start_index, uint32_t elements_to_read, uint32_t *elements_read,
+		struct memory_access_info access)
+{
+	const bool two_reads_per_element = access.element_size > 4;
+	const uint32_t reads_per_element = (two_reads_per_element ? 2 : 1);
+	assert(!two_reads_per_element || riscv_xlen(target) == 64);
+	assert(elements_to_read <= UINT32_MAX / reads_per_element);
+	const uint32_t nreads = elements_to_read * reads_per_element;
+	for (uint32_t curr_idx = start_index, read = 0; read < nreads; ++read) {
+		switch (riscv_batch_get_dmi_read_op(batch, read)) {
+		case DMI_STATUS_BUSY:
+			*elements_read = curr_idx - start_index;
+			return read_memory_progbuf_inner_on_dmi_busy(target, start_index, curr_idx
+					, access);
+		case DMI_STATUS_FAILED:
+			LOG_TARGET_DEBUG(target,
+					"Batch memory read encountered DMI_STATUS_FAILED on read %"
+					PRIu32, read);
+			return ERROR_FAIL;
+		case DMI_STATUS_SUCCESS:
+			break;
+		default:
+			assert(0);
+		}
+		const uint32_t value = riscv_batch_get_dmi_read_data(batch, read);
+		uint8_t * const curr_buff = access.buffer_address +
+			curr_idx * access.element_size;
+		const target_addr_t curr_addr = access.target_address +
+			curr_idx * access.increment;
+		const uint32_t size = access.element_size;
+
+		assert(size <= 8);
+		const bool is_odd_read = read % 2;
+
+		if (two_reads_per_element && !is_odd_read) {
+			buf_set_u32(curr_buff + 4, 0, (size * 8) - 32, value);
+			continue;
+		}
+		const bool is_second_read = two_reads_per_element;
+
+		buf_set_u32(curr_buff, 0, is_second_read ? 32 : (size * 8), value);
+		log_memory_access64(curr_addr, buf_get_u64(curr_buff, 0, size * 8),
+				size, /*is_read*/ true);
+		++curr_idx;
+	}
+	*elements_read = elements_to_read;
+	return ERROR_OK;
+}
+
+/**
+ * This function reads a batch of elements from memory.
+ * Prior to calling this function the folowing conditions should be met:
+ * - Appropriate program loaded to program buffer.
+ * - DM_ABSTRACTAUTO_AUTOEXECDATA is set.
+ */
+static int read_memory_progbuf_inner_run_and_process_batch(struct target *target,
+		struct riscv_batch *batch, struct memory_access_info access,
+		uint32_t start_index, uint32_t elements_to_read, uint32_t *elements_read)
 {
 	RISCV013_INFO(info);
 
-	int result = ERROR_OK;
-
-	/* Write address to S0. */
-	result = register_write_direct(target, GDB_REGNO_S0, address);
-	if (result != ERROR_OK)
-		return result;
-
-	if (increment == 0 &&
-			register_write_direct(target, GDB_REGNO_A0, 0) != ERROR_OK)
+	if (batch_run(target, batch) != ERROR_OK)
 		return ERROR_FAIL;
 
-	uint32_t command = access_register_command(target, GDB_REGNO_S1,
-			riscv_xlen(target),
-			AC_ACCESS_REGISTER_TRANSFER | AC_ACCESS_REGISTER_POSTEXEC);
-	if (execute_abstract_command(target, command) != ERROR_OK)
+	uint32_t abstractcs;
+
+	if (wait_for_idle(target, &abstractcs) != ERROR_OK)
 		return ERROR_FAIL;
 
-	/* First read has just triggered. Result is in s1. */
-	if (count == 1) {
-		uint64_t value;
-		if (register_read_direct(target, &value, GDB_REGNO_S1) != ERROR_OK)
+	uint32_t elements_to_extract_from_batch;
+
+	info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
+	switch (info->cmderr) {
+	case CMDERR_NONE:
+		LOG_TARGET_DEBUG(target, "successful (partial?) memory read [%"
+				PRIu32 ", %" PRIu32 ")", start_index, start_index + elements_to_read);
+		elements_to_extract_from_batch = elements_to_read;
+		break;
+	case CMDERR_BUSY:
+		LOG_TARGET_DEBUG(target, "memory read resulted in busy response");
+		if (read_memory_progbuf_inner_on_ac_busy(target, start_index,
+					elements_to_read, &elements_to_extract_from_batch, access)
+				!= ERROR_OK)
 			return ERROR_FAIL;
-		buf_set_u64(buffer, 0, 8 * size, value);
-		log_memory_access64(address, value, size, true);
-		return ERROR_OK;
+		break;
+	default:
+		LOG_TARGET_DEBUG(target, "error when reading memory, cmderr=0x%" PRIx32,
+				info->cmderr);
+		riscv013_clear_abstract_error(target);
+		return ERROR_FAIL;
 	}
 
-	if (dmi_write(target, DM_ABSTRACTAUTO,
-			1 << DM_ABSTRACTAUTO_AUTOEXECDATA_OFFSET) != ERROR_OK)
-		goto error;
-	/* Read garbage from dmi_data0, which triggers another execution of the
-	 * program. Now dmi_data0 contains the first good result, and s1 the next
-	 * memory value. */
-	if (dmi_read_exec(target, NULL, DM_DATA0) != ERROR_OK)
-		goto error;
-
-	/* read_addr is the next address that the hart will read from, which is the
-	 * value in s0. */
-	unsigned index = 2;
-	while (index < count) {
-		riscv_addr_t read_addr = address + index * increment;
-		LOG_DEBUG("i=%d, count=%d, read_addr=0x%" PRIx64, index, count, read_addr);
-		/* The pipeline looks like this:
-		 * memory -> s1 -> dm_data0 -> debugger
-		 * Right now:
-		 * s0 contains read_addr
-		 * s1 contains mem[read_addr-size]
-		 * dm_data0 contains[read_addr-size*2]
-		 */
-
-		struct riscv_batch *batch = riscv_batch_alloc(target, RISCV_BATCH_ALLOC_SIZE,
-				info->dmi_busy_delay + info->ac_busy_delay);
-		if (!batch)
-			return ERROR_FAIL;
-
-		unsigned reads = 0;
-		for (unsigned j = index; j < count; j++) {
-			if (size > 4)
-				riscv_batch_add_dmi_read(batch, DM_DATA1);
-			riscv_batch_add_dmi_read(batch, DM_DATA0);
-
-			reads++;
-			if (riscv_batch_full(batch))
-				break;
-		}
-
-		batch_run(target, batch);
-
-		/* Wait for the target to finish performing the last abstract command,
-		 * and update our copy of cmderr. If we see that DMI is busy here,
-		 * dmi_busy_delay will be incremented. */
-		uint32_t abstractcs;
-		if (dmi_read(target, &abstractcs, DM_ABSTRACTCS) != ERROR_OK)
-			return ERROR_FAIL;
-		while (get_field(abstractcs, DM_ABSTRACTCS_BUSY))
-			if (dmi_read(target, &abstractcs, DM_ABSTRACTCS) != ERROR_OK)
-				return ERROR_FAIL;
-		info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
-
-		unsigned next_index;
-		unsigned ignore_last = 0;
-		switch (info->cmderr) {
-			case CMDERR_NONE:
-				LOG_DEBUG("successful (partial?) memory read");
-				next_index = index + reads;
-				break;
-			case CMDERR_BUSY:
-				LOG_DEBUG("memory read resulted in busy response");
-
-				increase_ac_busy_delay(target);
-				riscv013_clear_abstract_error(target);
-
-				dmi_write(target, DM_ABSTRACTAUTO, 0);
-
-				uint32_t dmi_data0, dmi_data1 = 0;
-				/* This is definitely a good version of the value that we
-				 * attempted to read when we discovered that the target was
-				 * busy. */
-				if (dmi_read(target, &dmi_data0, DM_DATA0) != ERROR_OK) {
-					riscv_batch_free(batch);
-					goto error;
-				}
-				if (size > 4 && dmi_read(target, &dmi_data1, DM_DATA1) != ERROR_OK) {
-					riscv_batch_free(batch);
-					goto error;
-				}
-
-				/* See how far we got, clobbering dmi_data0. */
-				if (increment == 0) {
-					uint64_t counter;
-					result = register_read_direct(target, &counter, GDB_REGNO_A0);
-					next_index = counter;
-				} else {
-					uint64_t next_read_addr;
-					result = register_read_direct(target, &next_read_addr,
-												  GDB_REGNO_S0);
-					next_index = (next_read_addr - address) / increment;
-				}
-				if (result != ERROR_OK) {
-					riscv_batch_free(batch);
-					goto error;
-				}
-
-				uint64_t value64 = (((uint64_t)dmi_data1) << 32) | dmi_data0;
-				buf_set_u64(buffer + (next_index - 2) * size, 0, 8 * size, value64);
-				log_memory_access64(address + (next_index - 2) * size, value64, size, true);
-
-				/* Restore the command, and execute it.
-				 * Now DM_DATA0 contains the next value just as it would if no
-				 * error had occurred. */
-				dmi_write_exec(target, DM_COMMAND, command, true);
-				next_index++;
-
-				dmi_write(target, DM_ABSTRACTAUTO,
-						1 << DM_ABSTRACTAUTO_AUTOEXECDATA_OFFSET);
-
-				ignore_last = 1;
-
-				break;
-			default:
-				LOG_DEBUG("error when reading memory, abstractcs=0x%08lx", (long)abstractcs);
-				riscv013_clear_abstract_error(target);
-				riscv_batch_free(batch);
-				result = ERROR_FAIL;
-				goto error;
-		}
-
-		/* Now read whatever we got out of the batch. */
-		dmi_status_t status = DMI_STATUS_SUCCESS;
-		unsigned read_count = 0;
-		assert(index >= 2);
-		for (unsigned j = index - 2; j < index + reads; j++) {
-			assert(j < count);
-			LOG_DEBUG("index=%d, reads=%d, next_index=%d, ignore_last=%d, j=%d",
-				index, reads, next_index, ignore_last, j);
-			if (j + 3 + ignore_last > next_index)
-				break;
-
-			status = riscv_batch_get_dmi_read_op(batch, read_count);
-			uint64_t value = riscv_batch_get_dmi_read_data(batch, read_count);
-			read_count++;
-			if (status != DMI_STATUS_SUCCESS) {
-				/* If we're here because of busy count, dmi_busy_delay will
-				 * already have been increased and busy state will have been
-				 * cleared in dmi_read(). */
-				/* In at least some implementations, we issue a read, and then
-				 * can get busy back when we try to scan out the read result,
-				 * and the actual read value is lost forever. Since this is
-				 * rare in any case, we return error here and rely on our
-				 * caller to reread the entire block. */
-				LOG_WARNING("Batch memory read encountered DMI error %d. "
-						"Falling back on slower reads.", status);
-				riscv_batch_free(batch);
-				result = ERROR_FAIL;
-				goto error;
-			}
-			if (size > 4) {
-				status = riscv_batch_get_dmi_read_op(batch, read_count);
-				if (status != DMI_STATUS_SUCCESS) {
-					LOG_WARNING("Batch memory read encountered DMI error %d. "
-							"Falling back on slower reads.", status);
-					riscv_batch_free(batch);
-					result = ERROR_FAIL;
-					goto error;
-				}
-				value <<= 32;
-				value |= riscv_batch_get_dmi_read_data(batch, read_count);
-				read_count++;
-			}
-			riscv_addr_t offset = j * size;
-			buf_set_u64(buffer + offset, 0, 8 * size, value);
-			log_memory_access64(address + j * increment, value, size, true);
-		}
-
-		index = next_index;
-
-		riscv_batch_free(batch);
-	}
-
-	dmi_write(target, DM_ABSTRACTAUTO, 0);
-
-	if (count > 1) {
-		/* Read the penultimate word. */
-		uint32_t dmi_data0, dmi_data1 = 0;
-		if (dmi_read(target, &dmi_data0, DM_DATA0) != ERROR_OK)
-			return ERROR_FAIL;
-		if (size > 4 && dmi_read(target, &dmi_data1, DM_DATA1) != ERROR_OK)
-			return ERROR_FAIL;
-		uint64_t value64 = (((uint64_t)dmi_data1) << 32) | dmi_data0;
-		buf_set_u64(buffer + size * (count - 2), 0, 8 * size, value64);
-		log_memory_access64(address + size * (count - 2), value64, size, true);
-	}
-
-	/* Read the last word. */
-	uint64_t value;
-	result = register_read_direct(target, &value, GDB_REGNO_S1);
-	if (result != ERROR_OK)
-		goto error;
-	buf_set_u64(buffer + size * (count - 1), 0, 8 * size, value);
-	log_memory_access64(address + size * (count - 1), value, size, true);
+	if (read_memory_progbuf_inner_extract_batch_data(target, batch, start_index,
+				elements_to_extract_from_batch, elements_read, access) != ERROR_OK)
+		return ERROR_FAIL;
 
 	return ERROR_OK;
+}
 
-error:
-	dmi_write(target, DM_ABSTRACTAUTO, 0);
+static uint32_t read_memory_progbuf_inner_fill_batch(struct riscv_batch *batch,
+		uint32_t count, uint32_t size)
+{
+	assert(size <= 8);
+	const uint32_t two_regs_used[] = {DM_DATA1, DM_DATA0};
+	const uint32_t one_reg_used[] = {DM_DATA0};
+	const uint32_t reads_per_element = size > 4 ? 2 : 1;
+	const uint32_t * const used_regs = size > 4 ? two_regs_used : one_reg_used;
+	const uint32_t batch_capacity = riscv_batch_available_scans(batch) / reads_per_element;
+	const uint32_t end = MIN(batch_capacity, count);
 
+	for (uint32_t j = 0; j < end; ++j)
+		for (uint32_t i = 0; i < reads_per_element; ++i)
+			riscv_batch_add_dmi_read(batch, used_regs[i]);
+	return end;
+}
+
+static int read_memory_progbuf_inner_try_to_read(struct target *target,
+		struct memory_access_info access, uint32_t *elements_read,
+		uint32_t index, uint32_t loop_count)
+{
+	RISCV013_INFO(info);
+	struct riscv_batch *batch = riscv_batch_alloc(target, RISCV_BATCH_ALLOC_SIZE,
+			info->dmi_busy_delay + info->ac_busy_delay);
+	if (!batch)
+		return ERROR_FAIL;
+
+	const uint32_t elements_to_read = read_memory_progbuf_inner_fill_batch(batch,
+			loop_count - index, access.element_size);
+
+	int result = read_memory_progbuf_inner_run_and_process_batch(target, batch,
+			access, index, elements_to_read, elements_read);
+	riscv_batch_free(batch);
 	return result;
 }
 
-/* Only need to save/restore one GPR to read a single word, and the progbuf
- * program doesn't need to increment. */
-static int read_memory_progbuf_one(struct target *target, target_addr_t address,
-		uint32_t size, uint8_t *buffer)
+/**
+ * read_memory_progbuf_inner_startup() must be called before calling this function
+ * with the address argument equal to curr_target_address.
+ */
+static int read_memory_progbuf_inner_ensure_forward_progress(struct target *target,
+		struct memory_access_info access, uint32_t start_index)
 {
-	uint64_t mstatus = 0;
-	uint64_t mstatus_old = 0;
-	if (modify_privilege(target, &mstatus, &mstatus_old) != ERROR_OK)
+	LOG_TARGET_DEBUG(target,
+			"Executing one loop iteration to ensure forward progress (index=%"
+			PRIu32 ")", start_index);
+	const target_addr_t curr_target_address = access.target_address +
+		start_index * access.increment;
+	uint8_t * const curr_buffer_address = access.buffer_address +
+		start_index * access.element_size;
+	const struct memory_access_info curr_access = {
+		.buffer_address = curr_buffer_address,
+		.target_address = curr_target_address,
+		.element_size = access.element_size,
+		.increment = access.increment,
+	};
+	uint32_t elements_read;
+	if (read_memory_progbuf_inner_try_to_read(target, curr_access, &elements_read,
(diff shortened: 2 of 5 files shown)


en-sc:
This change passed testing on tests from `riscv-tests/debug` testsuite (the results can be seen below), as well as some testing on hardware.
<details><summary>spike32</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 187s ]:::::::::::::::::::::::::::
14 tests returned not_applicable
57 tests returned pass
</details>
<details><summary>spike64</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 198s ]:::::::::::::::::::::::::::
12 tests returned not_applicable
59 tests returned pass
</details>
<details><summary>spike32-2-hwthread</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 241s ]:::::::::::::::::::::::::::
12 tests returned not_applicable
59 tests returned pass
</details>
<details><summary>spike64-2</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 260s ]:::::::::::::::::::::::::::
14 tests returned not_applicable
57 tests returned pass
</details>
<details><summary>spike32-2</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 267s ]:::::::::::::::::::::::::::
13 tests returned not_applicable
58 tests returned pass
</details>
<details><summary>spike64-2-hwthread</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 320s ]:::::::::::::::::::::::::::
12 tests returned not_applicable
59 tests returned pass
</details>

JanMatCodasip (commented):
I will be able to review this change in the first half of the next week (the week of June-26).

src/target/riscv/program.c
```diff
@@ -119,6 +119,23 @@ int riscv_program_lbr(struct riscv_program *p, enum gdb_regno d, enum gdb_regno
...
+		return riscv_program_lbr(p, d, b, offset);
+	case 2:
+		return riscv_program_lhr(p, d, b, offset);
+	case 4:
+		return riscv_program_lwr(p, d, b, offset);
+	case 8:
+		return riscv_program_ldr(p, d, b, offset);
+	}
+	LOG_ERROR("Unsupported size: %d", size);
+	return ERROR_FAIL;
```
timsifive: This should probably be an assert, no? I can't imagine calling code being able to recover from this. It probably means something went really wrong.

src/target/riscv/riscv-013.c
```diff
@@ -3396,329 +3396,567 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+		return ERROR_FAIL;
+	set_buffer_and_log_access(access, index, value);
+	return ERROR_OK;
+}
 
-			reads++;
-			if (riscv_batch_full(batch))
-				break;
-		}
+static int read_memory_progbuf_inner_fill_progbuff(struct target *target,
```
```suggestion
static int read_memory_progbuf_inner_fill_progbuf(struct target *target,
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,329 +3396,567 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
 		return ERROR_FAIL;
 
-	/* First read has just triggered. Result is in s1. */
-	if (count == 1) {
-		uint64_t value;
-		if (register_read_direct(target, &value, GDB_REGNO_S1) != ERROR_OK)
+	return read_word_from_s1(target, access, 0);
+}
+
+static void set_buffer_and_log_access(struct memory_access_info access,
```
```suggestion
static void set_buffer_and_log_read(struct memory_access_info access,
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,329 +3396,567 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+static uint32_t read_memory_progbuf_inner_startup_command(struct target *target)
+{
+	/* AC_ACCESS_REGISTER_POSTEXEC is used to trigger first stage of the
+	 * pipeline (memory -> s1) whenever this command is executed
+	 * (see the comment for read_memory_progbuf_inner_startup())
+	 */
+	return access_register_command(target, GDB_REGNO_S1, riscv_xlen(target),
+			AC_ACCESS_REGISTER_TRANSFER | AC_ACCESS_REGISTER_POSTEXEC);
+}
+/**
```
```suggestion
}

/**
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,329 +3396,567 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+	if (wait_for_idle(target, &abstractcs))
+		return ERROR_FAIL;
 
-				ignore_last = 1;
+	info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
+	switch (info->cmderr) {
+	case CMDERR_NONE:
+		break;
+	default:
+		/*TODO: Here one could attempt to try again on CMDERR_BUSY */
```
timsifive: What happens if we encounter BUSY? It should not be fatal, but must result in a retry. That doesn't have to be right here. In the comment can you mention where it's handled?

src/target/riscv/riscv-013.c
```diff
@@ -3396,329 +3396,567 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
-				break;
-			default:
-				LOG_DEBUG("error when reading memory, abstractcs=0x%08lx", (long)abstractcs);
-				riscv013_clear_abstract_error(target);
-				riscv_batch_free(batch);
-				result = ERROR_FAIL;
-				goto error;
-		}
+static uint32_t read_memory_progbuf_inner_fill_batch(struct riscv_batch *batch,
+		uint32_t count, uint32_t size);
```
timsifive: For this and other prototypes, is it possible to reorder the functions to eliminate the prototypes?

Since this change already rewrites all of it, there's not downside to moving the functions around.

src/target/riscv/riscv-013.c
```diff
@@ -3396,329 +3396,567 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+static uint32_t read_memory_progbuf_inner_fill_batch(struct riscv_batch *batch,
+		uint32_t count, uint32_t size)
+{
+	assert(size <= 8);
+	const uint32_t two_regs_used[] = {DM_DATA1, DM_DATA0};
+	const uint32_t one_reg_used[] = {DM_DATA0};
+	const uint32_t reads_per_element = size > 4 ? 2 : 1;
+	const uint32_t * const used_regs = size > 4 ? two_regs_used : one_reg_used;
+	const uint32_t batch_capacity = riscv_batch_available_scans(batch) / reads_per_element;
+	const uint32_t end =  batch_capacity < count ? batch_capacity : count;
```
```suggestion
	const uint32_t end = MIN(batch_capacity, count);
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,329 +3396,567 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+/**
+ * This function extracts the data from the batch.
+ */
+static int read_memory_progbuf_inner_read_batch(struct target *target,
+		const struct riscv_batch *batch,
+		uint32_t start_index, uint32_t elements_to_read, uint32_t *elements_read,
+		struct memory_access_info access)
+{
+	const bool two_reads_per_element = access.element_size > 4;
+	const uint32_t reads_per_element = (two_reads_per_element ? 2 : 1);
```
timsifive, line 3561: I think you can eliminate `two_reads_per_element` and just use `reads_per_element == 2` where you need it.

src/target/riscv/riscv-013.c
```diff
@@ -3396,329 +3396,567 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+}
 /**
  * Read the requested memory, silently handling memory access errors.
  */
 static int read_memory_progbuf(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, uint8_t *buffer, uint32_t increment)
 {
 	if (riscv_xlen(target) < size * 8) {
-		LOG_ERROR("XLEN (%d) is too short for %d-bit memory read.",
+		LOG_ERROR("XLEN (%d) is too short for %" PRIu32 "-bit memory read.",
```
timsifive: Use LOG_TARGET_ERROR

timsifive (changes requested):
I've completed a superficial read-through of this code. Overall it feels good. The refactor makes the error handling more comprehensible, and possibly more thorough as well. I'll try to go through it again later this week, hopefully with a bit more of an understanding how the pieces fit togethere.

Can you write a test in riscv-tests that tests the misaligned failure case that you're fixing here? Basing it off of MemTest32 or something, but with a misaligned address.

General point: It would be good to modify spike to clobber dmdata after an access, to flush out any other places where OpenOCD relies on the value staying the same.

src/target/riscv/riscv-013.c
en-sc: These prototypes are not needed in most cases (for now I only see one "loop" : `read_memory_progbuf_inner_loop_iteration` calls `read_memory_progbuf_inner_ensure_forward_progress`, which calls `read_memory_progbuf_inner_loop_iteration`).

I have added these to reorder the code in such a way that when the code is ridden from the top, the use of a function is seen before the definition. IMHO, this makes code easier to understand.

One example of such a case is `read_memory_progbuf_inner_startup`. When functions are reordered, this one appears at the top. But it does not make much sense on its own (without the context of `read_memory_progbuf_inner`). Furthermore, it would be separated from `read_memory_progbuf_inner` by a plethora of the likes of `read_memory_progbuf_inner_on_dmi_busy` which would not make sense either, so by the time a reader reaches `read_memory_progbuf_inner` they might get confused.

However, I understand that this is not consistent with the codebase. So, if you don't like it and don't find my reasoning appealing enough, I will gladly reorder these functions.

src/target/riscv/riscv-013.c
en-sc: I would like to leave this function's name as is.

The only thing this function does is calling `buf_set_u64` and `log_memory_access64` with corresponding arguments. So I would like for the name to represent this.

src/target/riscv/riscv-013.c
en-sc: Handling this particular `CMDERR_BUSY` is quite complicated. If one retries the access, it is possible to get stuck in an infinite loop if the hardware hangs and always returns busy. Since, at this point, no element has been read from memory.

This one `CMDERR_BUSY` is rare. I've never encountered it during testing. It would be much easier to address it when there is a reproduction of this exact `CMDERR_BUSY`.

I can file an issue.

src/target/riscv/riscv-013.c
en-sc: This flag has quite a lot of uses in conditions in the function, so IMHO it is justified to make it a separate variable, since it makes all the conditions a bit simpler.
I will eliminate it, if you insist.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/program.c
en-sc: Addressed.

src/target/riscv/riscv-013.c
timsifive: Yes, but it only logs a memory read. It is not used for arbitrary memory accesses.

src/target/riscv/riscv-013.c
timsifive: OK. In that case please update the comment to say that (for now) we're not retrying on this busy condition.

src/target/riscv/riscv-013.c
timsifive: Whether it's easier to see users or callers first depends on the code and the reader. Usually my preference is to eliminate prototypes because they're unnecessary duplication.

When reading the code I assume people mostly rely on the tools their editors give them to navigate between functions. (Unfortunately in github reviews this is not easy.)

src/target/riscv/riscv-013.c
timsifive: OK. Leave it as is.

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,556 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+	const uint32_t loop_count = count - 2;
+
+	for (uint32_t index = 0; index < loop_count;)
+		if (read_memory_progbuf_inner_loop_iteration(target, access, &index,
+					loop_count) != ERROR_OK) {
+			dmi_write(target, DM_ABSTRACTAUTO, 0);
+			return ERROR_FAIL;
+		}
+	if (dmi_write(target, DM_ABSTRACTAUTO, 0) != ERROR_OK)
+		return ERROR_FAIL;
```
timsifive: Here you clear DM_ABSTRACTAUTO on one failure but not another. It might be simpler to clear it at the end of read_memory_progbuf() instead of going everywhere through the code where it might be necessary to clear it.

src/target/riscv/riscv-013.c
timsifive: I don't think CMDERR_BUSY can really occur here. execute_abstract_command() above already checks abstractcs.cmderr, and if it's non-zero then returns ERROR_FAIL. So there is no way that we see CMDERR_BUSY when you execute the abstract command in response to the DMI read, because we're certain that the previous abstract command has already completed.

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,556 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+		struct riscv_batch *batch, struct memory_access_info access,
+		uint32_t start_index, uint32_t elements_to_read, uint32_t *elements_read);
 
-		index = next_index;
+static int read_memory_progbuf_inner_ensure_forward_progress(struct target *target,
+		struct memory_access_info access, uint32_t start_index, uint32_t elements_read);
 
-		riscv_batch_free(batch);
-	}
+static int read_memory_progbuf_inner_loop_iteration(struct target *target,
```
timsifive: Needs a comment that it guarantees that when this function returns ERROR_OK, *index is increased so forward progress is guaranteed.

src/target/riscv/riscv-013.c
en-sc: Sorry, I didn't understand what you meant. The call `dmi_write(target, DM_ABSTRACTAUTO, 0)` inside the loop is in fact a part of error handling. The call after the loop is not.  DM_ABSTRACTAUTO needs to be cleared after the loop. If it is not the call to `read_word_from_dmi` after it will cause extra execution of the program in `progbuf` which will load an element past the end of the array which can cause an exception or in some other way set `abstractcs.cmderr` to a non-zero value.

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,556 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+		if (register_read_direct(target, &address_on_target, GDB_REGNO_S0) != ERROR_OK)
+			return ERROR_FAIL;
+		index_on_target = (address_on_target - access.target_address) /
+			access.increment;
+	}
+
+	/* According to the spec, if an abstract command fails, one can't make any
+	 * assumptions about dm_data registers, so all the values in the pipeline
+	 * are clobbered now and need to be reread.
+	 */
```
timsifive, line 3516: This means reads at a specific address will happen twice. If reading memory has side effects (e.g. while reading from a FIFO using increment=0) then the output won't be as expected.

Given what the spec says and the desire to read quickly, I'm not sure how to resolve that. In practice most people won't care, and busy conditions will be rare. Best is probably to add a LOG_TARGET_WARNING saying which address was read more than once. In a normal debug session those might show up once or twice before OpenOCD learns the correct timing for bulk reads.

What do you think?

src/target/riscv/riscv-013.c
en-sc: You are right. Changed the case to print an error message.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: I agree. Myself and @aap-sc are concerned about re-reading memory too. It is indeed best avoided, and the user should be warned about such occurrences. Will address.

en-sc: Addressed.

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+static int read_memory_progbuf_inner_loop_iteration(struct target *target,
+		struct memory_access_info access, uint32_t *index, uint32_t loop_count);
+
+static int read_memory_progbuf_inner_ensure_forward_progress(struct target *target,
+		struct memory_access_info access, uint32_t start_index, uint32_t elements_to_read)
+{
+	if (elements_to_read == 1) {
+		LOG_TARGET_DEBUG(target, "Busy on reading one element");
+		/* FIXME: Here it would be better to fail only if delay is greater then
+		 * some threashhold
```
```suggestion
		 * some threshold
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+}
 
-		riscv_batch_free(batch);
+static int read_memory_progbuf_inner_loop_iteration(struct target *target,
+		struct memory_access_info access, uint32_t *index, uint32_t loop_count);
+
+static int read_memory_progbuf_inner_ensure_forward_progress(struct target *target,
+		struct memory_access_info access, uint32_t start_index, uint32_t elements_to_read)
+{
+	if (elements_to_read == 1) {
```
timsifive: Can you comment that this check is to prevent infinite recursion between read_memory_progbuf_inner_ensure_forward_progress() and read_memory_progbuf_inner_loop_iteration()?

src/target/riscv/riscv-013.c
timsifive: I think I misread this.

timsifive (commented):
I think I've walked through most of the various code paths. Overall this looks good, and the error handling is more readable than the original code.

Note that it's currently failing one of the automated tests. I assume you're also running this against riscv-tests/debug `make all`. Aside from the targeted tests, that also ends up testing a number of the busy-retry scenarios.

en-sc:
> Note that it's currently failing one of the automated tests.

The reason for this failure is the warning message on re-reading memory.
the `riscv-tests/debug`. It can be seen in the log:
```
--------------------------------[ Traceback ]---------------------------------
Traceback (most recent call last):
  File "/home/runner/work/riscv-openocd/riscv-openocd/riscv-tests/debug/testlib.py", line 1205, in run
    result = self.test()    # pylint: disable=no-member
  File "/home/runner/work/riscv-openocd/riscv-openocd/riscv-tests/debug/./gdbserver.py", line 920, in test
    values.append(int(v, 16))
ValueError: invalid literal for int() with base 16: '[riscv.cpu0]'
...
(gdb) monitor riscv repeat_read 1024 0x200bff8 4
monitor riscv repeat_read 1024 0x200bff8 4
[riscv.cpu0] re-reading memory from addresses 0x200bff8 and 0x200bff8
[riscv.cpu0] re-reading memory from addresses 0x200bff8 and 0x200bff8
0062ddf6 00635d4e 0063eb1a 0063ee6c 0063f15a 0063f222 0063f31c
...
```

Will create a patch to `riscv-tests/debug` addressing this.

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
 	return result;
 }
 
+static uint32_t read_memory_progbuf_inner_startup_command(struct target *target)
+{
+	/* AC_ACCESS_REGISTER_POSTEXEC is used to trigger first stage of the
+	 * pipeline (memory -> s1) whenever this command is executed
+	 * (see the comment for read_memory_progbuf_inner_startup())
+	 */
```
JanMatCodasip: Just nitpicking - when there are longer comments, I prefer to use proper sentences for readability (capital letter at the beginning, period at the end).

```suggestion
	/* AC_ACCESS_REGISTER_POSTEXEC is used to trigger first stage of the
	 * pipeline (memory -> s1) whenever this command is executed.
	 * (See the comment for read_memory_progbuf_inner_startup().)
	 */
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
  */
-static int read_memory_progbuf_inner(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, uint8_t *buffer, uint32_t increment)
+static int read_memory_progbuf_inner_startup(struct target *target,
+		target_addr_t address, uint32_t increment, uint32_t index)
 {
 	RISCV013_INFO(info);
+	/* s0 holds the next address to read from
+	 * s1 holds the next data value read
+	 * a0 is a counter in case increment is 0
```
```suggestion
	/* s0 holds the next address to read from.
	 * s1 holds the next data value read.
	 * a0 is a counter in case increment is 0.
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
-		return result;
+	const uint32_t startup_command =
+		read_memory_progbuf_inner_startup_command(target);
+	if (execute_abstract_command(target, startup_command) != ERROR_OK)
+		return ERROR_FAIL;
 
-	if (increment == 0 &&
-			register_write_direct(target, GDB_REGNO_A0, 0) != ERROR_OK)
+	/* First read has just triggered. Result is in s1.
+	 * dm_data registers contain the previos value of s1 (garbage).
```
JanMatCodasip: Just a typo:

```suggestion
	 * dm_data registers contain the previous value of s1 (garbage).
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+	if (/*is_repeated_read*/ increment == 0 &&
+			register_write_direct(target, GDB_REGNO_A0, index) != ERROR_OK)
+		return ERROR_FAIL;
 
-	/* Write address to S0. */
-	result = register_write_direct(target, GDB_REGNO_S0, address);
-	if (result != ERROR_OK)
-		return result;
+	const uint32_t startup_command =
+		read_memory_progbuf_inner_startup_command(target);
```
JanMatCodasip: Is this the only place where this function is used? If so, you may consider inlining it here.

If you'd still prefer to keep this as a separate function, then you may rename it to `get_read_memory_progbuf_initial_command()` (or similar), to make it more clear that it only returns the command, and does not actually execute it.

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+		return ERROR_FAIL;
 
-	if (increment == 0 &&
-			register_write_direct(target, GDB_REGNO_A0, 0) != ERROR_OK)
+	/* First read has just triggered. Result is in s1.
+	 * dm_data registers contain the previos value of s1 (garbage).
+	 */
+	if (dmi_write(target, DM_ABSTRACTAUTO,
+				set_field(0, DM_ABSTRACTAUTO_AUTOEXECDATA, 1)) != ERROR_OK)
 		return ERROR_FAIL;
```
JanMatCodasip: The progbuf read algorithm in OpenOCD relies on `autoexecdata[0]` being supported. But per the debug spec, it is an optional feature of the hardware. This topic has already been discussed here: https://github.com/riscv/riscv-openocd/issues/587

When making these changes, you may (optionally, if you like), take this opportunity and improve this. It would mean adding a check if the setting of `abstractauto.autoexecdata[0]` succeeded -- by reading the register back. If not, an error message can be printed and the whole operation aborted.

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+	case CMDERR_NONE:
+		break;
+		LOG_TARGET_ERROR(target, "Unexpected busy error. This is probably a hardware bug.");
+		/* fall through */
+	default:
+		LOG_TARGET_DEBUG(target, "error when reading memory, cmderr=0x%" PRIx32,
+				info->cmderr);
+		riscv013_clear_abstract_error(target);
+		dmi_write(target, DM_ABSTRACTAUTO, 0);
+		return ERROR_FAIL;
```
JanMatCodasip: This looks as a typo to me - probably like a missing line `case CMDERR_BUSY:` 

```suggestion
	case CMDERR_NONE:
		break;
	case CMDERR_BUSY:
		LOG_TARGET_ERROR(target, "Unexpected busy error. This is probably a hardware bug.");
		/* fall through */
	default:
		LOG_TARGET_DEBUG(target, "error when reading memory, cmderr=0x%" PRIx32,
				info->cmderr);
		riscv013_clear_abstract_error(target);
		dmi_write(target, DM_ABSTRACTAUTO, 0);
		return ERROR_FAIL;
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
-					goto error;
-				}
-				if (size > 4 && dmi_read(target, &dmi_data1, DM_DATA1) != ERROR_OK) {
-					riscv_batch_free(batch);
-					goto error;
-				}
+/**
+ * This function extracts the data from the batch.
+ */
+static int read_memory_progbuf_inner_read_batch(struct target *target,
```
JanMatCodasip: You can rename this function to make it clearer what it is doing (that it only extracts data from a finished batch).

For example:

```suggestion
static int read_memory_progbuf_inner_read_data_from_batch(struct target *target,
```

or

```suggestion
static int read_memory_progbuf_inner_read_batch_data(struct target *target,
```

or 

```suggestion
static int read_memory_progbuf_inner_extract_batch_data(struct target *target,
```

... or something similar :)

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+		return ERROR_FAIL;
 
-		index = next_index;
+	return ERROR_OK;
+}
 
-		riscv_batch_free(batch);
+static int read_memory_progbuf_inner_loop_iteration(struct target *target,
+		struct memory_access_info access, uint32_t *index, uint32_t loop_count);
+
```
JanMatCodasip: I'd recommend to move this forward declaration to the top of the file.

src/target/riscv/riscv-013.c
JanMatCodasip: Or, even better, would it please be possible to re-organize the functions so that these functions do not call each other?

E.g. create one function to perform the actual operations, and that would be called from both "loop_iteration()" and "ensure_forward_progress()".

src/target/riscv/riscv-013.c
```diff
@@ -3396,311 +3396,532 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
-				buf_set_u64(buffer + (next_index - 2) * size, 0, 8 * size, value64);
-				log_memory_access64(address + (next_index - 2) * size, value64, size, true);
+/**
+ * This function reads a batch of elements from memory.
+ * Prior to calling this function the folowing conditions should be met:
+ * - Appropriate program loaded to program buffer.
+ * - DM_ABSTRACTAUTO_AUTOEXECDATA is set.
+ * At least one element is read.
+ */
+static int read_memory_progbuf_inner_process_batch(struct target *target,
```
JanMatCodasip: If you'd like:

```suggestion
static int read_memory_progbuf_inner_run_and_process_batch(struct target *target,
```

... The function names are getting a bit long, I know. But we will be able to drop the prefixes when we split the `riscv-013.c` file to multiple modules (e.g. `riscv-013-progbuf-read.c` + others).

JanMatCodasip (commented):
I have managed a first round of review. Overall, this looks fine to me and I am posting several comments.

src/target/riscv/riscv-013.c
en-sc: I would like to leave this declaration where it is, since it is used only there. IMHO, current code organization improves readability.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Thanks for the great suggestion! Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: If I understand correctly, it would be better to implement this in `mem_should_skip_progbuf()`, so I would like to address it in a separate commit.

src/target/riscv/riscv-013.c
en-sc: Addressed. Thanks for the catch!

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Not an issue anymore, if I'm correct.

src/target/riscv/riscv-013.c

en-sc:
I've added a [commit to CI](https://github.com/riscv/riscv-openocd/pull/871/commits/291433c9f3709b8ca93ca75750ba532987074278), so in testing the branch with [suggested changes for `riscv-tests`](https://github.com/riscv-software-src/riscv-tests/pull/486) is tracked.

In such configuration, full testing is passed (the same as described [here](https://github.com/riscv/riscv-openocd/pull/871#issuecomment-1602943071)).

@timsifive, @JanMatCodasip, would you kindly take a look once again?

timsifive:
We need to coordinate commits between riscv-openocd and riscv-tests such that mainline of either branch will always work with each other. Best is probably to modify riscv-tests to work with both, merge that, then merge this PR, and then update riscv-tests to no longer look for any old error messages.

.github/workflows/spike-openocd-tests.yml
```diff
@@ -6,7 +6,7 @@ env:
   SPIKE_REPO: https://github.com/riscv-software-src/riscv-isa-sim.git
   SPIKE_REV: master
   RISCV_TESTS_REPO: https://github.com/riscv-software-src/riscv-tests.git
-  RISCV_TESTS_REV: master
+  RISCV_TESTS_REV: refs/pull/486/head
```
timsifive: We can't merge this with this change here.

src/target/riscv/riscv-013.c
```diff
@@ -3396,310 +3396,524 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+		return ERROR_FAIL;
+	} else if (index_on_target == min_index_on_target) {
+		LOG_TARGET_DEBUG(target, "No forward progress");
+	}
+	const uint32_t next_index = (index_on_target - 2);
+	*elements_read = next_index - start_index;
+	LOG_TARGET_WARNING(target, "re-reading memory from addresses 0x%"
+			TARGET_PRIxADDR " and 0x%" TARGET_PRIxADDR,
+			access.target_address + access.increment * next_index,
+			access.target_address + access.increment * (next_index + 1));
```
timsifive: I was surprised to see that you mention 2 addresses here. I think that is because read_memory_progbuf_inner_startup() already does 2 reads, so by the time you get here there have always been 2 reads already. Is that right?

timsifive: User-visible messages should generally be capitalized and have punctuation like regular English sentences. (Note that this breaks your riscv-tests change. Simplest is to make that one ignore capitalization.)

timsifive (commented):
How does this change affect the performance when reading a lot of data?

en-sc:
> How does this change affect the performance when reading a lot of data?

From my testing, there is no measurable difference, since at the core the approach is the same. The change affects busy response handling, which is a rare occurrence.

The performance was tested by dumping an image (on spike) as follows:
```
openocd -f ... -c 'init' -c 'reset halt' -c 'dump_image dump.img 0x80000000 [expr 4*1024*1024]' -c 'shutdown'
```
Before the change:
```
Open On-Chip Debugger 0.12.0+dev-00573-g92c031926 (2023-07-07-13:53)
...
dumped 4194304 bytes in 60.885689s (67.274 KiB/s)
...
```
After the change:
```
Open On-Chip Debugger 0.12.0+dev-00575-g291433c9f (2023-07-07-13:51)
...
Warn : [riscv.cpu] re-reading memory from addresses 0x80000008 and 0x80000010
dumped 4194304 bytes in 55.689625s (73.551 KiB/s)
...
```

There seems to be a speed-up, but it is not reproduced on hardware - I have observed a similarly small decrease in read speed.

src/target/riscv/riscv-013.c
en-sc: > User-visible messages should generally be capitalized and have punctuation like regular English sentences. (Note that this breaks your riscv-tests change. Simplest is to make that one ignore capitalization.)

Addressed.

> I was surprised to see that you mention 2 addresses here. I think that is because read_memory_progbuf_inner_startup() already does 2 reads, so by the time you get here there have always been 2 reads already. Is that right?

Yes, that's right.
At all times in the loop, there are two values loaded from memory, but not yet read by the debugger. The first value is in `s1` and the other in `dm_data[0:1]`.
If the abstract command fails, both values can get clobbered, so it is necessary to re-read them.

.github/workflows/spike-openocd-tests.yml
en-sc: Addressed. Testing results can be seen in the [PR in my fork](https://github.com/en-sc/riscv-openocd/pull/2).

en-sc:
**Please, disregard the failing tests.**
This is due to the change in warning message.
Testing results for updated `riscv-tests` version can be seen in the [PR in my fork](https://github.com/en-sc/riscv-openocd/pull/2).

en-sc:
@timsifive, @JanMatCodasip,  would you kindly take a look once again?

src/target/riscv/riscv-013.c
timsifive: I'm in favor of doing that fix separately. If nothing else, it is much simpler to review small changes.

timsifive (dismissed):

src/target/riscv/riscv-013.c
JanMatCodasip: > I'm in favor of doing that fix separately.

Agreed, let's leave it for some other merge request.

src/target/riscv/riscv-013.c
```diff
@@ -3396,310 +3396,524 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
-		uint64_t value;
-		if (register_read_direct(target, &value, GDB_REGNO_S1) != ERROR_OK)
-			return ERROR_FAIL;
-		buf_set_u64(buffer, 0, 8 * size, value);
-		log_memory_access64(address, value, size, true);
-		return ERROR_OK;
+	uint32_t abstractcs;
+
+	if (wait_for_idle(target, &abstractcs))
+		return ERROR_FAIL;
```
JanMatCodasip: I think we should clear `abstractauto` register also if `wait_for_idle()` fails.

```suggestion
	if (wait_for_idle(target, &abstractcs)) {
		dmi_write(target, DM_ABSTRACTAUTO, 0);
		return ERROR_FAIL;
	}
```

Since the function is quite complex, you may, if you like, use the single return + goto scheme for the error handling. It won't save lines of code, but that way it is less likely that someone will forget about the cleanup when modifying the function in the future.

```
	// ...

	if (some_operation() != ERROR_OK) {
		retval = ERROR_FAIL;
		goto handle_fail;
	}
	
	// ...

	if (other_operation() != ERROR_OK) {
		retval = ERROR_FAIL;
		goto handle_fail;
	}

	// ...

	// success
	return ERROR_OK;

	// error handling
handle_fail:
	dmi_write(target, DM_ABSTRACTAUTO, 0);
	return retval;
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,310 +3396,524 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
-		unsigned next_index;
-		unsigned ignore_last = 0;
-		switch (info->cmderr) {
-			case CMDERR_NONE:
-				LOG_DEBUG("successful (partial?) memory read");
-				next_index = index + reads;
-				break;
-			case CMDERR_BUSY:
-				LOG_DEBUG("memory read resulted in busy response");
+		buf_set_u32(curr_buff, 0, is_second_read ? 32 : size * 8, value);
```
JanMatCodasip: Minor thing - for readability, you can put here extra parentheses:

```suggestion
		buf_set_u32(curr_buff, 0, is_second_read ? 32 : (size * 8), value);
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,310 +3396,524 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+		LOG_TARGET_DEBUG(target, "error when reading memory, cmderr=0x%" PRIx32,
+				info->cmderr);
+		riscv013_clear_abstract_error(target);
+		return ERROR_FAIL;
+	}
 
-				dmi_write(target, DM_ABSTRACTAUTO,
-						1 << DM_ABSTRACTAUTO_AUTOEXECDATA_OFFSET);
+	if (read_memory_progbuf_inner_extract_batch_data(target, batch, start_index,
+				elements_to_fetch_from_batch, elements_read, access) != ERROR_OK)
```
JanMatCodasip: Detail: Since the function is called "extract" batch data, you can also use the word "extract" instead of "fetch" in the variable name - to make it consistent.

```suggestion
				elements_to_extract_from_batch, elements_read, access) != ERROR_OK)
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,310 +3396,524 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+static uint32_t read_memory_progbuf_inner_fill_batch(struct riscv_batch *batch,
+		uint32_t count, uint32_t size)
+{
+	assert(size <= 8);
+	const uint32_t two_regs_used[] = {DM_DATA1, DM_DATA0};
+	const uint32_t one_reg_used[] = {DM_DATA0};
+	const uint32_t reads_per_element = size > 4 ? 2 : 1;
+	const uint32_t * const used_regs = size > 4 ? two_regs_used : one_reg_used;
+	const uint32_t batch_capacity = riscv_batch_available_scans(batch) / reads_per_element;
+	const uint32_t end =  MIN(batch_capacity, count);
```

```suggestion
	const uint32_t end = MIN(batch_capacity, count);
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,310 +3396,524 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
-		assert(index >= 2);
-		for (unsigned j = index - 2; j < index + reads; j++) {
-			assert(j < count);
-			LOG_DEBUG("index=%d, reads=%d, next_index=%d, ignore_last=%d, j=%d",
-				index, reads, next_index, ignore_last, j);
-			if (j + 3 + ignore_last > next_index)
-				break;
+static int read_memory_progbuf_inner_try_to_read(struct target *target,
+		struct memory_access_info access, uint32_t *elements_read,
+		uint32_t index, uint32_t loop_count)
```
JanMatCodasip, line 3675: Maybe `loop_count` could be renamed to something more descriptive. E.g. `max_elements_to_read` ?

```suggestion
		uint32_t index, uint32_t max_elements_to_read)
```

src/target/riscv/riscv-013.c
```diff
@@ -3396,310 +3396,524 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+			/*is_read*/ true);
+}
 
-	if (count > 1) {
-		/* Read the penultimate word. */
-		uint32_t dmi_data0, dmi_data1 = 0;
-		if (dmi_read(target, &dmi_data0, DM_DATA0) != ERROR_OK)
-			return ERROR_FAIL;
-		if (size > 4 && dmi_read(target, &dmi_data1, DM_DATA1) != ERROR_OK)
+static int read_word_from_dmi(struct target *target,
```
JanMatCodasip: I'd recommend to call this "read_word_from_data_regs()", as opposed to "..from_dmi()".

```suggestion
static int read_word_from_data_regs(struct target *target,
```

Or, even better, I think you can use the existing `read_abstract_arg()` function.

src/target/riscv/riscv-013.c
```diff
@@ -3396,310 +3396,524 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+
+	if (riscv_program_load(p, d, b, offset, size) != ERROR_OK)
 		return ERROR_FAIL;
 
-	int result = ERROR_FAIL;
+	if (mprven && riscv_program_csrrci(p, GDB_REGNO_ZERO, CSR_DCSR_MPRVEN,
+				GDB_REGNO_DCSR) != ERROR_OK)
+		return ERROR_FAIL;
+
+	return ERROR_OK;
```
JanMatCodasip: @timsifive, an aside side question: Is there a reason to handle dcsr.mprven before and after every load? Would it be possible to instead handle it at the beginning of the block read, and restore it only at the very end.

No need to change anything in this merge request, though. The change is already large enough.

src/target/riscv/riscv-013.c
```diff
@@ -3396,310 +3396,524 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+		return ERROR_FAIL;
+	if (riscv_program_write(&program) != ERROR_OK)
+		return ERROR_FAIL;
+
+	return ERROR_OK;
+}
+
+/**
+ * Read the requested memory, taking care to execute every read exactly once,
+ * even if cmderr=busy is encountered.
```
JanMatCodasip: I believe this comment is no longer accurate - that's because on abstract command busy, two items need to be re-read.

Maybe you can reformulate it like this (or similar):

```suggestion
 * Read the requested memory, taking care to re-read the remaining data if 
 * "abstract command busy" or "DMI busy" is encountered in the process.
```

JanMatCodasip (commented):
Overall, this looks fine. 

I have only several last items to look at. This was my last review round and believe I won't have any more findings than these.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: I have tried to come up with a better name.

IMHO, renaming loop_count to max_elements_to_read would make it seem like this is the maximum value of elements_to_read, but it is not the case -- elements_to_read are less or equal to loop_count - index.

The other approach is something like max_index, but in such case it is easy to assume that this maximum is reached by index, meaning on some iteration index is equal to loop_count, which is not the case -- it is always less and the loop ends when index == loop_count - 1.

If you have any other suggestions, I'd be glad to hear them.

src/target/riscv/riscv-013.c
en-sc: Thanks for the great suggestion! Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
JanMatCodasip: Thanks for explanation. In that case, let's leave it as is :)

src/target/riscv/riscv-013.c
```diff
@@ -3396,310 +3396,513 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
+		LOG_TARGET_ERROR(target, "Unexpected busy error. This is probably a hardware bug.");
+		/* fall through */
+	default:
+		LOG_TARGET_DEBUG(target, "error when reading memory, cmderr=0x%" PRIx32,
+				info->cmderr);
+		riscv013_clear_abstract_error(target);
+	}
+clear_abstractauto_and_fail:
+	dmi_write(target, DM_ABSTRACTAUTO, 0);
+	return ERROR_FAIL;
```
JanMatCodasip: As a defensive coding measure, and to clearly show what should happen, dere I'd recommend to not leave the "default" case un-terminated.

```suggestion
		riscv013_clear_abstract_error(target);
		goto clear_abstractauto_and_fail;
	}

clear_abstractauto_and_fail:
	dmi_write(target, DM_ABSTRACTAUTO, 0);
	return ERROR_FAIL;
```

JanMatCodasip (dismissed):
This whole change looks all right - thanks.

I have just one last little detail.

src/target/riscv/riscv-013.c

JanMatCodasip (approved):

src/target/riscv/riscv-013.c
timsifive: I don't think there's any reason to handle it for every memory access. In fact, it probably just could be set when the hart first halts and then not touched again (unless the user changes the setting).

timsifive:
@en-sc One justification for this change is that dmdata* values cannot be read by the reader (because they're message registers). Do you have a target where that is in fact the case? I'm looking to see if there are real-world implementations of that behavior.

aap-sc:
> @en-sc One justification for this change is that dmdata* values cannot be read by the reader (because they're message registers). Do you have a target where that is in fact the case? I'm looking to see if there are real-world implementations of that behavior.

@timsifive @en-sc is unavailable for a couple of days. So I'll take the liberty to answer for him. No, we don't actually have such a target. The fact that `dmdata` is not persistent was detected during internal code review (before pushing this patch here) - so @en-sc  adjusted the code to take this into account.

timsifive:
Thanks. I was just curious if anyone had implemented something like that, related to some debug spec discussions.

PR #872, opened 2023-06-22
[target/riscv] support for smp group manipulation

aap-sc:
this functionality allows to query if a target belongs to some smp group and to dynamically turn on/off smp-specific behavior

Change-Id: I469453d95e7c1640a91bc60d80c854404e508535

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10812,6 +10812,18 @@ When utilizing version 0.11 of the RISC-V Debug Specification,
 and DBUS registers, respectively.
 @end deffn
 
+@deffn {Command} {riscv smp} [on|off]
+Display, enable or disable SMP handling mode. This command is needed only if
+user wants to temporary @b{disable} SMP handling for an existing SMP group
+(see @code{aarch64 smp} for additional information). To define an SMP
+group the command @code{target smp} should be used.
+@end deffn
+
+@deffn {Command} {riscv smp_gdb} [core_id]
+Display/set the current core displayed in GDB. This is needed only if
+@code{riscv smp} was used.
+@end deffn
+
 @deffn {Command} {riscv use_bscan_tunnel} value
 Enable or disable use of a BSCAN tunnel to reach the Debug Module. Supply the
 width of the DM transport TAP's instruction register to enable. Supply a
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -4383,6 +4383,9 @@ static const struct command_registration riscv_command_handlers[] = {
 		.usage = "",
 		.chain = semihosting_common_handlers
 	},
+	{
+		.chain = smp_command_handlers
+	},
 	COMMAND_REGISTRATION_DONE
 };
 


timsifive:
Can you update `openocd.texi` as well to reflect the new commands?

aap-sc:
> Can you update openocd.texi as well to reflect the new commands?

I can try, though it seems to me that the current documentation is somewhat deficient and it's not quite clear how to structure the documentation properly without lots of copy-pasting.

The functionality itself is described in `6.3.4 Define CPU targets working in SMP` . There is an example there that uses cortex_a target as an example.

However, cortext_a-specific documentation is quite brief:

```
  @deffn {Command} {cortex_a smp} [on|off]                                                                      
  Display/set the current SMP mode                                                                              
  @end deffn                                                                                                    
                                                                                                                
  @deffn {Command} {cortex_a smp_gdb} [core_id]                                                                 
  Display/set the current core displayed in GDB                                                                 
  @end deffn  
```

The same functionality is also documented in aarch64-specific section. The documentation there is quite rich:

```
 Display, enable or disable SMP handling mode. The state of SMP handling                                       
 influences the way targets in an SMP group are handled by the run control. With                               
 SMP handling enabled, issuing halt or resume to one core will trigger halting                                 
 or resuming of all cores in the group. The command @code{target smp} defines                                  
 which targets are in the SMP group. With SMP handling disabled, all targets                                   
 need to be treated individually.
```
I like this variant more, however I'm reluctant to copy-paste this in riscv-specific section. So **I'm more inclined to add only a very brief documentation like for `cortex_a`**.

Note: the functionality is available for mips_m4k, aarch64, cortex_a, cortex_m and some espressif chips. Only aarch64 and cortex_a have this documented.

aap-sc:
> Can you update openocd.texi as well to reflect the new commands?

addressed, but please see my previos comment for some details (just in case)

JanMatCodasip:
@aap-sc, do I understand correctly that these commands are used for the "old" way debugging SMP systems - that is, GDB sees just one thread and the user has to manually switch between the cores using the dedicated commands?

I thought hardly anyone is using that approach now, and the "hwthread" pseudo-RTOS would be used instead, which presents the SMP cores to GDB in a much nicer form: as though they were multiple threads of a program.

Are you still using the old way of SMP in OpenOCD? Or do you use the `smp*` commands for some other purpose?

Thanks in advance for explanation.

aap-sc:
> Are you still using the old way of SMP in OpenOCD? Or do you use the smp* commands for some other purpose?

@JanMatCodasip these are needed for tcl scripts. You see, when targets are combined via `targets smp ...` command it changes the behvior of ordinary `resume` and `halt` commands. (`Resume` begins to affect all cores at once, `halt` to). 

Now, imagine that you have a tcl function (that you call via telnet connection) that does `resume_all` - that is resumes each openocd target. It could look something like this:

```
proc resume_all {} {
        foreach t [target names] {
                targets $t
                resume
        }
}
```

Now, the function above won't work properly if targets a members of the same smp group. This is because the first resume already affects all targets. So during the iteration you'll end up calling resume for already running target - this may lead to unexpected error messages (or even could render the system non-functional , see here: 8ca5c2fbe412e6ad42474ceda0956ed7726b634f)

The functionality in question allows one to query the status of `smp` (or even temporary turn it off) and adjust the behavior of the script to handle such situation gracefully.

JanMatCodasip:
@aap-sc I see, so you're using those commands mostly to detect if the targets are operating in SMP mode or not, so that you can use the proper commands. Understood, thank you for explaining it.

doc/openocd.texi
```diff
@@ -10812,6 +10812,14 @@ When utilizing version 0.11 of the RISC-V Debug Specification,
 and DBUS registers, respectively.
 @end deffn
 
+@deffn {Command} {riscv smp} [on|off]
+Display/set the current SMP mode
+@end deffn
+
+@deffn {Command} {riscv smp_gdb} [core_id]
+Display/set the current core displayed in GDB
+@end deffn
```
timsifive: Given the discussion you had with Jan, can you mention that it's generally preferable to use `-rtos hwthread` instead of using these commands to set SMP mode or current core? It makes sense to use the existing commands for detection, but I imagine this opens up more corner cases that I don't really want to test, because using this SMP mechanism (specifically for RISC-V) should be considered deprecated.

aap-sc: I think there is some confusion here (unless I'm missing something).

Probably you know all of this, but I just want to make sure that we are on the same page.

According to my understanding:

1. `-rtos whatever` does not enable SMP mode. It only defines an algorithm by which OpenOCD pupulates the list of threads whenever GDB requests it. GDB uses this list to construct the list of so-called "inferior" threads.

So, OpenOCD can report several threads even if we have only one hart (my understanding is this is what happens when we specify something like Zephyr, though I've never tested this)

2. There is a command called `target smp ...`. This command groups existing targets into an SMP group. Typically, the command is called during `init_targets` procedure. However, it can be called even after the configuration stage like this:

```

target smp {*}[target names]

```

When targets are combined into an SMP  group like that -  the behaviour of `resume` and `halt` functions changes (as well as some others).

If `-rtos hwthread` is specified - each OpenOCD target is reported as a separate thread when GDB requests the list of threads. Note: this happens only because `-rtos hwthread` is implemented this way. It is possible to specify `-rtos none` for all targets and still create an SMP group - one still gets all the behavior associated with an SMP group except for the thread list.

3. Now, we have the ordinary `smp` command (more specically, `<target_name> smp`) introduced in this PR. Don't confuse with the `target smp`. The naming is awkward. It allows:

- to query whether the SMP behavior is enabled for the specified OpenOCD target.
- to disable SMP behavior.

When SMP behavior is disabled it allows us to effectively update the list of threads maintained by gdb, allowing us to temporarily switch to a dedicated hart and to debug only it.

It does not define new SMP groups (if no SMP group is defined via `target smp ...` call - then `smp` command does nothing). It does not destroy existing smp group - it only renders SMP-associated behavior as disabled.. 

I don't think this is a deprecated mechanism/functionality. The context in which it is expected to be used is quite unique.

JanMatCodasip: Thank you for the summary above.

My understanding was that it is unusual to use SMP groups (defined by `target smp ...`) without `-rtos hwthread`. 

I even thought it was deprecated, but I was probably mistaken here, as I can't find a mention of this deprecation the doc or mailing list. Sorry for this misleading comment.

@aap-sc My questions would be: 

- Do you actively use the SMP groups (defined by `target smp ...`) without `-rtos hwthread`?
- If yes, are you using it also with GDB this way? Or only without GDB, that is in scripts or in the telnet console?

Thank you.

aap-sc: > My understanding was that it is unusual to use SMP groups (defined by target smp ...) without -rtos hwthread.

This my understanding too. 

> Do you actively use the SMP groups (defined by target smp ...) without -rtos hwthread?

No. Though, I see such configurations from time to time, but most likely this is a consequence of copy-pasting erroneous configuration - not the conscious choice of the user.

timsifive: @aap-sc That's an excellent description. I'm concerned that people will see `riscv smp` and confuse it with `target smp`. Most users will never want to use `riscv smp`. So can you add some kind of explanation in the documentation that makes it clear that this is a distinct command from `target smp`, and that most users will want the latter and not the former.

Something like:
> `riscv smp` allows fine control over which targets are in the SMP group exposed to gdb. For most use cases, `target smp` is recommended.

aap-sc: > I'm concerned that people will see riscv smp and confuse it with target smp

Ah...  I see. Ok, I'll try to add something to avoid potential confusion.

aap-sc: Tried to address... @timsifive , @JanMatCodasip would you kindly take a look once again?

timsifive (approved):

Issue #855, opened 2023-05-31
Segfault on riscv_batch_run when using bscan tunnel


eosea:
I am getting a segfault on the buffer_shr function call on line 119 in src/target/riscv/batch.c.  The cause appears to be the in_value being set to NULL in the riscv_batch_add_dmi_write function due to read_back being false.   The segfault can be avoided by setting a null pointer check before the buffer_shr call (shown below).  Beyond this check, the riscv_batch_run function might need to be aware of if the transactions are set to not read back.

```
	if (bscan_tunnel_ir_width != 0) {
		/* need to right-shift "in" by one bit, because of clock skew between BSCAN TAP and DM TAP */
		for (size_t i = 0; i < batch->used_scans; ++i) {
                        // EDITED CODE HERE: Batch may be set to read_back == false, so skip them
			if ((batch->fields + i)-> in_value != NULL) {
                            buffer_shr((batch->fields + i)->in_value, DMI_SCAN_BUF_SIZE, 1);
                        }
                }
	}
```

timsifive:
Your analysis sounds right. I don't have a straightforward way to test any possible fix, though.

If you make a PR and confirm that it works, I'll be happy to review it.

eosea:
This issue was solved with commit #873.  Closing Issue.

PR #873, opened 2023-06-22
Add null pointer check before right shift for bscan tunneling.

eosea:
Add a null pointer check to address issue #855.

Change-Id: I5d4764c777f33d48705b3e5273eb840c13cfbfb7


JanMatCodasip (commented):
Thank you for this fix, it looks all right.

To fully handle the `in_value == NULL` case, and avoid future issues, I would also recommend to add the following assert to `riscv_batch_get_dmi_read_op()` and `riscv_batch_get_dmi_read_data()`:

```cpp
assert(batch->fields[index].in_value != NULL);
```

timsifive (approved):

PR #875, opened 2023-06-27
target/riscv: support multiple DMs

zqb-all:
Support assign DMI address of the debug module by pass -dbgbase to the target create command

Change-Id: I774c3746567f6e6d77c43a62dea5e9e67bb25770

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10891,15 +10891,28 @@ Write the 32-bit value to authdata or authdata0 (index=0), or authdata1
 
 @subsection RISC-V DMI Commands
 
-The following commands allow direct access to the Debug Module Interface, which
-can be used to interact with custom debug features.
+The following commands allow for direct low-level access to the registers
+of the Debug Module (DM). They can be useful to access custom features in the DM.
+
+@deffn {Command} {riscv dm_read} reg_address
+Perform a 32-bit read from the register indicated by reg_address from the DM of the
+current target.
+@end deffn
+
+@deffn {Command} {riscv dm_write} reg_address value
+Write the 32-bit value to the register indicated by reg_address from the DM of the
+current target.
+@end deffn
+
+The following commands allow for direct low-level access to the Debug Module
+Interface (DMI). They can be useful to access any device that resides on the DMI.
 
 @deffn {Command} {riscv dmi_read} address
-Perform a 32-bit DMI read at address, returning the value.
+Perform a 32-bit read from the given DMI address, returning the value.
 @end deffn
 
 @deffn {Command} {riscv dmi_write} address value
-Perform a 32-bit DMI write of value at address.
+Perform a 32-bit write to the given DMI address.
 @end deffn
 
 @subsection RISC-V Trigger Commands
--- a/src/target/riscv/batch.h
+++ b/src/target/riscv/batch.h
@@ -58,14 +58,14 @@ bool riscv_batch_full(struct riscv_batch *batch);
 /* Executes this scan batch. */
 int riscv_batch_run(struct riscv_batch *batch);
 
-/* Adds a DMI write to this batch. */
-void riscv_batch_add_dmi_write(struct riscv_batch *batch, unsigned int address, uint64_t data,
+/* Adds a DM register write to this batch. */
+void riscv_batch_add_dm_write(struct riscv_batch *batch, unsigned int address, uint64_t data,
 	bool read_back);
 
-/* DMI reads must be handled in two parts: the first one schedules a read and
+/* DM register reads must be handled in two parts: the first one schedules a read and
  * provides a key, the second one actually obtains the result of the read -
  * status (op) and the actual data. */
-size_t riscv_batch_add_dmi_read(struct riscv_batch *batch, unsigned address);
+size_t riscv_batch_add_dm_read(struct riscv_batch *batch, unsigned int address);
 unsigned int riscv_batch_get_dmi_read_op(const struct riscv_batch *batch, size_t key);
 uint32_t riscv_batch_get_dmi_read_data(const struct riscv_batch *batch, size_t key);
 
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -57,6 +57,9 @@ static void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a,
 static void riscv013_fill_dmi_read_u64(struct target *target, char *buf, int a);
 static int riscv013_dmi_write_u64_bits(struct target *target);
 static void riscv013_fill_dmi_nop_u64(struct target *target, char *buf);
+static void riscv013_fill_dm_write_u64(struct target *target, char *buf, int a, uint64_t d);
+static void riscv013_fill_dm_read_u64(struct target *target, char *buf, int a);
+static void riscv013_fill_dm_nop_u64(struct target *target, char *buf);
 static unsigned int register_size(struct target *target, enum gdb_regno number);
 static int register_read_direct(struct target *target, riscv_reg_t *value,
 		enum gdb_regno number);
@@ -296,10 +299,10 @@ static uint32_t set_dmcontrol_hartsel(uint32_t initial, int hart_index)
 	return initial;
 }
 
-static void decode_dmi(char *text, unsigned address, unsigned data)
+static void decode_dm(char *text, unsigned int address, unsigned int data)
 {
 	static const struct {
-		unsigned address;
+		unsigned int address;
 		uint64_t mask;
 		const char *name;
 	} description[] = {
@@ -373,6 +376,11 @@ static void decode_dmi(char *text, unsigned address, unsigned data)
 			}
 		}
 	}
+}
+
+static void decode_dmi(char *text, unsigned int address, unsigned int data)
+{
+	decode_dm(text, address, data);
 }
 
 static void dump_field(int idle, const struct scan_field *field)
@@ -551,7 +559,7 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
  * @param dmi_busy_encountered
  *                 If non-NULL, will be updated to reflect whether DMI busy was
  *                 encountered while executing this operation or not.
- * @param dmi_op   The operation to perform (read/write/nop).
+ * @param op       The operation to perform (read/write/nop).
  * @param address  The address argument to that operation.
  * @param data_out The data to send to the target.
  * @param timeout_sec
@@ -562,7 +570,7 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
  *                 DMI operation succeeded.
  */
 static int dmi_op_timeout(struct target *target, uint32_t *data_in,
-		bool *dmi_busy_encountered, int dmi_op, uint32_t address,
+		bool *dmi_busy_encountered, int op, uint32_t address,
 		uint32_t data_out, int timeout_sec, bool exec, bool ensure_success)
 {
 	select_dmi(target);
@@ -573,7 +581,7 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 		*dmi_busy_encountered = false;
 
 	const char *op_name;
-	switch (dmi_op) {
+	switch (op) {
 		case DMI_OP_NOP:
 			op_name = "nop";
 			break;
@@ -584,7 +592,7 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 			op_name = "write";
 			break;
 		default:
-			LOG_ERROR("Invalid DMI operation: %d", dmi_op);
+			LOG_ERROR("Invalid DMI operation: %d", op);
 			return ERROR_FAIL;
 	}
 
@@ -594,7 +602,7 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 	/* This first loop performs the request.  Note that if for some reason this
 	 * stays busy, it is actually due to the previous access. */
 	while (1) {
-		status = dmi_scan(target, NULL, NULL, dmi_op, address, data_out,
+		status = dmi_scan(target, NULL, NULL, op, address, data_out,
 				exec);
 		if (status == DMI_STATUS_BUSY) {
 			increase_dmi_busy_delay(target);
@@ -650,10 +658,10 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 }
 
 static int dmi_op(struct target *target, uint32_t *data_in,
-		bool *dmi_busy_encountered, int dmi_op, uint32_t address,
+		bool *dmi_busy_encountered, int op, uint32_t address,
 		uint32_t data_out, bool exec, bool ensure_success)
 {
-	int result = dmi_op_timeout(target, data_in, dmi_busy_encountered, dmi_op,
+	int result = dmi_op_timeout(target, data_in, dmi_busy_encountered, op,
 			address, data_out, riscv_command_timeout_sec, exec, ensure_success);
 	if (result == ERROR_TIMEOUT_REACHED) {
 		LOG_TARGET_ERROR(target, "DMI operation didn't complete in %d seconds. The target is "
@@ -686,10 +694,47 @@ static int dmi_write_exec(struct target *target, uint32_t address,
 	return dmi_op(target, NULL, NULL, DMI_OP_WRITE, address, value, true, ensure_success);
 }
 
+static int dm_op_timeout(struct target *target, uint32_t *data_in,
+		bool *dmi_busy_encountered, int op, uint32_t address,
+		uint32_t data_out, int timeout_sec, bool exec, bool ensure_success)
+{
+	dmi_op_timeout(target, data_in, dmi_busy_encountered, op, address,
+		data_out, timeout_sec, exec, ensure_success);
+}
+
+static int dm_op(struct target *target, uint32_t *data_in,
+		bool *dmi_busy_encountered, int op, uint32_t address,
+		uint32_t data_out, bool exec, bool ensure_success)
+{
+	return dmi_op(target, data_in, dmi_busy_encountered, op, address,
+			data_out, exec, ensure_success);
+}
+
+static int dm_read(struct target *target, uint32_t *value, uint32_t address)
+{
+	return dmi_read(target, value, address);
+}
+
+static int dm_read_exec(struct target *target, uint32_t *value, uint32_t address)
+{
+	return dmi_read_exec(target, value, address);
+}
+
+static int dm_write(struct target *target, uint32_t address, uint32_t value)
+{
+	return dmi_write(target, address, value);
+}
+
+static int dm_write_exec(struct target *target, uint32_t address,
+		uint32_t value, bool ensure_success)
+{
+	return dmi_write_exec(target, address, value, ensure_success);
+}
+
 static int dmstatus_read_timeout(struct target *target, uint32_t *dmstatus,
 		bool authenticated, unsigned timeout_sec)
 {
-	int result = dmi_op_timeout(target, dmstatus, NULL, DMI_OP_READ,
+	int result = dm_op_timeout(target, dmstatus, NULL, DMI_OP_READ,
 			DM_DMSTATUS, 0, timeout_sec, false, true);
 	if (result != ERROR_OK)
 		return result;
@@ -750,7 +795,7 @@ static int wait_for_idle(struct target *target, uint32_t *abstractcs)
 	RISCV013_INFO(info);
 	time_t start = time(NULL);
 	while (1) {
-		if (dmi_read(target, abstractcs, DM_ABSTRACTCS) != ERROR_OK)
+		if (dm_read(target, abstractcs, DM_ABSTRACTCS) != ERROR_OK)
 			return ERROR_FAIL;
 
 		if (get_field(*abstractcs, DM_ABSTRACTCS_BUSY) == 0)
@@ -795,7 +840,7 @@ static int execute_abstract_command(struct target *target, uint32_t command)
 		}
 	}
 
-	if (dmi_write_exec(target, DM_COMMAND, command, false) != ERROR_OK)
+	if (dm_write_exec(target, DM_COMMAND, command, false) != ERROR_OK)
 		return ERROR_FAIL;
 
 	uint32_t abstractcs = 0;
@@ -805,7 +850,7 @@ static int execute_abstract_command(struct target *target, uint32_t command)
 	if (info->cmderr != 0 || result != ERROR_OK) {
 		LOG_TARGET_DEBUG(target, "command 0x%x failed; abstractcs=0x%x", command, abstractcs);
 		/* Clear the error. */
-		dmi_write(target, DM_ABSTRACTCS, DM_ABSTRACTCS_CMDERR);
+		dm_write(target, DM_ABSTRACTCS, DM_ABSTRACTCS_CMDERR);
 		return ERROR_FAIL;
 	}
 
@@ -823,11 +868,11 @@ static riscv_reg_t read_abstract_arg(struct target *target, unsigned index,
 			LOG_TARGET_ERROR(target, "Unsupported size: %d bits", size_bits);
 			return ~0;
 		case 64:
-			dmi_read(target, &v, DM_DATA0 + offset + 1);
+			dm_read(target, &v, DM_DATA0 + offset + 1);
 			value |= ((uint64_t) v) << 32;
 			/* falls through */
 		case 32:
-			dmi_read(target, &v, DM_DATA0 + offset);
+			dm_read(target, &v, DM_DATA0 + offset);
 			value |= v;
 	}
 	return value;
@@ -842,10 +887,10 @@ static int write_abstract_arg(struct target *target, unsigned index,
 			LOG_TARGET_ERROR(target, "Unsupported size: %d bits", size_bits);
 			return ERROR_FAIL;
 		case 64:
-			dmi_write(target, DM_DATA0 + offset + 1, value >> 32);
+			dm_write(target, DM_DATA0 + offset + 1, value >> 32);
 			/* falls through */
 		case 32:
-			dmi_write(target, DM_DATA0 + offset, value);
+			dm_write(target, DM_DATA0 + offset, value);
 	}
 	return ERROR_OK;
 }
@@ -1056,7 +1101,7 @@ static int examine_progbuf(struct target *target)
 	}
 
 	uint32_t written;
-	if (dmi_read(target, &written, DM_PROGBUF0) != ERROR_OK)
+	if (dm_read(target, &written, DM_PROGBUF0) != ERROR_OK)
 		return ERROR_FAIL;
 	if (written == (uint32_t) info->progbuf_address) {
 		LOG_TARGET_INFO(target, "progbuf is writable at 0x%" PRIx64,
@@ -1229,18 +1274,18 @@ static int scratch_read64(struct target *target, scratch_mem_t *scratch,
 	uint32_t v;
 	switch (scratch->memory_space) {
 		case SPACE_DM_DATA:
-			if (dmi_read(target, &v, DM_DATA0 + scratch->debug_address) != ERROR_OK)
+			if (dm_read(target, &v, DM_DATA0 + scratch->debug_address) != ERROR_OK)
 				return ERROR_FAIL;
 			*value = v;
-			if (dmi_read(target, &v, DM_DATA1 + scratch->debug_address) != ERROR_OK)
+			if (dm_read(target, &v, DM_DATA1 + scratch->debug_address) != ERROR_OK)
 				return ERROR_FAIL;
 			*value |= ((uint64_t) v) << 32;
 			break;
 		case SPACE_DMI_PROGBUF:
-			if (dmi_read(target, &v, DM_PROGBUF0 + scratch->debug_address) != ERROR_OK)
+			if (dm_read(target, &v, DM_PROGBUF0 + scratch->debug_address) != ERROR_OK)
 				return ERROR_FAIL;
 			*value = v;
-			if (dmi_read(target, &v, DM_PROGBUF1 + scratch->debug_address) != ERROR_OK)
+			if (dm_read(target, &v, DM_PROGBUF1 + scratch->debug_address) != ERROR_OK)
 				return ERROR_FAIL;
 			*value |= ((uint64_t) v) << 32;
 			break;
@@ -1268,12 +1313,12 @@ static int scratch_write64(struct target *target, scratch_mem_t *scratch,
 {
 	switch (scratch->memory_space) {
 		case SPACE_DM_DATA:
-			dmi_write(target, DM_DATA0 + scratch->debug_address, value);
-			dmi_write(target, DM_DATA1 + scratch->debug_address, value >> 32);
+			dm_write(target, DM_DATA0 + scratch->debug_address, value);
+			dm_write(target, DM_DATA1 + scratch->debug_address, value >> 32);
 			break;
 		case SPACE_DMI_PROGBUF:
-			dmi_write(target, DM_PROGBUF0 + scratch->debug_address, value);
-			dmi_write(target, DM_PROGBUF1 + scratch->debug_address, value >> 32);
+			dm_write(target, DM_PROGBUF0 + scratch->debug_address, value);
+			dm_write(target, DM_PROGBUF1 + scratch->debug_address, value >> 32);
 			riscv013_invalidate_cached_debug_buffer(target);
 			break;
 		case SPACE_DMI_RAM:
@@ -1680,10 +1725,10 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
 	assert(group <= 31);
 	write_val = set_field(write_val, DM_DMCS2_GROUP, group);
 	write_val = set_field(write_val, DM_DMCS2_GROUPTYPE, (grouptype == HALT_GROUP) ? 0 : 1);
-	if (dmi_write(target, DM_DMCS2, write_val) != ERROR_OK)
+	if (dm_write(target, DM_DMCS2, write_val) != ERROR_OK)
 		return ERROR_FAIL;
 	uint32_t read_val;
-	if (dmi_read(target, &read_val, DM_DMCS2) != ERROR_OK)
+	if (dm_read(target, &read_val, DM_DMCS2) != ERROR_OK)
 		return ERROR_FAIL;
 	if (supported)
 		*supported = (get_field(read_val, DM_DMCS2_GROUP) == group);
@@ -1722,20 +1767,20 @@ static int examine(struct target *target)
 	if (!dm)
 		return ERROR_FAIL;
 	if (!dm->was_reset) {
-		dmi_write(target, DM_DMCONTROL, 0);
-		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_DMACTIVE);
+		dm_write(target, DM_DMCONTROL, 0);
+		dm_write(target, DM_DMCONTROL, DM_DMCONTROL_DMACTIVE);
 		dm->was_reset = true;
 	}
 	/* We're here because we're uncertain about the state of the target. That
 	 * includes our progbuf cache. */
 	riscv013_invalidate_cached_debug_buffer(target);
 
-	dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_HARTSELLO |
+	dm_write(target, DM_DMCONTROL, DM_DMCONTROL_HARTSELLO |
 			DM_DMCONTROL_HARTSELHI | DM_DMCONTROL_DMACTIVE |
 			DM_DMCONTROL_HASEL);
 	dm->current_hartid = HART_INDEX_UNKNOWN;
 	uint32_t dmcontrol;
-	if (dmi_read(target, &dmcontrol, DM_DMCONTROL) != ERROR_OK)
+	if (dm_read(target, &dmcontrol, DM_DMCONTROL) != ERROR_OK)
 		return ERROR_FAIL;
 	/* Ensure the HART_INDEX_UNKNOWN is flushed out */
 	if (dm013_select_hart(target, 0) != ERROR_OK)
@@ -1772,7 +1817,7 @@ static int examine(struct target *target)
 	LOG_TARGET_DEBUG(target, "hartsellen=%d", info->hartsellen);
 
 	uint32_t hartinfo;
-	if (dmi_read(target, &hartinfo, DM_HARTINFO) != ERROR_OK)
+	if (dm_read(target, &hartinfo, DM_HARTINFO) != ERROR_OK)
 		return ERROR_FAIL;
 
 	info->datasize = get_field(hartinfo, DM_HARTINFO_DATASIZE);
@@ -1786,12 +1831,12 @@ static int examine(struct target *target)
 		return ERROR_FAIL;
 	}
 
-	if (dmi_read(target, &info->sbcs, DM_SBCS) != ERROR_OK)
+	if (dm_read(target, &info->sbcs, DM_SBCS) != ERROR_OK)
 		return ERROR_FAIL;
 
 	/* Check that abstract data registers are accessible. */
 	uint32_t abstractcs;
-	if (dmi_read(target, &abstractcs, DM_ABSTRACTCS) != ERROR_OK)
+	if (dm_read(target, &abstractcs, DM_ABSTRACTCS) != ERROR_OK)
 		return ERROR_FAIL;
 	info->datacount = get_field(abstractcs, DM_ABSTRACTCS_DATACOUNT);
 	info->progbufsize = get_field(abstractcs, DM_ABSTRACTCS_PROGBUFSIZE);
@@ -1830,7 +1875,7 @@ static int examine(struct target *target)
 			dm->hart_count = i + 1;
 
 			if (get_field(s, DM_DMSTATUS_ANYHAVERESET))
-				dmi_write(target, DM_DMCONTROL,
+				dm_write(target, DM_DMCONTROL,
 						set_dmcontrol_hartsel(DM_DMCONTROL_DMACTIVE | DM_DMCONTROL_ACKHAVERESET, i));
 		}
 
@@ -1988,7 +2033,7 @@ static int riscv013_authdata_read(struct target *target, uint32_t *value, unsign
 	if (wait_for_authbusy(target, NULL) != ERROR_OK)
 		return ERROR_FAIL;
 
-	return dmi_read(target, value, DM_AUTHDATA);
+	return dm_read(target, value, DM_AUTHDATA);
 }
 
 static int riscv013_authdata_write(struct target *target, uint32_t value, unsigned int index)
@@ -2002,7 +2047,7 @@ static int riscv013_authdata_write(struct target *target, uint32_t value, unsign
 	if (wait_for_authbusy(target, &before) != ERROR_OK)
 		return ERROR_FAIL;
 
-	dmi_write(target, DM_AUTHDATA, value);
+	dm_write(target, DM_AUTHDATA, value);
 
 	if (wait_for_authbusy(target, &after) != ERROR_OK)
 		return ERROR_FAIL;
@@ -2295,12 +2340,12 @@ static int sb_write_address(struct target *target, target_addr_t address,
 	unsigned int sbasize = get_field(info->sbcs, DM_SBCS_SBASIZE);
 	/* There currently is no support for >64-bit addresses in OpenOCD. */
 	if (sbasize > 96)
-		dmi_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS3, 0, false, false);
+		dm_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS3, 0, false, false);
 	if (sbasize > 64)
-		dmi_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS2, 0, false, false);
+		dm_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS2, 0, false, false);
 	if (sbasize > 32)
-		dmi_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS1, address >> 32, false, false);
-	return dmi_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS0, address,
+		dm_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS1, address >> 32, false, false);
+	return dm_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS0, address,
 				  false, ensure_success);
 }
 
@@ -2399,7 +2444,7 @@ static int sample_memory_bus_v1(struct target *target,
 						sbcs_write |= DM_SBCS_SBREADONDATA;
 					sbcs_write |= sb_sbaccess(config->bucket[i].size_bytes);
 					if (!sbcs_valid || sbcs_write != sbcs) {
-						riscv_batch_add_dmi_write(batch, DM_SBCS, sbcs_write, true);
+						riscv_batch_add_dm_write(batch, DM_SBCS, sbcs_write, true);
 						sbcs = sbcs_write;
 						sbcs_valid = true;
 					}
@@ -2408,18 +2453,18 @@ static int sample_memory_bus_v1(struct target *target,
 							(!sbaddress1_valid ||
 							sbaddress1 != config->bucket[i].address >> 32)) {
 						sbaddress1 = config->bucket[i].address >> 32;
-						riscv_batch_add_dmi_write(batch, DM_SBADDRESS1, sbaddress1, true);
+						riscv_batch_add_dm_write(batch, DM_SBADDRESS1, sbaddress1, true);
 						sbaddress1_valid = true;
 					}
 					if (!sbaddress0_valid ||
 							sbaddress0 != (config->bucket[i].address & 0xffffffff)) {
 						sbaddress0 = config->bucket[i].address;
-						riscv_batch_add_dmi_write(batch, DM_SBADDRESS0, sbaddress0, true);
+						riscv_batch_add_dm_write(batch, DM_SBADDRESS0, sbaddress0, true);
 						sbaddress0_valid = true;
 					}
 					if (config->bucket[i].size_bytes > 4)
-						riscv_batch_add_dmi_read(batch, DM_SBDATA1);
-					riscv_batch_add_dmi_read(batch, DM_SBDATA0);
+						riscv_batch_add_dm_read(batch, DM_SBDATA1);
+					riscv_batch_add_dm_read(batch, DM_SBDATA0);
 					result_bytes += 1 + config->bucket[i].size_bytes;
 				}
 			}
@@ -2430,7 +2475,7 @@ static int sample_memory_bus_v1(struct target *target,
 			break;
 		}
 
-		size_t sbcs_read_index = riscv_batch_add_dmi_read(batch, DM_SBCS);
+		size_t sbcs_read_index = riscv_batch_add_dm_read(batch, DM_SBCS);
 
 		int result = batch_run(target, batch);
 		if (result != ERROR_OK) {
@@ -2452,13 +2497,13 @@ static int sample_memory_bus_v1(struct target *target,
 			/* Discard this batch when we encounter "busy error" state on the System Bus level.
 			 * We'll try next time with a larger System Bus read delay. */
 			info->bus_master_read_delay += info->bus_master_read_delay / 10 + 1;
-			dmi_write(target, DM_SBCS, sbcs_read | DM_SBCS_SBBUSYERROR | DM_SBCS_SBERROR);
+			dm_write(target, DM_SBCS, sbcs_read | DM_SBCS_SBBUSYERROR | DM_SBCS_SBERROR);
 			riscv_batch_free(batch);
 			continue;
 		}
 		if (get_field(sbcs_read, DM_SBCS_SBERROR)) {
 			/* The memory we're sampling was unreadable, somehow. Give up. */
-			dmi_write(target, DM_SBCS, DM_SBCS_SBBUSYERROR | DM_SBCS_SBERROR);
+			dm_write(target, DM_SBCS, DM_SBCS_SBBUSYERROR | DM_SBCS_SBERROR);
 			riscv_batch_free(batch);
 			return ERROR_FAIL;
 		}
@@ -2520,7 +2565,7 @@ static int riscv013_get_hart_state(struct target *target, enum riscv_hart_state 
 		 */
 		if (target->state == TARGET_HALTED)
 			dmcontrol |= DM_DMCONTROL_HALTREQ;
-		dmi_write(target, DM_DMCONTROL, dmcontrol);
+		dm_write(target, DM_DMCONTROL, dmcontrol);
 	}
 	if (get_field(dmstatus, DM_DMSTATUS_ALLNONEXISTENT)) {
 		*state = RISCV_STATE_NON_EXISTENT;
@@ -2583,14 +2628,16 @@ static int init_target(struct command_context *cmd_ctx,
 	generic_info->write_debug_buffer = &riscv013_write_debug_buffer;
 	generic_info->execute_debug_buffer = &riscv013_execute_debug_buffer;
 	generic_info->invalidate_cached_debug_buffer = &riscv013_invalidate_cached_debug_buffer;
-	generic_info->fill_dmi_write_u64 = &riscv013_fill_dmi_write_u64;
-	generic_info->fill_dmi_read_u64 = &riscv013_fill_dmi_read_u64;
-	generic_info->fill_dmi_nop_u64 = &riscv013_fill_dmi_nop_u64;
+	generic_info->fill_dm_write_u64 = &riscv013_fill_dm_write_u64;
+	generic_info->fill_dm_read_u64 = &riscv013_fill_dm_read_u64;
+	generic_info->fill_dm_nop_u64 = &riscv013_fill_dm_nop_u64;
 	generic_info->dmi_write_u64_bits = &riscv013_dmi_write_u64_bits;
 	generic_info->authdata_read = &riscv013_authdata_read;
 	generic_info->authdata_write = &riscv013_authdata_write;
 	generic_info->dmi_read = &dmi_read;
 	generic_info->dmi_write = &dmi_write;
+	generic_info->dm_read = &dm_read;
+	generic_info->dm_write = &dm_write;
 	generic_info->read_memory = read_memory;
 	generic_info->hart_count = &riscv013_hart_count;
 	generic_info->data_bits = &riscv013_data_bits;
@@ -2645,7 +2692,7 @@ static int assert_reset(struct target *target)
 		control = set_field(control, DM_DMCONTROL_HALTREQ,
 				target->reset_halt ? 1 : 0);
 		control = set_field(control, DM_DMCONTROL_NDMRESET, 1);
-		result = dmi_write(target, DM_DMCONTROL, control);
+		result = dm_write(target, DM_DMCONTROL, control);
 		if (result != ERROR_OK)
 			return result;
 	}
@@ -2669,7 +2716,7 @@ static int deassert_reset(struct target *target)
 	control = set_field(control, DM_DMCONTROL_DMACTIVE, 1);
 	control = set_field(control, DM_DMCONTROL_HALTREQ, target->reset_halt ? 1 : 0);
 	control = set_dmcontrol_hartsel(control, info->index);
-	result = dmi_write(target, DM_DMCONTROL, control);
+	result = dm_write(target, DM_DMCONTROL, control);
 	if (result != ERROR_OK)
 		return result;
 
@@ -2723,7 +2770,7 @@ static int deassert_reset(struct target *target)
 	control = set_field(control, DM_DMCONTROL_DMACTIVE, 1);
 	control = set_field(control, DM_DMCONTROL_ACKHAVERESET, 1);
 	control = set_dmcontrol_hartsel(control, info->index);
-	return dmi_write(target, DM_DMCONTROL, control);
+	return dm_write(target, DM_DMCONTROL, control);
 }
 
 static int execute_fence(struct target *target)
@@ -2805,7 +2852,7 @@ static int read_memory_bus_word(struct target *target, target_addr_t address,
 	static int sbdata[4] = { DM_SBDATA0, DM_SBDATA1, DM_SBDATA2, DM_SBDATA3 };
 	assert(size <= 16);
 	for (int i = (size - 1) / 4; i >= 0; i--) {
-		result = dmi_op(target, &sbvalue[i], NULL, DMI_OP_READ, sbdata[i], 0, false, true);
+		result = dm_op(target, &sbvalue[i], NULL, DMI_OP_READ, sbdata[i], 0, false, true);
 		if (result != ERROR_OK)
 			return result;
 		buf_set_u32(buffer + i * 4, 0, 8 * MIN(size, 4), sbvalue[i]);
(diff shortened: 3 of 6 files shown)


zqb-all:
There is an issue about multiple DMs:  https://github.com/riscv/riscv-openocd/issues/377

This patch need to add some checks mentioned in that issue:
> There should be validity checks for the address.
> That it is unique.
> That such debug module really exists.

I sent this patch out because I had some questions and want to see if anyone could give me some advice:

1. I noticed that `select_dtmcontrol` in `dtmcontrol_scan`  https://github.com/riscv/riscv-openocd/blob/a45589d60aa6864475fddcded885c8ff47d50be1/src/target/riscv/riscv-013.c#L434 is init with value  `DTMCONTROL` https://github.com/riscv/riscv-openocd/blob/a45589d60aa6864475fddcded885c8ff47d50be1/src/target/riscv/riscv.c#L40-L44 for multiple DMs, can I just add `dbgbase` to `select_dtmcontrol.out_value` before do `jtag_add_ir_scan` ? 
2. there is a TODO comment about multiple DMs.  https://github.com/riscv/riscv-openocd/blob/a45589d60aa6864475fddcded885c8ff47d50be1/src/target/riscv/riscv-013.c#L1715 Any advice about how to fix it?

src/target/riscv/riscv-013.c
```diff
@@ -29,6 +29,9 @@
 #include "asm.h"
 #include "batch.h"
 
+#define DTMCONTROL					0x10
+#define DBUS						0x11
```
timsifive: DTMCONTROL is old. The register was renamed, and its value is defined in debug_defines.h as DTM_DTMCS.
Likewise DBUS is now DTM_DMI.

src/target/riscv/riscv-013.c
```diff
@@ -417,7 +424,21 @@ static void select_dmi(struct target *target)
...
+		tmp_ir_dbus[0] = (target->dbgbase + DBUS) & 0xff;
+		tmp_ir_dbus[1] = ((target->dbgbase + DBUS) >> 8) & 0xff;
+		tmp_ir_dbus[2] = ((target->dbgbase + DBUS) >> 16) & 0xff;
+		tmp_ir_dbus[3] = ((target->dbgbase + DBUS) >> 24) & 0xff;
+	} else {
+		tmp_ir_dbus[0] = DBUS;
+	}
+	struct scan_field tmp_select_dbus;
+	memcpy(&tmp_select_dbus, &select_dbus, sizeof(struct scan_field));
+	tmp_select_dbus.out_value = tmp_ir_dbus;
```
timsifive: It's simpler to get rid of the global ir_dbus and just create a new one here than it is to mess around with memcpy.

src/target/riscv/riscv-013.c
```diff
@@ -429,9 +450,20 @@ static uint32_t dtmcontrol_scan(struct target *target, uint32_t out)
...
+	if (target->dbgbase_set) {
+		tmp_ir_dtmcontrol[0] = (target->dbgbase + DTMCONTROL) & 0xff;
+		tmp_ir_dtmcontrol[1] = ((target->dbgbase + DTMCONTROL) >> 8) & 0xff;
+		tmp_ir_dtmcontrol[2] = ((target->dbgbase + DTMCONTROL) >> 16) & 0xff;
+		tmp_ir_dtmcontrol[3] = ((target->dbgbase + DTMCONTROL) >> 24) & 0xff;
+	} else {
+		tmp_ir_dtmcontrol[0] = DTMCONTROL;
+	}
+	struct scan_field tmp_select_dtmcontrol;
+	memcpy(&tmp_select_dtmcontrol, &select_dtmcontrol, sizeof(struct scan_field));
```
timsifive: Similar here.

src/target/riscv/riscv-013.c
```diff
@@ -565,6 +600,10 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 		bool *dmi_busy_encountered, int dmi_op, uint32_t address,
 		uint32_t data_out, int timeout_sec, bool exec, bool ensure_success)
 {
+	uint32_t address_with_dbgbase = address;
+	if (target->dbgbase_set)
+		address_with_dbgbase += target->dbgbase;
```
timsifive: This construct where you check for dbg_base_set and then maybe add something is kind of clunky.

When it's not set, can we simply ensure target->dbgbase is 0, and then then all the add/subtract will simply be a nop and leave the code much cleaner.

src/target/riscv/riscv-013.c
timsifive: There should be an error or something in case dbgbase is larger than can be accommodated by abits. We should probably check for that in examine(), assuming that dbgbase must be set by then.

timsifive (commented):
This looks like it's heading in the right direction.

timsifive:
1. I don't think you can update the global variable, because the same one is currently used for every target. You'll have to make a new one for every scan because you never know which DM you'll be asked to talk to. (If you really want you could store a computed value in dm013_info_t, but I doubt the performance difference would be noticeable.)
2. info->index stores the debug index in the DM that it is in. The first hart in each DM has debug index 0. target->coreid is globally unique in OpenOCD. So if there are 2 DMs on a single OpenOCD, then those numbers can't be the same. I think you'll have to track for each DM the highest index seen so far, and assume that the next created target is the next one.

timsifive:
How are you testing these changes?

src/target/riscv/riscv-013.c
zqb-all: ok, this is copied from riscv.c. I will fix it

src/target/riscv/riscv-013.c
zqb-all: > It's simpler to get rid of the global ir_dbus and just create a new one here than it is to mess around with memcpy.

As memcpy is for select_dbus,  I guess you mean remove both global ir_dbus and global select_dbus.
I use memcpy because I'm worried that global select_dbus may change in somewhere I haven't noticed.
I agree get rid of global and memcpy is better, I will do that

zqb-all:
> 2. info->index stores the debug index in the DM that it is in. The first hart in each DM has debug index 0. target->coreid is globally unique in OpenOCD. So if there are 2 DMs on a single OpenOCD, then those numbers can't be the same. I think you'll have to track for each DM the highest index seen so far, and assume that the next created target is the next one.

I see, my hardware happens to have globally unique coreid, so it doesn't bother me here, but it would be better to add some protection

zqb-all:
> How are you testing these changes?

I test it on an fpga.  With this patch, I can debug the cores on 2 different DMs

src/target/riscv/riscv-013.c
zqb-all: https://github.com/riscv/riscv-openocd/blob/4be6a5683f9de5705c5d8a28f4ed9ada2250dccb/src/target/target.c#L6229

 target->dbgbase should be 0 as target is init by calloc

src/target/riscv/riscv-013.c
```diff
@@ -4704,9 +4764,10 @@ void riscv013_fill_dmi_read_u64(struct target *target, char *buf, int a)
 void riscv013_fill_dmi_nop_u64(struct target *target, char *buf)
 {
 	RISCV013_INFO(info);
+	int a = target->dbgbase;
 	buf_set_u64((unsigned char *)buf, DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH, DMI_OP_NOP);
 	buf_set_u64((unsigned char *)buf, DTM_DMI_DATA_OFFSET, DTM_DMI_DATA_LENGTH, 0);
-	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, 0);
+	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, a);
 }
```
JanMatCodasip: For DMI NOP operation, the address and data are both ignored by the JTAG DTM. So I believe you don't need to take the `dbgbase` into account at all -- and therefore the code of riscv013_fill_dmi_nop_u64() can stay as is.

src/target/riscv/riscv-013.c
```diff
@@ -4688,6 +4746,7 @@ int riscv013_execute_debug_buffer(struct target *target)
 void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a, uint64_t d)
 {
 	RISCV013_INFO(info);
+	a += target->dbgbase;
 	buf_set_u64((unsigned char *)buf, DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH, DMI_OP_WRITE);
 	buf_set_u64((unsigned char *)buf, DTM_DMI_DATA_OFFSET, DTM_DMI_DATA_LENGTH, d);
 	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, a);
```
JanMatCodasip: This is just a detail, but for clarity of the code, I'd refrain from modifying the input argument (`a`) and would create a helper variable instead:

```suggestion
	const unsigned dmi_addr = a +target->dbgbase;
	buf_set_u64((unsigned char *)buf, DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH, DMI_OP_WRITE);
	buf_set_u64((unsigned char *)buf, DTM_DMI_DATA_OFFSET, DTM_DMI_DATA_LENGTH, d);
	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, dmi_addr);
```

src/target/riscv/riscv-013.c
```diff
@@ -4696,6 +4755,7 @@ void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a, uint64
 void riscv013_fill_dmi_read_u64(struct target *target, char *buf, int a)
 {
 	RISCV013_INFO(info);
+	a += target->dbgbase;
 	buf_set_u64((unsigned char *)buf, DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH, DMI_OP_READ);
 	buf_set_u64((unsigned char *)buf, DTM_DMI_DATA_OFFSET, DTM_DMI_DATA_LENGTH, 0);
 	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, a);
```
JanMatCodasip: (Same as above - I'd recommend to create a new variable)

src/target/riscv/riscv-013.c
```diff
@@ -1694,6 +1742,9 @@ static int examine(struct target *target)
 {
 	/* Don't need to select dbus, since the first thing we do is read dtmcontrol. */
 
+	if (target->dbgbase)
+		LOG_DEBUG("dbgbase=0x%x", target->dbgbase);
```
JanMatCodasip: Can we print this every time? Also use LOG_TARGET_DEBUG() which includes the target name with the print.

```suggestion
	LOG_TARGET_DEBUG(target, "dbgbase=0x%x", target->dbgbase);
```

src/target/riscv/riscv-013.c
```diff
@@ -2626,6 +2683,7 @@ static int init_target(struct command_context *cmd_ctx,
 
 	info->has_aampostincrement = YNM_MAYBE;
 
+
```
JanMatCodasip: Please, do not add any unneeded blank lines.

```suggestion
```

src/target/riscv/riscv-013.c
```diff
@@ -1726,6 +1777,12 @@ static int examine(struct target *target)
 		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_DMACTIVE);
 		dm->was_reset = true;
 	}
+
+	if (dbgbase_verify(target) == false) {
+		LOG_ERROR("dbgbase 0x%x verify fail\n", target->dbgbase);
```
JanMatCodasip: This error message can be improved, so that the users can understand it better. For example:

```suggestion
		LOG_TARGET_ERROR(target, "Could not find debug module with DMI base address (dbgbase) = 0x%x", target->dbgbase);
```

We should use `LOG_TARGET_*` instead of `LOG_*` where possible.

Also don't include `\n` at the end of the messages. The `LOG_*` functions print the newline automatically.

src/target/riscv/riscv-013.c
```diff
@@ -686,6 +715,25 @@ static int dmi_write_exec(struct target *target, uint32_t address,
 	return dmi_op(target, NULL, NULL, DMI_OP_WRITE, address, value, true, ensure_success);
 }
 
+static bool dbgbase_verify(struct target *target)
```
JanMatCodasip: If you like, you can give this function a more descriptive name:

```suggestion
static bool check_dbgbase_exists(struct target *target)
```

src/target/riscv/riscv-013.c
```diff
@@ -686,6 +715,25 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+{
+	uint32_t dbgbase_set = target->dbgbase;
```
JanMatCodasip: You can call this:

```suggestion
	uint32_t dbgbase_to_find = target->dbgbase;
```

src/target/riscv/riscv-013.c
```diff
@@ -686,6 +715,25 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+	uint32_t nextdm = 0;
```
```suggestion
	/* Start with the first debug module */
	uint32_t nextdm = 0;
```

src/target/riscv/riscv-013.c
```diff
@@ -686,6 +715,25 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+
+	while (1) {
+		target->dbgbase = nextdm;
+		if (target->dbgbase == dbgbase_set)
+			return true;
+		if (dmi_read(target, &nextdm, DM_NEXTDM) != ERROR_OK)
+			break;
+		LOG_DEBUG("dm_nextdm=0x%x", nextdm);
```
JanMatCodasip: I've tried to improve the error message - it would be good to display both the old and new address for easier troubleshooting.

```suggestion
		LOG_TARGET_DEBUG(target, "dm @ 0x%x --> nextdm=0x%x", target->dbgbase, nextdm);
```

src/target/riscv/riscv-013.c
```diff
@@ -686,6 +715,25 @@ static int dmi_write_exec(struct target *target, uint32_t address,
 	return dmi_op(target, NULL, NULL, DMI_OP_WRITE, address, value, true, ensure_success);
 }
 
+static bool dbgbase_verify(struct target *target)
+{
+	uint32_t dbgbase_set = target->dbgbase;
+	uint32_t nextdm = 0;
+
+	while (1) {
```
JanMatCodasip, line 768: If you like, you can add more prints to make the log analysis easier.

```suggestion
	LOG_TARGET_DEBUG(target, "Searching for DM with DMI base address (dbgbase) = 0x%x", target->dbgbase);
	while (1) {
```

src/target/riscv/riscv-013.c
```diff
@@ -1726,6 +1777,12 @@ static int examine(struct target *target)
 		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_DMACTIVE);
 		dm->was_reset = true;
 	}
+
+	if (dbgbase_verify(target) == false) {
```
JanMatCodasip: You can use `(!expr)` instead of `(expr == false)`:

```suggestion
	if (!dbgbase_verify(target)) {
```

src/target/riscv/riscv-013.c
```diff
@@ -386,12 +386,12 @@ static void dump_field(int idle, const struct scan_field *field)
 	uint64_t out = buf_get_u64(field->out_value, 0, field->num_bits);
 	unsigned int out_op = get_field(out, DTM_DMI_OP);
 	unsigned int out_data = get_field(out, DTM_DMI_DATA);
-	unsigned int out_address = out >> DTM_DMI_ADDRESS_OFFSET;
+	unsigned int out_address = (out >> DTM_DMI_ADDRESS_OFFSET) - target->dbgbase;
```
JanMatCodasip: I'd recommend to dump the real address used on the DMI interface, not just the relative one to the base.

```suggestion
	unsigned int out_address = out >> DTM_DMI_ADDRESS_OFFSET;
```

src/target/riscv/riscv-013.c
```diff
@@ -386,12 +386,12 @@ static void dump_field(int idle, const struct scan_field *field)
...
 
 	uint64_t in = buf_get_u64(field->in_value, 0, field->num_bits);
 	unsigned int in_op = get_field(in, DTM_DMI_OP);
 	unsigned int in_data = get_field(in, DTM_DMI_DATA);
-	unsigned int in_address = in >> DTM_DMI_ADDRESS_OFFSET;
+	unsigned int in_address = (in >> DTM_DMI_ADDRESS_OFFSET) - target->dbgbase;
```
JanMatCodasip: (Same as above)

```suggestion
	unsigned int in_address = in >> DTM_DMI_ADDRESS_OFFSET;
```

src/target/riscv/riscv-013.c
```diff
@@ -417,7 +417,19 @@ static void select_dmi(struct target *target)
...
+	uint32_t address = DTM_DMI + target->dbgbase;
+
+	uint8_t ir_dtm_dmi[4] = {address & 0xff, (address >> 8) & 0xff,
+		(address >> 16) & 0xff, (address >> 24) & 0xff};
+
+	struct scan_field select_dtm_dmi = {
+		.in_value = NULL,
+		.out_value = ir_dtm_dmi,
+		.num_bits = target->tap->ir_length
+	};
```
JanMatCodasip: I am afraid this is not correct. 

Even when there are multiple DMs on the DMI bus, there will still be just one JTAG DTM that contains a single`dmi` JTAG data register.

For that reason, the address of this `dmi` JTAG data register -- that is, the value written to the JTAG instruction register (IR) -- should remain the same, and definitely should not be incremented by `target->dbgbase`.

I believe there should be no change in `select_dmi()`.

src/target/riscv/riscv-013.c
```diff
@@ -429,9 +441,19 @@ static uint32_t dtmcontrol_scan(struct target *target, uint32_t out)
...
+		(address >> 16) & 0xff, (address >> 24) & 0xff};
+
+	struct scan_field select_dtm_dtmcs = {
+		.in_value = NULL,
+		.out_value = ir_dtm_dtmcs,
+		.num_bits = target->tap->ir_length
+	};
+
+	buf_set_u32(out_value, 0, 32, out);
+	jtag_add_ir_scan(target->tap, &select_dtm_dtmcs, TAP_IDLE);
```
JanMatCodasip: This is the exactly same situation as in `select_dmi()`: 

I believe the address of the DMTCS, written to the JTAG IR, should remain identical, regardless of `target->dbgbase`.

src/target/riscv/riscv-013.c
```diff
@@ -565,6 +589,11 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 		bool *dmi_busy_encountered, int dmi_op, uint32_t address,
 		uint32_t data_out, int timeout_sec, bool exec, bool ensure_success)
 {
+	if (!target->dbgbase_set)
+		assert(target->dbgbase == 0);
```
JanMatCodasip: This can be re-written as:

```suggestion
	/* Safety check for dbgbase */
	assert(target->dbgbase_set || target->dbgbase == 0);
```

src/target/riscv/riscv-013.c
```diff
@@ -565,6 +589,11 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
...
+
+	uint32_t address_with_dbgbase = address + target->dbgbase;
```
JanMatCodasip: If you like, this can be `const` (makes the life a bit easier for whoever reads the code).

```suggestion
	const uint32_t address_with_dbgbase = address + target->dbgbase;
```

JanMatCodasip: Please also move the variable definition close to the place of first use.

src/target/riscv/riscv-013.c
```diff
@@ -499,6 +521,8 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 
 	assert(info->abits != 0);
 
+	address_out += target->dbgbase;
```
JanMatCodasip: I'd recommend to create another variable and not modify the input parameter value (detail, but helps code readability):

```suggestion
	const unint32_t address_with_dbgbase = address_out + target->dbgbase;
```

JanMatCodasip (changes requested):
Hi @zqb-all, 
thank you for taking the time to prepare this change. I believe there are some items to fix - please, take a look at my comments.

zqb-all:
> Hi @zqb-all, thank you for taking the time to prepare this change. I believe there are some items to fix - please, take a look at my comments.

Thank you for your detailed review suggestions, which are very helpful. I will add them to this patch tomorrow

src/target/riscv/riscv-013.c
zqb-all: Oh, my mistake,  In the [previous version](https://github.com/riscv/riscv-openocd/commit/b2948d319e5768eb2543ec5c5ab1645c1c0ff496#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63L403-L404), I modified the address after the address was dump and before the decode_dmi, as decode_dmi need relative address.
In the later version, when I delete `if (target->dbgbase_set)` , I forgot the address dump above and put the address change directly to the top. 
Thanks for pointing it out, I'll put it before decode_dmi, so we can dump the real address.

src/target/riscv/riscv-013.c
zqb-all: Thanks. I really don't understand this function very well. In my tests, whether or not I modified select_dmi/dtmcontrol_scan did not affect my debugging of core on 2 DMs.
I will remove the changes to select_dmi/dtmcontrol_scan from this patch.

timsifive:
> I see, my hardware happens to have globally unique coreid, so it doesn't bother me here, but it would be better to add some protection

What do you mean by coreid? In the hardware there are 2 numbers that are relevant AFAIK:
1. The RISC-V hartid, readable in the mhartid CSR. These can be anything, but must be "unique within the execution environment."
2. The index of each hart on the Debug Module. For each Debug Module these must start at 0 and be contiguous.

zqb-all:
> What do you mean by coreid? In the hardware there are 2 numbers that are relevant AFAIK:
> 1. The RISC-V hartid, readable in the mhartid CSR. These can be anything, but must be "unique within the execution environment."
> 2. The index of each hart on the Debug Module. For each Debug Module these must start at 0 and be contiguous.

It seems that I misunderstood something, I thought you meant that the hart on different DMs might have same hartid.

Let me clarify my usage.

I mean the number which config as `-coreid` in `target create` is unique.
When I debug 2 DMs and each with 2 harts , my config file is like
```
# cluster 0
target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -rtos hwthread

# cluster 1
target create $_TARGETNAME_4 riscv -chain-position $_CHIPNAME.cpu -coreid 4 -dbgbase 0x400
target create $_TARGETNAME_5 riscv -chain-position $_CHIPNAME.cpu -coreid 5 -dbgbase 0x400

target smp $_TARGETNAME_0 $_TARGETNAME_1 $_TARGETNAME_4 $_TARGETNAME_5
```

this config file works with the code
https://github.com/riscv/riscv-openocd/blob/a45589d60aa6864475fddcded885c8ff47d50be1/src/target/riscv/riscv-013.c#L1715-L1716

timsifive:
So that then means that to select the first hart on cluster 1 (DM 1), you write 4 to dmcontrol.hartsel. If that works, then your implementation is not compliant with the debug spec. The first hart in a DM must be selected by dmcontrol.hartsel=0, the second by dmcontrol.hartsel=1, and so on. (This is so that it's possible for a debugger to automatically detect every hart on the DM. The fact that OpenOCD still requires you to manually list them is a problem with OpenOCD, but other debuggers exist.)

zqb-all:
> So that then means that to select the first hart on cluster 1 (DM 1), you write 4 to dmcontrol.hartsel. 

Yes
> If that works, then your implementation is not compliant with the debug spec. The first hart in a DM must be selected by dmcontrol.hartsel=0, the second by dmcontrol.hartsel=1, and so on. 

This DM implementation only handle dmcontrol.hartsel=0/1/2/3 now，so when I write 4/5 to dmcontrol.hartsel, the high bits are ignored, just like write 0/1 to dmcontrol.hartsel

So it's a coincidence that it just works, I get the point now and will fix it.

when config with `target create --coreid xxx`,  `xxx` means hartid and is unique. Then openocd should iterate over all harts and read out mhartid , if mhartid match xxx，set info->index. Is this the correct behavior we expect ?

zqb-all:
> So that then means that to select the first hart on cluster 1 (DM 1), you write 4 to dmcontrol.hartsel. 

Yes
> If that works, then your implementation is not compliant with the debug spec. The first hart in a DM must be selected by dmcontrol.hartsel=0, the second by dmcontrol.hartsel=1, and so on. 

This DM implementation only handle dmcontrol.hartsel=0/1/2/3 now，so when I write 4/5 to dmcontrol.hartsel, the high bits are ignored, just like write 0/1 to dmcontrol.hartsel

So it's a coincidence that it just works, I get the point now ans will fix it

src/target/riscv/riscv-013.c
zqb-all: new patch don't modify this funtion

zqb-all:
> when config with `target create --coreid xxx`, `xxx` means hartid and is unique. Then openocd should iterate over all harts and read out mhartid , if mhartid match xxx，set info->index. Is this the correct behavior we expect ?

Iterate over all harts (contain those  not examine yet) and read the mhartid CSR, that doesn't sounds like a good idea, 
Is there a better way for us to map hartid and info->index(The index of hart on the Debug Module) ?

timsifive:
> This DM implementation only handle dmcontrol.hartsel=0/1/2/3 now，so when I write 4/5 to dmcontrol.hartsel, the high bits are ignored, just like write 0/1 to dmcontrol.hartsel

That's a clever trick to make this work for your implementation. :)

> when config with target create --coreid xxx, xxx means hartid and is unique. Then openocd should iterate over all harts and read out mhartid , if mhartid match xxx，set info->index. Is this the correct behavior we expect ?

I don't think so. It's conceivable that OpenOCD is debugging 2 completely separate RISC-V chips (on different scan chains, or even merely different DMs). Since they're completely separate, they could have conflicting hart ids. So we shouldn't force OpenOCD coreid to be the same as RISC-V hartid. At the same time, if it is possible then we do want them to be the same. It just makes everything much easier for users. For the purposes of this pull request, I don't think you need to do anything special regarding hartids. It is a separate issue, that can be solved separately some day.

> Iterate over all harts (contain those not examine yet) and read the mhartid CSR, that doesn't sounds like a good idea,
Is there a better way for us to map hartid and info->index(The index of hart on the Debug Module) ?

I don't think it's as bad as you think. examine() already halts each hart and reads misa. Reading mhartid in addition is not a big change. Some day there will be a unified discovery mechanism, but that group is only just starting up again.

But I don't think we need to worry about hartids right now. We do need to worry about debug index and OpenOCD coreid no longer being equal.

zqb-all:
> But I don't think we need to worry about hartids right now. We do need to worry about debug index and OpenOCD coreid no longer being equal.

Make sense, ignore hartid for now would make things easier.
How about let `info->index = target->coreid - target->dbgbase` ? 

config ` target create  -coreid 0x0` --> `target->coreid=0, target->dbgbase=0, info->index = 0`
config ` target create  -coreid 0x1` --> `target->coreid=1, target->dbgbase=0, info->index = 1`
config ` target create  -coreid 0x400 -dbgbase 0x400` --> `target->core_d=0x400, target->dbgbase=0x400, info->index = 0`
config ` target create  -coreid 0x401 -dbgbase 0x400` --> `target->coreid=0x401, target->dbgbase=0x400, info->index = 1`

src/target/riscv/riscv-013.c
```diff
@@ -1712,8 +1720,12 @@ static int examine(struct target *target)
...
 
 	riscv013_info_t *info = get_info(target);
-	/* TODO: This won't be true if there are multiple DMs. */
-	info->index = target->coreid;
+	/* As info->index is the index of each hart on the DM (Debug Module),
+	 * for each DM these must start at 0 and be contiguous.
+	 * And target->coreid is target->coreid is globally unique in OpenOCD.
+	 * So for multiple DMs, info->index and target->coreid can't be the same.
+	 * Consider different DM have different dbgbase, let coreid = index + dbgbase */
+	info->index = target->coreid - target->dbgbase;
```
zqb-all: should I add  `if (target->coreid >= target->dbgbase)` for ` info->index = target->coreid - target->dbgbase;`

JanMatCodasip:
I will be able to re-review this change in the week of July 10.

src/target/riscv/riscv-013.c
```diff
@@ -1712,8 +1741,17 @@ static int examine(struct target *target)
 	}
 
 	riscv013_info_t *info = get_info(target);
-	/* TODO: This won't be true if there are multiple DMs. */
-	info->index = target->coreid;
+	/* As info->index is the index of each hart on the DM (Debug Module),
+	 * for each DM these must start at 0 and be contiguous.
+	 * And target->coreid is target->coreid is globally unique in OpenOCD.
```
MarekVCodasip: This looks like a word repeat typo

src/target/riscv/riscv-013.c
```diff
@@ -1712,8 +1741,17 @@ static int examine(struct target *target)
...
+	 * So for multiple DMs, info->index and target->coreid can't be the same.
+	 * Consider different DM have different dbgbase, let coreid = index + dbgbase */
+	if ((unsigned int)target->coreid < target->dbgbase) {
+		LOG_TARGET_ERROR(target, "coreid should not be smaller than DMI base address (dbgbase), coreid=0x%x dbgbase=0x%x",
+				target->coreid, target->dbgbase);
+		return ERROR_FAIL;
+	}
+	info->index = target->coreid - target->dbgbase;
```
MarekVCodasip: This does not seem right to me. I believe coreids and debug module base addresses should be independent.


For example, consider debug module which controls 1024 harts and base DMI address 0x0. Base DMI address of the next debug module could be for example 0x100. It does not have to be 0x400 (which would be 1024 in decimal).

src/target/riscv/riscv-013.c
zqb-all: oh, duplicate `target->coreid is`， thanks

src/target/riscv/riscv-013.c
zqb-all: Yes, coreid and dbgbase really don't have to be associated, I just want to find a simple way to map unique coreid to index. But as you said, this mapping method is not appropriate for some case, I will think about it again. 
Thank you for pointing it out.

zqb-all: Up to `2^20` harts can be connected to a single DM, how about use higher bits to distinguish between different DMs?

That means `info->index = target->coreid & (1<<20 - 1)`

timsifive: As long as the high bits are a DM index and not a DM address then that storage scheme seems fine.

It's a bit user-unfriendly to have to write:
`target create $_TARGETNAME_1 riscv -chain-position $_CHIPNAME.cpu -coreid 1048577`

But we could address that separately by introducing a new tcl command that takes separate DM index and debug index. It won't look pretty either, but again that's something we can address later.

Is there any code that creates arrays of stuff to fit every coreid?

zqb-all: > It's a bit user-unfriendly to have to write:
> `target create $_TARGETNAME_1 riscv -chain-position $_CHIPNAME.cpu -coreid 1048577`

`0x100001`  looks friendlier than `1048577`

> Is there any code that creates arrays of stuff to fit every coreid?

Sorry, I don't quite understand the question, could you explain it in more detail

timsifive: Is there any code that relies on the fact that coreids are usually small? E.g. somewhere maybe code creates an array indexed by coreid. That is fine if coreid numbers are small, but if all of a sudden that code needs to deal with core IDs like 0x100001, then memory is allocated for 0x100000 unused elements, which is excessively wasteful.

zqb-all: I found no such code in openocd. And when I config 1 core with coreid 0, Windows Task Manager show that openocd cost 2.7MB, when config 4 core with coreid 0/1/0x100000/0x100001, openocd cost 4.4MB.

src/target/riscv/riscv-013.c
```diff
@@ -1837,11 +1845,16 @@ static int examine(struct target *target)
...
+	if (dm->hart_count <= 0) {
 		LOG_ERROR("No harts found!");
 		return ERROR_FAIL;
 	}
 
+	if (info->index >= (unsigned int)dm->hart_count) {
+		LOG_ERROR("Index out of range, index=0x%x hartcount=0x%x", info->index, dm->hart_count);
+		return ERROR_FAIL;
+	}
+
```
zqb-all: if we use  `info->index %= dm->hart_count` here, then will allow multiple DMs avoid huge coreid，for example, when each DM have n harts, first DM can use coreid `0 - (n - 1)`, second DM can use coreid `n - (2n - 1)`, looks friendlier. 
Of course, use bits higher than 20 to distinguish DM still works, second DM can use coreid `0x100000 - (0x100000 + n - 1)`

src/target/riscv/riscv-013.c
JanMatCodasip: Actually, why try to squeeze the core id and debug module base into the `coreid` variable in the first place? :)

> Tim: ... target->coreid is globally unique in OpenOCD ...

I believe there is no such requirement in OpenOCD for target->coreid to be globally unique. 

In principle, the coreid only needs to be unique for given RISC-V debug module -- identified by combination of `dbgbase` and `JTAG chain_position`. (And if the user does not obey this rule, then that is a problem of the user-written configuration. We may add some checks for it one day, not necessarily in this merge request.)

JanMatCodasip: > Tim: But I don't think we need to worry about hartids right now. We do need to worry about debug index and OpenOCD coreid no longer being equal.

As indicated in the previous comment, I think this is not an issue. Core IDs can remain equal to Hart IDs.

I believe that what needs to be dealt with is the `get_dm()` function that allocates new debug module instances. Currently, only JTAG chain position (`target->tap->abs_chain_position`) is taken into account. The value of `dbgbase` needs to be taken into account now, too.
https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv-013.c#L232

zqb-all: When I try to config all harts(from different DMs) into one smp, coreid shoule be globally unique because threadid comes from coreid
https://github.com/riscv/riscv-openocd/blob/21d21408aa36ccb458ccfaad871f7dd1bb1f731a/src/rtos/hwthread.c#L40-L43

But then I realized that config smp with harts from different DMs is not that easy, more things need to deal with, such as haltgroup. So this PR can ignore it.

I don't know yet if there are other places that require coreid to be globally unique.

JanMatCodasip: > But then I realized that config smp with harts from different DMs is not that easy

I agree. If I understand the current state correctly, SMP for RISC-V targets on different DMs is not supported (or at least it is not supported well). RISC-V debug spec does not even provide mechanisms for simultaneous halting/resuming of harts across multiple DMs.

zqb-all: >  Core IDs can remain equal to Hart IDs.

This PR change ` info->index = target->coreid`  to ` info->index = target->coreid & (RISCV_MAX_HARTS - 1);`,
As coreid don't need to be globally unique, should I keep this change（like a protect, because up to 2^20 harts can be connected to a single DM）? or just drop the change, keep the original code  ` info->index = target->coreid` ?

zqb-all: > I believe that what needs to be dealt with is the get_dm() function that allocates new debug module instances.

Thanks, that's really a problem, I'll handle it. There was no problem when debugging only one DM at a time, so I didn't notice it in time.

src/target/riscv/riscv-013.c
```diff
@@ -1712,8 +1745,8 @@ static int examine(struct target *target)
 	}
 
 	riscv013_info_t *info = get_info(target);
-	/* TODO: This won't be true if there are multiple DMs. */
-	info->index = target->coreid;
+	/* Up to 2^20 harts can be connected to a single DM */
+	info->index = target->coreid & (RISCV_MAX_HARTS - 1);
```
```suggestion
	info->index = target->coreid % RISCV_MAX_HARTS;
```

Seems sloppy to assume the constant is a power of 2, even though it is. The performance difference doesn't matter.

timsifive (commented):
So `coreid` is commented to be "which device on the TAP?"

Both `coreid` and `dbgbase` look like they are purely interpreted by the target layer, so we can use them like that. As @JanMatCodasip points out, there is no need for a complex scheme where we use very high core ids since they don't have to be globally unique. The comment implies they have to be unique per TAP, but I would be surprised if that is also the case.

src/target/riscv/riscv-013.c
zqb-all: Thanks

src/target/riscv/riscv-013.c
JanMatCodasip: >  As coreid don't need to be globally unique, should I keep this change（like a protect, because up to 2^20 harts can be connected to a single DM）? or just drop the change, keep the original code  info->index = target->coreid ?

I think it is OK to keep the current state: `info->index = target->coreid`

JanMatCodasip:
> Tim: The comment implies they [core IDs] have to be unique per TAP, but I would be surprised if that is also the case.

It seems to me that such comment is inaccurate. Most likely it applied to architectures where `one TAP == one DM` (that is, where `one TAP --> multiple DMs` was not a possibility).

src/target/riscv/riscv-013.c
zqb-all: ok, keep the current state make this PR simpler

zqb-all:
Is there anything unclear about this patch that I need to confirm? Or do I just need to wait for further review?

JanMatCodasip:
> Is there anything unclear about this patch that I need to confirm? Or do I just need to wait for further review?

Thank you for the reminder. I will be happy to re-review this, and will be able to do it later this week.

src/target/riscv/riscv-013.c
```diff
@@ -120,6 +120,7 @@ typedef enum slot {
 typedef struct {
 	struct list_head list;
 	int abs_chain_position;
+	uint32_t dbgbase;
```
timsifive: Can you add a comment describing what this is for?

src/target/riscv/riscv-013.c
```diff
@@ -236,11 +237,13 @@ static dm013_info_t *get_dm(struct target *target)
 		return info->dm;
 
 	int abs_chain_position = target->tap->abs_chain_position;
+	uint32_t dbgbase = target->dbgbase;
```
timsifive: I'd remove the local `dbgbase` and just refer to target->dbgbase directly where it's needed.

src/target/riscv/riscv-013.c
```diff
@@ -499,9 +503,11 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 
 	assert(info->abits != 0);
 
+	const uint32_t address_out_with_dbgbase = address_out + target->dbgbase;
```
timsifive: Seems unnecessary here, especially given it's used only once.

src/target/riscv/riscv-013.c
```diff
@@ -686,6 +696,27 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
 
+static bool check_dbgbase_exists(struct target *target)
+{
+	uint32_t dbgbase_to_find = target->dbgbase;
+	/* Start with the first debug module */
+	uint32_t nextdm = 0;
+
+	LOG_TARGET_DEBUG(target, "Searching for DM with DMI base address (dbgbase) = 0x%x", target->dbgbase);
+	while (1) {
+		target->dbgbase = nextdm;
```
timsifive: Can you rewrite this function not to change target at all? Using `target->dbgbase` as a temporary variable, only to restore it at the end, makes the code hard to follow.

src/target/riscv/riscv-013.c
```diff
@@ -686,6 +696,27 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
 }
 
+static bool check_dbgbase_exists(struct target *target)
+{
+	uint32_t dbgbase_to_find = target->dbgbase;
+	/* Start with the first debug module */
+	uint32_t nextdm = 0;
+
+	LOG_TARGET_DEBUG(target, "Searching for DM with DMI base address (dbgbase) = 0x%x", target->dbgbase);
+	while (1) {
```
timsifive, line 768: Can you limit this loop to 100 DMs or so? We should try to avoid an infinite loop if the hardware is broken.

Ideally we'd check for duplicates, but that seems more complicated to implement and unnecessary.

src/target/riscv/riscv-013.c
```diff
@@ -1726,6 +1759,12 @@ static int examine(struct target *target)
 		dmi_write(target, DM_DMCONTROL, DM_DMCONTROL_DMACTIVE);
 		dm->was_reset = true;
 	}
+
+	if (!check_dbgbase_exists(target)) {
```
timsifive: This should happen before we attempt any other accesses to the DM.

timsifive (changes requested):
I'm pleasantly surprised this is all it takes to support multiple DMs. Overall this looks quite straightforward.

src/target/riscv/riscv-013.c
zqb-all: Change `target->dbgbase` is because `dmi_read`(which eventually call `dmi_scan`) will use ` target->dbgbase` to access the target DM

src/target/riscv/riscv-013.c
zqb-all: done

src/target/riscv/riscv-013.c
zqb-all: done

src/target/riscv/riscv-013.c
zqb-all: done

src/target/riscv/riscv-013.c
zqb-all: done

src/target/riscv/riscv-013.c
zqb-all: done

src/target/riscv/riscv-013.c
```diff
@@ -120,7 +120,8 @@ typedef enum slot {
 typedef struct {
 	struct list_head list;
 	int abs_chain_position;
-
+	/* The base address to access this DM on DMI */
+	uint32_t dbgbase;
```
timsifive: Is this used anywhere? You set it, but then everywhere dbgbase is needed you get it from target instead of dm.

src/target/riscv/riscv-013.c
timsifive: I see. I think that highlights a problem with this implementation, where layers of abstraction are mixed together where they really ought to be separate.

timsifive (changes requested):
The existing code basically has `target`s and DMI. It calls e.g. dmi_read() with address that assume it's talking to the first DM. That works if there's one DM, but otherwise it falls apart.

The real layers that exist are:
* OpenOCD target
* Debug Module (DM). Each target has a DM.
* Debug Module Interface (DMI). Each DM has a base address on the DMI.

You addressed this by adding an offset to the target. Effectively you changed every function from one that takes a DMI address to one that takes an offset into a DM.

The "right" way to address the 3 layers is to have a set of functions for each one:
1. target functions: e.g. wait_for_idle()
2. DM functions, e.g. dm_read() (which does not currently exist)
3. DMI functions, e.g. dmi_read() (which does exist)

The way to make the code be nice:
1. Create dm_read(), dm_write(), and a few others.
2. Change every existing call to dmi_read() etc. to dm_read().
3. Now you have the primitives required to implement check_dbgbase_exists(), or other possible future functions that want to access things that are not DMs.

That is going to make the diff a lot larger, but it should be a very straightforward conversion.

Does that make sense? @JanMatCodasip, am I going overboard here?

src/target/riscv/riscv-013.c
zqb-all: use in `get_dm`
https://github.com/riscv/riscv-openocd/blob/be70adca6f20ea20d906027a01eda8b2526f21ef/src/target/riscv/riscv-013.c#L244-L245

src/target/riscv/riscv-013.c
zqb-all: I need to go through all the DMs starting at 0.
The current implementation is to pass the `target` to `dmi_read`, so I need to keep changing the `target->dbgbase`.
If change to pass `dm` to `dmi_read`, then I need to keep changing `dm->dbgbase`.
I'll try that.

zqb-all:
A quick look at the code, it seems can not just change
https://github.com/riscv/riscv-openocd/blob/be70adca6f20ea20d906027a01eda8b2526f21ef/src/target/riscv/riscv-013.c#L675-L678
 to
```c
  static int dm_read(struct target *target, uint32_t *value, uint32_t address)
  {
      dm013_info_t *dm = get_dm(target);
      if (!dm)
          return ERROR_FAIL;
      dmi_read(dm, value, address);
  }

  static int dmi_read(struct dm013_info_t *dm, uint32_t *value, uint32_t address)
  {
      return dmi_op(dm, value, NULL, DMI_OP_READ, address, 0, false, true);
  }
```

because `dmi_read-->dmi_op-->dmi_scan` still need `target` to get `info`
https://github.com/riscv/riscv-openocd/blob/be70adca6f20ea20d906027a01eda8b2526f21ef/src/target/riscv/riscv-013.c#L478-L480


I'll look at the code another time and think about how to make the layers target/dm/dmi clearer.

src/target/riscv/riscv-013.c
```diff
@@ -686,6 +693,30 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+		target->dbgbase = nextdm;
+		if (target->dbgbase == dbgbase_to_find)
+			return true;
+		if (dmi_read(target, &nextdm, DM_NEXTDM) != ERROR_OK)
+			break;
+		LOG_TARGET_DEBUG(target, "dm @ 0x%x --> nextdm=0x%x", target->dbgbase, nextdm);
+		if (nextdm == 0) /* last one in the chain */
+			break;
+		if (count++ >= RISCV_MAX_DMS)
+			break;
```
JanMatCodasip: Here you may add a comment, explaining why there is the loop counter - it may not be clear to the reader why it is there.

```suggestion
		/* Safety: Avoid looping forever in case of buggy nexdm values in the hardware. */
		if (count++ >= RISCV_MAX_DMS)
			break;
```

JanMatCodasip:
> Tim: (...) The "right" way to address the 3 layers is to have a set of functions for each one (...)

@timsifive I agree. This separation of the layers and the corresponding functions makes perfect sense to me. 

At the moment, there will be just a small difference between the `dm_*()` and `dmi_*()` layer: only the value of `dbgbase` is added to the address. In future, there can be more differences, for example different of logging the transactions.

JanMatCodasip (commented):
I have only a single last minor review comment, otherwise the implementation looks fine!

I believe the only remaining thing prior to the merge is to address Tim's suggestion about the `dm_*()` vs. `dmi_*()` functions. I agree that this would improve the logical structure of the code.

JanMatCodasip:
> zqb-all: A quick look at the code, it seems can not just change (....)

@zqb-all, I believe Tim's idea was to:

- Don't add the `dbgbase` to the address in `dmi_scan()` - leave the address as is.
- Instead, introduce new functions `dm_read()` and `dm_write()`. Apply the `dbgbase` in these functions, as shown below.
- Replace existing calls `dmi_read()` and `dmi_write()` by the new `dm_read()` and `dm_write()`. Except for the special case of `check_dbgbase_exists()`, where `dmi_read()` would be used.

```
static int dm_read(struct target *target, uint32_t *value, uint32_t address)
{
	return dmi_read(target, value, address + target->dbgbase);
}

// dm_write() - analogous
```

Hope this helps :)

JanMatCodasip (dismissed):

zqb-all:
@JanMatCodasip Thanks, very helpful

src/target/riscv/riscv-013.c
zqb-all: done

src/target/riscv/riscv-013.c
zqb-all: done

src/target/riscv/riscv.c
```diff
@@ -4272,18 +4271,18 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
 	},
 	{
-		.name = "dmi_write",
-		.handler = riscv_dmi_write,
+		.name = "dm_write",
+		.handler = riscv_dm_write,
 		.mode = COMMAND_ANY,
 		.usage = "address value",
-		.help = "Perform a 32-bit DMI write of value at address."
+		.help = "Perform a 32-bit DM write of value at address."
```
timsifive: We can't simply remove dmi_{read,write}. There are people relying on these commands. I agree that having dm_{read,write} makes sense. Usually when renaming a command we set up an alias so the old one still works for a while. For this case I think it makes more sense to simply have both, since they're useful for different scenarios.

Finally, you don't even have to change dmi_{read,write} at all for this PR (but the behavior must remain the same).

src/target/riscv/riscv-013.c
```diff
@@ -686,10 +704,88 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+		uint32_t current_dm = next_dm;
+		if (current_dm == target->dbgbase)
+			return true;
+		if (dmi_read(target, &next_dm, DM_NEXTDM + current_dm) != ERROR_OK)
+			break;
+		LOG_TARGET_DEBUG(target, "dm @ 0x%x --> nextdm=0x%x", current_dm, next_dm);
+		/* Check if it's last one in the chain. */
+		if (next_dm == 0)
+			break;
+		/* Safety: Avoid looping forever in case of buggy nexdm values in the hardware. */
```
```suggestion
		/* Safety: Avoid looping forever in case of buggy nextdm values in the hardware. */
```

src/target/riscv/riscv-013.c
```diff
@@ -686,10 +704,88 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+	while (1) {
+		uint32_t current_dm = next_dm;
+		if (current_dm == target->dbgbase)
+			return true;
+		if (dmi_read(target, &next_dm, DM_NEXTDM + current_dm) != ERROR_OK)
+			break;
+		LOG_TARGET_DEBUG(target, "dm @ 0x%x --> nextdm=0x%x", current_dm, next_dm);
+		/* Check if it's last one in the chain. */
+		if (next_dm == 0)
+			break;
```
timsifive: LOG_TARGET_ERROR here to indicate the failure.

src/target/riscv/riscv-013.c
```diff
@@ -686,10 +704,88 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+		/* Safety: Avoid looping forever in case of buggy nexdm values in the hardware. */
+		if (count++ >= RISCV_MAX_DMS)
+			break;
```
timsifive: LOG_TARGET_ERROR here to indicate the failure.

timsifive (changes requested):

src/target/riscv/riscv.c
zqb-all: done

timsifive (dismissed):

doc/openocd.texi
```diff
@@ -10902,6 +10902,14 @@ Perform a 32-bit DMI read at address, returning the value.
...
 @end deffn
 
+@deffn {Command} {riscv dm_read} address
+Perform a 32-bit DM read at address, returning the value.
+@end deffn
+
+@deffn {Command} {riscv dm_write} address value
+Perform a 32-bit DM write of value at address.
+@end deffn
+
```
JanMatCodasip, line 10917: Here I'd like to suggest little changes in the documentation:

The `riscv dm_read` and `riscv dm_write` commands should go first, because they are expected to be used more often.

The text should be updated to clearly show the difference between these commands. My proposal is below. 

@timsifive Does that wording work for you, too?

```
The following commands allow for direct low-level access to the registers 
of the Debug Module (DM) via the Debug Module Interface (DMI) bus. 
These commands can be used to interact with custom debug features.

@deffn {Command} {riscv dm_read} reg_address
Perform a 32-bit read from a DM register at reg_address, returning the value. 
The reg_address should be relative to the base DMI address of the DM. 
OpenOCD adds the DM base address (dbgbase) automatically.
@end deffn

@deffn {Command} {riscv dm_write} reg_address value
Perform a 32-bit write to a DM register at reg_address. 
The reg_address should be relative to the base DMI address of the DM. 
OpenOCD adds the DM base address (dbgbase) automatically.
@end deffn

Note: The DM base address (dbgbase) is typically zero unless there 
are multiple DMs on the DMI bus.

@deffn {Command} {riscv dmi_read} address
Perform a 32-bit read from the given DMI address, returning the value.
@end deffn

@deffn {Command} {riscv dmi_write} address value
Perform a 32-bit write to the given DMI address.
@end deffn
```

src/target/riscv/batch.h
```diff
@@ -58,14 +58,14 @@ bool riscv_batch_full(struct riscv_batch *batch);
 /* Executes this scan batch. */
 int riscv_batch_run(struct riscv_batch *batch);
 
-/* Adds a DMI write to this batch. */
-void riscv_batch_add_dmi_write(struct riscv_batch *batch, unsigned int address, uint64_t data,
+/* Adds a DM write to this batch. */
```
JanMatCodasip: Detail, for clarity:

```suggestion
/* Adds a DM register write to this batch. */
```

src/target/riscv/batch.h
```diff
@@ -58,14 +58,14 @@ bool riscv_batch_full(struct riscv_batch *batch);
...
+void riscv_batch_add_dm_write(struct riscv_batch *batch, unsigned int address, uint64_t data,
 	bool read_back);
 
-/* DMI reads must be handled in two parts: the first one schedules a read and
+/* DM reads must be handled in two parts: the first one schedules a read and
```
JanMatCodasip:
```suggestion
/* DM register reads must be handled in two parts: the first one schedules a read and
```

src/target/riscv/riscv-013.c
```diff
@@ -686,10 +704,92 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+	while (1) {
+		uint32_t current_dm = next_dm;
+		if (current_dm == target->dbgbase)
+			return true;
+		if (dmi_read(target, &next_dm, DM_NEXTDM + current_dm) != ERROR_OK)
+			break;
+		LOG_TARGET_DEBUG(target, "dm @ 0x%x --> nextdm=0x%x", current_dm, next_dm);
+		/* Check if it's last one in the chain. */
+		if (next_dm == 0) {
+			LOG_TARGET_ERROR(target, "Reach end of DM chain");
```
JanMatCodasip: I believe you can change this error print to debug. There is already a good and easy-to-understand error message in the function that calls `check_dbgbase_exists()`. 

If someone needs to dig deeper and troubleshoot the situation, they will most likely enable the verbose log (`-d`) and will be able to easily find out what happened.

```
LOG_TARGET_DEBUG(target, "Reach end of DM chain");
```

src/target/riscv/riscv.c
```diff
@@ -3524,6 +3523,76 @@ COMMAND_HANDLER(riscv_dmi_write)
...
+	}
+
+	if (r->dm_read) {
+		uint32_t address, value;
+		COMMAND_PARSE_NUMBER(u32, CMD_ARGV[0], address);
+		if (r->dm_read(target, &value, address) != ERROR_OK)
+			return ERROR_FAIL;
+		command_print(CMD, "0x%" PRIx32, value);
+	} else {
+		LOG_ERROR("dm_read is not implemented for this target.");
```
JanMatCodasip: I believe you can use LOG_TARGET_ERROR here:

```suggestion
		LOG_TARGET_ERROR(target, "dm_read is not implemented for this target.");
```

src/target/riscv/riscv.c
```diff
@@ -3524,6 +3523,76 @@ COMMAND_HANDLER(riscv_dmi_write)
...
+		bool dm_deactivated = (address == DM_DMCONTROL && (value & DM_DMCONTROL_DMACTIVE) == 0);
+		if (progbuf_touched || dm_deactivated) {
+			if (r->invalidate_cached_debug_buffer)
+				r->invalidate_cached_debug_buffer(target);
+		}
+
+		return retval;
+	}
+
+	LOG_ERROR("dm_write is not implemented for this target.");
```
JanMatCodasip: Same as above - LOG_TARGET_ERROR can be used here:

```suggestion
	LOG_TARGET_ERROR(target, "dm_write is not implemented for this target.");
```

JanMatCodasip (changes requested):
This looks fine! I have only few last suggestions regarding the documentation, comments and prints.

@zqb-all Thanks for bearing with all the review comments.

doc/openocd.texi
zqb-all: Thanks, and if @timsifive  has no other comments, I'll use this version

src/target/riscv/riscv.c
```diff
@@ -4285,6 +4354,20 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.usage = "address value",
 		.help = "Perform a 32-bit DMI write of value at address."
 	},
+	{
+		.name = "dm_read",
+		.handler = riscv_dm_read,
+		.mode = COMMAND_ANY,
+		.usage = "address",
+		.help = "Perform a 32-bit DM read at address, returning the value."
```
JanMatCodasip: One more detail - relates to the already-changed documentation:

```suggestion
		.handler = riscv_dm_read,
		.mode = COMMAND_ANY,
		.usage = "reg_address",
		.help = "Perform a 32-bit read from DM register at reg_address, returning the value."
```

src/target/riscv/riscv.c
```diff
@@ -4285,6 +4354,20 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
+	},
+	{
+		.name = "dm_write",
+		.handler = riscv_dm_write,
+		.mode = COMMAND_ANY,
+		.usage = "address value",
+		.help = "Perform a 32-bit DM write of value at address."
```
```suggestion
		.name = "dm_write",
		.handler = riscv_dm_write,
		.mode = COMMAND_ANY,
		.usage = "reg_address value",
		.help = "Write a 32-bit value to the DM register at reg_address."
```

JanMatCodasip (dismissed):
Two last details, otherwise I consider the review complete from my side.

Thanks!

zqb-all:
> Two last details, otherwise I consider the review complete from my side.
> 
> Thanks!

Thank you for your detailed and helpful review

JanMatCodasip (dismissed):

src/target/riscv/riscv-013.c
timsifive: I disagree. Both of these should be LOG_TARGET_ERROR. These errors will be very rare, so there's no risk of clutter. When the error in the caller does get printed, the user will always want to know immediately which of the two cases were hit. If you only want a single error message (which is not important to me in this case), then the correct way is to remove the error message in the calling function and add more details to these ones here.

doc/openocd.texi
timsifive: I've rephrased this to talk about DMs and other devices, instead of about base addresses.
I find it easier to read. Do you agree?

```
The following commands allow for direct low-level access to the registers 
of the Debug Module (DM). They can be useful to access custom features in the DM.

@deffn {Command} {riscv dm_read} reg_address
Perform a 32-bit read from the register indicated by reg_address from the DM of the
current target.
@end deffn

@deffn {Command} {riscv dm_write} reg_address value
Write the 32-bit value to the register indicated by reg_address from the DM of the
current target.
@end deffn

The following commands allow for direct low-level access to the Debug Module
Interface (DMI). They can be useful to access any device that resides on the DMI.

@deffn {Command} {riscv dmi_read} address
Perform a 32-bit read from the given DMI address, returning the value.
@end deffn

@deffn {Command} {riscv dmi_write} address value
Perform a 32-bit write to the given DMI address.
@end deffn
```

zqb-all: I agree, it looks easier to read now

JanMatCodasip: Looks fine to me.

src/target/riscv/riscv-013.c
JanMatCodasip: I don't mind both of those being LOG_TARGET_ERROR. It is a detail. 

(Still, the message about reaching the end of the DM chain feels redundant: If the search failed, it implicitly means we went through the list and reached the end without finding the item. On the other hand, if it is the very rare case hitting of maximum number of DMs on the DMI bus, then that second error print is justified.)

zqb-all: Let's just use LOG_TARGET_ERROR.
And to make it slightly more useful, I add count of DMs

https://github.com/riscv/riscv-openocd/compare/5cc1a4307e7e7981371530dcaf9c8aecef0192aa..b98698ac7c40125a72de14f5c67329a19c4b92d1

src/target/riscv/riscv-013.c
```diff
@@ -686,10 +704,92 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+			break;
+		LOG_TARGET_DEBUG(target, "dm @ 0x%x --> nextdm=0x%x", current_dm, next_dm);
+		/* Check if it's last one in the chain. */
+		if (next_dm == 0) {
+			LOG_TARGET_ERROR(target, "Reach end of DM chain, %d DMs in total", count);
+			break;
+		}
+		/* Safety: Avoid looping forever in case of buggy nextdm values in the hardware. */
+		if (count++ > RISCV_MAX_DMS) {
+			LOG_TARGET_ERROR(target, "Only support %d DMs in one chain now, abort", RISCV_MAX_DMS);
```
JanMatCodasip: I have tried to improve the wording here, too:

```suggestion
			LOG_TARGET_ERROR(target, "Supporting no more than %d DMs on a DMI bus. Aborting.", RISCV_MAX_DMS);
```

src/target/riscv/riscv-013.c
```diff
@@ -686,10 +704,92 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+	while (1) {
+		uint32_t current_dm = next_dm;
+		if (current_dm == target->dbgbase)
+			return true;
+		if (dmi_read(target, &next_dm, DM_NEXTDM + current_dm) != ERROR_OK)
+			break;
+		LOG_TARGET_DEBUG(target, "dm @ 0x%x --> nextdm=0x%x", current_dm, next_dm);
+		/* Check if it's last one in the chain. */
+		if (next_dm == 0) {
+			LOG_TARGET_ERROR(target, "Reach end of DM chain, %d DMs in total", count);
```
JanMatCodasip: I have tried to improve the wording.

Also if `count` is changed to be `unsigned int`, there should be `%u` in the formatting string of the printf. Again, these things are often pointed out in reviews at openocd.org.

```suggestion
			LOG_TARGET_ERROR(target, "Reached the end of DM chain (detected %u DMs in total).", count);
```

src/target/riscv/riscv-013.c
```diff
@@ -686,10 +704,92 @@ static int dmi_write_exec(struct target *target, uint32_t address,
...
+	dm013_info_t *dm = get_dm(target);
+	if (!dm)
+		return ERROR_FAIL;
+	return dmi_write_exec(target, address + dm->base, value, ensure_success);
+}
+
+static bool check_dbgbase_exists(struct target *target)
+{
+	uint32_t next_dm = 0;
+	uint32_t count = 1;
```
JanMatCodasip: You can use `unsigned int` here. I'm suggesting it because in vanilla upstream (openocd.org), the reviewers tend to pay more attention to the types (and similar things).

```suggestion
	unsigned int count = 1;
```

src/target/riscv/riscv-013.c
JanMatCodasip: Sounds good :)

src/target/riscv/riscv-013.c
zqb-all: Thanks, this will help with synchronizing with upstream

JanMatCodasip (approved):

timsifive (approved):

PR #876, opened 2023-07-04
target/riscv: Fix the trigger writing sequence

MarekVCodasip:
According to section 5.6 in the RISC-V debug specification, the previous way to set triggers was incorrect, as was discussed as part of https://github.com/riscv/riscv-openocd/issues/870. This commit fixes the sequence to be in line with the specification as well as adds some comments to clarify for any future reader as to what is actually done.

Change-Id: Iffc5cc0f866a466a7aaa72a4c53ee95c9080ac9d


timsifive (approved):
Looks good. Thanks.

PR #878, opened 2023-07-05
target/riscv: cleanup trigger setup

en-sc:
* Add a warning when eq trigger is setup and it's behavior is different from other triggers.

* Make eq trigger's behavior consistent with other triggers in case of length == 1.

* Fix a bug in setting chained triggers (LT, GT case).

* Improve logging.

Change-Id: Id1ed0d11971b8ed875afbb979e6c8a8b51dd3818

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -655,9 +655,18 @@ struct trigger_request_info {
 	riscv_reg_t tdata1_ignore_mask;
 };
 
+static void log_trigger_request_info(struct trigger_request_info trig_info)
+{
+	LOG_DEBUG("tdata1=%" PRIx64 ", tdata2=%" PRIx64 ", tdata1_ignore_mask=%" PRIx64,
+			trig_info.tdata1, trig_info.tdata2, trig_info.tdata1_ignore_mask);
+};
+
 static int try_setup_single_match_trigger(struct target *target,
 		struct trigger *trigger, struct trigger_request_info trig_info)
 {
+	LOG_TARGET_DEBUG(target, "trying to set up a match trigger");
+	log_trigger_request_info(trig_info);
+
 	int trigger_type =
 		get_field(trig_info.tdata1, CSR_MCONTROL_TYPE(riscv_xlen(target)));
 	int ret = ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
@@ -683,6 +692,9 @@ static int try_setup_chained_match_triggers(struct target *target,
 		struct trigger *trigger, struct trigger_request_info t1,
 		struct trigger_request_info t2)
 {
+	LOG_TARGET_DEBUG(target, "trying to set up a chain of match triggers");
+	log_trigger_request_info(t1);
+	log_trigger_request_info(t2);
 	int trigger_type =
 		get_field(t1.tdata1, CSR_MCONTROL_TYPE(riscv_xlen(target)));
 	int ret = ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
@@ -717,8 +729,10 @@ static int try_setup_chained_match_triggers(struct target *target,
 struct match_triggers_tdata1_fields {
 	riscv_reg_t common;
 	struct {
+		/* Other values are available for this field,
+		 * but currently only `any` is needed.
+		 */
 		riscv_reg_t any;
-		riscv_reg_t s8bit;
 	} size;
 	struct {
 		riscv_reg_t enable;
@@ -752,10 +766,7 @@ static struct match_triggers_tdata1_fields fill_match_triggers_tdata1_fields_t2(
 		.size = {
 			.any =
 				field_value(CSR_MCONTROL_SIZELO, CSR_MCONTROL_SIZELO_ANY & 3) |
-				field_value(CSR_MCONTROL_SIZEHI, (CSR_MCONTROL_SIZELO_ANY >> 2) & 3),
-			.s8bit =
-				field_value(CSR_MCONTROL_SIZELO, CSR_MCONTROL_SIZELO_8BIT & 3) |
-				field_value(CSR_MCONTROL_SIZEHI, (CSR_MCONTROL_SIZELO_8BIT >> 2) & 3)
+				field_value(CSR_MCONTROL_SIZEHI, (CSR_MCONTROL_SIZELO_ANY >> 2) & 3)
 		},
 		.chain = {
 			.enable = field_value(CSR_MCONTROL_CHAIN, CSR_MCONTROL_CHAIN_ENABLED),
@@ -793,8 +804,7 @@ static struct match_triggers_tdata1_fields fill_match_triggers_tdata1_fields_t6(
 			field_value(CSR_MCONTROL6_LOAD, trigger->is_read) |
 			field_value(CSR_MCONTROL6_STORE, trigger->is_write),
 		.size = {
-			.any = field_value(CSR_MCONTROL6_SIZE, CSR_MCONTROL6_SIZE_ANY),
-			.s8bit = field_value(CSR_MCONTROL6_SIZE, CSR_MCONTROL6_SIZE_8BIT)
+			.any = field_value(CSR_MCONTROL6_SIZE, CSR_MCONTROL6_SIZE_ANY)
 		},
 		.chain = {
 			.enable = field_value(CSR_MCONTROL6_CHAIN, CSR_MCONTROL6_CHAIN_ENABLED),
@@ -852,13 +862,13 @@ static int maybe_add_trigger_t2_t6(struct target *target,
 		struct trigger_request_info lt_1 = {
 			.tdata1 = fields.common | fields.size.any | fields.chain.enable |
 				fields.match.lt,
-			.tdata2 = trigger->address,
+			.tdata2 = trigger->address + trigger->length,
 			.tdata1_ignore_mask = fields.tdata1_ignore_mask
 		};
 		struct trigger_request_info ge_2 = {
 			.tdata1 = fields.common | fields.size.any | fields.chain.disable |
 				fields.match.ge,
-			.tdata2 = trigger->address + trigger->length,
+			.tdata2 = trigger->address,
 			.tdata1_ignore_mask = fields.tdata1_ignore_mask
 		};
 		ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv.c
```diff
@@ -718,7 +730,6 @@ struct match_triggers_tdata1_fields {
 	riscv_reg_t common;
 	struct {
 		riscv_reg_t any;
-		riscv_reg_t s8bit;
 	} size;
```
timsifive: Worth a quick comment explaining why there's a struct with just a single variable.

src/target/riscv/riscv.c
```diff
@@ -852,24 +859,26 @@ static int maybe_add_trigger_t2_t6(struct target *target,
 		struct trigger_request_info lt_1 = {
 			.tdata1 = fields.common | fields.size.any | fields.chain.enable |
 				fields.match.lt,
-			.tdata2 = trigger->address,
+			.tdata2 = trigger->address + trigger->length,
```
timsifive, line 865: Could you add a test case for this code in riscv-tests/debug? I believe spike supports these trigger types already.

src/target/riscv/riscv.c
```diff
@@ -852,24 +859,26 @@ static int maybe_add_trigger_t2_t6(struct target *target,
...
 			.tdata1_ignore_mask = fields.tdata1_ignore_mask
 		};
 		ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
 		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
 			return ret;
 	}
 	LOG_TARGET_DEBUG(target, "trying to setup equality match trigger");
+	if (trigger->length > 1)
+		LOG_TARGET_WARNING(target, "trigger will only match accesses at address 0x%"
+				TARGET_PRIxADDR, trigger->address);
```
timsifive: Is that true? I think the spec says that a size any address trigger at 0x1001 will (should?) match on a 32-bit access at 0x1000.

select=0 says:
> There is at least one compare value and it contains the lowest virtual address of the access. In addition, it is recommended that there are additional compare values for the other accessed virtual addresses match. (E.g. on a 32-bit read from 0x4000, the lowest address is 0x4000 and the other addresses are 0x4001, 0x4002, and 0x4003.)

timsifive (commented):
Looks like you removed matching on 8-bit accesses for 8-bit triggers. Can you share the reasoning behind that? (It makes sense to me, because neither OpenOCD nor gdb support specifying access size on a watchpoint.)

src/target/riscv/riscv.c
en-sc: Thanks for the catch! I didn't consider this difference in 1.0 and 0.13.2 versions of the spec.

I would like to ask, is there a way for a debugger to test which of those additional compare values are supported for a particular access? As far as I can see, the only way to do so is to write a program to some scratch memory which tries to trigger a watchpoint. However, this is quite complex and not available on some targets.

src/target/riscv/riscv.c
en-sc: To test this, it is necessary to specify the exact supported value of `mcontrol.match` for each trigger. In particular, there should be no consecutive pair of triggers, such as the first one supports `ge`, and the second - `lt`. As far as I know, such a level of trigger description is not supported in `spike`.
I can add an issue to document this.
I can also add a test for `ge` `lt` trigger pair to `riscv-tests`, though, IMHO, it is only somewhat related to this particular commit.

src/target/riscv/riscv.c

en-sc:
> Looks like you removed matching on 8-bit accesses for 8-bit triggers. Can you share the reasoning behind that? (It makes sense to me, because neither OpenOCD nor gdb support specifying access size on a watchpoint.)

The reasoning for matching accesses with any size for 8-bit equality (`mcontrol.match = 0`) triggers is the following:

In other triggers (`mcontrol.match = 1 (napot)`, or a chain of `mcontrol.match = 2, 3 (lt, ge)`) access with an address in the inclusive range `[trigger->address, trigger->address + trigger->length]` matches, even if `access_address + access_size > trigger->address + trigger->length`.

If `s8bit` is used (`trigger->length == 1`), this will not be the case for the resulting trigger.
It will not be triggered on access `access_address = trigger->address, access_size = 2`.

src/target/riscv/riscv.c
timsifive: There is not. :-( Trying to discover this with a test program feels like trying too hard, and it's OK to leave this up for the user to discover/read in their documentation.

We keep thinking that the Unified Discovery Group will address this. That group is supposed to start up again later this month.

src/target/riscv/riscv.c
timsifive: On spike at least, all triggers are the same. Every trigger supports all the features.

Agreed that the test is separate, but it would have caught the bug you're fixing so it seems good to have coverage, given how simple the test (presumably) will be.

src/target/riscv/riscv.c
en-sc: Addressed (Changed the message).

src/target/riscv/riscv.c
```diff
@@ -852,24 +862,28 @@ static int maybe_add_trigger_t2_t6(struct target *target,
...
 		ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
 		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
 			return ret;
 	}
 	LOG_TARGET_DEBUG(target, "trying to setup equality match trigger");
+	if (trigger->length > 1)
+		LOG_TARGET_WARNING(target, "Trigger will match accesses at address 0x%"
+				TARGET_PRIxADDR ", but may not match accesses at addresses in the inclusive range from 0x%"
+				TARGET_PRIxADDR " to 0x%" TARGET_PRIxADDR ".", trigger->address,
+				trigger->address + 1, trigger->address + trigger->length - 1);
```
timsifive: I originally misread this check. I thought we were talking about triggers at 0x101 not firing when an access is made to 0x100. But this is the opposite, which is way more common.

I don't think we can print out a warning every time a trigger with length>1 is set. Users will see this message every time they resume a target when such breakpoints are set, because then gdb clears all hardware berakpoints, steps, and sets them all again.

I'm inclined to leave the warning out altogether. If you do think it's important, then please ensure it's displayed only once per session.

en-sc: I would propose to leave the warning on display each time such trigger is setup.
The reasoning is this case is rare. Most platforms also support other types of triggers (not only equality), so this message won't get displayed.
Since the behavior is fundamentally different for other triggers, I think it is justified to print this warning, even if it is quite obnoxious.
What do you think?

timsifive: It may be rare among platforms, but for platforms where this case is hit, it will be hit all the time. I would prefer the warning pop up once, saying something like "on this platform ..." instead of every time the trigger is set (which will be often because gdb clears/sets them for every resume).

en-sc:
Please, disregard the mention in `riscv-tests/debug` PR  -- it was a typo.

src/target/riscv/riscv.c

timsifive (dismissed):

src/target/riscv/riscv.c
```diff
@@ -852,24 +862,35 @@ static int maybe_add_trigger_t2_t6(struct target *target,
...
 		ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
 		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
 			return ret;
 	}
 	LOG_TARGET_DEBUG(target, "trying to setup equality match trigger");
+	if (trigger->length > 1) {
+		/* This message is displayed only once per session to avoid
+		 * overwhelming the user with such messages on resume.
+		 */
+		static bool warning_was_displayed;
```
JanMatCodasip: Please, do not use static local variables - because that's in fact almost a global variable, just hidden inside a function :-)

My recommendation would be to: 
- Add a bool flag to riscv013_info_t
- Print a more generic warning message (only once, based on the flag). Just say that it was not possible to add a trigger for range of addresses, and some triggers may match only on the beginning of the respective range. No need to mention concrete addresses in this generic warning. That's because may be other triggers facing the same limitation, so the user just needs to be informed that such thing has happened, and the single concrete address is not that relevant since there can be arbitrary number of such cases with different addresses.
- Turn the current "warning" message to "debug" message and print it every time. It can be very useful for troubleshooting.

JanMatCodasip (changes requested):

src/target/riscv/riscv.c

src/target/riscv/riscv.c
```diff
@@ -852,24 +862,38 @@ static int maybe_add_trigger_t2_t6(struct target *target,
...
+				TARGET_PRIxADDR " to 0x%" TARGET_PRIxADDR ".", trigger->address,
+				trigger->address + 1, trigger->address + trigger->length - 1);
+		RISCV_INFO(info);
+		if (!info->inconsistent_trigger_warning_was_displayed)
+			/* This message is displayed only once per target to avoid
+			 * overwhelming the user with such messages on resume.
+			 */
+			LOG_TARGET_WARNING(target,
+					"This and some other triggers will match accesses at the specifyed address, "
+					"but may not match accesses into the range [address, address + length - 1]");
```
JanMatCodasip: I have tried to re-word it to make it even more clear for the user. Would you be OK with this wording?

```suggestion
					"Could not set a trigger that will match a whole address range. "
					"As a fallback, this trigger (and maybe others) will only match "
					"against the first address of the range.");
```

src/target/riscv/riscv.c
```diff
@@ -852,24 +862,38 @@ static int maybe_add_trigger_t2_t6(struct target *target,
...
+		info->inconsistent_trigger_warning_was_displayed = true;
```
JanMatCodasip: If you'd like, the name of the variable can e.g. be shortened to this (or something similar):

```suggestion
		info->range_trigger_fallback_encountered = true;
```

src/target/riscv/riscv.c
```diff
@@ -852,24 +862,38 @@ static int maybe_add_trigger_t2_t6(struct target *target,
...
-			.tdata2 = trigger->address + trigger->length,
+			.tdata2 = trigger->address,
 			.tdata1_ignore_mask = fields.tdata1_ignore_mask
 		};
 		ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
 		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
 			return ret;
 	}
 	LOG_TARGET_DEBUG(target, "trying to setup equality match trigger");
+	if (trigger->length > 1) {
```
JanMatCodasip, line 889: Please, consider displaying the debug & warn messages only if the setting of the exact-match trigger succeeds.

If `try_setup_single_match_trigger()` fails, I believe it does not make sense to display this message. In such case, there will be already a different error message printed.

```
LOG_TARGET_DEBUG(target, "trying to setup equality match trigger");
struct trigger_request_info eq = { ... };
retval = try_setup_single_match_trigger(target, trigger, eq);

if (retval == ERROR_OK && trigger->length > 1) {
	/* The exact-match trigger was successfully set up but the original 
	request was for a range-match trigger, which the target could not 
	accommodate. Warn the user about this limitation of the target. */
	LOG_TARGET_DEBUG(...);
	if (!info->range_trigger_fallback_encountered) {
		/* ... */
		LOG_TARGET_WARNING(...);
	}
	info->range_trigger_fallback_encountered = true;
}

return retval;
```

JanMatCodasip (changes requested):
Thank you for the changes so far. I have one very last set of comments, otherwise this looks fine.

src/target/riscv/riscv.c
en-sc: Thanks for the suggestion! This is much better! Addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: Addressed.

JanMatCodasip (approved):

src/target/riscv/riscv.c
```diff
@@ -867,13 +877,32 @@ static int maybe_add_trigger_t2_t6(struct target *target,
...
-			(trigger->length == 1 ? fields.size.s8bit : fields.size.any) |
-			fields.chain.disable | fields.match.eq,
+		.tdata1 = fields.common | fields.size.any | fields.chain.disable |
+			fields.match.eq,
 		.tdata2 = trigger->address,
 		.tdata1_ignore_mask = fields.tdata1_ignore_mask
 	};
-	return try_setup_single_match_trigger(target, trigger, eq);
+	ret = try_setup_single_match_trigger(target, trigger, eq);
+	if (ret != ERROR_OK)
```
erhankur, line 886: @en-sc don't we need to handle ERROR_TARGET_RESOURCE_NOT_AVAILABLE return code here?

en-sc: Thanks for the catch! Addressed in #899

PR #879, opened 2023-07-06
target/breakpoints: Clear software breakpoints from available targets

timsifive:
If a target where a software breakpoint was set is not currently available, but there are other targets in the same SMP group that are available, then we can use those to remove the software breakpoint.

### DIFF
--- a/src/target/breakpoints.c
+++ b/src/target/breakpoints.c
@@ -220,6 +220,8 @@ int breakpoint_add(struct target *target,
 
 		return ERROR_OK;
 	} else {
+		/* For software breakpoints on SMP targets, only set them on a
+		 * single target. We assume that SMP targets share memory. */
 		return breakpoint_add_internal(target, address, length, type);
 	}
 }
@@ -270,11 +272,17 @@ int hybrid_breakpoint_add(struct target *target,
 		return hybrid_breakpoint_add_internal(target, address, asid, length, type);
 }
 
-/* free up a breakpoint */
-static void breakpoint_free(struct target *target, struct breakpoint *breakpoint_to_remove)
+/* Free the data structures we use to track a breakpoint on data_target.
+ * Remove the actual breakpoint from breakpoint_target.
+ * This separation is useful when a software breakpoint is tracked on a target
+ * that is currently unavailable, but the breakpoint also affects a target that
+ * is available.
+ */
+static void breakpoint_free(struct target *data_target, struct target *breakpoint_target,
+		struct breakpoint *breakpoint_to_remove)
 {
-	struct breakpoint *breakpoint = target->breakpoints;
-	struct breakpoint **breakpoint_p = &target->breakpoints;
+	struct breakpoint *breakpoint = data_target->breakpoints;
+	struct breakpoint **breakpoint_p = &data_target->breakpoints;
 	int retval;
 
 	while (breakpoint) {
@@ -287,33 +295,12 @@ static void breakpoint_free(struct target *target, struct breakpoint *breakpoint
 	if (!breakpoint)
 		return;
 
-	retval = target_remove_breakpoint(target, breakpoint);
+	retval = target_remove_breakpoint(breakpoint_target, breakpoint);
 
 	LOG_DEBUG("free BPID: %" PRIu32 " --> %d", breakpoint->unique_id, retval);
 	(*breakpoint_p) = breakpoint->next;
 	free(breakpoint->orig_instr);
 	free(breakpoint);
-}
-
-static int breakpoint_remove_internal(struct target *target, target_addr_t address)
-{
-	struct breakpoint *breakpoint = target->breakpoints;
-
-	while (breakpoint) {
-		if ((breakpoint->address == address) ||
-		    (breakpoint->address == 0 && breakpoint->asid == address))
-			break;
-		breakpoint = breakpoint->next;
-	}
-
-	if (breakpoint) {
-		breakpoint_free(target, breakpoint);
-		return 1;
-	} else {
-		if (!target->smp)
-			LOG_ERROR("no breakpoint at address " TARGET_ADDR_FMT " found", address);
-		return 0;
-	}
 }
 
 static void breakpoint_remove_all_internal(struct target *target)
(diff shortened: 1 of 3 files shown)


src/target/breakpoints.c
```diff
@@ -270,11 +272,17 @@ int hybrid_breakpoint_add(struct target *target,
 		return hybrid_breakpoint_add_internal(target, address, asid, length, type);
 }
 
-/* free up a breakpoint */
-static void breakpoint_free(struct target *target, struct breakpoint *breakpoint_to_remove)
+/* Free the data structures we use to track a breakpoint on data_target.
+ * Remove the actual breakpoint from breakpoint_target.
+ * This is separation is useful when a software breakpoint is tracked on a
```
JanMatCodasip: Just a typo:

```suggestion
 * This separation is useful when a software breakpoint is tracked on a
```

src/target/breakpoints.c
```diff
@@ -287,60 +295,89 @@ static void breakpoint_free(struct target *target, struct breakpoint *breakpoint
...
+
+		if (!found) {
 			LOG_ERROR("no breakpoint at address " TARGET_ADDR_FMT " found", address);
+			return;
+		}
+
+		/* TODO: What if the target state changed since the last time we
+		 * polled? */
+		struct target *remove_target = halted_target ? halted_target : available_target;
+		if (software_breakpoint && remove_target) {
```
JanMatCodasip: In case `software_breakpoint != NULL && remove_target == NULL`, should a warning be printed that we have no way of removing the breakpoint?

src/target/breakpoints.c
```diff
@@ -287,60 +295,89 @@ static void breakpoint_free(struct target *target, struct breakpoint *breakpoint
...
+			found++;
+
+			if (breakpoint->type == BKPT_SOFT) {
+				/* Software breakpoints are set on only one of
+				 * the SMP targets.  We can remove them through
+				 * any of the SMP targets. */
+				assert(!software_breakpoint_target);
+				software_breakpoint_target = curr;
+				software_breakpoint = breakpoint;
+				continue;
```
JanMatCodasip: At the first glance, this `continue` seemed suspicious and I thought we should `break` instead. Then I realized there are actually multiple independent searches going on in this single loop, so we have to continue.

Also for `BKPT_HARD`, all those extra operations and variables like available_target & halted_target are not meaningful at all, which may not be immediately apparent to the reader.

I'd recommend to re-structure the code to make it more clear. Quick proposal - for your consideration:

```c
void breakpoint_remove(...)
{
    /* Handle the easy case first, then return. 
       This is to keep the nesting and indents low, therefore keeping the function more readable.  */
    if (!target->smp) {
        /* ... */
        return;
    }

    /* Handle the SMP case */
    if (breakpoint->type == BKPT_SOFT) {
        /* loop 1: find available_target, halted_target */
        foreach_smp_target(head, target->smp_targets) {
            /* ... */
        }
        /* loop 2: find the target the breakpoint is on, remove it, break the loop */
        foreach_smp_target(head, target->smp_targets) {
            /* ... */
        }        
    }
    else {
        /* Not a SW breakpoint = simpler handling. */
        /* No need to deal with unavailable targets, etc. */
        /* Just go through each target and call breakpoint_find() + breakpoint_free(). */
        /* ... */
    }
}

```

src/target/breakpoints.c
```diff
@@ -287,60 +295,89 @@ static void breakpoint_free(struct target *target, struct breakpoint *breakpoint
...
 
 void breakpoint_remove(struct target *target, target_addr_t address)
 {
 	if (target->smp) {
-		unsigned int num_breakpoints = 0;
+		unsigned int found = 0;
 		struct target_list *head;
+		/* Target where we found a software breakpoint. */
+		struct target *software_breakpoint_target = NULL;
+		struct breakpoint *software_breakpoint = NULL;
```
JanMatCodasip: If you like, these identifiers can be shortened without making them unclear:

```suggestion
		struct target *sw_breakpoint_target = NULL;
		struct breakpoint *sw_breakpoint = NULL;
```

src/target/breakpoints.c
```diff
@@ -287,60 +295,89 @@ static void breakpoint_free(struct target *target, struct breakpoint *breakpoint
...
-		if (!num_breakpoints)
+
+		if (!found) {
 			LOG_ERROR("no breakpoint at address " TARGET_ADDR_FMT " found", address);
+			return;
+		}
+
+		/* TODO: What if the target state changed since the last time we
+		 * polled? */
+		struct target *remove_target = halted_target ? halted_target : available_target;
```
JanMatCodasip: For robustness, does it make sense to to use `sw_breakpoint_target` for the removal as the first choice, provided it is in a favorable state?

Something like (pseudocode):

```
if sw_breakpoint_target && (sw_breakpoint_target is available and halted):
    remove_target = sw_breakpoint_target
else if halted_target:
    remove_target = halted_target
else:
    remove_target = available_target
```

src/target/breakpoints.c
```diff
@@ -287,60 +295,89 @@ static void breakpoint_free(struct target *target, struct breakpoint *breakpoint
...
+		if (software_breakpoint && remove_target) {
+			LOG_DEBUG("removing software breakpoint found on %s using %s (address=" TARGET_ADDR_FMT ")",
+					target_name(software_breakpoint_target),
+					target_name(remove_target),
+					address);
+			/* Remove the software breakpoint through
+			 * remove_target, but update the breakpoints structure
+			 * of software_breakpoint_target. */
+			/* TODO: What if the target has become unavailable since
+			 * we last polled? */
```
JanMatCodasip: If we are working with a system where the the hart can change its state on us suddenly (become unavailable anytime), then I am afraid there is not (in principle) anything more that can be done about it. There will always be some window of time between the state check and the action, during which things can change.

JanMatCodasip (commented):
Overall, this looks all right to me. I have posted few suggestions.

src/target/breakpoints.c
timsifive: Done.

src/target/breakpoints.c
timsifive: Good idea. Done.

src/target/breakpoints.c
timsifive: I like the !target->smp check, but I don't think the other refactor works. We don't have a breakpoint object until we call breakpoint_find(). Each target might have different breakpoints set on the same address. I don't think you're supposed to be able to get into that state, but it seems better to be a bit defensive about that. I prefer to keep the single loop, which will delete any hardware breakpoints it finds, and also keeps track of available target, halted target, and software breakpoint target.

src/target/breakpoints.c
timsifive: I don't think shorter is better in this case.

src/target/breakpoints.c
timsifive: Good point.

src/target/breakpoints.c
timsifive: That makes sense. At a minimum it will most closely preserve the existing behavior.

src/target/breakpoints.c
JanMatCodasip: Ok, let's leave it as is.

src/target/breakpoints.c
JanMatCodasip: >  I don't think the other refactor works ... We don't have a breakpoint object until we call breakpoint_find() ...

Ah, I missed that we don't know the breakpoint type upfront (only the address). In that case, let's leave it as is.

Another thing is that the current algorithm assumes that the SW breakpoint will be found on a single target only. To stay defensive, would it make sense to warn if that is not the case (instead of the assert)?

timsifive: Yes. That makes sense. Especially with the `smp` commands that let you modify SMP groups at run time, you could end up in a situation where this assert is fired due to user configuration.

GregSavin (approved):
Nice, good catch for that corner case of sw breakpoint left in place by a hart that becomes unavailable

JanMatCodasip (approved):

src/target/breakpoints.c
```diff
@@ -385,7 +438,8 @@ struct breakpoint *breakpoint_find(struct target *target, target_addr_t address)
 	struct breakpoint *breakpoint = target->breakpoints;
 
 	while (breakpoint) {
-		if (breakpoint->address == address)
+		if (breakpoint->address == address ||
+				(breakpoint->address == 0 && breakpoint->asid == address))
```
aap-sc, line 442: @timsifive I have a few questions regarding this commit.

1. My understanding is that since this is a commit to the platform-independent part, we may have difficulties merging this upstream. Any chance we already have some MR to the upstream openocd? I searched on Gerrit, but could not find anything.

2. The reason I'm asking is that this condition `(breakpoint->address  0 && breakpoint->asid  address)` seems strange to me (since I think that `address` and `asid` should be independent entities). But it looks like this may be a defect of the existing OpenOCD interface, rather than this commit - if it is then it should be fixed upstream. So I was wondering if it makes sense to revise existing interface/data structures for setting breakpoints and if it makes sense to do it now, or if it is better to wait till these changes are upstreamed.

timsifive: There is no MR upstream. This change only makes sense when targets are unavailable, and that code is not present in upstream either.

Generally we make target-independent changes in this branch if they resolve issues that we see with RISC-V targets. (The change also needs to be reasonable so that some day it might be accepted upstream.)

I agree that comparing address to asid is odd, but that's also true upstream in breakpoint_remove_internal(). If this is causing problems on RISC-V, then feel to fix it here. If it's just a hypothetical bug that's not affecting anyone right now, I prefer for it to be fixed upstream.

aap-sc: Got it, thanks for clarification!

> If it's just a hypothetical bug that's not affecting anyone right now, I prefer for it to be fixed upstream.

Currently, this is mostly a hypothetical bug... It's just I was scanning recent patches, and I was a little confused by these lines.

PR #882, opened 2023-07-12
target/riscv: Add target name logging to most log prints.

MarekVCodasip:
This patch adds target name (LOG_TARGET_XXXX) to logging calls where it makes sense. This is especially useful when debugging multiple targets at once, such as multicore systems.

Change-Id: Ia9861f3fa0e6e5908b683c2a8280659c3c264395

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -88,7 +88,7 @@ bool riscv_batch_full(struct riscv_batch *batch)
 int riscv_batch_run(struct riscv_batch *batch)
 {
 	if (batch->used_scans == 0) {
-		LOG_DEBUG("Ignoring empty batch.");
+		LOG_TARGET_DEBUG(batch->target, "Ignoring empty batch.");
 		return ERROR_OK;
 	}
 
@@ -107,7 +107,7 @@ int riscv_batch_run(struct riscv_batch *batch)
 	keep_alive();
 
 	if (jtag_execute_queue() != ERROR_OK) {
-		LOG_ERROR("Unable to execute JTAG queue");
+		LOG_TARGET_ERROR(batch->target, "Unable to execute JTAG queue");
 		return ERROR_FAIL;
 	}
 
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -247,7 +247,7 @@ static dm013_info_t *get_dm(struct target *target)
 	}
 
 	if (!dm) {
-		LOG_DEBUG("[%d] Allocating new DM", target->coreid);
+		LOG_TARGET_DEBUG(target, "Coreid [%d] Allocating new DM", target->coreid);
 		dm = calloc(1, sizeof(dm013_info_t));
 		if (!dm)
 			return NULL;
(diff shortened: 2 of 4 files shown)


timsifive (commented):
I'm not aware of any multi-hart riscv-011 targets. If they don't exist, let's not touch this code.

Otherwise, this touches a lot of code which will make conflict resolution more painful. But it also makes the code better, so overall I think this is good?

JanMatCodasip:
> Tim: Otherwise, this touches a lot of code which will make conflict resolution more painful. But it also makes the code better, so overall I think this is good?

I thought about the conflicts, but I believe that any current feature branches should be easy to rebase since these are trivial line-by-line changes.

MarekVCodasip:
> I'm not aware of any multi-hart riscv-011 targets. If they don't exist, let's not touch this code.

I have removed those changes then.

Also, it seems one of the tests does not pass thanks to the changes. You want me to submit a pull request to riscv-tests repo right away or after merging?

timsifive:
The goal is that the latest spike, riscv-tests, and openocd combination always pass. In this case that means fix riscv-tests so it works both with current openocd and with this PR.

en-sc:
> Also, it seems one of the tests does not pass thanks to the changes. You want me to submit a pull request to riscv-tests repo right away or after merging?

The test will pass after https://github.com/riscv-software-src/riscv-tests/pull/491 is merged.

timsifive:
> The test will pass after riscv-software-src/riscv-tests#491 is merged.

I don't understand. That change will no longer ignore lines starting with "Batch memory". Can you elaborate on what causes the test failure and how riscv-software-src/riscv-tests#491 fixes it?

JanMatCodasip:
It looks like we have here at least two parallel merge requests that have impact on the tests: 
- this merge request 
- and https://github.com/riscv/riscv-openocd/pull/871

My recommendation is to merge all pull requests that are currently submitted and close to ready (approved), including https://github.com/riscv/riscv-openocd/pull/871. Then this merge request with message changes can be rebased - and we will see if there is still something to adjust in the tests due to the changed messages or not.

en-sc:
> > The test will pass after [riscv-software-src/riscv-tests#491](https://github.com/riscv-software-src/riscv-tests/pull/491) is merged.
> 
> I don't understand. That change will no longer ignore lines starting with "Batch memory". Can you elaborate on what causes the test failure and how [riscv-software-src/riscv-tests#491](https://github.com/riscv-software-src/riscv-tests/pull/491) fixes it?

After this change, the lines start with `[riscv.cpu] `, so the check `line.startswith("Batch memory")` dose not filter these warnings and the following attempt to convert them to integer values causes an exception.

My previous comment was poorly worded. The idea was that according to the strategy described [here](https://github.com/riscv/riscv-openocd/pull/871#issuecomment-1623962270), https://github.com/riscv-software-src/riscv-tests/pull/491 will be merged after #871, so after rebase (and resolving merge conflicts between this PR and #871), the test will pass.

The merge of  https://github.com/riscv-software-src/riscv-tests/pull/491 is not necessary for this PR to pass testing, it would suffice to rebase on top of #871.

timsifive:
> The merge of riscv-software-src/riscv-tests#491 is not necessary for this PR to pass testing, it would suffice to rebase on top of #871.

OK. I just merged #871 so please go ahead and rebase this on top of `riscv` then to make the tests pass.

timsifive (dismissed):

JanMatCodasip (approved):

PR #883, opened 2023-07-12
target/riscv: update some macro

zqb-all:
1. update RISCV_MAX_HARTS to 2^20 according to SPEC
2. remove RISCV_MAX_REGISTERS, it's not used anywhere anymore

Change-Id: Iadf0fa1ba3bbe5b9420b8430883e140db87f4f9e

### DIFF
--- a/src/target/riscv/riscv.h
+++ b/src/target/riscv/riscv.h
@@ -12,12 +12,11 @@ struct riscv_program;
 #include "target/register.h"
 #include "target/semihosting_common.h"
 #include <helper/command.h>
+#include <helper/bits.h>
 
 #define RISCV_COMMON_MAGIC	0x52495356U
 
-/* The register cache is statically allocated. */
-#define RISCV_MAX_HARTS 1024
-#define RISCV_MAX_REGISTERS 5000
+#define RISCV_MAX_HARTS  ((int)BIT(20))
 #define RISCV_MAX_TRIGGERS 32
 #define RISCV_MAX_HWBPS 16
 
@@ -30,7 +29,7 @@ struct riscv_program;
 #define RISCV_HGATP_PPN(xlen)  ((xlen) == 32 ? HGATP32_PPN : HGATP64_PPN)
 #define RISCV_PGSHIFT 12
 
-# define PG_MAX_LEVEL 4
+#define PG_MAX_LEVEL 4
 
 #define RISCV_NUM_MEM_ACCESS_METHODS  3
 
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10596,11 +10596,9 @@ addreg rtest 0x1234 org.gnu.gdb.or1k.group0 system
 @section RISC-V Architecture
 
 @uref{http://riscv.org/, RISC-V} is a free and open ISA. OpenOCD supports JTAG
-debug of RV32 and RV64 cores in heterogeneous multicore systems of up to 32
-harts. (It's possible to increase this limit to 1024 by changing
-RISCV_MAX_HARTS in riscv.h.) OpenOCD primarily supports 0.13 of the RISC-V
-Debug Specification, but there is also support for legacy targets that
-implement version 0.11.
+debug of RV32 and RV64 cores in heterogeneous multicore systems of up to 2^20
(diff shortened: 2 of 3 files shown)


timsifive (commented):
Have you checked that we correctly modify hartselhi where needed, and that the code to select multiple harts will also work with >1024 harts? (I mean by reading the code, making a test case for this would be hard.

zqb-all:
> Have you checked that we correctly modify hartselhi where needed

All the places that use hartsello now have hartselhi, so I think there's no problem

> and that the code to select multiple harts will also work with >1024 harts? 


https://github.com/riscv/riscv-openocd/blob/21d21408aa36ccb458ccfaad871f7dd1bb1f731a/src/target/riscv/riscv-013.c#L4450-L4453
The length of this array is variable and I think it need space from stack, if hart count is huge, maybe we should change it to get space from calloc.

src/target/riscv/riscv.h
```diff
@@ -12,12 +12,12 @@ struct riscv_program;
...
 #include "target/semihosting_common.h"
 #include <helper/command.h>
+#include <helper/bits.h>
 
 #define RISCV_COMMON_MAGIC	0x52495356U
 
 /* The register cache is statically allocated. */
-#define RISCV_MAX_HARTS 1024
-#define RISCV_MAX_REGISTERS 5000
+#define RISCV_MAX_HARTS  (int)BIT(20)
```
JanMatCodasip: For additional safety when used in expressions, please wrap the whole thing with brackets:

```suggestion
#define RISCV_MAX_HARTS  ((int)BIT(20))
```

JanMatCodasip:
> zqb-all: The length of this array is variable and I think it need space from stack, if hart count is huge, maybe we should change it to get space from calloc.

Agreed, I think using dynamically allocated memory (malloc, calloc) makes sense in that case.

JanMatCodasip (commented):
Overall, this looks all right to me.

As a last suggestion, I'd recommend to put `assert(dm->hart_count > 0)` to places where `hart_count` is used and expected to be already filled in.

That's because `hart_count` is set to `-1` initially before the real count is known, but we do not wish to see the negative value anytime after the harts have been examined.

src/target/riscv/riscv.h
zqb-all: done

zqb-all:
> That's because `hart_count` is set to `-1` initially before the real count is known, but we do not wish to see the negative value anytime after the harts have been examined.

There is a check in examine

https://github.com/riscv/riscv-openocd/blob/21d21408aa36ccb458ccfaad871f7dd1bb1f731a/src/target/riscv/riscv-013.c#L1840-L1843

src/target/riscv/riscv.h
```diff
@@ -12,29 +12,29 @@ struct riscv_program;
 #include "target/register.h"
 #include "target/semihosting_common.h"
 #include <helper/command.h>
+#include <helper/bits.h>
 
-#define RISCV_COMMON_MAGIC	0x52495356U
+#define RISCV_COMMON_MAGIC	(0x52495356U)
```
timsifive: Simple constants don't need parens. It's only important if your #define contains more than one token, such as in RISCV_SATP_MODE below, and the new version of RISCV_MAX_HARTS.

timsifive (changes requested):
Looks good overall.

src/target/riscv/riscv.h
zqb-all: done

src/target/riscv/riscv.h
```diff
@@ -12,12 +12,12 @@ struct riscv_program;
 #include "target/register.h"
 #include "target/semihosting_common.h"
 #include <helper/command.h>
+#include <helper/bits.h>
 
 #define RISCV_COMMON_MAGIC	0x52495356U
 
 /* The register cache is statically allocated. */
```
timsifive: Can you remove this comment also, along with RISCV_MAX_REGISTERS?

timsifive (changes requested):

src/target/riscv/riscv.h
zqb-all: done

timsifive (approved):

PR #885, opened 2023-07-13
Fix Checkpatch and Sparse tool warnings

erhankur:
We have fixed several warnings in our fork. https://github.com/espressif/openocd-esp32

I would like to fix them also here.  This will be helpful when/if you want to upstream your commits. 

Additionally cherry-picked a commit from upstream to fix build errors on the Clang compiler.

Remaining Sparse warnings:
https://github.com/erhankur/riscv-openocd/actions/runs/5548680099/jobs/10131935209?pr=3#step:5:3064

Most of them can be fixed by adding ULL or LL to 64-bit macros inside encoding.h
I can fix them after getting your opinion.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -28,7 +28,7 @@
 
 #define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
 #define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
-#define field_value(mask, val) set_field((riscv_reg_t) 0, mask, val)
+#define field_value(mask, val) set_field((riscv_reg_t)0, mask, val)
 
 /*** JTAG registers. ***/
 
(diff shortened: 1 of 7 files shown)


JanMatCodasip (dismissed):
Thanks!

timsifive (dismissed):
I'm not a fan of all this non-functional churn, which will only generate more changes between riscv-openocd and upstream. But I guess it needs to be done, and I assume every change here is because a tool is enforcing that style now.

If it possible to make a github workflow that automatically runs Sparse, similar to how we currently run checkpatch?

erhankur:
> I'm not a fan of all this non-functional churn, which will only generate more changes between riscv-openocd and upstream. But I guess it needs to be done, and I assume every change here is because a tool is enforcing that style now.
> 
> If it possible to make a github workflow that automatically runs Sparse, similar to how we currently run checkpatch?

Yes possible. The easy way, just need a reference for the current head.  It might be a counter or a log file that we can compare. 
Another way, not easy, might be comparing git diff lines with the created log file. 
We will make it for our fork. Also for the clang scan-build tool. Then I can come up here with another PR

timsifive:
Looks like this PR now has a conflict.

erhankur:
> Looks like this PR now has a conflict.

Rebased and fixed.

src/target/riscv/riscv.c
```diff
@@ -3969,7 +3969,7 @@ COMMAND_HANDLER(handle_memory_sample_command)
 	uint32_t bucket;
 	COMMAND_PARSE_NUMBER(u32, CMD_ARGV[0], bucket);
 	if (bucket > ARRAY_SIZE(r->sample_config.bucket)) {
-		LOG_ERROR("Max bucket number is %d.", (unsigned) ARRAY_SIZE(r->sample_config.bucket));
+		LOG_ERROR("Max bucket number is %d.", (unsigned int)ARRAY_SIZE(r->sample_config.bucket));
```
en-sc: I would suggest:
```suggestion
		LOG_ERROR("Max bucket number is %zu.", ARRAY_SIZE(r->sample_config.bucket));
```

erhankur: This PR is not targeted to correct format specifiers or remove unnecessary castings but since I touched that line, ok let's fix it. Thanks for the suggestion.

erhankur:
@timsifive can we merge this?

timsifive (approved):

PR #891, opened 2023-07-18
Run spike tests on `riscv` branch.

timsifive:
The main reason is to populate the cache, so that pull requests can benefit from a spike cache even the very first time. (Caches populated in one pull request are inaccessible to another one.)

Change-Id: If894f2ccfaadc740bd52e34be3024153626b9fbd


JanMatCodasip (approved):
This actually has two benefits:

- 1) everything in `riscv` will be tested (sanity-checked once more after each merge)
- 2) the dependencies get cached

PR #892, opened 2023-07-19
target/riscv: define register printers

en-sc:
This commit implements facilities to decode register fields for each register described in debug_defines.h.

Namely:
* define `<REG_NAME>_FORMAT_STR`
* macro `<REG_NAME>_FORMAT_ARGS(value)`

The use case is demonstarted by changing debug message in `execute_abstract_command()`

Change-Id: I63733d8d36385d89ca15de1a43139134bc488c4f


en-sc:
This commit fails `checkpatch`, since the macro argument is reused.
```
CHECK:MACRO_ARG_REUSE: Macro argument reuse 'value' - possible side-effects?
```

IMHO, this is an insignificant drawback, since there are other macro which also reuse their's arguments (e.g. `get_field`)

If you have any suggestions on how to fix or disable (locally) this warning, I would be glad to hear them.

en-sc:
The changes to `debug_defines.h` are generated by applying the patch https://github.com/riscv/riscv-debug-spec/pull/858 on top of previously used version of `riscv-debug-spec` (`d749752`).

en-sc:
`debug_register_printers.c` implements decoder for register values based on the new interface of `debug_defines.h`.

`riscv_debug_reg_to_s()` can be used to decode register value.  If the
pointer to buffer is `NULL` it does not print anything, just returns the
length of the string.

The format is:
`<register_value> { <field_name>=<field_value_name or field_value>, ..., }`

e.g:

`0x400382 { version=2, confstrptrvalid=invalid, hasresethaltreq=0, authbusy=ready, authenticated=true, anyhalted=1, allhalted=1, anyrunning=0, allrunning=0, anyunavail=0, allunavail=0, anynonexistent=0, allnonexistent=0, anyresumeack=0, allresumeack=0, anyhavereset=0, allhavereset=0, impebreak=1, stickyunavail=current, ndmresetpending=false, }`

`0x321009 { regno=0x1009, write=arg0, transfer=enabled, postexec=disabled, aarpostincrement=disabled, aarsize=64bit, cmdtype=0, }`

src/target/riscv/debug_register_printers.c
```diff
@@ -0,0 +1,79 @@
...
+	unsigned int length = 0;
+
+	length += print_value(buf, length, value);
+	length += get_len_or_sprintf(buf, length, " { ");
+
+	const unsigned int width = riscv_debug_reg_width(reg_ordinal, context);
+
+	for (unsigned int offset = 0; offset < width; ++offset) {
+		if (!riscv_debug_reg_field_is_present(reg_ordinal, context, offset))
+			continue;
```
timsifive: This looks really strange to me. You're iterating over every bit, and then you call a function for every bit offset to check if a field starts there, just to visit every field. Instead, it seems much better to have a list of fields for each register and then you can iterate over that.

Something like:
```
struct reg_field_info {
   const char *name;
   unsigned int lsb; // inclusive
   unsigned int msb; // inclusive
   const char **values; // If non-NULL, array of human-readable string for each possible value
};

struct reg_info {
   const char *name;
   struct reg_field_info fields[];
};
```

src/target/riscv/debug_register_printers.c
```diff
@@ -0,0 +1,79 @@
+#include <stdio.h>
+#include <inttypes.h>
+#include <assert.h>
+#include <stdarg.h>
+
+#include "debug_register_printers.h"
+
+unsigned int get_len_or_sprintf(char *buf, unsigned int curr, const char *format, ...)
```
timsifive: This makes things a lot more complex, just so we can avoid allocating an extra 30 or so bytes on the stack. It's simpler and safer to let the user of this file allocate a static buffer and to pass that in, along with the size of the buffer. Then use snprintf to not write any bytes outside that buffer. If the caller is really concerned that it might not pass enough buffer, then they can check the length of the resulting string (which must still be null-terminated).

timsifive (changes requested):

src/target/riscv/debug_register_printers.c
en-sc: Please, consider the following:
* The length of the buffer required can differ a lot, depending on the register.
```
0x400382 { version=2, confstrptrvalid=invalid, hasresethaltreq=0, authbusy=ready, authenticated=true, anyhalted=1, allhalted=1, anyrunning=0, allrunning=0, anyunavail=0, allunavail=0, anynonexistent=0, allnonexistent=0, anyresumeack=0, allresumeack=0, anyhavereset=0, allhavereset=0, impebreak=1, stickyunavail=current, ndmresetpending=false, }

0x321009 { regno=0x1009, write=arg0, transfer=enabled, postexec=disabled, aarpostincrement=disabled, aarsize=64bit, cmdtype=0, }
```
So it is not `30` and more like `O(<average_n_fields> * <average_field_name_l + average_field_value_name_l>)`.

There is an option to calculate the maximum required length of the buffer during code generation, but this value depends on the format of the output, so, to be safe, it would be required to specify the format in `debug_defines.h`.

 > If the caller is really concerned that it might not pass enough buffer, then they can check the length of the resulting string (which must still be null-terminated).

This requires the user to perform additional actions (apart from allocation and the call to the function), so the resulting client code is not as simple as it could be.

To conclude:
```
char buf[riscv_debug_reg_to_s(NULL, reg, context, value) + 1];
riscv_debug_reg_to_s(buf, reg, context, value);
```
vs
```
char buf[<SOME_ARBITRARY_CONSTANT>];
riscv_debug_reg_to_s(buf, <SOME_ARBITRARY_CONSTANT>, reg, context, value);
```
IMHO the second approach is not better.

As a compromise, it is possible to add the length of the buffer to `riscv_debug_reg_to_s`'s arguments (with corresponding checks inside). However, this would require the user to write more code (if one does not wants to use the `<SOME_ARBITRARY_CONSTANT>`):
```
size_t buf_len = riscv_debug_reg_to_s(NULL, reg, context, value) + 1;
char buf[buf_len];
riscv_debug_reg_to_s(buf, buf_len, reg, context, value);
```

en-sc:
The only warning from `checkpatch` left is:
```
WARNING:SPDX_LICENSE_TAG: Missing or malformed SPDX-License-Identifier tag in line 1
#27: FILE: src/target/riscv/debug_defines.c:1:
```
I'm not shure what is the best approach.
Should I leave it as-is (there is no license tag in `debug_defines.h` also) or should I add it?

src/target/riscv/debug_register_printers.c
en-sc: Addressed.

src/target/riscv/debug_register_printers.h
```diff
@@ -0,0 +1,8 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+
+#include <stdbool.h>
+
+#include "debug_defines.h"
+
+unsigned int riscv_debug_reg_to_s(char *buf, enum riscv_debug_reg_ordinal reg_ordinal,
+		riscv_debug_reg_ctx_t context, uint64_t value);
```
timsifive: Since this is the programmer-facing function, can you add a comment that explains how to call this function, with a tiny example as well?

src/target/riscv/debug_register_printers.h
```diff
@@ -0,0 +1,8 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+
+#include <stdbool.h>
```
timsifive: Is that used in this file?

src/target/riscv/debug_defines.c
```diff
@@ -0,0 +1,4194 @@
...
+			.values = NULL
+		},
+		.get_next = NULL
+	};
+	return result;
+}
+
+riscv_debug_reg_info_t get_riscv_debug_reg_info(enum riscv_debug_reg_ordinal reg_ordinal)
+{
+	switch (reg_ordinal) {
```
timsifive: Can this be an index into an array instead of a giant switch statement?

src/target/riscv/riscv-013.c
```diff
@@ -285,6 +286,31 @@ static dm013_info_t *get_dm(struct target *target)
...
+	const riscv_debug_reg_ctx_t context = {
+		.XLEN = { .value = r->xlen, .is_set = true },
+		.DXLEN = { .value = r->xlen, .is_set = true },
+		.abits = { .value = info->abits, .is_set = true },
+	};
+	return context;
+}
+
+#define RISCV_LOG_REG(target, reg, value) \
+do {\
```
timsifive: Can this be a function? I don't think the performance is so critical, and it makes the code easier to deal with.

src/target/riscv/riscv-013.c
```diff
@@ -306,91 +332,42 @@ static uint32_t set_dmcontrol_hartsel(uint32_t initial, int hart_index)
...
-		{ DM_SBCS, DM_SBCS_SBAUTOINCREMENT, "sbautoincrement" },
-		{ DM_SBCS, DM_SBCS_SBREADONDATA, "sbreadondata" },
-		{ DM_SBCS, DM_SBCS_SBERROR, "sberror" },
-		{ DM_SBCS, DM_SBCS_SBASIZE, "sbasize" },
-		{ DM_SBCS, DM_SBCS_SBACCESS128, "sbaccess128" },
-		{ DM_SBCS, DM_SBCS_SBACCESS64, "sbaccess64" },
-		{ DM_SBCS, DM_SBCS_SBACCESS32, "sbaccess32" },
-		{ DM_SBCS, DM_SBCS_SBACCESS16, "sbaccess16" },
-		{ DM_SBCS, DM_SBCS_SBACCESS8, "sbaccess8" },
+#define DECL_DMI_DESC(ADDR) {ADDR, ADDR##_ORDINAL}
```
timsifive: This just makes the code harder to read, even if it makes it a little bit smaller.

timsifive (changes requested):

timsifive:
> WARNING:SPDX_LICENSE_TAG: Missing or malformed SPDX-License-Identifier tag in line 1

We can ignore that warning and merge it anyway when the time comes.

src/target/riscv/debug_register_printers.c
timsifive: OK. Let's just leave it as is.

src/target/riscv/riscv-013.c

src/target/riscv/debug_register_printers.h
en-sc: Addressed.

src/target/riscv/debug_register_printers.h
en-sc: Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/debug_defines.c
en-sc: Addressed.

src/target/riscv/debug_register_printers.h
```diff
@@ -0,0 +1,28 @@
...
+#include "debug_defines.h"
+
+/**
+ * This function is used to fill a buffer with a decoded string representation
+ * of register's value.
+ * @param buf      If non-NULL, the buffer to write the string into.
+ *                 Otherwise, the function does not perform any writes,
+ *                 it just calculates the number of characters used.
+ * @param reg_ordinal
+ *                 The ordinal of the regiter.
```
timsifive:
```suggestion
 *                 The ordinal of the register.
```

timsifive (changes requested):
This code looks good. I'll head over to https://github.com/riscv/riscv-debug-spec/pull/858 now.

src/target/riscv/debug_register_printers.h
en-sc: Addressed.

en-sc:
@timsifive, @JanMatCodasip, can this be merged (since https://github.com/riscv/riscv-debug-spec/pull/858 is already merged), or should I make some adjustments?

src/target/riscv/debug_defines.c
```diff
@@ -0,0 +1,4145 @@
+/*
+ * This file is auto-generated by running 'make debug_defines' in
+ * https://github.com/riscv/riscv-debug-spec/ (d749752-dirty)
```
JanMatCodasip: Please, can you re-generate these files from a non-dirty revision of the debug spec repo.

src/target/riscv/debug_defines.c
```diff
@@ -0,0 +1,4145 @@
...
+			.lsb = 0,
+			.msb = 0,
+			.values = NULL
+		},
+		.get_next = dmi_sercs_get_valid0
+	};
+	return result;
+}
+
+riscv_debug_reg_field_list_t dmi_sertx_get_data(riscv_debug_reg_ctx_t context)
```
JanMatCodasip: I assume that this function (and the similar ones) is only expected to be called from within debug_defines.c, not by an external user. Is that correct understanding?

If so, why is it not marked as `static`? To avoid naming clashes and polluting the global namespace.

(The change needs of course to be made in the riscv-debug-spec repo in the code generator.)

src/target/riscv/debug_defines.h
```diff
@@ -3059,3 +3061,93 @@
...
+	DM_SBDATA3_ORDINAL,
+	SHORTNAME_ORDINAL,
+	AC_ACCESS_REGISTER_ORDINAL,
+	AC_QUICK_ACCESS_ORDINAL,
+	AC_ACCESS_MEMORY_ORDINAL,
+	VIRT_PRIV_ORDINAL,
+	DMI_SERCS_ORDINAL,
+	DMI_SERTX_ORDINAL,
+	DMI_SERRX_ORDINAL
+};
```
JanMatCodasip, line 3229: Here I would like to (politely) restate my opinion that creating this enumeration is incorrect from architectural and API point of view -- for these reasons:

- Mixing of different classes of registers (JTAG, DM, CSR) in one enumeration and in one function (`get_riscv_debug_reg_info()`) is principally incorrect.
- The registers already have their enumerations or indexes defined, so defining another one is redundant and brings the need to convert between them.

Ideally, the API would use the existing indexes and either have one function per the register class (JTAG, DM or CSR); or one function with an additional argument to distinguish the register class. If there are special cases (like tdata1 or textra*), they should be handled by special means (e.g. specific function for those cases) and they IMO do not justify the creation of the enum mix.

Since https://github.com/riscv/riscv-debug-spec/pull/858 has already been reviewed & merged, I respect the decision to leave the current code as is. (No code change asked for here.)

src/target/riscv/debug_register_printers.h
```diff
@@ -0,0 +1,28 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
```
JanMatCodasip: How about calling the files just `reg_printer.h, .c` or `debug_reg_printer.h, .c`?

That is shorter and, IMO, equally clear.

src/target/riscv/riscv-013.c
```diff
@@ -417,13 +390,13 @@ static void dump_field(struct target *target, int idle, const struct scan_field
 			field->num_bits, op_string[out_op], out_data, out_address,
 			status_string[in_op], in_data, in_address, idle);
 
-	char out_text[500];
-	char in_text[500];
-	decode_dmi(target, out_text, out_address, out_data);
-	decode_dmi(target, in_text, in_address, in_data);
+	char out_text[decode_dmi(target, NULL, out_address, out_data) + 1];
+	unsigned int out_len  = decode_dmi(target, out_text, out_address, out_data);
```
JanMatCodasip: Detail: Extra space

```suggestion
	unsigned int out_len = decode_dmi(target, out_text, out_address, out_data);
```

src/target/riscv/debug_register_printers.c
```diff
@@ -0,0 +1,94 @@
...
+	if (buf)
+		length = vsprintf(buf + curr, format, args);
+	else
+		length = vsnprintf(NULL, 0, format, args);
+	va_end(args);
+	assert(length >= 0);
+	return (unsigned int)length;
+}
+
+static unsigned int print_value(char *buf, unsigned int offset, uint64_t value)
```
JanMatCodasip: If you like, a suggestion for a more descriptive name:

```suggestion
static unsigned int print_number(char *buf, unsigned int offset, uint64_t value)
```

JanMatCodasip (changes requested):
Despite having somewhat different opinion on what the API of debug_defines.h could be, I like this change overall and it looks to me like a nice step forward in unification of the log prints. Thanks!

I believe there are few minor items to address prior to the merge.

src/target/riscv/debug_defines.c
en-sc: This is intentional. I would like to avoid updating the version of the spec used to generate this file, since it would require a lot of changes in code (some constants were renamed and so on). The version is marked as dirty to indicate that the generators were cherry-picked on top of `d749752`. If you insist, I will update `debug_defines.h` to current version, though I'm not sure if it is the best time to do so.

src/target/riscv/debug_defines.c
en-sc: Addressed. Please, see https://github.com/riscv/riscv-debug-spec/pull/876/files#diff-48716077ff5d92cc3ba425f22306df3b5b5c5001c3213b6f307f0df1865a2da4R121

src/target/riscv/debug_register_printers.h
en-sc: Addressed. The corresponding change in `riscv-debug-spec`: https://github.com/riscv/riscv-debug-spec/pull/876

src/target/riscv/riscv-013.c

src/target/riscv/debug_register_printers.c
en-sc: Addressed. The corresponding change in `riscv-debug-spec`: https://github.com/riscv/riscv-debug-spec/pull/876

src/target/riscv/debug_defines.c
JanMatCodasip: > I would like to avoid updating the version of the spec used to generate this file, since it would require a lot of changes in code (some constants were renamed and so on). 

Understood. In that case, it is my recommendation to leave this merge request as is, but please make a proper upgrade of the debug_defines (from a clean commit) in a subsequent merge request.

> The version is marked as dirty to indicate that the generators were cherry-picked on top of d749752

This is exactly the reason why I suggest to use a clean commit -- we can't expect other developers to figure out what needs to be cherry-picked where (what mixture of commits to make) in order to replicate your work, e.g. during troubleshooting.

Thanks.

JanMatCodasip (dismissed):
This looks fine to me. 

I have only one remark https://github.com/riscv/riscv-openocd/pull/892#discussion_r1328269512 (which can be addressed as a separate merge request to keep the changes easier to review).

Thank you.

src/target/riscv/debug_defines.c
en-sc: I have re-checked and the changes to the code are limited (only one function), so I updated the version of `debug_defines.h` in this MR.

en-sc:
@timsifive, @JanMatCodasip can this be merged or is there some action required to be done?

timsifive (approved):

PR #895, opened 2023-07-25
target/riscv: OpenOCD fails with assert during running "reset" command

kr-sc:
OpenOCD fails in the presence of inactive/unresponsive cores

I faced with case when inactive core returns 0 while reading dtmcontrol. This leads to failure on assert: "addrbits != 0" in "dbus_scan".

Also change "read_bits","poll_target" funcs to avoid a lot lines in logs

Change-Id: If852126755317789602b7372c5c5732183fff6c5

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -426,7 +426,10 @@ static dbus_status_t dbus_scan(struct target *target, uint16_t *address_in,
 		.in_value = in
 	};
 
-	assert(info->addrbits != 0);
+	if (info->addrbits == 0) {
+		LOG_TARGET_ERROR(target, "Can't access DMI because addrbits=0.");
+		return DBUS_STATUS_FAILED;
+	}
 
 	buf_set_u64(out, DBUS_OP_START, DBUS_OP_SIZE, op);
 	buf_set_u64(out, DBUS_DATA_START, DBUS_DATA_SIZE, data_out);
@@ -680,17 +683,12 @@ static void dram_write32(struct target *target, unsigned int index, uint32_t val
 }
 
 /** Read the haltnot and interrupt bits. */
-static bits_t read_bits(struct target *target)
+static int read_bits(struct target *target, bits_t *result)
 {
 	uint64_t value;
 	dbus_status_t status;
 	uint16_t address_in;
 	riscv011_info_t *info = get_info(target);
-
-	bits_t err_result = {
-		.haltnot = 0,
-		.interrupt = 0
-	};
 
 	do {
 		unsigned i = 0;
@@ -700,26 +698,23 @@ static bits_t read_bits(struct target *target)
 				if (address_in == (1<<info->addrbits) - 1 &&
 						value == (1ULL<<DBUS_DATA_SIZE) - 1) {
 					LOG_ERROR("TDO seems to be stuck high.");
-					return err_result;
+					return ERROR_FAIL;
 				}
 				increase_dbus_busy_delay(target);
-			} else if (status == DBUS_STATUS_FAILED) {
-				/* TODO: return an actual error */
-				return err_result;
 			}
 		} while (status == DBUS_STATUS_BUSY && i++ < 256);
 
-		if (i >= 256) {
+		if (status != DBUS_STATUS_SUCCESS) {
 			LOG_ERROR("Failed to read from 0x%x; status=%d", address_in, status);
-			return err_result;
+			return ERROR_FAIL;
 		}
 	} while (address_in > 0x10 && address_in != DMCONTROL);
 
-	bits_t result = {
-		.haltnot = get_field(value, DMCONTROL_HALTNOT),
-		.interrupt = get_field(value, DMCONTROL_INTERRUPT)
-	};
-	return result;
+	if (result) {
+		result->haltnot = get_field(value, DMCONTROL_HALTNOT);
+		result->interrupt = get_field(value, DMCONTROL_INTERRUPT);
+	}
+	return ERROR_OK;
 }
 
 static int wait_for_debugint_clear(struct target *target, bool ignore_first)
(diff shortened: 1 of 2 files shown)


timsifive (commented):
At a glance this looks OK.

What hardware are you connecting to that implements 0.11? I didn't realize there were implementations beyond HiFive1.

What does "inactive" mean in this case? Is that core held in reset or something? And I assume that condition gets resolved after time, allowing examine() to succeed eventually.

kr-sc:
> What hardware are you connecting to that implements 0.11? I didn't realize there were implementations beyond HiFive1.

OpenOCD detect spec version using `version` field in `dtmcs` register, When it try to read this register, it get 0 value (because hart is inactive) and interpret it as `version = 0` . According to spec, it means that hardware implement 0.11

>What does "inactive" mean in this case? Is that core held in reset or something? And I assume that condition gets resolved after time, allowing examine() to succeed eventually.

Something like that, yes. Some harts in the system are turned off by default, unless a proper initialization procedure is executed. Once a hart is up (if user initiates the sequence), the examine will eventually succeeds

timsifive:
Thanks for explaining. I think it would be simpler to check if dtmcs (erroneously called dtmcontrol) is 0 in riscv_examine(). That would cover both versions, and lets you share the code/check. Mostly I'm saying that because 0.11 is rarely used, we and we don't have as good test coverage for it. I'd rather not change it unless there's a compelling reason.

JanMatCodasip:
> I think it would be simpler to check if dtmcs (erroneously called dtmcontrol) is 0 in riscv_examine()

Actually, the checks for `dtmcs (dtmcontrol) != 0` are already present on the top of examine() functions for both 0.11 and 0.13:

- https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv-011.c#L1466
- https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv-013.c#L1704

@kr-sc Are these existing checks insufficient to cover your case? Should they be extended or improved somehow?

kr-sc:
> Actually, the checks for dtmcs (dtmcontrol) != 0 are already present on the top of examine() functions for both 0.11 and 0.13:

Yes, but if I understand corretcly there are no any checks in assert_reset() function, for example. Even if target was not examined, OpenOCD try to reset and it leads to failure in assertion. So, as for me, it is better to replace this assertion with this check to avoid such situations for other functions where may not be such checks too

src/target/riscv/riscv-011.c
```diff
@@ -426,7 +426,10 @@ static dbus_status_t dbus_scan(struct target *target, uint16_t *address_in,
 		.in_value = in
 	};
 
-	assert(info->addrbits != 0);
+	if (info->addrbits == 0) {
+		LOG_ERROR("scan failed; addrbits=0");
```
timsifive:
```suggestion
		LOG_TARGET_ERROR(target, "Can't access DMI because addrbits=0.");
```

src/target/riscv/riscv-011.c
```diff
@@ -730,10 +725,17 @@ static int wait_for_debugint_clear(struct target *target, bool ignore_first)
...
+		read_bits(target, NULL);
 	}
 	while (1) {
-		bits_t bits = read_bits(target);
+		bits_t bits = {
+			.haltnot = 0,
+			.interrupt = 0
+		};
+		if (read_bits(target, &bits) != ERROR_OK) {
+			LOG_ERROR("read_bits fails");
```
timsifive: We don't need this generic error message, since read_bits() already provides a more specific error message.

src/target/riscv/riscv-011.c
```diff
@@ -1893,7 +1895,15 @@ static int poll_target(struct target *target, bool announce)
 	int old_debug_level = debug_level;
 	if (debug_level >= LOG_LVL_DEBUG)
 		debug_level = LOG_LVL_INFO;
-	bits_t bits = read_bits(target);
+	bits_t bits = {
+		.haltnot = 0,
+		.interrupt = 0
+	};
+	if (read_bits(target, &bits) != ERROR_OK) {
+		LOG_ERROR("read_bits fails");
```
timsifive: Remove also.

src/target/riscv/riscv-013.c
```diff
@@ -497,7 +497,10 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 	memset(in, 0, num_bytes);
 	memset(out, 0, num_bytes);
 
-	assert(info->abits != 0);
+	if (info->abits == 0) {
+		LOG_ERROR("scan failed; addrbits=0");
```
```suggestion
		LOG_TARGET_ERROR(target, "Can't access DMI because addrbits=0.");
```

timsifive (changes requested):
OK. These changes do make the code better, so let's go with them.

I've checked that everything still passes on HiFive1.

src/target/riscv/riscv-011.c
kr-sc: Done

src/target/riscv/riscv-011.c
kr-sc: I've decided to add those for consistency. Meaning, if we decide to refactor/change this code in the future, we can have a higher chance that at least some error message is printed.  Excessive printing should not be a problem - since these kinds of messages indicate that something is terribly wrong anyway. I don't insist however, so if you think that it's better to remove them, I'll do it. Nevertheless, could you revise your comments?

src/target/riscv/riscv-011.c
kr-sc: Same as above

src/target/riscv/riscv-013.c
kr-sc: Done

src/target/riscv/riscv-011.c
timsifive: What are you being consistent with? Generally the approach is to print one error as specific as possible, ideally in the place that the error is first generated. Then the other layers can simply propagate the error up. There are exceptions where the caller has more information to add, but in this case the caller error message doesn't add any extra information.

kr-sc: I just want to add more information in log (for example, we can understand what `read_bits` was called and failed in this function because of this error message), but if it's not necessary, I'll remove it

kr-sc: Done

src/target/riscv/riscv-011.c
kr-sc: Done

timsifive (approved):

timsifive:
These changes are failing checkpatch because of a merge from upstream. I believe if you rebase the changes on top of riscv, then that problem will be resolved.

timsifive:
These changes are failing checkpatch because of a merge from upstream. I believe if you rebase the changes on top of riscv, then that problem will be resolved.

kr-sc:
> These changes are failing checkpatch because of a merge from upstream. I believe if you rebase the changes on top of riscv, then that problem will be resolved.

Rebase done

MarekVCodasip:
I checked this visually. It looks alright.

MarekVCodasip (approved):

PR #896, opened 2023-07-31
target/riscv: Adding register tables to make register names consiste

AnastasiyaChernikova:
Motivation for the change is inconsistent output of register names in logs. 
For example:
```
...
command - wp 0x0 4
...
15627 Debug: 19313 10173 riscv.c:4757 riscv_get_register(): [riscv.cpu0] Reading tinfo from target                  
15628 Debug: 19314 10173 riscv-013.c:4233 riscv013_get_register(): [riscv.cpu0] reading register csr1956            
15629 Debug: 19315 10173 riscv-013.c:1508 register_read_direct(): [riscv.cpu0] Reading csr195
```
This problem is fixed and memory management for register names refactored.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -488,6 +488,21 @@ static void riscv_free_registers(struct target *target)
 	}
 }
 
+static void free_reg_names(struct target *target);
+
+static void free_custom_register_names(struct target *target)
+{
+	RISCV_INFO(info);
+
+	if (!info->custom_register_names.reg_names)
+		return;
+
+	for (unsigned int i = 0; i < info->custom_register_names.num_entries; i++)
+		free(info->custom_register_names.reg_names[i]);
+	free(info->custom_register_names.reg_names);
+	info->custom_register_names.reg_names = NULL;
+}
+
 static void riscv_deinit_target(struct target *target)
 {
 	LOG_TARGET_DEBUG(target, "riscv_deinit_target()");
@@ -524,7 +539,7 @@ static void riscv_deinit_target(struct target *target)
 		free(entry);
 	}
 
-	free(info->reg_names);
+	free_reg_names(target);
 	free(target->arch_info);
 
 	target->arch_info = NULL;
@@ -2670,7 +2685,8 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 				riscv_reg_t reg_value;
 				if (riscv_get_register(target, &reg_value, regno) != ERROR_OK)
 					break;
-				LOG_TARGET_ERROR(target, "%s = 0x%" PRIx64, gdb_regno_name(regno), reg_value);
+
+				LOG_TARGET_ERROR(target, "%s = 0x%" PRIx64, gdb_regno_name(target, regno), reg_value);
 			}
 			return ERROR_TARGET_TIMEOUT;
 		}
@@ -5011,7 +5027,7 @@ static int riscv_set_or_write_register(struct target *target,
 		assert(!target_was_examined(target));
 		LOG_TARGET_DEBUG(target,
 				"No cache, writing to target: %s <- 0x%" PRIx64,
-				gdb_regno_name(regid), value);
+				gdb_regno_name(target, regid), value);
 		return r->set_register(target, regid, value);
 	}
 
@@ -5106,7 +5122,7 @@ int riscv_get_register(struct target *target, riscv_reg_t *value,
 	if (!target->reg_cache) {
 		assert(!target_was_examined(target));
 		LOG_TARGET_DEBUG(target, "No cache, reading %s from target",
-				gdb_regno_name(regid));
+				gdb_regno_name(target, regid));
 		return r->get_register(target, value, regid);
 	}
 
@@ -5143,7 +5159,7 @@ int riscv_save_register(struct target *target, enum gdb_regno regid)
 {
 	if (target->state != TARGET_HALTED) {
 		LOG_TARGET_ERROR(target, "Can't save register %s on a hart that is not halted.",
-				 gdb_regno_name(regid));
+				 gdb_regno_name(target, regid));
 		return ERROR_FAIL;
 	}
 	assert(gdb_regno_cacheable(regid, /* is write? */ false) &&
@@ -5322,7 +5338,7 @@ static int disable_trigger_if_dmode(struct target *target, riscv_reg_t tdata1)
  * something.
  * Disable any hardware triggers that have dmode set. We can't have set them
  * ourselves. Maybe they're left over from some killed debug session.
- * */
+ */
 int riscv_enumerate_triggers(struct target *target)
 {
 	RISCV_INFO(r);
@@ -5381,188 +5397,174 @@ int riscv_enumerate_triggers(struct target *target)
 	return ERROR_OK;
 }
 
-const char *gdb_regno_name(enum gdb_regno regno)
+static char *init_reg_name(const char *name)
 {
-	static char buf[32];
+	const int size_buf = strlen(name) + 1;
 
-	switch (regno) {
-		case GDB_REGNO_ZERO:
-			return "zero";
-		case GDB_REGNO_RA:
-			return "ra";
-		case GDB_REGNO_SP:
-			return "sp";
-		case GDB_REGNO_GP:
-			return "gp";
-		case GDB_REGNO_TP:
-			return "tp";
-		case GDB_REGNO_T0:
-			return "t0";
-		case GDB_REGNO_T1:
-			return "t1";
-		case GDB_REGNO_T2:
-			return "t2";
-		case GDB_REGNO_S0:
-			return "s0";
-		case GDB_REGNO_S1:
-			return "s1";
-		case GDB_REGNO_A0:
-			return "a0";
-		case GDB_REGNO_A1:
-			return "a1";
-		case GDB_REGNO_A2:
-			return "a2";
-		case GDB_REGNO_A3:
-			return "a3";
-		case GDB_REGNO_A4:
-			return "a4";
-		case GDB_REGNO_A5:
-			return "a5";
-		case GDB_REGNO_A6:
-			return "a6";
-		case GDB_REGNO_A7:
-			return "a7";
-		case GDB_REGNO_S2:
-			return "s2";
-		case GDB_REGNO_S3:
-			return "s3";
-		case GDB_REGNO_S4:
-			return "s4";
-		case GDB_REGNO_S5:
-			return "s5";
-		case GDB_REGNO_S6:
-			return "s6";
-		case GDB_REGNO_S7:
-			return "s7";
-		case GDB_REGNO_S8:
-			return "s8";
-		case GDB_REGNO_S9:
-			return "s9";
-		case GDB_REGNO_S10:
-			return "s10";
-		case GDB_REGNO_S11:
-			return "s11";
-		case GDB_REGNO_T3:
-			return "t3";
-		case GDB_REGNO_T4:
-			return "t4";
-		case GDB_REGNO_T5:
-			return "t5";
-		case GDB_REGNO_T6:
-			return "t6";
-		case GDB_REGNO_PC:
-			return "pc";
-		case GDB_REGNO_FPR0:
-			return "fpr0";
-		case GDB_REGNO_FPR31:
-			return "fpr31";
-		case GDB_REGNO_CSR0:
-			return "csr0";
-		case GDB_REGNO_TSELECT:
-			return "tselect";
-		case GDB_REGNO_TDATA1:
-			return "tdata1";
-		case GDB_REGNO_TDATA2:
-			return "tdata2";
-		case GDB_REGNO_MISA:
-			return "misa";
-		case GDB_REGNO_DPC:
-			return "dpc";
-		case GDB_REGNO_DCSR:
-			return "dcsr";
-		case GDB_REGNO_DSCRATCH0:
-			return "dscratch0";
-		case GDB_REGNO_MSTATUS:
-			return "mstatus";
-		case GDB_REGNO_MEPC:
-			return "mepc";
-		case GDB_REGNO_MCAUSE:
-			return "mcause";
-		case GDB_REGNO_PRIV:
-			return "priv";
-		case GDB_REGNO_SATP:
-			return "satp";
-		case GDB_REGNO_VTYPE:
-			return "vtype";
-		case GDB_REGNO_VL:
-			return "vl";
-		case GDB_REGNO_V0:
-			return "v0";
-		case GDB_REGNO_V1:
-			return "v1";
-		case GDB_REGNO_V2:
-			return "v2";
-		case GDB_REGNO_V3:
-			return "v3";
-		case GDB_REGNO_V4:
-			return "v4";
-		case GDB_REGNO_V5:
-			return "v5";
-		case GDB_REGNO_V6:
-			return "v6";
-		case GDB_REGNO_V7:
-			return "v7";
-		case GDB_REGNO_V8:
-			return "v8";
-		case GDB_REGNO_V9:
-			return "v9";
-		case GDB_REGNO_V10:
-			return "v10";
-		case GDB_REGNO_V11:
-			return "v11";
-		case GDB_REGNO_V12:
-			return "v12";
-		case GDB_REGNO_V13:
-			return "v13";
-		case GDB_REGNO_V14:
-			return "v14";
-		case GDB_REGNO_V15:
-			return "v15";
-		case GDB_REGNO_V16:
-			return "v16";
-		case GDB_REGNO_V17:
-			return "v17";
-		case GDB_REGNO_V18:
-			return "v18";
-		case GDB_REGNO_V19:
-			return "v19";
-		case GDB_REGNO_V20:
-			return "v20";
-		case GDB_REGNO_V21:
-			return "v21";
-		case GDB_REGNO_V22:
-			return "v22";
-		case GDB_REGNO_V23:
-			return "v23";
-		case GDB_REGNO_V24:
-			return "v24";
-		case GDB_REGNO_V25:
-			return "v25";
-		case GDB_REGNO_V26:
-			return "v26";
-		case GDB_REGNO_V27:
-			return "v27";
-		case GDB_REGNO_V28:
-			return "v28";
-		case GDB_REGNO_V29:
-			return "v29";
-		case GDB_REGNO_V30:
-			return "v30";
-		case GDB_REGNO_V31:
-			return "v31";
-		default:
-			if (regno <= GDB_REGNO_XPR31)
-				sprintf(buf, "x%d", regno - GDB_REGNO_ZERO);
-			else if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095)
-				sprintf(buf, "csr%d", regno - GDB_REGNO_CSR0);
-			else if (regno >= GDB_REGNO_FPR0 && regno <= GDB_REGNO_FPR31)
-				sprintf(buf, "f%d", regno - GDB_REGNO_FPR0);
-			else
-				sprintf(buf, "gdb_regno_%d", regno);
-			return buf;
+	char * const buf = calloc(size_buf, sizeof(char));
+	if (!buf) {
+		LOG_ERROR("Failed to allocate memory for a register name.");
+		return NULL;
+	}
+	strcpy(buf, name);
+	return buf;
+}
+
+static char *init_reg_name_with_prefix(const char *name_prefix,
+	unsigned int num)
+{
+	const int size_buf = snprintf(NULL, 0, "%s%d", name_prefix, num) + 1;
+
+	char * const buf = calloc(size_buf, sizeof(char));
+	if (!buf) {
+		LOG_ERROR("Failed to allocate memory for a register name.");
+		return NULL;
+	}
+	int result = snprintf(buf, size_buf, "%s%d", name_prefix, num);
+	assert(result > 0 && result <= (size_buf - 1));
+	return buf;
+}
+
+static const char * const default_reg_names[GDB_REGNO_COUNT] = {
+	[GDB_REGNO_ZERO] = "zero",
+	[GDB_REGNO_RA] = "ra",
+	[GDB_REGNO_SP] = "sp",
+	[GDB_REGNO_GP] = "gp",
+	[GDB_REGNO_TP] = "tp",
+	[GDB_REGNO_T0] = "t0",
+	[GDB_REGNO_T1] = "t1",
+	[GDB_REGNO_T2] = "t2",
+	[GDB_REGNO_FP] = "fp",
+	[GDB_REGNO_S1] = "s1",
+	[GDB_REGNO_A0] = "a0",
+	[GDB_REGNO_A1] = "a1",
+	[GDB_REGNO_A2] = "a2",
+	[GDB_REGNO_A3] = "a3",
+	[GDB_REGNO_A4] = "a4",
+	[GDB_REGNO_A5] = "a5",
+	[GDB_REGNO_A6] = "a6",
+	[GDB_REGNO_A7] = "a7",
+	[GDB_REGNO_S2] = "s2",
+	[GDB_REGNO_S3] = "s3",
+	[GDB_REGNO_S4] = "s4",
+	[GDB_REGNO_S5] = "s5",
+	[GDB_REGNO_S6] = "s6",
+	[GDB_REGNO_S7] = "s7",
+	[GDB_REGNO_S8] = "s8",
+	[GDB_REGNO_S9] = "s9",
+	[GDB_REGNO_S10] = "s10",
+	[GDB_REGNO_S11] = "s11",
+	[GDB_REGNO_T3] = "t3",
+	[GDB_REGNO_T4] = "t4",
+	[GDB_REGNO_T5] = "t5",
+	[GDB_REGNO_T6] = "t6",
+	[GDB_REGNO_PC] = "pc",
+	[GDB_REGNO_CSR0] = "csr0",
+	[GDB_REGNO_PRIV] = "priv",
+	[GDB_REGNO_FT0] = "ft0",
+	[GDB_REGNO_FT1] = "ft1",
+	[GDB_REGNO_FT2] = "ft2",
+	[GDB_REGNO_FT3] = "ft3",
+	[GDB_REGNO_FT4] = "ft4",
+	[GDB_REGNO_FT5] = "ft5",
+	[GDB_REGNO_FT6] = "ft6",
+	[GDB_REGNO_FT7] = "ft7",
+	[GDB_REGNO_FS0] = "fs0",
+	[GDB_REGNO_FS1] = "fs1",
+	[GDB_REGNO_FA0] = "fa0",
+	[GDB_REGNO_FA1] = "fa1",
+	[GDB_REGNO_FA2] = "fa2",
+	[GDB_REGNO_FA3] = "fa3",
+	[GDB_REGNO_FA4] = "fa4",
+	[GDB_REGNO_FA5] = "fa5",
+	[GDB_REGNO_FA6] = "fa6",
+	[GDB_REGNO_FA7] = "fa7",
+	[GDB_REGNO_FS2] = "fs2",
+	[GDB_REGNO_FS3] = "fs3",
+	[GDB_REGNO_FS4] = "fs4",
+	[GDB_REGNO_FS5] = "fs5",
+	[GDB_REGNO_FS6] = "fs6",
+	[GDB_REGNO_FS7] = "fs7",
+	[GDB_REGNO_FS8] = "fs8",
+	[GDB_REGNO_FS9] = "fs9",
+	[GDB_REGNO_FS10] = "fs10",
+	[GDB_REGNO_FS11] = "fs11",
+	[GDB_REGNO_FT8] = "ft8",
+	[GDB_REGNO_FT9] = "ft9",
+	[GDB_REGNO_FT10] = "ft10",
+	[GDB_REGNO_FT11] = "ft11",
+
+	#define DECLARE_CSR(csr_name, number)[(number) + GDB_REGNO_CSR0] = #csr_name,
+	#include "encoding.h"
+	#undef DECLARE_CSR
+};
+
+static void free_reg_names(struct target *target)
+{
+	RISCV_INFO(info);
+
+	if (!info->reg_names)
+		return;
+
+	for (unsigned int i = 0; i < GDB_REGNO_COUNT; ++i)
+		free(info->reg_names[i]);
+	free(info->reg_names);
+	info->reg_names = NULL;
+
+	free_custom_register_names(target);
+}
+
+static void init_custom_csr_names(struct target *target)
+{
+	RISCV_INFO(info);
+	range_list_t *entry;
+
+	list_for_each_entry(entry, &info->expose_csr, list) {
+		if (!entry->name)
+			continue;
+		assert(entry->low == entry->high);
+		const unsigned int regno = entry->low + GDB_REGNO_CSR0;
+		assert(regno <= GDB_REGNO_CSR4095);
+		if (info->reg_names[regno])
+			return;
+		info->reg_names[regno] = init_reg_name(entry->name);
 	}
 }
 
+const char *gdb_regno_name(struct target *target, unsigned int regno)
+{
+	RISCV_INFO(info);
+
+	if (regno >= GDB_REGNO_COUNT) {
+		assert(info->custom_register_names.reg_names);
+		assert(regno - GDB_REGNO_COUNT <= info->custom_register_names.num_entries);
+		return info->custom_register_names.reg_names[regno - GDB_REGNO_COUNT];
+	}
+
+	if (!info->reg_names)
+		info->reg_names = calloc(GDB_REGNO_COUNT, sizeof(char *));
+
+	if (info->reg_names[regno])
+		return info->reg_names[regno];
+	if (default_reg_names[regno])
+		return default_reg_names[regno];
+	if (regno <= GDB_REGNO_XPR31) {
+		info->reg_names[regno] = init_reg_name_with_prefix("x", regno - GDB_REGNO_ZERO);
+		return info->reg_names[regno];
+	}
+	if (regno <= GDB_REGNO_V31 && regno >= GDB_REGNO_V0) {
+		info->reg_names[regno] = init_reg_name_with_prefix("v", regno - GDB_REGNO_V0);
+		return info->reg_names[regno];
+	}
+	if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095) {
+		init_custom_csr_names(target);
+		info->reg_names[regno] = init_reg_name_with_prefix("csr", regno - GDB_REGNO_CSR0);
+		return info->reg_names[regno];
+	}
+	assert(!"Encountered uninitialized entry in reg_names table");
+
+	return NULL;
+}
 
 /**
  * This function is the handler of user's request to read a register.
@@ -5662,14 +5664,54 @@ static struct reg_arch_type riscv_reg_arch_type = {
 	.set = register_set
 };
 
-struct csr_info {
-	unsigned number;
-	const char *name;
-};
-
-static int cmp_csr_info(const void *p1, const void *p2)
+static int init_custom_register_names(struct list_head *expose_custom,
+		struct reg_name_table *custom_register_names)
 {
-	return (int) (((struct csr_info *)p1)->number) - (int) (((struct csr_info *)p2)->number);
+	unsigned int custom_regs_num = 0;
+	if (!list_empty(expose_custom)) {
+		range_list_t *entry;
+		list_for_each_entry(entry, expose_custom, list)
+			custom_regs_num += entry->high - entry->low + 1;
+	}
+
+	if (!custom_regs_num)
+		return ERROR_OK;
+
+	custom_register_names->reg_names = calloc(custom_regs_num, sizeof(char *));
+	if (!custom_register_names->reg_names) {
+		LOG_ERROR("Failed to allocate memory for custom_register_names->reg_names");
+		return ERROR_FAIL;
+	}
+	custom_register_names->num_entries = custom_regs_num;
+	char **reg_names = custom_register_names->reg_names;
+	range_list_t *range;
+	unsigned int next_custom_reg_index = 0;
+	list_for_each_entry(range, expose_custom, list) {
+		for (unsigned int custom_number = range->low; custom_number <= range->high; ++custom_number) {
+			if (range->name)
+				reg_names[next_custom_reg_index] = init_reg_name(range->name);
+			else
+				reg_names[next_custom_reg_index] =
+					init_reg_name_with_prefix("custom", custom_number);
+
+			if (!reg_names[next_custom_reg_index])
+				return ERROR_FAIL;
+			++next_custom_reg_index;
+		}
+	}
+	return ERROR_OK;
+}
+
+static bool is_known_standard_csr(unsigned int csr_num)
+{
+	static const bool is_csr_in_buf[GDB_REGNO_CSR4095 - GDB_REGNO_CSR0 + 1] = {
+		#define DECLARE_CSR(csr_name, number)[number] = true,
+		#include "encoding.h"
+		#undef DECLARE_CSR
+	};
+	assert(csr_num < ARRAY_SIZE(is_csr_in_buf));
+
+	return is_csr_in_buf[csr_num];
 }
 
 int riscv_init_registers(struct target *target)
@@ -5679,32 +5721,27 @@ int riscv_init_registers(struct target *target)
 	riscv_free_registers(target);
 
 	target->reg_cache = calloc(1, sizeof(*target->reg_cache));
-	if (!target->reg_cache)
+	if (!target->reg_cache) {
+		LOG_TARGET_ERROR(target, "Failed to allocate memory for target->reg_cache");
 		return ERROR_FAIL;
+	}
 	target->reg_cache->name = "RISC-V Registers";
-	target->reg_cache->num_regs = GDB_REGNO_COUNT;
 
-	if (!list_empty(&info->expose_custom)) {
-		range_list_t *entry;
-		list_for_each_entry(entry, &info->expose_custom, list)
-			target->reg_cache->num_regs += entry->high - entry->low + 1;
+	if (init_custom_register_names(&info->expose_custom, &info->custom_register_names) != ERROR_OK) {
+		LOG_TARGET_ERROR(target, "init_custom_register_names failed");
+		return ERROR_FAIL;
 	}
 
+	target->reg_cache->num_regs = GDB_REGNO_COUNT + info->custom_register_names.num_entries;
 	LOG_TARGET_DEBUG(target, "create register cache for %d registers",
 			target->reg_cache->num_regs);
 
 	target->reg_cache->reg_list =
 		calloc(target->reg_cache->num_regs, sizeof(struct reg));
-	if (!target->reg_cache->reg_list)
+	if (!target->reg_cache->reg_list) {
+		LOG_TARGET_ERROR(target, "Failed to allocate memory for target->reg_cache->reg_list");
 		return ERROR_FAIL;
-
-	const unsigned int max_reg_name_len = 12;
-	free(info->reg_names);
-	info->reg_names =
-		calloc(target->reg_cache->num_regs, max_reg_name_len);
-	if (!info->reg_names)
-		return ERROR_FAIL;
-	char *reg_name = info->reg_names;
+	}
 
 	static struct reg_feature feature_cpu = {
 		.name = "org.gnu.gdb.riscv.cpu"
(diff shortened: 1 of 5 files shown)


AnastasiyaChernikova:
This change passed testing on tests from `riscv-tests/debug` testsuite (the results can be seen below), as well as some testing on hardware.
<details><summary>spike32</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 187s ]:::::::::::::::::::::::::::
14 tests returned not_applicable
57 tests returned pass
</details>
<details><summary>spike64</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 198s ]:::::::::::::::::::::::::::
12 tests returned not_applicable
59 tests returned pass
</details>
<details><summary>spike32-2-hwthread</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 241s ]:::::::::::::::::::::::::::
12 tests returned not_applicable
59 tests returned pass
</details>
<details><summary>spike64-2</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 260s ]:::::::::::::::::::::::::::
14 tests returned not_applicable
57 tests returned pass
</details>
<details><summary>spike32-2</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 267s ]:::::::::::::::::::::::::::
13 tests returned not_applicable
58 tests returned pass
</details>
<details><summary>spike64-2-hwthread</summary>
<details><summary> test list </summary>
CeaseMultiTest CeaseRunTest CeaseStepiTest CheckMisa CustomRegisterTest DebugBreakpoint DebugChangeString DebugCompareSections DebugExit DebugFunctionCall DebugSymbols DebugTurbostep DisconnectTest DownloadTest EbreakTest EtriggerTest FreeRtosTest Hwbp1 Hwbp2 HwbpManual IcountTest InfoTest InstantChangePc InstantHaltTest InterruptTest ItriggerTest JumpHbreak MemTest16 MemTest32 MemTest64 MemTest8 MemTestBlock0 MemTestBlock1 MemTestBlock2 MemTestBlockReadInvalid MemTestReadInvalid MemorySampleMixed MemorySampleSingle MulticoreRegTest MulticoreRtosSwitchActiveHartTest MulticoreRunAllHaltOne PrivChange PrivRw ProgramHwWatchpoint ProgramSwWatchpoint Registers RepeatReadTest Semihosting SemihostingFileio SimpleF18Test SimpleNoExistTest SimpleS0Test SimpleS1Test SimpleT0Test SimpleT1Test SimpleV13Test SmpSimultaneousRunHalt StepTest StepThread2Test Sv32Test Sv39Test Sv48Test TooManyHwbp TriggerDmode TriggerExecuteInstant TriggerLoadAddressInstant TriggerStoreAddressInstant UserInterrupt VectorTest WriteCsrs WriteGprs
</details>

:::::::::::::::::::::::::::[ ran 71 tests in 320s ]:::::::::::::::::::::::::::
12 tests returned not_applicable
59 tests returned pass
</details>

src/target/riscv/riscv.c
```diff
@@ -5084,189 +5086,139 @@ int riscv_enumerate_triggers(struct target *target)
...
+
+struct csr_info {
+	unsigned int number;
+	const char *name;
+};
+
+struct csr_info_table {
+	unsigned int records_num;
+	const struct csr_info *csr_records;
+};
```
timsifive: These structures aren't limited to CSRs. Probably call them reg_ instead of csr_

src/target/riscv/riscv.c
```diff
@@ -36,6 +36,7 @@
 #define DTMCONTROL_VERSION			(0xf)
 
 #define DBUS						0x11
+#define CSR_NAME_MAX				32u
```
```suggestion
#define REG_NAME_MAX_LENGTH				32u
```

src/target/riscv/riscv.c
```diff
@@ -5084,189 +5086,139 @@ int riscv_enumerate_triggers(struct target *target)
...
-				sprintf(buf, "x%d", regno - GDB_REGNO_ZERO);
-			else if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095)
-				sprintf(buf, "csr%d", regno - GDB_REGNO_CSR0);
-			else if (regno >= GDB_REGNO_FPR0 && regno <= GDB_REGNO_FPR31)
-				sprintf(buf, "f%d", regno - GDB_REGNO_FPR0);
+	int read_symbols = snprintf(buf->name, size_buf, "%s", name_prefix);
+	assert(read_symbols > 0 && "register name is too long");
+}
+
+static void initialize_csr_name_with_prefixed_name(struct csr_name *buf, const char *name_prefix,
```
timsifive: Not just CSR

src/target/riscv/riscv.c
```diff
@@ -5084,189 +5086,139 @@ int riscv_enumerate_triggers(struct target *target)
...
+	unsigned int num_in_name)
+{
+	const unsigned int size_buf = ARRAY_SIZE(buf->name);
+
+	int read_symbols = snprintf(buf->name, size_buf, "%s%d", name_prefix, num_in_name);
+	assert(read_symbols > 0 && "register name is too long");
+}
+
+const char *gdb_regno_name(struct target *target, enum gdb_regno regno)
```
timsifive: I don't see where target is used in this function.

src/target/riscv/riscv.c
```diff
@@ -5382,32 +5359,27 @@ int riscv_init_registers(struct target *target)
 	riscv_free_registers(target);
 
 	target->reg_cache = calloc(1, sizeof(*target->reg_cache));
-	if (!target->reg_cache)
+	if (!target->reg_cache) {
+		LOG_ERROR("Failed to allocate memory for target->reg_cache");
```
```suggestion
		LOG_TARGET_ERROR(target, "Failed to allocate memory for target->reg_cache");
```

src/target/riscv/riscv.c
```diff
@@ -5382,32 +5359,27 @@ int riscv_init_registers(struct target *target)
...
+	}
 	target->reg_cache->name = "RISC-V Registers";
-	target->reg_cache->num_regs = GDB_REGNO_COUNT;
 
-	if (!list_empty(&info->expose_custom)) {
-		range_list_t *entry;
-		list_for_each_entry(entry, &info->expose_custom, list)
-			target->reg_cache->num_regs += entry->high - entry->low + 1;
+	if (initialize_custom_register_names(&info->expose_custom, &info->custom_register_names) != ERROR_OK) {
+		LOG_ERROR("initialize_custom_register_names failed");
```
```suggestion
		LOG_TARGET_ERROR(target, "initialize_custom_register_names failed");
```

timsifive (changes requested):
Overall concept looks good. I need to come back to this for a more detailed review.

src/target/riscv/riscv.c
AnastasiyaChernikova: To set custom registers. We get info from target. (riscv.c:5123)

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed

src/target/riscv/riscv.c
```diff
@@ -5084,189 +5086,139 @@ int riscv_enumerate_triggers(struct target *target)
...
+
+	int read_symbols = snprintf(buf->name, size_buf, "%s%d", name_prefix, num_in_name);
+	assert(read_symbols > 0 && "register name is too long");
+}
+
+const char *gdb_regno_name(struct target *target, enum gdb_regno regno)
+{
+	if (regno >= GDB_REGNO_COUNT) {
+		RISCV_INFO(info);
+		assert(regno - GDB_REGNO_COUNT <= info->custom_register_names.names_num);
```
aap-sc: assert that `info->custom_registers_names.reg_names` is not zero

AnastasiyaChernikova: Addressed

src/target/riscv/riscv.c
```diff
@@ -5084,189 +5086,140 @@ int riscv_enumerate_triggers(struct target *target)
...
+struct csr_info {
+	unsigned int number;
+	const char *name;
+};
+
+struct csr_info_table {
+	unsigned int records_num;
+	const struct csr_info *csr_records;
+};
+
```
timsifive: Are these used anywhere?

src/target/riscv/riscv.c
timsifive: Thanks.

src/target/riscv/riscv.c
```diff
@@ -5084,189 +5086,140 @@ int riscv_enumerate_triggers(struct target *target)
...
-			return "v31";
-		default:
-			if (regno <= GDB_REGNO_XPR31)
-				sprintf(buf, "x%d", regno - GDB_REGNO_ZERO);
-			else if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095)
-				sprintf(buf, "csr%d", regno - GDB_REGNO_CSR0);
-			else if (regno >= GDB_REGNO_FPR0 && regno <= GDB_REGNO_FPR31)
-				sprintf(buf, "f%d", regno - GDB_REGNO_FPR0);
+	int read_symbols = snprintf(buf->name, size_buf, "%s", name_prefix);
+	assert(read_symbols > 0 && "register name is too long");
```
timsifive: This function is weird.

snprintf returns "the number of characters printed". So read_symbols should maybe be called characters_printed, or simply `result`.

Then the assert just makes sure that any characters were printed at all. It has nothing to do with anything being too long.

src/target/riscv/riscv.c
```diff
@@ -5084,189 +5086,140 @@ int riscv_enumerate_triggers(struct target *target)
...
+}
+
+static void initialize_reg_name_with_prefixed_name(struct reg_name *buf, const char *name_prefix,
```
timsifive: Same here. Just get rid of this function and the `initialize_reg_name`.

src/target/riscv/riscv.c
```diff
@@ -5084,189 +5086,140 @@ int riscv_enumerate_triggers(struct target *target)
...
+	if (!initialized) {
+		for (unsigned int i = 0; i < GDB_REGNO_COUNT; ++i) {
+			if (buf[i].name[0])
+				continue;
+			else if (i <= GDB_REGNO_XPR31)
+				initialize_reg_name_with_prefixed_name(&buf[i], "x", i - GDB_REGNO_ZERO);
+			else if (i <= GDB_REGNO_V31 && i >= GDB_REGNO_V0)
+				initialize_reg_name_with_prefixed_name(&buf[i], "v", i - GDB_REGNO_V0);
+			else if (i >= GDB_REGNO_CSR0 && i <= GDB_REGNO_CSR4095)
+				initialize_reg_name_with_prefixed_name(&buf[i], "csr", i - GDB_REGNO_CSR0);
```
timsifive: Replace with the appropriate snprintf(). Then at the end of the for() block check the return value and assert it's less than the max reg name size minus one. (snprintf does not count the null byte in its return value)

src/target/riscv/riscv.c
```diff
@@ -5365,14 +5318,39 @@ static struct reg_arch_type riscv_reg_arch_type = {
...
+	struct reg_name *additional_regs = custom_register_names->reg_names;
+	range_list_t *range;
+	unsigned int custom_registers_index = 0;
+	list_for_each_entry(range, expose_custom, list) {
+		for (unsigned int custom_number = range->low; custom_number <= range->high; ++custom_number) {
+			if (range->name)
+				initialize_reg_name(&additional_regs[custom_registers_index], range->name);
+			else
+				initialize_reg_name_with_prefixed_name(&additional_regs[custom_registers_index],
+						"custom", custom_number);
```
timsifive: Similar here.

src/target/riscv/riscv.c
```diff
@@ -5365,14 +5318,39 @@ static struct reg_arch_type riscv_reg_arch_type = {
...
+		range_list_t *entry;
+		list_for_each_entry(entry, expose_custom, list)
+			custom_regs_num += entry->high - entry->low + 1;
+	}
+
+	if (!custom_regs_num)
+		return ERROR_OK;
+
+	custom_register_names->reg_names = calloc(custom_regs_num, sizeof(struct reg_name));
+	custom_register_names->names_num = custom_regs_num;
```
timsifive: Move this after the reg_names check, so if the error happens we don't falsely lately believe names_num>0

src/target/riscv/riscv.c
```diff
@@ -5365,14 +5318,39 @@ static struct reg_arch_type riscv_reg_arch_type = {
...
+		LOG_ERROR("Failed to allocate memory for custom_register_names->reg_names");
+		return ERROR_FAIL;
+	}
+	struct reg_name *additional_regs = custom_register_names->reg_names;
+	range_list_t *range;
+	unsigned int custom_registers_index = 0;
+	list_for_each_entry(range, expose_custom, list) {
+		for (unsigned int custom_number = range->low; custom_number <= range->high; ++custom_number) {
+			if (range->name)
+				initialize_reg_name(&additional_regs[custom_registers_index], range->name);
```
timsifive: It would be better to point directly at range->name. That way you're not limiting the user to any register length.

It might be simpler to just malloc() each register. A bit wasteful, but less wasteful than statically allocating 32 bytes, and then changing the pointer is trivial. Needs slightly more free() code, but should also be straightforward.

src/target/riscv/riscv.c
```diff
@@ -36,6 +36,7 @@
 #define DTMCONTROL_VERSION			(0xf)
 
 #define DBUS						0x11
+#define REG_NAME_MAX_LENGTH			32u
```
timsifive: Play with this value to see what the smallest we can get away with. Even 8 might be big enough, which would prevent a fair amount of memory waste.

timsifive (changes requested):

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Changed the assert call.

src/target/riscv/riscv.c
AnastasiyaChernikova: Changed the assert call. This function was written to make calling snprintf safe. It seems to me that it is better not to remove them for possible future use.

src/target/riscv/riscv.c
AnastasiyaChernikova: Сhanged the assert call in the used functions.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: In this case, the memory consumption is about 128 kilobytes, which is very small for current devices. I tried to keep the old behavior and not add dynamic memory, so as not to create potential memory bugs. If you insist I can change it.

timsifive: I really do prefer one malloc() per register. It will remove a limit on the register size a user can type, and the code complexity won't change much. I also think that version of the code will be easier to upstream.

src/target/riscv/riscv.c
```diff
@@ -5084,189 +5086,140 @@ int riscv_enumerate_triggers(struct target *target)
...
+const char *gdb_regno_name(struct target *target, enum gdb_regno regno)
+{
+	if (regno >= GDB_REGNO_COUNT) {
+		RISCV_INFO(info);
+		assert(info->custom_register_names.reg_names);
+		assert(regno - GDB_REGNO_COUNT <= info->custom_register_names.names_num);
+		return info->custom_register_names.reg_names[regno - GDB_REGNO_COUNT].name;
+	}
+
+	static struct reg_name buf[GDB_REGNO_COUNT] = {
```
JanMatCodasip: Please, do not use static local variables (whose lifetime and point of initialization does not match their scope).

Instead, please move this buffer to `struct riscv_info`. 

Initialize it fully in `riscv_init_target()`, and de-initialize it fully in `riscv_deinit_target()`. 

This way the lifetime of the data becomes well defined:
- data created at the moment of target initialization (which is a clear point in the OpenOCD life cycle; as opposed to first call to gdb_regno_get(), which we're not sure when it exactly takes place)
- then valid all the way until target de-initialization (again, that is another clear point in OpenOCD life cycle)

Thanks.

JanMatCodasip (changes requested):
Overall, I like the idea of having a single list of all register names, used everywhere (single source of truth).

src/target/riscv/riscv.c
AnastasiyaChernikova: In this case, local variables are used so as not to allocate memory for each target, since the data is the same.

timsifive: static makes sense to me. Really this is data that could be hard-coded in the source, but instead it's computed once during program execution.

JanMatCodasip: > AnastasiyaChernikova: [...] for each target, since the data is the same.

> timsifive: [...] Really this is data that could be hard-coded in the source, but instead it's computed once during program execution.

Thank you for explanation. 

In that case, please use a global const array, initialized with the constant data. The array can even go to a separate `.h / .c` files if you wish (e.g. `riscv_reg_names.h`, `riscv_reg_names.c`). 

That way:

- the register names will be available to any user, if needed (not just `gdb_regno_name()`);
- the register names will be constant (immutable);
- it will be more clear to the reader that these are fixed constants (known at compile-time, same as GDB_REGNO_* numbers), not something that needs to be dynamically computed;
- the somewhat weird pattern of using static variables inside functions (whose lifetime and initialization does not correspond to their scope, as already noted above) is avoided
- the fact that the first initialization happens and not-very-well defined point in time (that is, the first call to gdb_regno_get()) is avoided, too

src/target/riscv/riscv.h
```diff
@@ -111,6 +111,13 @@ typedef struct {
 	char *name;
 } range_list_t;
 
+struct reg_name;
+
+struct reg_name_table {
+	unsigned int names_num;
+	struct reg_name *reg_names;
+};
+
```
JanMatCodasip, line 120: For linked list, please consider using the already existing resources from `list.h`.

It would be slightly more code, but there are benefits to it:

- The code would be unified with the rest of OpenOCD.
- Higher chance of being accepted in upstream as is.

src/target/riscv/riscv.h
```diff
@@ -111,6 +111,13 @@ typedef struct {
 	char *name;
 } range_list_t;
 
+struct reg_name;
+
+struct reg_name_table {
+	unsigned int names_num;
```
JanMatCodasip: For clarity, this could be called e.g.:

```suggestion
	unsigned int num_entries;
```

src/target/riscv/riscv.c
```diff
@@ -5698,119 +5565,20 @@ int riscv_init_registers(struct target *target)
...
+		} else if (reg_num >= GDB_REGNO_CSR0 && reg_num <= GDB_REGNO_CSR4095) {
 			r->group = "csr";
 			r->feature = &feature_csr;
-			unsigned csr_number = number - GDB_REGNO_CSR0;
+			unsigned int csr_number = reg_num - GDB_REGNO_CSR0;
+			static const bool is_csr_in_buf[GDB_REGNO_CSR4095 - GDB_REGNO_CSR0 + 1] = {
+				#define DECLARE_CSR(csr_name, number)[number] = true,
+				#include "encoding.h"
+				#undef DECLARE_CSR
+			};
```
JanMatCodasip: This is actually very similar case to the one already discussed in https://github.com/riscv/riscv-openocd/pull/896/files/85cde238b39f38707f08e9ebd4077faf709ecad2#r1322490745: 

I'd recommend to 
- make the array global (not static local)
- preferably move it to the same place where the global array of register names will be declared
- and give it a better name: `is_known_standard_csr`, `is_known_gdb_csr` or similar

If you do not wish to do that, alternatively, this could be encapsulated into a small function, called `is_known_standard_csr()` or similar. 

But the first mentioned solution -- having both the register name array and the is_known_standard_csr array defined as const globals in the same place -- is in my opinion preferred.

src/target/riscv/riscv.c
```diff
@@ -5698,119 +5565,20 @@ int riscv_init_registers(struct target *target)
...
-					break;
-			}
 			r->group = "float";
 			r->feature = &feature_fpu;
-		} else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095) {
+		} else if (reg_num >= GDB_REGNO_CSR0 && reg_num <= GDB_REGNO_CSR4095) {
 			r->group = "csr";
 			r->feature = &feature_csr;
-			unsigned csr_number = number - GDB_REGNO_CSR0;
+			unsigned int csr_number = reg_num - GDB_REGNO_CSR0;
```
JanMatCodasip: This can be const. Helps to improve readability.

Also please prefer consistent naming -  `csr_num` instead of `csr_number` (for consistency with `reg_num`).

```suggestion
			const unsigned int csr_num = reg_num - GDB_REGNO_CSR0;
```

src/target/riscv/riscv.c
```diff
@@ -6031,45 +5798,44 @@ int riscv_init_registers(struct target *target)
...
 			r->size = info->vlenb * 8;
-			sprintf(reg_name, "v%d", number - GDB_REGNO_V0);
 			r->group = "vector";
 			r->feature = &feature_vector;
 			r->reg_data_type = &info->type_vector;
 
-		} else if (number >= GDB_REGNO_COUNT) {
+		} else if (reg_num >= GDB_REGNO_COUNT) {
 			/* Custom registers. */
+			struct reg_name *additional_regs = info->custom_register_names.reg_names;
```
JanMatCodasip: For readability, please move the definition of `additional_regs` right before the only place where it is used -- before the line 5836.

src/target/riscv/riscv.c
```diff
@@ -6031,45 +5798,44 @@ int riscv_init_registers(struct target *target)
...
+			const unsigned int custom_register_index = reg_num - GDB_REGNO_COUNT;
+
 			assert(!list_empty(&info->expose_custom));
+			assert(custom_register_index < info->custom_register_names.names_num);
 
 			range_list_t *range = list_first_entry(&info->expose_custom, range_list_t, list);
 
-			unsigned custom_number = range->low + custom_within_range;
+			unsigned int custom_number = range->low + custom_within_range;
```
JanMatCodasip: This can also be const.

```suggestion
			const unsigned int custom_number = range->low + custom_within_range;
```

src/target/riscv/riscv.c
```diff
@@ -6031,45 +5798,44 @@ int riscv_init_registers(struct target *target)
...
-			sprintf(reg_name, "v%d", number - GDB_REGNO_V0);
 			r->group = "vector";
 			r->feature = &feature_vector;
 			r->reg_data_type = &info->type_vector;
 
-		} else if (number >= GDB_REGNO_COUNT) {
+		} else if (reg_num >= GDB_REGNO_COUNT) {
 			/* Custom registers. */
+			struct reg_name *additional_regs = info->custom_register_names.reg_names;
+			const unsigned int custom_register_index = reg_num - GDB_REGNO_COUNT;
```
JanMatCodasip: It's better to use name `custom_reg_index` -- for consistency with `reg_num`.

```suggestion
			const unsigned int custom_reg_index = reg_num - GDB_REGNO_COUNT;
```

src/target/riscv/riscv.c
JanMatCodasip: This is related to the original code (not to the changes in this merge request):

@timsifive Can you please help me check the computation of `custom_number`. This line does not look right. The use of always the first range (list_first_entry) seems incorrect.

JanMatCodasip (dismissed):

src/target/riscv/riscv.c
timsifive: I agree this looks wrong. The code should walk the linked list to find the range that actually includes reg_num, and use that base.

But this is a pre-existing bug, so doesn't have to be fixed in this PR (mostly to keep the scope limited). I filed #914 to track it.

JanMatCodasip: Thank you for looking at it and recording the issue.

Marking this thread as resolved.

en-sc: This is not a bug. Please, take a look at [the call  to `list_rotate_left()` on line 5843](
https://github.com/riscv/riscv-openocd/pull/896/files/85cde238b39f38707f08e9ebd4077faf709ecad2#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR5843).

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.h
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: I would like to avoid creating such a file, since it will be hard to maintain. Please, take a look at the new approach.

timsifive:
Note that none of the automated checks pass.

src/target/riscv/riscv.c
```diff
@@ -5207,188 +5221,154 @@ int riscv_enumerate_triggers(struct target *target)
...
-			else if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095)
-				sprintf(buf, "csr%d", regno - GDB_REGNO_CSR0);
-			else if (regno >= GDB_REGNO_FPR0 && regno <= GDB_REGNO_FPR31)
-				sprintf(buf, "f%d", regno - GDB_REGNO_FPR0);
-			else
-				sprintf(buf, "gdb_regno_%d", regno);
-			return buf;
+	char * const buf = calloc(size_buf, sizeof(char));
+	if (!buf) {
+		LOG_ERROR("Failed to allocate memory for a regester name.");
```
JanMatCodasip: Typo:

```suggestion
		LOG_ERROR("Failed to allocate memory for a register name.");
```

src/target/riscv/riscv.c
```diff
@@ -5207,188 +5221,154 @@ int riscv_enumerate_triggers(struct target *target)
...
+		
+	if (info->reg_names[regno])
+		return info->reg_names[regno];
+	if (regno <= GDB_REGNO_XPR31)
+		return info->reg_names[regno] = initialize_reg_name_with_prefixed_name("x", regno - GDB_REGNO_ZERO);
+	if (regno <= GDB_REGNO_V31 && regno >= GDB_REGNO_V0)
+		return info->reg_names[regno] = initialize_reg_name_with_prefixed_name("v", regno - GDB_REGNO_V0);
+	if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095)
+		return info->reg_names[regno] = initialize_reg_name_with_prefixed_name("csr", regno - GDB_REGNO_CSR0);
+	assert(!"The table must already be filled out");
```
JanMatCodasip: This could be:

```suggestion
	assert(!"Encountered uninitialized entry in reg_names table");
```

src/target/riscv/riscv.c
```diff
@@ -5207,188 +5221,154 @@ int riscv_enumerate_triggers(struct target *target)
...
+	}
+		
+	if (info->reg_names[regno])
+		return info->reg_names[regno];
+	if (regno <= GDB_REGNO_XPR31)
+		return info->reg_names[regno] = initialize_reg_name_with_prefixed_name("x", regno - GDB_REGNO_ZERO);
+	if (regno <= GDB_REGNO_V31 && regno >= GDB_REGNO_V0)
+		return info->reg_names[regno] = initialize_reg_name_with_prefixed_name("v", regno - GDB_REGNO_V0);
+	if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095)
+		return info->reg_names[regno] = initialize_reg_name_with_prefixed_name("csr", regno - GDB_REGNO_CSR0);
```
JanMatCodasip: If the user exposes its own named CSR via command `riscv expose_csrs <number>=<name>`, it seems to me that the user-provided name will not be reflected in the reg_names table. Or did I just miss that logic somewhere?

AnastasiyaChernikova: The name of the registers entered by the user sets above (lines 5348-5352)
```
        if (regno >= GDB_REGNO_COUNT) {
                assert(info->custom_register_names.reg_names);
                assert(regno - GDB_REGNO_COUNT <= info->custom_register_names.num_entries);
                return info->custom_register_names.reg_names[regno - GDB_REGNO_COUNT];
        }
```

AnastasiyaChernikova: Corrected.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
```diff
@@ -5207,188 +5223,175 @@ int riscv_enumerate_triggers(struct target *target)
...
+			free((void *)info->reg_names[i]);
+		} else {
+			if (!default_reg_names[i])
+				free((void *)info->reg_names[i]);
+		}
+	}
+	free(info->reg_names);
+	info->reg_names = NULL;
+
+
```
JanMatCodasip: Detail: An extra blank line

```suggestion
```

src/target/riscv/riscv.c
```diff
@@ -482,6 +482,21 @@ static void riscv_free_registers(struct target *target)
...
+static void free_custom_register_names(struct target *target)
+{
+	RISCV_INFO(info);
+
+	if (!info->custom_register_names.reg_names)
+		return;
+
+	for (unsigned int i = 0; i < info->custom_register_names.num_entries; i++)
+		free((void *)info->custom_register_names.reg_names[i]);
```
JanMatCodasip: Curious - is the typecast to `void *` need here? 

If no tool asks for it (compilers, linters, etc.) I recommend to drop it.

```suggestion
		free(info->custom_register_names.reg_names[i]);
```

src/target/riscv/riscv.c
```diff
@@ -5207,188 +5223,175 @@ int riscv_enumerate_triggers(struct target *target)
...
+	if (!buf) {
+		LOG_ERROR("Failed to allocate memory for a register name.");
+		return NULL;
 	}
+	strcpy(buf, name);
+	return buf;
 }
 
+static const char *initialize_reg_name_with_prefixed_name(const char *name_prefix,
+	unsigned int num_in_name)
```
JanMatCodasip: I believe you can call this parameter just `num` without any loss of clarity.

```suggestion
	unsigned int num)
```

src/target/riscv/riscv.c
```diff
@@ -5207,188 +5223,175 @@ int riscv_enumerate_triggers(struct target *target)
...
+	char * const buf = calloc(size_buf, sizeof(char));
+	if (!buf) {
+		LOG_ERROR("Failed to allocate memory for a register name.");
+		return NULL;
 	}
+	strcpy(buf, name);
+	return buf;
 }
 
+static const char *initialize_reg_name_with_prefixed_name(const char *name_prefix,
```
JanMatCodasip: Unless anyone objects, I recommend to shorten this name to:

```suggestion
static const char *initialize_reg_name_prefixed(const char *name_prefix,
```

or 

```suggestion
static const char *initialize_reg_name_with_prefix(const char *name_prefix,
```

src/target/riscv/riscv.c
```diff
@@ -5207,188 +5223,175 @@ int riscv_enumerate_triggers(struct target *target)
...
+		}
+	}
+	free(info->reg_names);
+	info->reg_names = NULL;
+
+
+	free_custom_register_names(target);
+}
+
+static void init_user_specified_csr_names(struct target *target)
```

```suggestion
static void init_custom_csr_names(struct target *target)
```

Also some of the functions are named `initialize_*` whereas this one is `init_*`. I'd recommend to unify the names.

src/target/riscv/riscv.c
JanMatCodasip: Unless there are objections, I'd still recommend to put this into a separate function, named e.g. `is_known_standard_csr()`.

src/target/riscv/riscv.c
```diff
@@ -5207,188 +5223,175 @@ int riscv_enumerate_triggers(struct target *target)
...
+	if (!info->reg_names)
+		return;
+
+	for (unsigned int i = 0; i < GDB_REGNO_COUNT; ++i) {
+		if (i > ARRAY_SIZE(default_reg_names)) {
+			free((void *)info->reg_names[i]);
+		} else {
+			if (!default_reg_names[i])
+				free((void *)info->reg_names[i]);
+		}
```
JanMatCodasip: I am not happy that 
- some items in reg_names are **a)** pointers to constant data (no free() is needed)
- whereas others are **b)** pointers to dynamically allocated strings that need to be free'd.

I would like everything in one array to be either a) or b), not a mix.

I can think of these options:

Option A:

- Allocate everything dynamically. This increases memory consumption but I still prefer simplicity and consistency of the implementation over (insignificant?) memory saving.

Option B:

- Split reg_names to multiple arrays, some containining a), others b).

@timsifive What is your opinion on this, please?

src/target/riscv/riscv.c
```diff
@@ -5488,14 +5491,42 @@ static struct reg_arch_type riscv_reg_arch_type = {
...
+	if (!custom_regs_num)
+		return ERROR_OK;
+
+	custom_register_names->reg_names = calloc(custom_regs_num, sizeof(char *));
+	if (!custom_register_names->reg_names) {
+		LOG_ERROR("Failed to allocate memory for custom_register_names->reg_names");
+		return ERROR_FAIL;
+	}
+	custom_register_names->num_entries = custom_regs_num;
+	const char **additional_regs = custom_register_names->reg_names;
```
JanMatCodasip: Is this variable just a shortcut to avoid typing the whole `custom_register_names->reg_names`? 

In that case you can use just:

```suggestion
	const char **reg_names = custom_register_names->reg_names;
```

In another words, please do not introduce new names for the same thing, if not needed (it complicates the life of people reading the code).

src/target/riscv/riscv.c
```diff
@@ -5488,14 +5491,42 @@ static struct reg_arch_type riscv_reg_arch_type = {
...
+	range_list_t *range;
+	unsigned int custom_registers_index = 0;
```
JanMatCodasip: This can probably be called:

```suggestion
	unsigned int next_custom_reg_index = 0;
```

JanMatCodasip (commented):
Overall, this looks fine to me.

I have :

- one suggestion that I consider important (riscv.c, line 5341)
- and a bunch of cosmetic details.

src/target/riscv/riscv.c
JanMatCodasip: Marking a resolved - outdated thread.

src/target/riscv/riscv.c
AnastasiyaChernikova: Without the cast any compiler will issue a warning about implicit cast from a const pointer to a non-const one.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

AnastasiyaChernikova:
@timsifive  Hi! Could you take a looks once again? Looks like all comments are addressed

src/target/riscv/riscv.h
```diff
@@ -113,6 +113,11 @@ typedef struct {
 
 #define DTM_DTMCS_VERSION_UNKNOWN ((unsigned int)-1)
 
+struct reg_name_table {
+	unsigned int num_entries;
+	const char **reg_names;
```
JanMatCodasip: Can you please check if this should be without the `const`?

src/target/riscv/riscv.h
```diff
@@ -121,9 +126,8 @@ struct riscv_info {
 	struct command_context *cmd_ctx;
 	void *version_specific;
 
-	/* Single buffer that contains all register names, instead of calling
-	 * malloc for each register. Needs to be freed when reg_list is freed. */
-	char *reg_names;
+	struct reg_name_table custom_register_names;
+	const char **reg_names;
```
JanMatCodasip: Same as in the previous case - after your latest changes, now that all reg_names are dynamically allocated, shouldn't this be without `const`?

```suggestion
	char **reg_names;
```

src/target/riscv/riscv.c
JanMatCodasip: See my two preceding comments. I believe the pointers should not be declared as const. Therefore the extra `(void *)` cast would not need to be here.

src/target/riscv/riscv.c
```diff
@@ -5381,188 +5397,169 @@ int riscv_enumerate_triggers(struct target *target)
...
+
+static void free_reg_names(struct target *target)
+{
+	RISCV_INFO(info);
+
+	if (!info->reg_names)
+		return;
+
+	for (unsigned int i = 0; i < GDB_REGNO_COUNT; ++i)
+		free((void *)info->reg_names[i]);
```
JanMatCodasip: Same as the previous -- I believe the pointer should not be declared as `const`, and the `(void *)` cast here should be removed.

```suggestion
		free(info->reg_names[i]);
```

src/target/riscv/riscv.h
en-sc: The elements of the array are allocated and filled dynamically, but not changed throughout the code. There is no point in changing them. I guess this is the intention behind making it an array of constant strings, which seems perfectly fine to me.

JanMatCodasip: > The elements of the array are allocated and filled dynamically

For this exact reason, I strongly believe this should not be const. Also the need to cast away the const-ness in free() to avoid compiler warnings highlights it.

src/target/riscv/riscv.c
JanMatCodasip: @AnastasiyaChernikova Can you please check the function naming. There are still inconsistent prefixes: `init_` vs. `initialize_`. Thanks.

JanMatCodasip:
I have two last minor things:

- Several related comments about the use of `const` for non-const data.
- One remaining cosmetic comment about function naming: https://github.com/riscv/riscv-openocd/pull/896#discussion_r1374542383

Otherwise the code looks good to me. I don't expect to have any further comments. Thank you.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.h
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.h
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

AnastasiyaChernikova:
@timsifive Can you please take a look? Everything should be fixed.

src/target/riscv/riscv.c
```diff
@@ -5381,188 +5397,169 @@ int riscv_enumerate_triggers(struct target *target)
...
+		return info->reg_names[regno];
+	if (default_reg_names[regno])
+		return default_reg_names[regno];
+	if (regno <= GDB_REGNO_XPR31)
+		return info->reg_names[regno] = init_reg_name_with_prefix("x", regno - GDB_REGNO_ZERO);
+	if (regno <= GDB_REGNO_V31 && regno >= GDB_REGNO_V0)
+		return info->reg_names[regno] = init_reg_name_with_prefix("v", regno - GDB_REGNO_V0);
+	if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095) {
+		init_custom_csr_names(target);
+		return info->reg_names[regno] = init_reg_name_with_prefix("csr", regno - GDB_REGNO_CSR0);
```
JanMatCodasip: Cosmetic - coding style:

Please, separate the assignment and the return to two different statements. It is more lines of code, but easier to read and less error prone. 


```suggestion
	if (regno <= GDB_REGNO_XPR31) {
		info->reg_names[regno] = init_reg_name_with_prefix("x", regno - GDB_REGNO_ZERO);
		return info->reg_names[regno];
	}
	if (regno <= GDB_REGNO_V31 && regno >= GDB_REGNO_V0) {
		info->reg_names[regno] = init_reg_name_with_prefix("v", regno - GDB_REGNO_V0);
		return info->reg_names[regno];
	}
	if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095) {
		init_custom_csr_names(target);
		info->reg_names[regno] = init_reg_name_with_prefix("csr", regno - GDB_REGNO_CSR0);
		return info->reg_names[regno];
```

JanMatCodasip (dismissed):
The change looks good to me. I have one last cosmetic suggestion.

@AnastasiyaChernikova Thank you.

src/target/riscv/riscv.c
```diff
@@ -5381,188 +5397,169 @@ int riscv_enumerate_triggers(struct target *target)
...
+		assert(entry->low == entry->high);
+		const unsigned int regno = entry->low + GDB_REGNO_CSR0;
+		assert(regno <= GDB_REGNO_CSR4095);
+		if (info->reg_names[regno])
+			return;
+		info->reg_names[regno] = init_reg_name(entry->name);
+	}
+}
+
+const char *gdb_regno_name(struct target *target, unsigned int regno)
```
timsifive, line 5534: I think this goes back to previous feedback from @JanMatCodasip, but why do you need a target pointer here? It makes calling the function larger, makes this diff huge, and every target has the exact same register names.

It looks like you're generating the exact same names for each target?

JanMatCodasip: Custom CSR names (`riscv expose_csrs`) and full-custom register names (`riscv expose_custom`) may differ per target. So the `target` parameter makes sense to me here.

IIUC, we could get away without the `target`, but at the cost of losing the per-target custom register/CSR names.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

JanMatCodasip (approved):

src/target/riscv/riscv.c
timsifive: Good point about the custom register names. Let's keep the target parameter.

timsifive (approved):

PR #898, opened 2023-08-01
target/riscv: prevent the fence.i from triggering the exception

wxjstz:
On platforms that do not support Zifencei, executing fence.i will trigger an exception, this patch eliminates the impact of the exception

Change-Id: I459cff096374064fdb6e18a9d95170510ae1f32e


wxjstz:
please review it @timsifive @TommyMurphyTM1234

wxjstz:
The patch requires progbufsize > 1, is this a issue?

src/target/riscv/riscv-013.c
```diff
@@ -2837,16 +2837,44 @@ static int deassert_reset(struct target *target)
...
+
+	/* backup mtvec */
+	result = riscv_get_register(target, &saved_mtvec, GDB_REGNO_MTVEC);
+	if (result != ERROR_OK)
+		return result;
+
+	/* set mtvec after fence.i */
+	result = riscv_set_register(target, GDB_REGNO_MTVEC, info->progbuf_address + 4);
+	if (result != ERROR_OK)
+		return result;
```
MarekVCodasip, line 2859: This isn't great.
This basically loads an address as an instruction into the program buffer (if it is even memory mapped). It is then rewritten when the fence_i is written into the program buffer and executed.

MarekVCodasip (changes requested):
Thanks for the patch. I am afraid this is not a proper solution. The exceptions don't trigger in program buffer. There is therefore no reason to change the mtvec. Also, it is not mandatory to have program buffer memory mapped, therefore info->program_address does not have to be a meaningful value.

The exception caused by unsupported fence.i should not have any adverse effect. So there should be no need to handle it in a special way.

I do agree that with the separation of the fences into two program buffer executions. This will make it compatible with targets where `progbufsize == 1`

wxjstz:
> The exceptions don't trigger in program buffer

why?  Can you tell me which spce this information comes from？

MarekVCodasip:
> why? Can you tell me which spce this information comes from？

Alright. It is a bit more general. They don't trigger in debug mode in general.

https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-stable.pdf its currently on page 50 in section 4.1

wxjstz:
> > why? Can you tell me which spce this information comes from？
> 
> Alright. It is a bit more general. They don't trigger in debug mode in general.
> 
> https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-stable.pdf its currently on page 50 in section 4.1

Thank you!

MarekVCodasip:
Don't close this, separating the two fences into two separate program buffer executions is still good and useful as not all targets implementing the program buffer have to have a progbufsize > 1

PR #899, opened 2023-08-03
target/riscv: improve error handling in trigger setup

en-sc:
Change-Id: I235973a3c44fb3d934925c74ffee47f8bd96de0d

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -897,7 +897,7 @@ static int maybe_add_trigger_t2_t6(struct target *target,
 	};
 	ret = try_setup_single_match_trigger(target, trigger, eq);
 	if (ret != ERROR_OK)
-		return ERROR_FAIL;
+		return ret;
 
 	if (trigger->length > 1) {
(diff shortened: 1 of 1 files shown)


en-sc:
@erhankur, please, take a look.

erhankur:
@en-sc Thanks for the fix. Looks good now.

erhankur (approved):

src/target/riscv/riscv.c
```diff
@@ -1042,7 +1042,9 @@ static int add_trigger(struct target *target, struct trigger *trigger)
 			break;
 	} while (0);
 
-	riscv_set_register(target, GDB_REGNO_TSELECT, tselect);
+	if (riscv_set_register(target, GDB_REGNO_TSELECT, tselect) != ERROR_OK &&
+			ret == ERROR_OK)
+		return ERROR_FAIL;
```
MarekVCodasip, line 1047: While making changes here: It may not be clear to the reader why TSELECT must be restored here. I would recommend adding a short comment if you like.

MarekVCodasip (approved):

timsifive (approved):

PR #900, opened 2023-08-04
riscv: simplify state management during examine

aap-sc:
This also fixes a bug when, after `examine` completion, the target still
has  `unknown` status. To reproduce this one spike, it is enough to do
the following:

~~~
// make sure spike harts are halted
openocd ... -c init -c 'echo "[targets]"'
~~~

this behavior is quite dangerous and leads to segfaults in some cases

Change-Id: I13915f7038ad6d0251d56d2d519fbad9a2f13c18

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1870,8 +1870,12 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
 
 static int examine(struct target *target)
 {
-	/* Don't need to select dbus, since the first thing we do is read dtmcontrol. */
+	/* We reset target state in case if something goes wrong during examine:
+	 * DTM/DM scans could fail or hart may fail to halt. */
+	target->state = TARGET_UNKNOWN;
+	target->debug_reason = DBG_REASON_UNDEFINED;
 
+	/* Don't need to select dbus, since the first thing we do is read dtmcontrol. */
 	LOG_TARGET_DEBUG(target, "dbgbase=0x%x", target->dbgbase);
 
 	uint32_t dtmcontrol = dtmcontrol_scan(target, 0);
@@ -2033,34 +2037,21 @@ static int examine(struct target *target)
 	if (dm013_select_hart(target, info->index) != ERROR_OK)
 		return ERROR_FAIL;
 
-	enum riscv_hart_state state;
-	if (riscv_get_hart_state(target, &state) != ERROR_OK)
+	enum riscv_hart_state state_at_examine_start;
+	if (riscv_get_hart_state(target, &state_at_examine_start) != ERROR_OK)
 		return ERROR_FAIL;
-	bool halted = (state == RISCV_STATE_HALTED);
-	if (!halted) {
+	const bool hart_halted_at_examine_start = state_at_examine_start == RISCV_STATE_HALTED;
+	if (!hart_halted_at_examine_start) {
 		r->prepped = true;
 		if (riscv013_halt_go(target) != ERROR_OK) {
 			LOG_TARGET_ERROR(target, "Fatal: Hart %d failed to halt during %s",
 					info->index, __func__);
 			return ERROR_FAIL;
 		}
-		target->state = TARGET_HALTED;
-		target->debug_reason = DBG_REASON_DBGRQ;
 	}
-	/* FIXME: This is needed since register_read_direct relies on target->state
-	 * to work correctly, so, if target->state does not represent current state
-	 * of target, e.g. if a target is halted, but target->state is
-	 * TARGET_UNKNOWN, it can fail early, (e.g. accessing registers via program
-	 * buffer can not be done atomically on a running hart becuse mstatus can't
-	 * be prepared for a register access and then restored)
-	 * See https://github.com/riscv/riscv-openocd/pull/842#discussion_r1179414089
-	 */
-	const enum target_state saved_tgt_state = target->state;
-	const enum target_debug_reason saved_dbg_reason = target->debug_reason;
-	if (target->state != TARGET_HALTED) {
-		target->state = TARGET_HALTED;
-		target->debug_reason = DBG_REASON_DBGRQ;
-	}
+
+	target->state = TARGET_HALTED;
+	target->debug_reason = hart_halted_at_examine_start ? DBG_REASON_UNDEFINED : DBG_REASON_DBGRQ;
 
 	/* Without knowing anything else we can at least mess with the
 	 * program buffer. */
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1972,20 +1975,9 @@ static int examine(struct target *target)
 		target->state = TARGET_HALTED;
 		target->debug_reason = DBG_REASON_DBGRQ;
```
MarekVCodasip, line 1976: I feel like this can be removed now as well.

MarekVCodasip: Seeing as the target-> state is set here and also in riscv013_halt_go and debug_reason later. I would just delete these two lines.

MarekVCodasip (commented):
I think this is an improvement, especially later when the state is restored.

src/target/riscv/riscv-013.c
JanMatCodasip: Agreed, this seems redundant. The change of the recorded target state will be covered by lines 1979 and 1980.

src/target/riscv/riscv-013.c
```diff
@@ -1958,11 +1958,14 @@ static int examine(struct target *target)
...
-	if (riscv_get_hart_state(target, &state) != ERROR_OK)
+	target->state = TARGET_UNKNOWN;
+	target->debug_reason = DBG_REASON_UNDEFINED;
+
+	enum riscv_hart_state state_at_examine_start;
+	if (riscv_get_hart_state(target, &state_at_examine_start) != ERROR_OK)
 		return ERROR_FAIL;
-	bool halted = (state == RISCV_STATE_HALTED);
-	if (!halted) {
+	const bool hart_halted_at_examine_start = state_at_examine_start == RISCV_STATE_HALTED;
```
JanMatCodasip, line 2043: Just a side note: This part of code seems to not handle the case when the hart is unavailable (RISCV_STATE_UNAVAILABLE) - and neither did the the original code. 

No need to solve this in the current merge request. Just a remark for a future improvement.

JanMatCodasip (dismissed):
This looks all right to me.

src/target/riscv/riscv-013.c
```diff
@@ -1958,11 +1958,14 @@ static int examine(struct target *target)
 	if (dm013_select_hart(target, info->index) != ERROR_OK)
 		return ERROR_FAIL;
 
-	enum riscv_hart_state state;
-	if (riscv_get_hart_state(target, &state) != ERROR_OK)
+	target->state = TARGET_UNKNOWN;
+	target->debug_reason = DBG_REASON_UNDEFINED;
```
timsifive: What does this achieve? I don't think it's accessed below, it's just unconditionally overwritten.

src/target/riscv/riscv-013.c
timsifive: Agreed. Remove these 2 lines.

src/target/riscv/riscv-013.c
```diff
@@ -2059,13 +2051,13 @@ static int examine(struct target *target)
...
-	target->debug_reason = saved_dbg_reason;
-
-	if (!halted) {
+	if (state_at_examine_start == RISCV_STATE_RUNNING) {
 		riscv013_step_or_resume_current_hart(target, false);
 		target->state = TARGET_RUNNING;
 		target->debug_reason = DBG_REASON_NOTHALTED;
+	} else if (state_at_examine_start == RISCV_STATE_HALTED) {
+		target->state = TARGET_HALTED;
+		target->debug_reason = DBG_REASON_UNDEFINED;
```
timsifive, line 2134: Why is this necessary? We know that the hart is halted on line 2054, and the state should reflect that.

timsifive (changes requested):

src/target/riscv/riscv-013.c
EccoTheDolphin: > Why is this necessary? We know that the hart is halted on line 2054, and the state should reflect that.

Yes, we know that the hart is halted. And we have `target->state = TARGET_HALTED` to indicate that. However, we don't know **why** hart was halted (since we've just connected) and thus we set the reason to UNDEFINED.

src/target/riscv/riscv-013.c
EccoTheDolphin: `What does this achieve? I don't think it's accessed below, it's just unconditionally overwritten.`

This is for the case when examine fails. Imagine we could not examine/could not finish examination process. This allows us just to reset the state from whatever it was. My understanding is that examine (regardless of the current implementation) can actually be invoked several times (via `arp_examine` for example). See here:

https://sourceforge.net/p/openocd/mailman/message/33948884/

src/target/riscv/riscv-013.c
EccoTheDolphin: Well... actually it does. Kind of. Upon entry to examine routine we unconditionally set:

```
	target->state = TARGET_UNKNOWN;
	target->debug_reason = DBG_REASON_UNDEFINED;
```

And then the assumption is that if the hart is UNAVAILABLE - then all operations should fail (like halt_go).

src/target/riscv/riscv-013.c
EccoTheDolphin: I'll take look. on the first glance, yes it looks like we should remove this.

src/target/riscv/riscv-013.c
EccoTheDolphin: corrected the message. `arp_init` -> `arp_examine`

src/target/riscv/riscv-013.c
JanMatCodasip: Ok, thank you for explanation. (No need to change anything at the moment.)

src/target/riscv/riscv-013.c

JanMatCodasip (dismissed):

src/target/riscv/riscv-013.c
timsifive: OK. Can you add a short comment, like `/* In case e.g. riscv013_halt_go() fails. */`

Generally I would like to see these status updates happen where we know the status changed, but that's a much larger change.

src/target/riscv/riscv-013.c
timsifive: Got it. Thanks.

timsifive (changes requested):
Just a minor change request. Otherwise good to go.

src/target/riscv/riscv-013.c
EccoTheDolphin: > OK. Can you add a short comment, like /* In case e.g. riscv013_halt_go() fails. */

Addressed. Please do note that I moved these to lines way higher. I think this is more appropriate place.

timsifive (approved):

JanMatCodasip (approved):

PR #901, opened 2023-08-06
[riscv] refactor functions that register read/write via progbuf

aap-sc:
The motivation for this refactor is to fixup error handling for some corner cases. These *_progbuf functions unconditionally attempt to cache S0 register and only then perform a bunch of extra checks to figure out if the requested register is a valid one. The problem is that there are few corner cases when _*progbuf functions could receive a GPR as an input. For example, an abstract read could fail (for whatever reason) leading to infinite recursion:

```
save S0 -> read S0 -> save S0 -> read S0 -> ...
```

The case described above could be fixed by adding extra sanitity checks, however I decided to make these functions more modular since I find self-contained functions easier to read.

Change-Id: I01f57bf474ca45ebb67a30cd4d8fdef21f307c7d

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1448,6 +1448,57 @@ static int internal_register_read64_progbuf_scratch(struct target *target,
 	return result;
 }
 
+static int fpr_read_progbuf(struct target *target, uint64_t *value,
+		enum gdb_regno number)
+{
+	assert(target->state == TARGET_HALTED);
+	assert(number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31);
+
+	const unsigned int freg = number - GDB_REGNO_FPR0;
+
+	if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
+		return ERROR_FAIL;
+
+	struct riscv_program program;
+	riscv_program_init(&program, target);
+	if (riscv_supports_extension(target, 'D') && riscv_xlen(target) < 64) {
+		/* There are no instructions to move all the bits from a
+		 * register, so we need to use some scratch RAM.
+		 */
+		if (riscv_program_insert(&program, fsd(freg, S0, 0)) != ERROR_OK)
+			return ERROR_FAIL;
+		return internal_register_read64_progbuf_scratch(target, &program, value);
+	}
+	if (riscv_program_insert(&program,
+				riscv_supports_extension(target, 'D') ?
+				fmv_x_d(S0, freg) : fmv_x_w(S0, freg)) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (riscv_program_exec(&program, target) != ERROR_OK)
+		return ERROR_FAIL;
+
+	return register_read_abstract(target, value, GDB_REGNO_S0) != ERROR_OK;
+}
+
+static int csr_read_progbuf(struct target *target, uint64_t *value,
+		enum gdb_regno number)
+{
+	assert(target->state == TARGET_HALTED);
+	assert(number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095);
+
+	if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
+		return ERROR_FAIL;
+
+	struct riscv_program program;
+	riscv_program_init(&program, target);
+	if (riscv_program_csrr(&program, S0, number) != ERROR_OK)
+		return ERROR_FAIL;
+	if (riscv_program_exec(&program, target) != ERROR_OK)
+		return ERROR_FAIL;
+
+	return register_read_abstract(target, value, GDB_REGNO_S0) != ERROR_OK;
+}
+
 /**
  * This function reads a register by writing a program to program buffer and
  * executing it.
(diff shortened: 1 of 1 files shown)


aap-sc:
The issue mentioned above can be reproduced on spike if #900 is NOT applied (apparently, #900 hides this problem).

Reproduction (I've tried to reduce it and this is what I've got):

1. Target configuration file:

```
proc init_targets {} {
        jtag newtap riscv tap -irlen 5
        target create cpu0.riscv riscv -endian little -chain-position riscv.tap -coreid 0
        target create cpu1.riscv riscv -endian little -chain-position riscv.tap -coreid 1
}
```

2. Test script:

```
proc run_test {} {
  foreach t [target names] {
    targets $t
    halt
  }
  foreach t [target names] {
    targets $t
    resume
  }
  foreach t [target names] {
    targets $t
    halt
  }
  # entry point
  reg pc 0x80000000
  shutdown
}

```

3. Spike command line:

```
# make sure that entry point of the elf is at 0x80000000
spike \
  --rbb-port=33627 \
  --isa=rv64gc \
  --dm-progsize=6 \
  --dm-sba=0 \
  --dm-abstract-rti=0 \
  -p4 \
  -m0x80000000:0xa00000 \
  <some_elf_file>
```

4. OpenOCD invocation:

```
openocd -c "adapter driver remote_bitbang" -c "remote_bitbang port 33271" -c "remote_bitbang host 127.0.0.1" \
  -f config.cfg  -d2 -c init -f run_test.tcl -c run_test
openocd -c "adapter driver remote_bitbang" -c "remote_bitbang port 33271" -c "remote_bitbang host 127.0.0.1" \
  -f config.cfg  -d2 -c init -f run_test.tcl -c run_test
```

During the second invocation OpenOCD will crash due to stack overflow (because of infinite recursion).

aap-sc:
Note: the intention of this patch is only to fix infinite recursion. The reason we have such a situation on spike with no apparent reason is another matter.

src/target/riscv/riscv-013.c
```diff
@@ -1522,73 +1546,124 @@ static int internal_register_write64_progbuf_scratch(struct target *target,
...
+	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31)
+		return register_write_to_fpr_progbuf(target, number, value);
+	else if (number == GDB_REGNO_VTYPE)
+		return register_write_to_vtype_progbuf(target, value);
+	else if (number == GDB_REGNO_VL)
+		return register_write_to_vl_progbuf(target, value);
+	else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095)
+		return register_write_to_csr_progbuf(target, number, value);
+
+	LOG_ERROR("Unsupported register: %s", gdb_regno_name(number));
```
JanMatCodasip: I tried to reword this to make it more understandable for users:

```suggestion
	LOG_TARGET_ERROR(target, "Write to register %s via program buffer not supported.", gdb_regno_name(number));
```

src/target/riscv/riscv-013.c
```diff
@@ -1448,51 +1448,75 @@ static int internal_register_read64_progbuf_scratch(struct target *target,
...
+		enum gdb_regno number)
+{
+	assert(target->state == TARGET_HALTED);
+
+	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31)
+		return register_read_from_fpr_progbuf(target, value, number);
+	else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095)
+		return register_read_from_csr_progbuf(target, value, number);
+
+	LOG_TARGET_ERROR(target, "Unsupported register: %s", gdb_regno_name(number));
```
JanMatCodasip: (For clarity of the message:)

```suggestion
	LOG_TARGET_ERROR(target, "Reading register %s via program buffer is not supported.", gdb_regno_name(number));
```

JanMatCodasip (commented):
I will need to return to this one again for a more thorough review. However, overall, this looks very good.

> These *_progbuf functions unconditionally attempt to cache S0 register and only then perform a bunch of extra checks to figure out if the requested register is a valid one. The problem is that there are few corner cases when _*progbuf functions could receive a GPR as an input.

I fully agree. The `*_progbuf` functions should refuse invalid input argument (in this case, unsupported register type) at the very beginning, before attempting any further actions.

I also support the split of the original `register_{read,write}_progbuf` to the smaller functions.

src/target/riscv/riscv-013.c
EccoTheDolphin: yes, will address. Thanks!

src/target/riscv/riscv-013.c
```diff
@@ -1448,51 +1448,75 @@ static int internal_register_read64_progbuf_scratch(struct target *target,
 	return result;
 }
 
-/**
- * This function reads a register by writing a program to program buffer and
- * executing it.
- */
-static int register_read_progbuf(struct target *target, uint64_t *value,
-		enum gdb_regno number)
+static int register_read_from_fpr_progbuf(struct target *target,
```
```suggestion
static int fpr_read_progbuf(struct target *target,
```

src/target/riscv/riscv-013.c
```diff
@@ -1448,51 +1448,75 @@ static int internal_register_read64_progbuf_scratch(struct target *target,
...
-	if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
+	if (riscv_program_exec(&program, target) != ERROR_OK)
 		return ERROR_FAIL;
 
-	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31) {
-		const unsigned int freg = number - GDB_REGNO_FPR0;
+	return register_read_abstract(target, value, GDB_REGNO_S0) != ERROR_OK;
+}
+
+static int register_read_from_csr_progbuf(struct target *target,
```
```suggestion
static int csr_read_progbuf(struct target *target,
```

src/target/riscv/riscv-013.c
```diff
@@ -1522,73 +1546,124 @@ static int internal_register_write64_progbuf_scratch(struct target *target,
 	return result;
 }
 
-/**
- * This function writes a register by writing a program to program buffer and
- * executing it.
- */
-static int register_write_progbuf(struct target *target, enum gdb_regno number,
-		riscv_reg_t value)
+static int register_write_to_fpr_progbuf(struct target *target,
```
```suggestion
static int fpr_write_progbuf(struct target *target,
```

src/target/riscv/riscv-013.c
```diff
@@ -1522,73 +1546,124 @@ static int internal_register_write64_progbuf_scratch(struct target *target,
...
-			return ERROR_FAIL;
-	} else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095) {
-		if (riscv_program_csrw(&program, S0, number) != ERROR_OK)
-			return ERROR_FAIL;
-	} else {
-		LOG_TARGET_ERROR(target, "Unsupported register (enum gdb_regno)(%d)", number);
+	return riscv_program_exec(&program, target);
+}
+
+static int register_write_to_vtype_progbuf(struct target *target,
```
```suggestion
static int vtype_write_progbuf(struct target *target,
```

src/target/riscv/riscv-013.c
```diff
@@ -1522,73 +1546,124 @@ static int internal_register_write64_progbuf_scratch(struct target *target,
...
+	riscv_program_init(&program, target);
+	if (riscv_program_insert(&program, csrr(S1, CSR_VL)) != ERROR_OK)
+		return ERROR_FAIL;
+	if (riscv_program_insert(&program, vsetvl(ZERO, S1, S0)) != ERROR_OK)
+		return ERROR_FAIL;
+
+	return riscv_program_exec(&program, target);
+}
+
+static int register_write_to_vl_progbuf(struct target *target,
```
```suggestion
static int vl_write_progbuf(struct target *target,
```

src/target/riscv/riscv-013.c
```diff
@@ -1522,73 +1546,124 @@ static int internal_register_write64_progbuf_scratch(struct target *target,
...
+	riscv_program_init(&program, target);
+	if (riscv_program_insert(&program, csrr(S1, CSR_VTYPE)) != ERROR_OK)
+		return ERROR_FAIL;
+	if (riscv_program_insert(&program, vsetvl(ZERO, S0, S1)) != ERROR_OK)
+		return ERROR_FAIL;
+
 	return riscv_program_exec(&program, target);
 }
 
+static int register_write_to_csr_progbuf(struct target *target,
```
```suggestion
static int csr_write_progbuf(struct target *target,
```

timsifive (changes requested):
Looks good overall.

src/target/riscv/riscv-013.c
aap-sc: addressed, though, I've changed the message a little bit.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
aap-sc: this one is addressed too

timsifive (approved):

PR #903, opened 2023-08-08
target/riscv: fix execute_fence

wxjstz:
This patch improves the following issues:
1. Makes it compatible with targets with progbufsize == 1.
2. Although exceptions don’t update any registers, but  do end execution of the progbuf. This will make fence rw, rw impossible to execute.

Change-Id: I2208fd31ec6a7dae6e61c5952f90901568caada6

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2845,15 +2845,35 @@ static int execute_fence(struct target *target)
 
 	/* FIXME: For non-coherent systems we need to flush the caches right
 	 * here, but there's no ISA-defined way of doing that. */
+	int result;
 	struct riscv_program program;
-	riscv_program_init(&program, target);
-	riscv_program_fence_i(&program);
-	riscv_program_fence_rw_rw(&program);
-	int result = riscv_program_exec(&program, target);
-	if (result != ERROR_OK)
-		LOG_TARGET_DEBUG(target, "Unable to execute pre-fence");
 
-	return ERROR_OK;
+	if (has_sufficient_progbuf(target, 3)) {
+		riscv_program_init(&program, target);
+		riscv_program_fence_rw_rw(&program);
+		riscv_program_fence_i(&program);
+		result = riscv_program_exec(&program, target);
+		if (result != ERROR_OK)
+			LOG_TARGET_DEBUG(target, "Unable to execute pre-fence");
+		return ERROR_OK;
+	}
+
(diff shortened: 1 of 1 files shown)


wxjstz:
@MarekVCodasip

wxjstz:
[riscv-debug-stable.pdf](https://github.com/riscv/riscv-debug-spec/blob/master/riscv-debug-stable.pdf) page 50 section 4.1
>3. Exceptions don’t update any registers. That includes cause, epc, tval, dpc, and mstatus. They do end execution of the Program Buffer.

MarekVCodasip (dismissed):
Visually, it looks fine.

MarekVCodasip:
Good catch.

MarekVCodasip (dismissed):

timsifive (changes requested):
I think it's worth making two code paths: one which executes both instructions at once (old way), and another which executes them individually (new way).

The reason is that execute_fence() is called from read_memory_progbuf() and write_memory_progbuf(), which can be called quite a lot when a debugger is accessing complex structs from memory.

JanMatCodasip:
> I think it's worth making two code paths: one which executes both instructions at once (old way), and another which executes them individually (new way).

In that case, the algorithm could be:

```
if (progbufsize >= 2 && fence_i_supported):
    execute_fence_i_and_fence_rw_rw()
elif (progbufsize != 0):
    if fence_i_supported:
        execute_fence_i()
    execute_fence_rw_rw()

# The support for fence_i can be detected in examine().
```

@timsifive Is it what you have in mind?

wxjstz:
> I think it's worth making two code paths: one which executes both instructions at once (old way), and another which executes them individually (new way).
> 
> The reason is that execute_fence() is called from read_memory_progbuf() and write_memory_progbuf(), which can be called quite a lot when a debugger is accessing complex structs from memory.

done

timsifive (approved):

src/target/riscv/riscv-013.c
```diff
@@ -2845,15 +2845,35 @@ static int execute_fence(struct target *target)
 
 	/* FIXME: For non-coherent systems we need to flush the caches right
 	 * here, but there's no ISA-defined way of doing that. */
+	int result;
```
EccoTheDolphin, line 2848: it looks like we don't really need this `result` variable here. @wxjstz maybe we could cleanup this?

PR #905, opened 2023-08-15
fix crash when we try to read vector register on a running target

aap-sc:
When vector register is accessed we should try to preserve S0 only when `prep_for_vector_access` is successul. Otherwise we'll end up triggering

```
assert(target->state == TARGET_HALTED &&
             "Doesn't make sense to populate register cache on non-halted targets.");
```

Change-Id: I0e140d69faa67f8817310cf18a4db3c581013de2

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2369,14 +2369,14 @@ static int riscv013_get_register_buf(struct target *target,
 	if (dm013_select_target(target) != ERROR_OK)
 		return ERROR_FAIL;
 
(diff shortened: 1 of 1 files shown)


timsifive:
Can you rebase this now that #901 has merged, if it's still necessary?

EccoTheDolphin:
> Can you rebase this now that #901 has merged, if it's still necessary?

yes, unfortunately it is.  The thing is that without this patch we try to save/cache S0 without checking that the target is halted. This check is performed during `prep_for_vector_access`. This commit swaps caching of S0 with `prep_for_vector_access` invocation.

timsifive (approved):

PR #906, opened 2023-08-21
target/riscv: Don't write to zero.

MarekVCodasip:
During a previous patch, the ignoring of writes to register zero was deleted. This patch restores it to the original.

Change-Id: Ieb028a5b2e3f691e4847713c7bc809e10726e18c

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -4718,6 +4718,9 @@ unsigned int riscv_count_harts(struct target *target)
  */
 static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
 {
+	if (regno == GDB_REGNO_ZERO)
+		return !is_write;
+
 	/* GPRs, FPRs, vector registers are just normal data stores. */
 	if (regno <= GDB_REGNO_XPR31 ||
 			(regno >= GDB_REGNO_FPR0 && regno <= GDB_REGNO_FPR31) ||


en-sc:
I would like to ask, what is the motivation for this change? Is this a fix for some regression?

This was [an intentional change](https://github.com/riscv/riscv-openocd/commit/8f3a617dc74b867735aaa78bddd92b8a2fe57579#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aL5124).
The motivation is:
1. This code makes an assumption, that the hardware is working correctly. What if it is not the case? The write to `zero` is a well-defined operation, so it should be correctly performed on hardware.
2. IMHO this is a micro optimization, that mostly pessimizes the performance (the writes to `zero` are quite rare) and increases code complexity due to an extra branch). If a user wants to write to `zero` why shouldn't they be able to do so?

There is an issue, that [`zero` is considered cacheble](https://github.com/riscv/riscv-openocd/blame/5cb60e3f7d1b2eca5e376a6a829bdd3659c35f54/src/target/riscv/riscv.c#L4722), which can cause a bug (misrepresentation of `zero` due non-zero value written to cache). Maybe it is sufficient to only fix this without adding an extra branch to `register_set()`?

MarekVCodasip:
> I would like to ask, what is the motivation for this change? Is this a fix for some regression?

Yes, consider the sequence:

```
reg zero 1
reg zero
```

The last one could result in a non-zero value (and in our testing it does). From user perspective, this is wrong.

>  This code makes an assumption, that the hardware is working correctly. What if it is not the case? The write to `zero` is a well-defined operation, so it should be correctly performed on hardware.

Yes exactly, but currently it is cached before write and the operations could result in wrong values being presented to the user.

> IMHO this is a micro optimization, that mostly pessimizes the performance (the writes to `zero` are quite rare) and increases code complexity due to an extra branch). If a user wants to write to `zero` why shouldn't they be able to do so?

Any write to register zero will result in 0 on correctly functioning hardware, as already assume correctly functioning hardware in most of OpenOCD. The moment anything else happens, i think you have bigger trouble than misbehaving register zero. But I do agree that the user should be able to force write to it.

> There is an issue, that [`zero` is considered cacheble](https://github.com/riscv/riscv-openocd/blame/5cb60e3f7d1b2eca5e376a6a829bdd3659c35f54/src/target/riscv/riscv.c#L4722), which can cause a bug (misrepresentation of `zero` due non-zero value written to cache). Maybe it is sufficient to only fix this without adding an extra branch to `register_set()`?

I agree this might be a better solution. I did test it and it does work. @timsifive what you think about it? I did push the new solution suggested @en-sc as I do like it better.

src/target/riscv/riscv.c
```diff
@@ -4718,6 +4718,10 @@ unsigned int riscv_count_harts(struct target *target)
  */
 static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
 {
+	/* Writes to register zero are rare and should always result in 0 */
```
JanMatCodasip: If you settle on the solution of making x0 non-cacheable, I'd just use a different comment:

```suggestion
	/* Don't cache values of x0 -- this prevents storing non-zero (bogus) values in the cache. */
```

JanMatCodasip (commented):
As Marek explained, the motivation for this change is to never show the user that `x0 == non-zero value`.

Whichever of the two presented solutions to pick is up to you (I am fine with either).

src/target/riscv/riscv.c
```diff
@@ -4718,6 +4718,10 @@ unsigned int riscv_count_harts(struct target *target)
...
+	if (regno == GDB_REGNO_ZERO)
+		return false;
```
```suggestion
		return !is_write;
```

MarekVCodasip: Done

src/target/riscv/riscv.c
MarekVCodasip: Done

JanMatCodasip (dismissed):

src/target/riscv/riscv.c
```diff
@@ -4718,6 +4718,10 @@ unsigned int riscv_count_harts(struct target *target)
  */
 static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
 {
+	/* Don't cache values of x0 -- this prevents storing non-zero (bogus) values in the cache. */
```
en-sc: There seems to be a miss-match between the comment and the action -- currently, `x0` is cached on reads.
I would suggest to delete the comment, since, IMHO, the code is self-explanatory.

en-sc (approved):

timsifive (approved):

PR #908, opened 2023-08-25
target/riscv: Reject size 2 soft breakpoints when C extension not supported

MarekVCodasip:
This patch disables software breakpoints of size 2 for targets which don't support compressed instructions.

Change-Id: I8200b22a51c97ba2aa89e6328beadde8dd35cdd5

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1198,15 +1198,17 @@ static int riscv_add_breakpoint(struct target *target, struct breakpoint *breakp
 	LOG_TARGET_DEBUG(target, "@0x%" TARGET_PRIxADDR, breakpoint->address);
 	assert(breakpoint);
 	if (breakpoint->type == BKPT_SOFT) {
-		/** @todo check RVC for size/alignment */
-		if (!(breakpoint->length == 4 || breakpoint->length == 2)) {
-			LOG_TARGET_ERROR(target, "Invalid breakpoint length %d", breakpoint->length);
+		const bool c_extension_supported = riscv_supports_extension(target, 'C');
+		if (!(breakpoint->length == 4 || (breakpoint->length == 2 && c_extension_supported))) {
+			LOG_TARGET_ERROR(target, "Invalid breakpoint length %d, supported lengths: %s", breakpoint->length,
+				c_extension_supported ? "2, 4" : "4");
 			return ERROR_FAIL;
 		}
 
-		if (0 != (breakpoint->address % 2)) {
-			LOG_TARGET_ERROR(target, "Invalid breakpoint alignment for address 0x%" TARGET_PRIxADDR,
-				breakpoint->address);
+		const unsigned int required_align = c_extension_supported ? 2 : 4;
+		if ((breakpoint->address % required_align) != 0) {
+			LOG_TARGET_ERROR(target, "Invalid breakpoint alignment for address 0x%" TARGET_PRIxADDR
+				", required alignment: %u", breakpoint->address, required_align);
 			return ERROR_FAIL;
 		}
 


src/target/riscv/riscv.c
```diff
@@ -1199,7 +1199,7 @@ static int riscv_add_breakpoint(struct target *target, struct breakpoint *breakp
 	assert(breakpoint);
 	if (breakpoint->type == BKPT_SOFT) {
 		/** @todo check RVC for size/alignment */
```
JanMatCodasip: This ToDo can now be removed.

```suggestion
```

src/target/riscv/riscv.c
```diff
@@ -1199,7 +1199,7 @@ static int riscv_add_breakpoint(struct target *target, struct breakpoint *breakp
...
-		if (!(breakpoint->length == 4 || breakpoint->length == 2)) {
+		if (!(breakpoint->length == 4 || (breakpoint->length == 2 && riscv_supports_extension(target, 'C')))) {
 			LOG_TARGET_ERROR(target, "Invalid breakpoint length %d", breakpoint->length);
 			return ERROR_FAIL;
 		}
```
JanMatCodasip, line 1206: Please, update also the alignment check below. 

Something along these lines:

```
		const unsigned int required_align = riscv_supports_extension(target, 'C') ? 2 : 4;
		if (0 != (breakpoint->address % required_align)) {
			LOG_TARGET_ERROR(target, "Invalid breakpoint alignment for address 0x% (required alignment: %d)" TARGET_PRIxADDR,
				breakpoint->address, required_align);
			return ERROR_FAIL;
		}
```

src/target/riscv/riscv.c
```diff
@@ -1199,7 +1199,7 @@ static int riscv_add_breakpoint(struct target *target, struct breakpoint *breakp
 	assert(breakpoint);
 	if (breakpoint->type == BKPT_SOFT) {
 		/** @todo check RVC for size/alignment */
-		if (!(breakpoint->length == 4 || breakpoint->length == 2)) {
+		if (!(breakpoint->length == 4 || (breakpoint->length == 2 && riscv_supports_extension(target, 'C')))) {
 			LOG_TARGET_ERROR(target, "Invalid breakpoint length %d", breakpoint->length);
```
JanMatCodasip: Please, consider improving also the error message. Thanks.

Example:

```suggestion
			LOG_TARGET_ERROR(target, "Invalid breakpoint length %d (supported lengths: %s)", breakpoint->length, riscv_supports_extension(target, 'C') ? "2, 4" : "4");
```

JanMatCodasip (changes requested):

src/target/riscv/riscv.c
MarekVCodasip: Done

src/target/riscv/riscv.c
MarekVCodasip: Done

src/target/riscv/riscv.c
MarekVCodasip: Done

src/target/riscv/riscv.c
```diff
@@ -1198,15 +1198,17 @@ static int riscv_add_breakpoint(struct target *target, struct breakpoint *breakp
...
 			return ERROR_FAIL;
 		}
 
-		if (0 != (breakpoint->address % 2)) {
-			LOG_TARGET_ERROR(target, "Invalid breakpoint alignment for address 0x%" TARGET_PRIxADDR,
-				breakpoint->address);
+		const unsigned int required_align = c_extension_supported ? 2 : 4;
+		if (0 != (breakpoint->address % required_align)) {
+			LOG_TARGET_ERROR(target, "Invalid breakpoint alignment for address 0x%" TARGET_PRIxADDR
+				", allowed alignment: %u", breakpoint->address, required_align);
```
JanMatCodasip: Detail: I'd recommend to say "required" instead of "allowed".

```suggestion
				", required alignment: %u", breakpoint->address, required_align);
```

JanMatCodasip (dismissed):
Thanks.

src/target/riscv/riscv.c
MarekVCodasip: Done

JanMatCodasip (dismissed):

src/target/riscv/riscv.c
```diff
@@ -1198,15 +1198,17 @@ static int riscv_add_breakpoint(struct target *target, struct breakpoint *breakp
...
+			LOG_TARGET_ERROR(target, "Invalid breakpoint length %d, supported lengths: %s", breakpoint->length,
+				c_extension_supported ? "2, 4" : "4");
 			return ERROR_FAIL;
 		}
 
-		if (0 != (breakpoint->address % 2)) {
-			LOG_TARGET_ERROR(target, "Invalid breakpoint alignment for address 0x%" TARGET_PRIxADDR,
-				breakpoint->address);
+		const unsigned int required_align = c_extension_supported ? 2 : 4;
+		if (0 != (breakpoint->address % required_align)) {
```
timsifive: While you're here, can you change the comparison to put the constant at the end?

The OpenOCD project is very picky about this. See https://review.openocd.org/c/openocd/+/6354 and http://openocd.zylin.com/6350.

Maybe comparisons against 0 are OK even when comparisons against NULL are not.

timsifive (changes requested):
Looks good overall. Just have an OpenOCD style comment.

src/target/riscv/riscv.c
MarekVCodasip: Done

timsifive (approved):

src/target/riscv/riscv.c
```diff
@@ -1198,15 +1198,17 @@ static int riscv_add_breakpoint(struct target *target, struct breakpoint *breakp
 	LOG_TARGET_DEBUG(target, "@0x%" TARGET_PRIxADDR, breakpoint->address);
 	assert(breakpoint);
 	if (breakpoint->type == BKPT_SOFT) {
-		/** @todo check RVC for size/alignment */
-		if (!(breakpoint->length == 4 || breakpoint->length == 2)) {
-			LOG_TARGET_ERROR(target, "Invalid breakpoint length %d", breakpoint->length);
+		const bool c_extension_supported = riscv_supports_extension(target, 'C');
```
fanghuaqi, line 1201: There is a code size reduction isa called Zc*(ratified), see https://github.com/riscv/riscv-code-size-reduction

If Zc extension is used, the instruction can be also 16 bit, so this check by just `riscv_supports_extension(target, 'C')` maybe not correct, since as described here https://github.com/riscv/riscv-code-size-reduction/issues/144#issuecomment-1034370915, this misa.c only set when all normal compress instruction are implemented, but in some case, for example, `rv32imaf_zca`, it will not set misa.c bit, but it still contains 16 bit instruction, so I think sw breakpoint can still set here, so the check by misa.c bit is not enough.

TommyMurphyTM1234:
Are there specific reasons not to simply always use the "lowest common denominator" of 32-bit software breakpoints regardless of the target's 16-bit instruction capabilities? Off the top of my head the only one that I can think of is the inability to write one at the instruction address 2 bytes from the "top" of a memory region. Are there other reasons not to just take the simpler approach?

src/target/riscv/riscv.c
fanghuaqi: also here https://github.com/riscv-software-src/riscv-isa-sim/issues/1172

fanghuaqi: Hi @MarekVCodasip @timsifive @tariqkurd-repo , is there any feedback on my comments regarding the new added Zc* extensions will affect this compress instructions existing checking.

tariqkurd-repo: Implementing Zca and not C will not set misa.C
I think you actually need to specify C to set it
Is that your understanding @aswaterman ?

fanghuaqi: > Implementing Zca and not C will not set misa.C I think you actually need to specify C to set it Is that your understanding @aswaterman ?

If I implement rv32imaf_zca_zcb_zcf_zcmp_zcmt it contains all C required zca_zcf will the misa.C bit be set? This `riscv_supports_extension` function only use the misa.C bit to check whether compressed instructions are supported.

JanMatCodasip:
> TommyMurphyTM1234: Are there specific reasons not to simply always use the "lowest common denominator" of 32-bit software breakpoints regardless of the target's 16-bit instruction capabilities?

By placing a SW breakpoint (`ebreak` or `c.ebreak`), we must not overwrite any other instruction than the one on which the break is placed. That's because the subsequent instruction after the break can be a target of a branch. And as such, it must remain intact.

Using always the 32-bit SW breaks is therefore not an option.

TommyMurphyTM1234:
> > TommyMurphyTM1234: Are there specific reasons not to simply always use the "lowest common denominator" of 32-bit software breakpoints regardless of the target's 16-bit instruction capabilities?
> 
> By placing a SW breakpoint (`ebreak` or `c.ebreak`), we must not overwrite any other instruction than the one on which the break is placed. That's because the subsequent instruction after the break can be a target of a branch. And as such, it must remain intact.
> 
> Using always the 32-bit SW breaks is therefore not an option.

Thanks for that explanation @JanMatCodasip 👍

fanghuaqi:
Hi @MarekVCodasip @timsifive @JanMatCodasip , using this commit, I am not able to set breakpoint at 2 bytes aligned address, with Zc* extension enabled which also provided compressed instructions.

image

JanMatCodasip:
@fanghuaqi 

Is there a reliable way how OpenOCD can detect the `c.ebreak` support in the target hardware, please?

- If so, we may consider implementing it in OpenOCD.
- Otherwise we should probably revert/rework this commit, and simply obey the breakpoint length as asked by GDB (or by the user).

fanghuaqi:
> @fanghuaqi
> 
> Is there a reliable way how OpenOCD can detect the `c.ebreak` support in the target hardware, please?
> 
> * If so, we may consider implementing it in OpenOCD.
> * Otherwise we should probably revert/rework this commit, and simply obey the breakpoint length as asked by GDB (or by the user).

Hi @JanMatCodasip , I didn't know whether there is a way to check whether `c.ebreak` supported in hardware, don't DM provide some information about it? If not, may be make some notes in the doc will be good.

TommyMurphyTM1234:
Earlier:

* https://github.com/riscv/riscv-openocd/pull/908#discussion_r1349493892

it was stated that:

> There is a code size reduction isa called Zc*(ratified), see https://github.com/riscv/riscv-code-size-reduction
> 
> If Zc extension is used, the instruction can be also 16 bit

But, as far as I can see, support for `c.ebreak` is orthogonal to support for the `Zc*` extension/sub-extensions.

* Broken link: ~~https://wiki.riscv.org/display/HOME/Recently+Ratified+Extensions#:~:text=RISC%2DV%20Code,Zcf%2C%20Zcmp%2C%20Zcmt~~
* https://drive.google.com/file/d/1zmWDmfbtVY9I6hn0vuLTbk5rsSPc44sL/view

The `Zc*` extension/sub-extensions make no mention of `c.ebreak` - only The RISC-V Instruction Set Manual
Volume I: Unprivileged ISA mentions/specifies it. 

* https://riscv.org/technical/specifications/

That being the case shouldn't a check of `$misa` for the `C` bit being set should be sufficient for checking if the target supports `c.ebreak`?

TommyMurphyTM1234:
Perhaps this is another option?

* https://sourceware.org/legacy-ml/gdb-patches/2018-09/msg00815.html

> this patch instead examines
> the existing instruction at a breakpoint target to determine which
> breakpoint instruction to use.  If the existing instruction is a
> compressed instruction, C.EBREAK is used, otherwise EBREAK is used.

timsifive:
The simple option is to revert this change. gdb already has good logic to decide what breakpoint size to use, because it knows what size instruction is being replaced.

What was the original reason for this change? I don't see a rationale in the description.

timsifive:
From what I can tell, c.ebreak is defined in the unprivileged spec, where it is part of the "C" Standard Extension chapter. That means to me that if c.ebreak is implemented then misa.C must be set.

That feels like a spec bug, though. If you are allowed to implement 16-bit instructions without setting misa.C, which the Zc* extensions clearly do, then obviously you want c.ebreak as well, without being required to implement all the other C extension instructions. I filed https://github.com/riscv/riscv-code-size-reduction/issues/228 for this.

I would expect the outcome to be something that allows c.ebreak even without C, and the result for OpenOCD is that we should revert this change.

fanghuaqi:
`c.break` is part of Zca which is the only part of C without(compressed F/D instructions).

src/target/riscv/riscv.c
```diff
@@ -1198,15 +1198,17 @@ static int riscv_add_breakpoint(struct target *target, struct breakpoint *breakp
 	LOG_TARGET_DEBUG(target, "@0x%" TARGET_PRIxADDR, breakpoint->address);
 	assert(breakpoint);
 	if (breakpoint->type == BKPT_SOFT) {
-		/** @todo check RVC for size/alignment */
```
fanghuaqi, line 1201: @timsifive I think this patch is added due to there is a todo here, is it possible to check whether 16bit instruction execute ok without any exception, such as test `c.nop`, if no exception, compress instruction is supported, otherwise no.

JanMatCodasip:
> Tim: The simple option is to revert this change.

I am in agreement.

> Tim: What was the original reason for this change? I don't see a rationale in the description.

The reason was to prevent user mistakes and improve robustness (an attempt that in the end has the opposite effect: additional issues).

TommyMurphyTM1234:
> `c.break` is part of Zca which is the only part of C without(compressed F/D instructions).

Did you mean `c.ebreak`?
Can you clarify where in the `Zc*` spec that is stated?
I couldn't see it.

fanghuaqi:
> > `c.break` is part of Zca which is the only part of C without(compressed F/D instructions).
> 
> Did you mean `c.ebreak`? Can you clarify where in the `Zc*` spec that is stated? I couldn't see it.

Hi @TommyMurphyTM1234 , please check it here https://github.com/riscv/riscv-code-size-reduction/commit/4abffe77b3eb866a6256011bd150a1b20e87c8f1

TommyMurphyTM1234:
> > > `c.break` is part of Zca which is the only part of C without(compressed F/D instructions).
> > 
> > 
> > Did you mean `c.ebreak`? Can you clarify where in the `Zc*` spec that is stated? I couldn't see it.
> 
> Hi @TommyMurphyTM1234 , please check it here [riscv/riscv-code-size-reduction@4abffe7](https://github.com/riscv/riscv-code-size-reduction/commit/4abffe77b3eb866a6256011bd150a1b20e87c8f1)

The ratified `Zc*` extension is listed here:

* https://wiki.riscv.org/display/HOME/Recently+Ratified+Extensions#:~:text=RISC%2DV%20Code%20Size%20Reduction

The ratified specification does not mention `c.ebreak` or the text in the commit that you linked to.

* https://drive.google.com/file/d/1zmWDmfbtVY9I6hn0vuLTbk5rsSPc44sL/view?usp=drive_link

Are you referring to some proposed draft update?
I still don't see how what you linked to clarifies that `Zc*` necessarily implies support for `c.ebreak`.

timsifive:
@TommyMurphyTM1234, see https://github.com/riscv/riscv-code-size-reduction/issues/228.

TLDR: Every Zc* requires Zca, and Zca is all of the existing C extension, except some 16-bit floating point ops.

PR #909, opened 2023-08-25
target/riscv: cleanup riscv_enumerate_triggers()

en-sc:
1. Propagate error codes.
2. Disable leftover triggers in case `tinfo` is supported.

Change-Id: Ie20edb4d8b9245e13ac8757bf6afe549ac99c4f1

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -4162,13 +4162,23 @@ error:
 	return result;
 }
 
-COMMAND_HELPER(riscv_print_info_line, const char *section, const char *key,
-			   unsigned int value)
+static COMMAND_HELPER(riscv_print_info_line_if_available, const char *section,
+		const char *key, unsigned int value, bool is_available)
 {
 	char full_key[80];
 	snprintf(full_key, sizeof(full_key), "%s.%s", section, key);
-	command_print(CMD, "%-21s %3d", full_key, value);
+	if (is_available)
+		command_print(CMD, "%-21s %3d", full_key, value);
+	else
+		command_print(CMD, "%-21s unavailable", full_key);
 	return 0;
+}
+
+COMMAND_HELPER(riscv_print_info_line, const char *section, const char *key,
+			   unsigned int value)
+{
+	return CALL_COMMAND_HANDLER(riscv_print_info_line_if_available, section,
+			key, value, /*is_available*/ true);
 }
 
 COMMAND_HANDLER(handle_info)
@@ -4179,10 +4189,11 @@ COMMAND_HANDLER(handle_info)
 	/* This output format can be fed directly into TCL's "array set". */
 
 	riscv_print_info_line(CMD, "hart", "xlen", riscv_xlen(target));
-	riscv_enumerate_triggers(target);
-	riscv_print_info_line(CMD, "hart", "trigger_count",
-						  r->trigger_count);
 
+	const bool trigger_count_available =
+		riscv_enumerate_triggers(target) == ERROR_OK;
+	riscv_print_info_line_if_available(CMD, "hart", "trigger_count",
+				r->trigger_count, trigger_count_available);
 	if (r->print_info)
 		return CALL_COMMAND_HANDLER(r->print_info, target);
 
@@ -5029,6 +5040,78 @@ int riscv_dmi_write_u64_bits(struct target *target)
 	return r->dmi_write_u64_bits(target);
 }
 
+static int check_if_trigger_exists(struct target *target, unsigned int index)
+{
+	/* If we can't write tselect, then this hart does not support triggers. */
+	if (riscv_set_register(target, GDB_REGNO_TSELECT, index) != ERROR_OK)
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	riscv_reg_t tselect_rb;
+	if (riscv_get_register(target, &tselect_rb, GDB_REGNO_TSELECT) != ERROR_OK)
+		return ERROR_FAIL;
+	/* Mask off the top bit, which is used as tdrmode in legacy RISC-V Debug Spec
+	 * (old revisions of v0.11 spec). */
+	tselect_rb &= ~(1ULL << (riscv_xlen(target) - 1));
+	if (tselect_rb != index)
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	return ERROR_OK;
+}
+
+/**
+ * This function reads `tinfo` or `tdata1`, when reading `tinfo` fails,
+ * to determine trigger types supported by a trigger.
+ * It is assumed that the trigger is already selected via writing `tselect`.
+ */
+static int get_trigger_types(struct target *target, unsigned int *trigger_tinfo,
+		riscv_reg_t tdata1)
+{
+	assert(trigger_tinfo);
+	riscv_reg_t tinfo;
+	if (riscv_get_register(target, &tinfo, GDB_REGNO_TINFO) == ERROR_OK) {
+		/* tinfo.INFO == 1: trigger doesn’t exist
+		 * tinfo == 0 or tinfo.INFO != 1 and tinfo LSB is set: invalid tinfo */
+		if (tinfo == 0 || tinfo & 0x1)
+			return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+		*trigger_tinfo = tinfo;
+		return ERROR_OK;
+	}
+	const unsigned int type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
+	if (type == 0)
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	*trigger_tinfo = 1 << type;
+	return ERROR_OK;
+}
+
+static int disable_trigger_if_dmode(struct target *target, riscv_reg_t tdata1)
+{
+	bool dmode_is_set = false;
+	switch (get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)))) {
+		case CSR_TDATA1_TYPE_LEGACY:
+			/* On these older cores we don't support software using
+			 * triggers. */
+			dmode_is_set = true;
+			break;
+		case CSR_TDATA1_TYPE_MCONTROL:
+			dmode_is_set = tdata1 & CSR_MCONTROL_DMODE(riscv_xlen(target));
+			break;
+		case CSR_TDATA1_TYPE_MCONTROL6:
+			dmode_is_set = tdata1 & CSR_MCONTROL6_DMODE(riscv_xlen(target));
+			break;
+		case CSR_TDATA1_TYPE_ICOUNT:
+			dmode_is_set = tdata1 & CSR_ICOUNT_DMODE(riscv_xlen(target));
+			break;
+		case CSR_TDATA1_TYPE_ITRIGGER:
+			dmode_is_set = tdata1 & CSR_ITRIGGER_DMODE(riscv_xlen(target));
+			break;
+		case CSR_TDATA1_TYPE_ETRIGGER:
+			dmode_is_set = tdata1 & CSR_ETRIGGER_DMODE(riscv_xlen(target));
+			break;
+	}
+	if (!dmode_is_set)
+		/* Nothing to do */
+		return ERROR_OK;
+	return riscv_set_register(target, GDB_REGNO_TDATA1, 0);
+}
+
 /**
  * Count triggers, and initialize trigger_count for each hart.
  * trigger_count is initialized even if this function fails to discover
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv.c
```diff
@@ -5073,59 +5072,65 @@ int riscv_enumerate_triggers(struct target *target)
...
+		LOG_TARGET_DEBUG(target, "Trigger %u: supported types (mask) = 0x%08x",
+				t, trigger_tinfo[t]);
+	}
 
-	LOG_TARGET_INFO(target, "Found %d triggers.", r->trigger_count);
+	if (riscv_set_register(target, GDB_REGNO_TSELECT, tselect) != ERROR_OK)
+		return ERROR_FAIL;
 
+	r->triggers_enumerated = true;
+	r->trigger_count = t + 1;
```
aap-sc: This is incorrect. Why do you do this increment?

src/target/riscv/riscv.c
```diff
@@ -5043,89 +5121,54 @@ int riscv_enumerate_triggers(struct target *target)
 	if (r->triggers_enumerated)
 		return ERROR_OK;
 
-	r->triggers_enumerated = true;	/* At the very least we tried. */
```
JanMatCodasip, line 5046: This is a change of the original logic:

- The original code attempted the trigger evaluation only once, and if failed, it did not retry the process.
- The current code seems to allow to retry the trigger enumeration - until it succeeds.

I am not against that, but please if you can explain the motivation for this change.

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,73 @@ int riscv_dmi_write_u64_bits(struct target *target)
...
+	if (tselect_rb != t)
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	return ERROR_OK;
+}
+
+static int get_trigger_types(struct target *target, unsigned int *trigger_tinfo,
+		riscv_reg_t tdata1)
+{
+	riscv_reg_t tinfo;
+	const bool may_support_multiple_trigger_types =
```
JanMatCodasip: If you like, this can be called simply `tinfo_exists` or `tinfo_readable`, which is much shorter and still very clear.

```suggestion
	const bool tinfo_exists =
```

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,73 @@ int riscv_dmi_write_u64_bits(struct target *target)
...
+		riscv_get_register(target, &tinfo, GDB_REGNO_TINFO) == ERROR_OK;
+	if (may_support_multiple_trigger_types) {
+		/* tinfo == 0: invalid
+		 * tinfo == 1: trigger doesn’t exist */
+		if (tinfo == 0 || tinfo == 1)
+			return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
```
JanMatCodasip: I know this is the original code, just moved. But if you like, you can be more defensive here: you can end the enumeraton if the LSB of tinfo is set, regardless of the other bits. This is to defend against potential buggy HW implementations of tinfo.

```suggestion
		 * tinfo has LSB set: trigger doesn’t exist */
		if (tinfo == 0 || tinfo & 0x1)
			return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
```

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,73 @@ int riscv_dmi_write_u64_bits(struct target *target)
...
+	tselect_rb &= ~(1ULL << (riscv_xlen(target) - 1));
+	if (tselect_rb != t)
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	return ERROR_OK;
+}
+
+static int get_trigger_types(struct target *target, unsigned int *trigger_tinfo,
+		riscv_reg_t tdata1)
+{
+	riscv_reg_t tinfo;
```
JanMatCodasip, line 5068: If you like:

```suggestion
	assert(trigger_tinfo);
	riscv_reg_t tinfo;
```

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,73 @@ int riscv_dmi_write_u64_bits(struct target *target)
...
+		case CSR_TDATA1_TYPE_ITRIGGER:
+			is_leftover = tdata1 & CSR_ITRIGGER_DMODE(riscv_xlen(target));
+			break;
+		case CSR_TDATA1_TYPE_ETRIGGER:
+			is_leftover = tdata1 & CSR_ETRIGGER_DMODE(riscv_xlen(target));
+			break;
+	}
+	if (is_leftover && riscv_set_register(target, GDB_REGNO_TDATA1, 0) != ERROR_OK)
+		return ERROR_FAIL;
+	return ERROR_OK;
```

```suggestion
	if (!is_leftover)
		/* Nothing to do */
		return ERROR_OK;
		
	return riscv_set_register(target, GDB_REGNO_TDATA1, 0);
```

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,73 @@ int riscv_dmi_write_u64_bits(struct target *target)
 	return r->dmi_write_u64_bits(target);
 }
 
+static int check_if_trigger_is_availible(struct target *target, unsigned int t)
```
JanMatCodasip: There is a typo in the original name.

You can also rename it to the following (which is easier to read and equally clear):

```suggestion
static int check_if_trigger_exists(struct target *target, unsigned int t)
```

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,73 @@ int riscv_dmi_write_u64_bits(struct target *target)
...
+{
+	/* If we can't write tselect, then this hart does not support triggers. */
+	if (riscv_set_register(target, GDB_REGNO_TSELECT, t) != ERROR_OK)
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	riscv_reg_t tselect_rb;
+	if (riscv_get_register(target, &tselect_rb, GDB_REGNO_TSELECT) != ERROR_OK)
+		return ERROR_FAIL;
+	/* Mask off the top bit, which is used as tdrmode in old
+	 * implementations. */
+	tselect_rb &= ~(1ULL << (riscv_xlen(target) - 1));
```
JanMatCodasip: @timsifive Side question: Is this a relict of the 0.11 debug implementation? Just for my personal curiosity.

(No need to change it anyhow in this merge request.)

src/target/riscv/riscv.c
```diff
@@ -5043,89 +5121,54 @@ int riscv_enumerate_triggers(struct target *target)
...
-		* implemented. There are no triggers to enumerate then and no error
-		* should be thrown. */
+	 * implemented. */
 	if (result != ERROR_OK) {
-		LOG_TARGET_DEBUG(target, "Cannot access tselect register. "
-				"Assuming that triggers are not implemented.");
-		r->trigger_count = 0;
-		return ERROR_OK;
+		LOG_DEBUG("[%s] Cannot access tselect register. "
+				"Assuming that triggers are not implemented.", target_name(target));
```
JanMatCodasip: If tselect is not accessible, I think we should interpret it as a successful enumeration during which we did not find any triggers. In another words:

```
r->triggers_enumerated = true;
r->trigger_count = t;
return ERROR_OK;
```

Also, can we use LOG_TARGET_DEBUG here? 
Or even LOG_TARGET_INFO?

src/target/riscv/riscv.c
```diff
@@ -5043,89 +5121,54 @@ int riscv_enumerate_triggers(struct target *target)
...
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
 	}
 
-	for (unsigned int t = 0; t < RISCV_MAX_TRIGGERS; ++t) {
-		r->trigger_count = t;
-
-		/* If we can't write tselect, then this hart does not support triggers. */
-		if (riscv_set_register(target, GDB_REGNO_TSELECT, t) != ERROR_OK)
+	unsigned int t = 0;
+	unsigned int trigger_tinfo[ARRAY_SIZE(r->trigger_tinfo)];
```
JanMatCodasip: Here you can zero-initialize the whole array -- so that anyone looking at it via debugger does not see any random garbage in the unused array items.

```suggestion
	unsigned int trigger_tinfo[ARRAY_SIZE(r->trigger_tinfo)] = {0};
```

src/target/riscv/riscv.c
```diff
@@ -5043,89 +5121,54 @@ int riscv_enumerate_triggers(struct target *target)
...
+	for (; t < ARRAY_SIZE(r->trigger_tinfo); ++t) {
+		result = check_if_trigger_is_availible(target, t);
+		if (result == ERROR_FAIL)
+			return ERROR_FAIL;
```
JanMatCodasip, line 5150: Should we try to restore the `orig_tselect` even if the enumeration fails? I believe so.

In that case, you can use the `goto fail;` pattern instead of direct `return ERROR_FAIL`.

JanMatCodasip (changes requested):

src/target/riscv/riscv.c
en-sc: `riscv_enumerate_triggers` returns `ERROR_FAIL` in case `tselect` or `tdata1` is not accessible. It looks like something is seriously wrong with trigger module and there is no benefit in attempting a restore of `tselect`, unless I'm missing something.

src/target/riscv/riscv.c
en-sc: Is this really beneficial to implement workarounds for an implementation that does not behave as specified in advance? I'm afraid this adds unnecessary complexity. If there is a HW implementation that needs a workaround, I would like to see it applied as a separate commit with a solid reasoning and a reproduction. IMHO this makes the intention clearer and results in a more supportable codebase.

src/target/riscv/riscv.c
en-sc: The motivation is:
Before the change if the enumeration failed (e.g. it was attempted while the hart was running), there would be no triggers available until restart of OpenOCD (there is no code that unsets `triggers_enumerated` flag), so the software representation of the target in the OpenOCD will be incorrect for the whole session.

src/target/riscv/riscv.c
en-sc: Addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: You are right. Thanks for the catch! Addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: Addressed. This variable is used only once, so I decided to just propagate the definition.

src/target/riscv/riscv.c
JanMatCodasip: Ok, understood, that makes sense.

src/target/riscv/riscv.c
JanMatCodasip: Defensive approach is about pro-actively finding and fixing points which can be made safer - in advance, not waiting until something breaks.

In this case the implementation is trivial (and can be clearly described in a comment), so I wouldn't mind making it as part of the change. However, I don't insist either.

Actually, the `tinfo == 0` part is already a similar situation - it is not legal per the spec (compliant hardware should never say tinfo==0), but the implementation in OpenOCD already checks for it.

src/target/riscv/riscv.c
JanMatCodasip: Ok, agreed. Let's leave the current implementation, then.

src/target/riscv/riscv.c
JanMatCodasip: @en-sc Can you please check if the change has been pushed? I don't see it in the diff on github.

src/target/riscv/riscv.c
JanMatCodasip: @en-sc Has this change been pushed already, please?

src/target/riscv/riscv.c
sifiverobert: Hi, I noticed this:
check_if_trigger_is_availible
which should be:
check_if_trigger_is_available

src/target/riscv/riscv.c
en-sc: Please, see [the diff](https://github.com/riscv/riscv-openocd/compare/d4cdb997e455801e1c63572f113fc600e8d0fad2..ea43a461a42681651f30d07b3e8ff03457ac1f70#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR5064). I did forget to push the changes. Sorry.

src/target/riscv/riscv.c
en-sc: Please, see [the diff](https://github.com/riscv/riscv-openocd/compare/d4cdb997e455801e1c63572f113fc600e8d0fad2..ea43a461a42681651f30d07b3e8ff03457ac1f70#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR5133). I did forget to push the changes. Sorry.

src/target/riscv/riscv.c
en-sc: > Hi, I noticed this: check_if_trigger_is_availible which should be: check_if_trigger_is_available

Addressed.

src/target/riscv/riscv.c
```diff
@@ -5043,89 +5120,56 @@ int riscv_enumerate_triggers(struct target *target)
...
 		return ERROR_OK;
 	}
 
-	for (unsigned int t = 0; t < RISCV_MAX_TRIGGERS; ++t) {
-		r->trigger_count = t;
-
-		/* If we can't write tselect, then this hart does not support triggers. */
-		if (riscv_set_register(target, GDB_REGNO_TSELECT, t) != ERROR_OK)
+	unsigned int t = 0;
+	unsigned int trigger_tinfo[ARRAY_SIZE(r->trigger_tinfo)] = {};
```
JanMatCodasip: Here I'd recommend to use:

```suggestion
	unsigned int trigger_tinfo[ARRAY_SIZE(r->trigger_tinfo)] = {0};
```

Empty initializers seem to be a non-standard GNU extension in GCC; otherwise they are only available starting from C23: https://en.cppreference.com/w/c/language/initialization

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,72 @@ int riscv_dmi_write_u64_bits(struct target *target)
...
+
+static int get_trigger_types(struct target *target, unsigned int *trigger_tinfo,
+		riscv_reg_t tdata1)
+{
+	assert(trigger_tinfo);
+	riscv_reg_t tinfo;
+	if (riscv_get_register(target, &tinfo, GDB_REGNO_TINFO) == ERROR_OK) {
+		if (get_field(tinfo, CSR_TINFO_INFO) == 1)
+			/* trigger doesn’t exist */
+			return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
```
JanMatCodasip: This check has been made less defensive instead of more defensive.

The original code checked for  `tinfo==1` and `tinfo==0`.

Please, if you do not wish to extend the condition to the suggested `tinfo & 0x1`, leave at least the current state, unless there is very good reason to make the check weaker.

src/target/riscv/riscv.c
JanMatCodasip: No problem - thanks.

src/target/riscv/riscv.c
JanMatCodasip: No problem - thanks.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
```diff
@@ -4179,10 +4189,11 @@ COMMAND_HANDLER(handle_info)
...
 
 	riscv_print_info_line(CMD, "hart", "xlen", riscv_xlen(target));
-	riscv_enumerate_triggers(target);
-	riscv_print_info_line(CMD, "hart", "trigger_count",
-						  r->trigger_count);
 
+	const bool trigger_count_available =
+		riscv_enumerate_triggers(target) == ERROR_OK;
+	riscv_print_info_line_if_available(CMD, "hart", "trigger_count",
+				r->trigger_count, trigger_count_available);
```
timsifive, line 4196: You shouldn't print "unavailable" if the problem is that the number of triggers could not be detected, because it could also be interpreted to mean that there are no triggers present. Something like "detection failed" is much more clear.

Instead of creating riscv_print_info_line_if_available, I prefer adding riscv_print_info_line_str and renaming riscv_print_info_line to riscv_print_info_line_int. Then the caller, which knows much more about the details, can decide what to show the user.

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,73 @@ int riscv_dmi_write_u64_bits(struct target *target)
 	return r->dmi_write_u64_bits(target);
 }
 
+static int check_if_trigger_exists(struct target *target, unsigned int t)
```
```suggestion
static int check_if_trigger_exists(struct target *target, unsigned int index)
```

src/target/riscv/riscv.c
timsifive: I think this is even older than "latest" 0.11. This code works with such ancient implementations, at the cost of limiting the target to 2 million triggers, which noone will implement anyway.

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,73 @@ int riscv_dmi_write_u64_bits(struct target *target)
...
+		return ERROR_FAIL;
+	/* Mask off the top bit, which is used as tdrmode in old
+	 * implementations. */
+	tselect_rb &= ~(1ULL << (riscv_xlen(target) - 1));
+	if (tselect_rb != t)
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	return ERROR_OK;
+}
+
+static int get_trigger_types(struct target *target, unsigned int *trigger_tinfo,
```
timsifive, line 5064: This function doesn't just look at the tdata1 field. It also reads tinfo and assumes tselect is already set. That's far from obvious from the prototype. Can you add a comment?

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,73 @@ int riscv_dmi_write_u64_bits(struct target *target)
...
+		return ERROR_OK;
+	}
+	const unsigned int type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
+	if (type == 0)
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	*trigger_tinfo = 1 << type;
+	return ERROR_OK;
+}
+
+static int disable_trigger_if_leftover(struct target *target, riscv_reg_t tdata1)
```
```suggestion
static int disable_trigger_if_dmode(struct target *target, riscv_reg_t tdata1)
```

Leftover isn't very clear what it means.

src/target/riscv/riscv.c
timsifive: Can you check that a user doesn't get tons of errors on systems which don't implement triggers at all? What's the behavior if you try to set a hwbp through gdb with this change?

src/target/riscv/riscv.c
```diff
@@ -5043,89 +5121,56 @@ int riscv_enumerate_triggers(struct target *target)
...
 
-	riscv_set_register(target, GDB_REGNO_TSELECT, tselect);
-
-	LOG_TARGET_INFO(target, "Found %d triggers.", r->trigger_count);
+	if (riscv_set_register(target, GDB_REGNO_TSELECT, orig_tselect) != ERROR_OK)
+		return ERROR_FAIL;
 
+	r->triggers_enumerated = true;
+	r->trigger_count = t;
+	memcpy(r->trigger_tinfo, trigger_tinfo, r->trigger_count * sizeof(*trigger_tinfo));
```
timsifive: Can you modify r->trigger_tinfo directly instead of doing this copy?
r->triggers_enumerated already guards against r->trigger_tinfo being partially populated.

timsifive (changes requested):
This does make things a bit more readable.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: According to [this comment](https://github.com/riscv/riscv-openocd/pull/909/files#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR4189), the format of the `riscv info` command is intended to be used directly with TCL's `array set`.
So I have assumed that it is intended for the output to be machine-readable. Providing a programmer with a way to output an arbitrary string will greatly complicate the parsing of the output.
I can change the string that is currently used.
Currently:
```
hart.trigger_count           unavailable
```
Reports that `trigger_count` is not available.
I can document this in `doc/openocd.texi`.

src/target/riscv/riscv.c
en-sc: Addressed.

src/target/riscv/riscv.c
JanMatCodasip: Thank you for explanation.

src/target/riscv/riscv.c
```diff
@@ -5029,6 +5040,78 @@ int riscv_dmi_write_u64_bits(struct target *target)
...
+static int check_if_trigger_exists(struct target *target, unsigned int index)
+{
+	/* If we can't write tselect, then this hart does not support triggers. */
+	if (riscv_set_register(target, GDB_REGNO_TSELECT, index) != ERROR_OK)
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	riscv_reg_t tselect_rb;
+	if (riscv_get_register(target, &tselect_rb, GDB_REGNO_TSELECT) != ERROR_OK)
+		return ERROR_FAIL;
+	/* Mask off the top bit, which is used as tdrmode in old
+	 * implementations. */
```
JanMatCodasip: It'd be good to adjust this comment per [Tim's reply](https://github.com/riscv/riscv-openocd/pull/909/files#r1312044451):

```suggestion
	/* Mask off the top bit, which is used as tdrmode in legacy RISC-V Debug Spec
	 * (old revisions of v0.11 spec). */
```

JanMatCodasip (dismissed):
Looks good to me.

src/target/riscv/riscv.c

JanMatCodasip (dismissed):

src/target/riscv/riscv.c
timsifive: Good point about the TCL array command. Please update the documentation and then leave the code as-is.

timsifive (changes requested):
Just the documentation change left.

en-sc:
> Just the documentation change left.

Addressed.

timsifive (approved):

Issue #907, opened 2023-08-25
Support for smp in multiple DMs


lz-bro:
How to support for smp in multiple DMs? It appears from the code that harts  across multiple DMs can be haled/resumed, but the same threadid in multiple DMs, and gdb can only debug harts in sigle DM

config with
```
set _TARGETNAME_0 $_CHIPNAME.dm0cpu0
set _TARGETNAME_1 $_CHIPNAME.dm0cpu1
set _TARGETNAME_2 $_CHIPNAME.dm1cpu0
set _TARGETNAME_3 $_CHIPNAME.dm1cpu1

# cluster 0
target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -rtos hwthread
target create $_TARGETNAME_1 riscv -chain-position $_CHIPNAME.cpu -coreid 1

# cluster 1
target create $_TARGETNAME_2 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -dbgbase 0x400 -rtos hwthread
target create $_TARGETNAME_3 riscv -chain-position $_CHIPNAME.cpu -coreid 1 -dbgbase 0x400

target smp $_TARGETNAME_0 $_TARGETNAME_1 $_TARGETNAME_2 $_TARGETNAME_3
```
The index of each hart on the DM define
https://github.com/riscv/riscv-openocd/blob/a45589d60aa6864475fddcded885c8ff47d50be1/src/target/riscv/riscv-013.c#L1715-L1716
openocd log shown
```
Info :  [riscv.dm0cpu0] became available (halted)
Info :  [riscv.dm0cpu1] became available (halted)
Info :  [riscv.dm1cpu0] became available (halted)
Info :  [riscv.dm1cpu1] became available (halted)
```
threadid define
https://github.com/riscv/riscv-openocd/blob/21d21408aa36ccb458ccfaad871f7dd1bb1f731a/src/rtos/hwthread.c#L40-L43

so 1 gdb can only see 2 core in dm1
```
(gdb) info threads
  Id   Target Id                                                            Frame
* 1    Thread 1 "riscv.dm1cpu0" (Name: riscv.dm1cpu0, state: debug-request) 0x0000000000001c6c in ?? ()
  2    Thread 2 "riscv.dm1cpu1" (Name: riscv.dm1cpu1, state: debug-request) 0x000000000000769a in ?? ()
```
Can anyone give me some advice?

en-sc:
[...]

PR #910, opened 2023-08-28
Support for smp in multiple DMs

lz-bro:
```
Support for smp in multiple DMs by dealing with riscv hartid and index of each hart simply
Change-Id: 7c1b89f89d0f5ccb6031e754fc4bbec2d70d53c4
```
There is an issue about Support for smp in multiple DMs: [#907 ](https://github.com/riscv/riscv-openocd/issues/907)

When I debug 2 DMs and each with 2 harts , hart count my config file is like
```
set _TARGETNAME_0 $_CHIPNAME.dm0cpu0
set _TARGETNAME_1 $_CHIPNAME.dm0cpu1
set _TARGETNAME_2 $_CHIPNAME.dm1cpu0
set _TARGETNAME_3 $_CHIPNAME.dm1cpu1

# cluster 0
target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -rtos hwthread

# cluster 1
target create $_TARGETNAME_2 riscv -chain-position $_CHIPNAME.cpu -coreid 4 -dbgbase 0x400 -rtos hwthread
target create $_TARGETNAME_3 riscv -chain-position $_CHIPNAME.cpu -coreid 5 -dbgbase 0x400

target smp $_TARGETNAME_0 $_TARGETNAME_1 $_TARGETNAME_2 $_TARGETNAME_3
```
this config file works with the code
`info->index = target->coreid % dm->hart_count;`
dm->hart_count was 4, so when I write 4/5 to coreid,  I can write 0/1 to info->index. For each Debug Module, the index of each hart can start at 0 and be contiguous.
openocd log shown
```
(gdb) info threads
  Id   Target Id                                                            Frame
* 1    Thread 1 "riscv.dm0cpu0" (Name: riscv.dm0cpu0, state: debug-request) 0x0000000000001c6c in ?? ()
  2    Thread 2 "riscv.dm0cpu1" (Name: riscv.dm0cpu1, state: debug-request) 0x000000000000769a in ?? ()
  3    Thread 5 "riscv.dm1cpu0" (Name: riscv.dm1cpu0, state: debug-request) 0x000000000000769a in ?? ()
  4    Thread 6 "riscv.dm1cpu1" (Name: riscv.dm1cpu1, state: debug-request) 0x0000000000007698 in ?? ()
(gdb) c
Continuing.
```


JanMatCodasip:
I'm sorry, but my understanding is that the SMP support in RISC-V is not designed to work across multiple debug modules. I am afraid that this simple change won't make it work. (@timsifive please correct me if I am mistaken.)

In any case, if harts are managed by multiple different DMs, it is not even possible to accomplish the near-instataneous halt & resume -- which is one of the main features of SMP debugging. So this is not just OpenOCD limitation, this is also a limitation of the given HW.

lz-bro:
> I'm sorry, but my understanding is that the SMP support in RISC-V is not designed to work across multiple debug modules. I am afraid that this simple change won't make it work. (@timsifive please correct me if I am mistaken.)
> 
> In any case, if harts are managed by multiple different DMs, it is not even possible to accomplish the near-instataneous halt & resume -- which is one of the main features of SMP debugging. So this is not just OpenOCD limitation, this is also a limitation of the given HW.

Thank you for your reply, this is not a perfect solution, because it can only accomplish the instataneous halt&resume of the harts on each cluster through hartwindow. But even if the external trigger in the debug 1.0 Spec supports hardware to realize simultaneous halt/resume, this problem will still exist. So my idea is to let openocd report error when user config cross-cluster smp (by another patch), or to support similar imperfect cross-cluster smp implementations(by this patch).

src/target/riscv/riscv-013.c
```diff
@@ -2027,6 +2026,10 @@ static int examine(struct target *target)
 		return ERROR_FAIL;
 	}
 
+	/* The RISC-V hartid is sequential, and the index of each hart
+	 * on the Debug Module should start at 0 and be contiguous. */
+	info->index = target->coreid % dm->hart_count;
```
timsifive, line 2035: Doesn't this assume that each DM has the same number of harts?

timsifive (commented):
I think you can pass `-coreid` to `target create` commands, which means you don't need this change. Does that not work?

lz-bro:
> I think you can pass `-coreid` to `target create` commands, which means you don't need this change. Does that not work?

This patch about Support for smp in multiple DMs: #907, cross-cluster smp cannot be realized by passing `-coreid` to `target create` commands.

src/target/riscv/riscv-013.c
lz-bro: Thanks for your reminder,  I didn't consider the situation that each DM has different number of harts, if this situation happens, user configuration will become unfriendly. So I currently have two ideas to solve it:
**Config modification**
config like
```
# cluster 0
target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -index 0 -rtos hwthread
target create $_TARGETNAME_1 riscv -chain-position $_CHIPNAME.cpu -coreid 1 -index 1

# cluster 1
target create $_TARGETNAME_2 riscv -chain-position $_CHIPNAME.cpu -coreid 2 -index 0 -dbgbase 0x400 -rtos hwthread
target create $_TARGETNAME_3 riscv -chain-position $_CHIPNAME.cpu -coreid 3 -index 1 -dbgbase 0x400

target smp $_TARGETNAME_0 $_TARGETNAME_1 $_TARGETNAME_2 $_TARGETNAME_3
```
Modify the 'jim_target_create' function to get the target->index.

info->index get
https://github.com/riscv/riscv-openocd/blob/699eecaab434337dc3915171606b0548c48c6d51/src/target/riscv/riscv-013.c#L1900
changed to

```
info->index = target->index;
```
**Get different threadids through hartid**
config like
```
# cluster 0
target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -rtos hwthread

# cluster 1
target create $_TARGETNAME_2 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -dbgbase 0x400 -rtos hwthread
target create $_TARGETNAME_3 riscv -chain-position $_CHIPNAME.cpu -coreid 1 -dbgbase 0x400

target smp $_TARGETNAME_0 $_TARGETNAME_1 $_TARGETNAME_2 $_TARGETNAME_3
```
Make appropriate modifications to the "threadid_from_target' function, I still don’t know how to modify it
```
static inline threadid_t threadid_from_target(const struct target *target)
{
	#return target->coreid + 1;
        ...
}
```
thread id should be 0/1/2/3

timsifive:
> This patch about Support for smp in multiple DMs: #907, cross-cluster smp cannot be realized by passing -coreid to target create commands.

That's not obvious to me. What happens when you pass coreid like this:
```
# cluster 0
target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -rtos hwthread

# cluster 1
target create $_TARGETNAME_2 riscv -chain-position $_CHIPNAME.cpu -coreid 2 -dbgbase 0x400 -rtos hwthread
target create $_TARGETNAME_3 riscv -chain-position $_CHIPNAME.cpu -coreid 3 -dbgbase 0x400
```

lz-bro:
> > This patch about Support for smp in multiple DMs: #907, cross-cluster smp cannot be realized by passing -coreid to target create commands.
> 
> That's not obvious to me. What happens when you pass coreid like this:
> 
> ```
> # cluster 0
> target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -rtos hwthread
> target create $_TARGETNAME_1 riscv -chain-position $_CHIPNAME.cpu -coreid 1
> 
> # cluster 1
> target create $_TARGETNAME_2 riscv -chain-position $_CHIPNAME.cpu -coreid 2 -dbgbase 0x400 -rtos hwthread
> target create $_TARGETNAME_3 riscv -chain-position $_CHIPNAME.cpu -coreid 3 -dbgbase 0x400
> ```

info->index stores the index of each hart on the DM. For each DM these must start at 0 and be contiguous. So that then means that to select the first hart on cluster 1 (DM 1), I write 2 to dmcontrol.hartsel, it won't work. The first hart in a DM must be selected by dmcontrol.hartsel=0, the second by dmcontrol.hartsel=1, and so on.

timsifive:
OK, so what happens if you do:
```
# cluster 0
target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -rtos hwthread

# cluster 1
target create $_TARGETNAME_2 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -dbgbase 0x400 -rtos hwthread
target create $_TARGETNAME_3 riscv -chain-position $_CHIPNAME.cpu -coreid 1 -dbgbase 0x400
```

lz-bro:
> OK, so what happens if you do:
> 
> ```
> # cluster 0
> target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -rtos hwthread
> target create $_TARGETNAME_1 riscv -chain-position $_CHIPNAME.cpu -coreid 1
> 
> # cluster 1
> target create $_TARGETNAME_2 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -dbgbase 0x400 -rtos hwthread
> target create $_TARGETNAME_3 riscv -chain-position $_CHIPNAME.cpu -coreid 1 -dbgbase 0x400
> ```

```
target smp $_TARGETNAME_0 $_TARGETNAME_1
target smp $_TARGETNAME_2 $_TARGETNAME_3
```
If I only smp harts in the a DM, 2 gdb were used to debug, nothing happens

`target smp $_TARGETNAME_0 $_TARGETNAME_1$_TARGETNAME_2 $_TARGETNAME_3`
But If I smp them all together, 1 gdb can only see 2 core in dm1
```
(gdb) info threads
  Id   Target Id                                                            Frame
* 1    Thread 1 "riscv.dm1cpu0" (Name: riscv.dm1cpu0, state: debug-request) 0x0000000000001c6c in ?? ()
  2    Thread 2 "riscv.dm1cpu1" (Name: riscv.dm1cpu1, state: debug-request) 0x000000000000769a in ?? ()
```
So the harts in multiple DMs cannot be smp to halt/resume together, and I want to get some advice about this

timsifive:
Hmm.... the comment for target.core_id says `which device on the TAP?`. That implies that you could have multiple targets with the same coreid if they are on different TAPs. `struct gdb_service` in target.h then implies that gdb requires a unique coreid for each target. That seems like a bug because there can be multiple TAPs, although perhaps I'm missing some code. I only took a quick look.

Is that fixable? That seems like a better fix than RISC-V-specific heuristics.

lz-bro:
> Hmm.... the comment for target.core_id says `which device on the TAP?`. That implies that you could have multiple targets with the same coreid if they are on different TAPs. `struct gdb_service` in target.h then implies that gdb requires a unique coreid for each target. That seems like a bug because there can be multiple TAPs, although perhaps I'm missing some code. I only took a quick look.
> 
> Is that fixable? That seems like a better fix than RISC-V-specific heuristics.

As I understand, target->coreid is used to specify which device on the TAP and define a unique threadid (dgb)  for each target. If I smp all target on multiple clusters, there is a bug. 
I gave some ideas in previous comments.

lz-bro:
thanks, we will try to fix it in upstream openocd

PR #911, opened 2023-08-29
From upstream

timsifive:
Merge up to commit 'dfbbfac4d72e247e8094a49c8573b2f49689b6d5' from upstream


src/rtos/rtos_nuttx_stackings.c
```diff
@@ -0,0 +1,110 @@
...
+	{ GDB_REGNO_PC, 0x00, 32 },
+};
+
+const struct rtos_register_stacking nuttx_riscv_stacking = {
+	.stack_registers_size = 33 * 4,
+	.stack_growth_direction = -1,
+	.num_output_registers = 33,
+	.calculate_process_stack = rtos_generic_stack_align8,
+	.register_offsets = nuttx_stack_offsets_riscv,
+};
```
MarekVCodasip, line 110: I know this came from upstream, not from riscv-openocd, but this seems to assume only RV32 cores not RV64. nuttx says it is scaleable from 8-bit to 64-bit. Just FYI.

MarekVCodasip:
It seems to be that the only thing affecting RISC-V in non-marginal way is the riscv stacking info.

See the comment below. Otherwise it looks alright.

src/rtos/rtos_nuttx_stackings.c
TommyMurphyTM1234: > I know this came from upstream, not from riscv-openocd, but this seems to assume only RV32 cores not RV64. 

Or RV32E?

timsifive: Good point. Maybe something for the nuttx team. Possibly 64-bit is already supported upstream, just not yet in these changes.

aap-sc: > Good point. Maybe something for the nuttx team. Possibly 64-bit is already supported upstream, just not yet in these changes.

There is an explicit comment in `src/rtos/nuttx.c` : 

> /* Only 32-bit CPUs are supported by the current implementation.  Supporting
>  * other CPUs will require reading this information from the target and
>  * adapting the code accordingly.
> */

just FYI

timsifive:
@JanMatCodasip can you look at this?

JanMatCodasip:
> @JanMatCodasip can you look at this?

We have already checked it together with @MarekVCodasip (above), and it looks all right. 

Marking as "approved".

JanMatCodasip (approved):

timsifive:
> We have already checked it together with @MarekVCodasip (above), and it looks all right.

I saw that, but Marek's comment doesn't count for github as satisfying the merge requirement.

Thanks.

PR #912, opened 2023-09-06
target/riscv_semihosting: Make the unknown operation number an error

MarekVCodasip:
Previously, an unknown semihosting operation number was logged as debug. This patch changes it and few other places to be logged as error instead.

Change-Id: I83cae5ca1e3daed440f92b08bd372bfffbbad63c

### DIFF
--- a/src/target/riscv/riscv_semihosting.c
+++ b/src/target/riscv/riscv_semihosting.c
@@ -109,13 +109,13 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
 
 		result = riscv_get_register(target, &r0, GDB_REGNO_A0);
 		if (result != ERROR_OK) {
-			LOG_TARGET_DEBUG(target, "   -> ERROR (couldn't read a0)");
+			LOG_TARGET_ERROR(target, "Could not read semihosting operation code (register a0)");
 			return SEMIHOSTING_ERROR;
 		}
 
 		result = riscv_get_register(target, &r1, GDB_REGNO_A1);
 		if (result != ERROR_OK) {
-			LOG_TARGET_DEBUG(target, "   -> ERROR (couldn't read a1)");
+			LOG_TARGET_ERROR(target, "Could not read semihosting operation code (register a1)");
 			return SEMIHOSTING_ERROR;
 		}
 
@@ -134,7 +134,7 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
 			}
 		} else {
 			/* Unknown operation number, not a semihosting call. */
-			LOG_TARGET_DEBUG(target, "   -> NONE (unknown operation number)");
+			LOG_TARGET_ERROR(target, "Unknown semihosting operation requested (op = 0x%x)", semihosting->op);
 			return SEMIHOSTING_NONE;
 		}
 	}


src/target/riscv/riscv_semihosting.c
```diff
@@ -109,13 +109,13 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
 
 		result = riscv_get_register(target, &r0, GDB_REGNO_A0);
 		if (result != ERROR_OK) {
-			LOG_TARGET_DEBUG(target, "   -> ERROR (couldn't read a0)");
+			LOG_TARGET_ERROR(target, "   -> ERROR (couldn't read a0)");
```
JanMatCodasip: Now that these messages become more user-visible, please consider printing a full sentence. For example:

```suggestion
			LOG_TARGET_ERROR(target, "Could not read semihosting operation code (register a0)");
```

src/target/riscv/riscv_semihosting.c
```diff
@@ -109,13 +109,13 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
...
 			return SEMIHOSTING_ERROR;
 		}
 
 		result = riscv_get_register(target, &r1, GDB_REGNO_A1);
 		if (result != ERROR_OK) {
-			LOG_TARGET_DEBUG(target, "   -> ERROR (couldn't read a1)");
+			LOG_TARGET_ERROR(target, "   -> ERROR (couldn't read a1)");
```
JanMatCodasip:
```suggestion
			LOG_TARGET_ERROR(target, "Could not read semihosting address (register a1)");
```

src/target/riscv/riscv_semihosting.c
```diff
@@ -134,7 +134,7 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
 			}
 		} else {
 			/* Unknown operation number, not a semihosting call. */
-			LOG_TARGET_DEBUG(target, "   -> NONE (unknown operation number)");
+			LOG_TARGET_ERROR(target, "   -> NONE (unknown operation number)");
```
JanMatCodasip:
```suggestion
			LOG_TARGET_ERROR(target, "Unknown semihosting operation requested (op = 0x%x)", semihosting->op);
```

JanMatCodasip (changes requested):
This looks OK. Please just make the error messages more clear to the user.

JanMatCodasip:
The main reason for the change is to notify the user in case an unknown semihosting operation is requested -- that is the third error message. 

The previous two errors (unable to read a0 or a1) should be very unlikely to happen.

TommyMurphyTM1234:
> The main reason for the change is to notify the user in case an unknown semihosting operation is requested -- that is the third error message.
> 
> The previous two errors (unable to read a0 or a1) should be very unlikely to happen.

Surely the failure to read `a0` or `a1` should have an error message that reflects that situation rather than implying that it's an "unknown" semihosting request?

Although - as you imply - if those registers cannot be read then there are most likely much bigger problems with the target and this code may never even be hit!

JanMatCodasip:
> Surely the failure to read `a0` or `a1` should have an error message [...]

And yes, those error messages are part of this change, too.

src/target/riscv/riscv_semihosting.c
MarekVCodasip: Done

src/target/riscv/riscv_semihosting.c
MarekVCodasip: Done

src/target/riscv/riscv_semihosting.c
MarekVCodasip: Done

JanMatCodasip (approved):

timsifive (approved):

PR #913, opened 2023-09-12
From upstream

timsifive:
Merge up to commit 'ee31f1578a333a75737bc5b183cd4ae98cdaf798' from upstream
    
Conflicts:
* Makefile.am
* jimtcl
* src/helper/Makefile.am
* src/rtos/rtos.c
* src/rtos/rtos.h
* src/rtos/rtos_standard_stackings.c


JanMatCodasip:
@timsifive I have started looking at this and have not come across anything suspicious so far. I would still like to spend little more time on this. Please expect the final review from me on Monday.

JanMatCodasip:
@timsifive I have noticed differences in `.travis.yml` when comparing against the upstream. I assume riscv-openocd no longer uses Travis CI.

If that's the case, I recommend to simply replace `.travis.yml` with the upstream version so that there are less diffs to look at next time.

JanMatCodasip:
@timsifive Can we get rid of the hardcoded `build_remote_bitbang=yes` in configure.ac? Instead, we should be passing `--enable-remote-bitbang` explicitly to `./configure`, when needed.

My goal is to reduce unnecessary differences against the upstream.

```diff
--- ../openocd-offic/configure.ac	2023-09-18 09:44:01.021065136 +0200
+++ configure.ac	2023-09-18 09:43:27.411748377 +0200
@@ -374,7 +375,7 @@
 
 AC_ARG_ENABLE([remote-bitbang],
   AS_HELP_STRING([--enable-remote-bitbang], [Enable building support for the Remote Bitbang jtag driver]),
-  [build_remote_bitbang=$enableval], [build_remote_bitbang=no])
+  [build_remote_bitbang=$enableval], [build_remote_bitbang=yes])
 
 AS_CASE(["${host_cpu}"],
   [i?86|x86*], [],

```

JanMatCodasip:
@timsifive Now that the checkpatch jobs in the CI are more clever (filterdiff used), do we still need to comment out the below piece of code in checkpatch.pl? 

Change: https://github.com/riscv/riscv-openocd/commit/aa8d30a4467182205f5bb1f9a33f99b67b230800

Could you please try checkpatch without that custom change?

JanMatCodasip (changes requested):
This looks fine to me. I have just three smaller suggestions (see my previous comments) where we could shake off several diffs against the vanilla upstream.

timsifive:
> @timsifive Can we get rid of the hardcoded `build_remote_bitbang=yes` in configure.ac? Instead, we should be passing `--enable-remote-bitbang` explicitly to `./configure`, when needed.

I would like to keep remote-bitbang enabled by default in this branch, because that's how people use it with spike. There's no real downside to keeping it besides the difference and it reduces the number of user complaints we get.

timsifive:
> I recommend to simply replace `.travis.yml` with the upstream version so that there are less diffs to look at next time.

> Now that the checkpatch jobs in the CI are more clever (filterdiff used), do we still need to comment out the below piece of code in checkpatch.pl?

Agreed. I'd like to make those changes separately, to keep the merge as plain as possible in case there's some unforeseen fall-out.

JanMatCodasip:
Thank you for explaining, understood. Let's keep it as is, then.

JanMatCodasip:
Sure, please feel free to do these in a separate merge request.

Please consider this merge request approved from me (due to being OoO until
Oct 9, I cannot submit the usual github approval).

I am glad that riscv-openocd is getting more synchronized with the vanilla
upstream, and will happily continue to support this process.

timsifive:
I filed #922 for the cleanup issues.

PR #915, opened 2023-09-13
target/riscv: Remove duplicate read PC message

timsifive:
Change-Id: Ie085758e3cf193f2671ea53fb82fd401d0c52d86

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -4702,7 +4702,6 @@ static int riscv013_get_register(struct target *target,
 	if (rid == GDB_REGNO_PC) {
 		/* TODO: move this into riscv.c. */
 		result = register_read_direct(target, value, GDB_REGNO_DPC);
-		LOG_TARGET_DEBUG(target, "read PC from DPC: 0x%" PRIx64, *value);
 	} else if (rid == GDB_REGNO_PRIV) {
 		uint64_t dcsr;
 		/* TODO: move this into riscv.c. */


src/target/riscv/riscv-013.c
```diff
@@ -4702,7 +4702,8 @@ static int riscv013_get_register(struct target *target,
 	if (rid == GDB_REGNO_PC) {
 		/* TODO: move this into riscv.c. */
 		result = register_read_direct(target, value, GDB_REGNO_DPC);
-		LOG_TARGET_DEBUG(target, "read PC from DPC: 0x%" PRIx64, *value);
+		if (result == ERROR_OK)
+			LOG_TARGET_DEBUG(target, "read PC from DPC: 0x%" PRIx64, *value);
```
JanMatCodasip: If the read succeeds, the register name and value is already printed inside `register_read_direct()` for any register (riscv013.c on line 1722).

So you may consider removing this extra duplicate print of PC value altogether. I would actually prefer that -- for the sake of consistency.

However, if decide to keep the print and have the PC value in the log twice (for extra visibility), then the change looks all right to me.

timsifive: Removed. Thanks.

en-sc:
Is this neccesary? `GDB_REGNO_PC != GDB_REGNO_DPC`. I would like for it to be represented in `-d3` log in some form. Maybe it would be sufficient to only remove the printing of the value, but leave the message?

timsifive:
If we're printing the message, we should also print the value.

I don't really care. I just don't want to print the value when it might be uninitialized.

JanMatCodasip (approved):

PR #917, opened 2023-09-14
provide riscv-specific controls to disable triggers from being used for watchpoints

kr-sc:
Add a new riscv specific commands to disable triggers

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -10871,6 +10871,21 @@ Control dcsr.ebreaku. When on (default), U-mode ebreak instructions trap to
 OpenOCD. When off, they generate a breakpoint exception handled internally.
 @end deffn
 
+The commands below can be used to prevent OpenOCD from using certain RISC-V trigger features.
+For example in cases when there are known issues in the target hardware.
+
+@deffn {Command} {riscv set_enable_eq_match_trigger} [on|off]
+When on (default), allow OpenOCD to use exact-match triggers in watchpoints.
+@end deffn
+
+@deffn {Command} {riscv set_enable_napot_trigger} [on|off]
+When on (default), allow OpenOCD to use NAPOT trigger in watchpoints.
+@end deffn
+
+@deffn {Command} {riscv set_enable_ge_lt_trigger} [on|off]
+When on (default), allow OpenOCD to use a pair of chained less-than & greater-than triggers in watchpoints.
+@end deffn
+
 @subsection RISC-V Authentication Commands
 
 The following commands can be used to authenticate to a RISC-V system. Eg.  a
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -857,15 +857,16 @@ static struct match_triggers_tdata1_fields fill_match_triggers_tdata1_fields_t6(
 	return result;
 }
 
-static int maybe_add_trigger_t2_t6(struct target *target,
+static int maybe_add_trigger_t2_t6_for_wp(struct target *target,
 		struct trigger *trigger, struct match_triggers_tdata1_fields fields)
 {
-	int ret = ERROR_OK;
+	RISCV_INFO(r);
+	int ret = ERROR_FAIL;
 
-	if (!trigger->is_execute && trigger->length > 1) {
+	if (trigger->length > 0) {
 		/* Setting a load/store trigger ("watchpoint") on a range of addresses */
 
-		if (can_use_napot_match(trigger)) {
+		if (r->enable_napot_trigger && can_use_napot_match(trigger)) {
 			LOG_TARGET_DEBUG(target, "trying to setup NAPOT match trigger");
 			struct trigger_request_info napot = {
 				.tdata1 = fields.common | fields.size.any |
@@ -877,52 +878,58 @@ static int maybe_add_trigger_t2_t6(struct target *target,
 			if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
 				return ret;
 		}
-		LOG_TARGET_DEBUG(target, "trying to setup GE+LT chained match trigger pair");
-		struct trigger_request_info ge_1 = {
-			.tdata1 = fields.common | fields.size.any | fields.chain.enable |
-				fields.match.ge,
-			.tdata2 = trigger->address,
-			.tdata1_ignore_mask = fields.tdata1_ignore_mask
-		};
-		struct trigger_request_info lt_2 = {
-			.tdata1 = fields.common | fields.size.any | fields.chain.disable |
-				fields.match.lt,
-			.tdata2 = trigger->address + trigger->length,
-			.tdata1_ignore_mask = fields.tdata1_ignore_mask
-		};
-		ret = try_setup_chained_match_triggers(target, trigger, ge_1, lt_2);
-		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
-			return ret;
 
-		LOG_TARGET_DEBUG(target, "trying to setup LT+GE chained match trigger pair");
-		struct trigger_request_info lt_1 = {
-			.tdata1 = fields.common | fields.size.any | fields.chain.enable |
-				fields.match.lt,
-			.tdata2 = trigger->address + trigger->length,
-			.tdata1_ignore_mask = fields.tdata1_ignore_mask
-		};
-		struct trigger_request_info ge_2 = {
-			.tdata1 = fields.common | fields.size.any | fields.chain.disable |
-				fields.match.ge,
-			.tdata2 = trigger->address,
-			.tdata1_ignore_mask = fields.tdata1_ignore_mask
-		};
-		ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
-		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
-			return ret;
+		if (r->enable_ge_lt_trigger) {
+			LOG_TARGET_DEBUG(target, "trying to setup GE+LT chained match trigger pair");
+			struct trigger_request_info ge_1 = {
+				.tdata1 = fields.common | fields.size.any | fields.chain.enable |
+					fields.match.ge,
+				.tdata2 = trigger->address,
+				.tdata1_ignore_mask = fields.tdata1_ignore_mask
+			};
+			struct trigger_request_info lt_2 = {
+				.tdata1 = fields.common | fields.size.any | fields.chain.disable |
+					fields.match.lt,
+				.tdata2 = trigger->address + trigger->length,
+				.tdata1_ignore_mask = fields.tdata1_ignore_mask
+			};
+			ret = try_setup_chained_match_triggers(target, trigger, ge_1, lt_2);
+			if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+				return ret;
+
+			LOG_TARGET_DEBUG(target, "trying to setup LT+GE chained match trigger pair");
+			struct trigger_request_info lt_1 = {
+				.tdata1 = fields.common | fields.size.any | fields.chain.enable |
+					fields.match.lt,
+				.tdata2 = trigger->address + trigger->length,
+				.tdata1_ignore_mask = fields.tdata1_ignore_mask
+			};
+			struct trigger_request_info ge_2 = {
+				.tdata1 = fields.common | fields.size.any | fields.chain.disable |
+					fields.match.ge,
+				.tdata2 = trigger->address,
+				.tdata1_ignore_mask = fields.tdata1_ignore_mask
+			};
+			ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
+			if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+				return ret;
+		}
 	}
-	LOG_TARGET_DEBUG(target, "trying to setup equality match trigger");
-	struct trigger_request_info eq = {
-		.tdata1 = fields.common | fields.size.any | fields.chain.disable |
-			fields.match.eq,
-		.tdata2 = trigger->address,
-		.tdata1_ignore_mask = fields.tdata1_ignore_mask
-	};
-	ret = try_setup_single_match_trigger(target, trigger, eq);
-	if (ret != ERROR_OK)
-		return ret;
 
-	if (trigger->length > 1) {
+	if (r->enable_equality_match_trigger) {
+		LOG_TARGET_DEBUG(target, "trying to setup equality match trigger");
+		struct trigger_request_info eq = {
+			.tdata1 = fields.common | fields.size.any | fields.chain.disable |
+				fields.match.eq,
+			.tdata2 = trigger->address,
+			.tdata1_ignore_mask = fields.tdata1_ignore_mask
+		};
+		ret = try_setup_single_match_trigger(target, trigger, eq);
+		if (ret != ERROR_OK)
+			return ERROR_FAIL;
+	}
+
+	if (ret == ERROR_OK && trigger->length > 1) {
 		LOG_TARGET_DEBUG(target, "Trigger will match accesses at address 0x%" TARGET_PRIxADDR
 				", but may not match accesses at addresses in the inclusive range from 0x%"
 				TARGET_PRIxADDR " to 0x%" TARGET_PRIxADDR ".", trigger->address,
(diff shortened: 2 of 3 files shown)


src/target/riscv/riscv.c
```diff
@@ -141,6 +141,10 @@ bool riscv_ebreaku = true;
 
 bool riscv_enable_virtual;
 
+bool enable_equality_match_trigger = true;
+bool enable_napot_trigger = true;
+bool enable_ge_lt_trigger = true;
+
```
JanMatCodasip: Ideally, these configuration flags should be per-target and not global -- that is, they should go into `struct riscv_info`. 

(The same applies to the already existing `riscv_ebreakX` bools above. They are global not on purpose, but because no one has moved them yet inside the riscv_info structure.)

JanMatCodasip:
I have done only a superficial review so far.

What is the motivation for having these new configuration options, please? If the target supports the given trigger types, why not to always use them? Why would someone force-disable that functionality when the target provides it?

kr-sc:
> What is the motivation for having these new configuration options, please? If the target supports the given trigger types, why not to always use them? Why would someone force-disable that functionality when the target provides it?

I think, it will be usefull for hardware debugging (for testing particular type of trigger).

JanMatCodasip:
> I think, it will be useful for hardware debugging (for testing particular type of trigger).

If you wish to test the HW with the help of OpenOCD, I recommend to stick with the existing low-level commands (`poll off`, `riscv dmi_read`, `riscv dmi_write`, `reg`, ...). You can for example write own TCL script that will perform the requested operations (e.g. set some triggers) and check the outcome. The tests can even be scripted in python -- you can connect through the telnet or TCL channels and execute commands this way. This approach provides more flexibility and allows to test virtually any debug feature. 

I would rather not add any TCL commands that have no other use case than testing.

kr-sc:
> If you wish to test the HW with the help of OpenOCD, I recommend to stick with the existing low-level commands (poll off, riscv dmi_read, riscv dmi_write, reg, ...). You can for example write own TCL script that will perform the requested operations (e.g. set some triggers) and check the outcome. The tests can even be scripted in python -- you can connect through the telnet or TCL channels and execute commands this way. This approach provides more flexibility and allows to test virtually any debug feature.

If we think like that we can write everything with low-level jtag commands (and it is possible, I think), but I think that we want to extend existing codebase to simplify control of debugged unit. 

We have a problem with our hardware: `equality_match trigger` is supported, but it compare only first byte of address of memory access request: for example, if you set a wp on `0x100` with size 4 and try to read from `0x98` with size 4, wp will not hit. So, we want to add an option to turn off usage of this trigger for wp (and at the same time we made a same options for other types of triggers).

JanMatCodasip:
Thank you for sharing the motivation for this feature.

> If we think like that we can write everything with low-level jtag commands (and it is possible, I think) ...

If it would be a hw-testing-only feature, I maintain the opinion that these should not be added to OpenOCD. Test cases for hardware should be developed separately. And if someone wishes to execute them through OpenOCD, the interface to do so is already available (as mentioned in the previous comment).

> ... but I think that we want to extend existing codebase to simplify control of debugged unit.

I agree as long as there are real use cases for those commands or options (beyond testing).

> We have a problem with our hardware: equality_match trigger is supported, but ...

Known issues in existing hardware - that is IMO a good argument to include the feature :+1:

src/target/riscv/riscv.c
kr-sc: Addressed

src/target/riscv/riscv.c
```diff
@@ -880,52 +881,58 @@ static int maybe_add_trigger_t2_t6(struct target *target,
...
+				.tdata2 = trigger->address,
+				.tdata1_ignore_mask = fields.tdata1_ignore_mask
+			};
+			ret = try_setup_chained_match_triggers(target, trigger, lt_1, ge_2);
+			if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+				return ret;
+		}
+	}
+
+	if (r->enable_equality_match_trigger) {
```
JanMatCodasip, line 919: I am wondering if we wish (need) to have a configuration option for enabling the equality match trigger. It is the most basic trigger type. 

What would be the use case for disabling the basic equality match trigger and leave some of the more advanced triggers enabled? If the user does so, almost the whole watchpoint feature will still remain non-functional. The user may very well just stop using watchpoints to achieve the same effect.

src/target/riscv/riscv.c
```diff
@@ -4501,6 +4583,27 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
+		.mode = COMMAND_CONFIG,
+		.usage = "[on|off]",
+		.help = "Allow OpenOCD to use NAPOT trigger in bp/wp"
+	},
+	{
+		.name = "ge_lt_trigger_enable",
+		.handler = riscv_ge_lt_trigger_enable,
+		.mode = COMMAND_CONFIG,
+		.usage = "[on|off]",
+		.help = "Allow OpenOCD to use GE/LT triggers in bp/wp"
```
JanMatCodasip: There is inconsistency between the help and the implementation: The help say `Allow OpenOCD to use ... in bp/wp` but only watchpoints are affected. 

What is your intention? To affect just watchpoints or also breakpoints?

src/target/riscv/riscv.c
```diff
@@ -4501,6 +4583,27 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
+	},
```
JanMatCodasip: When commands are added or changed, the manual of OpenOCD (`doc/openocd.texi`) should be updated, too.

src/target/riscv/riscv.c
```diff
@@ -941,7 +948,31 @@ static int maybe_add_trigger_t2_t6(struct target *target,
...
+	return try_setup_single_match_trigger(target, trigger, eq);
+}
+
+static int maybe_add_trigger_t2_t6(struct target *target,
+		struct trigger *trigger, struct match_triggers_tdata1_fields fields)
+{
+	if (trigger->is_execute)
+		return maybe_add_trigger_t2_t6_for_bp(target, trigger, fields);
+
+	return maybe_add_trigger_t2_t6_for_wp(target, trigger, fields);
```

To make the code easier to read, you can add an `else` branch to clearly show the reader that both of the calls are logically on the same level of importance.

Assertions can be added to catch programmer mistakes and help readers to understand the code.

```suggestion
	if (trigger->is_execute) {
		assert(!trigger->is_load && !trigger->is_store);
		return maybe_add_trigger_t2_t6_for_bp(target, trigger, fields);
	} else {
		assert(trigger->is_load || trigger->is_store);
		return maybe_add_trigger_t2_t6_for_wp(target, trigger, fields);	
	}
```

src/target/riscv/riscv.c
```diff
@@ -4501,6 +4583,27 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
+	},
+	{
+		.name = "napot_trigger_enable",
+		.handler = riscv_napot_trigger_enable,
+		.mode = COMMAND_CONFIG,
+		.usage = "[on|off]",
+		.help = "Allow OpenOCD to use NAPOT trigger in bp/wp"
+	},
+	{
+		.name = "ge_lt_trigger_enable",
```
JanMatCodasip: The existing commands follow the naming scheme `set_something` or `set_enable_something` but the new commands are named differently.

I am fine with any naming convention but would like this to be consistent for all RISC-V commands, and would like this to be accepted in upstream without the need to redo it.

@timsifive This is now a recurring topic. What is your opinion or preference?

src/target/riscv/riscv.c
```diff
@@ -4501,6 +4583,27 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.help = "Execute a sequence of 32-bit instructions using the program buffer. "
 			"The final ebreak instruction is added automatically, if needed."
 	},
+	{
+		.name = "equality_match_trigger_enable",
+		.handler = riscv_equality_match_trigger_enable,
```
JanMatCodasip: This name is already quite long, so if you like, I believe you can use just `eq` instead of the full word `equality`.

```suggestion
		.name = "eq_match_trigger_enable",
		.handler = riscv_eq_match_trigger_enable,
```

JanMatCodasip (dismissed):

src/target/riscv/riscv.c
kr-sc: > I am wondering if we wish (need) to have a configuration option for enabling the equality match trigger.

We need this option, because we have problem with this type of trigger (as I write above).

> If the user does so, almost the whole watchpoint feature will still remain non-functional.

No, we can still use watchpoints with size more than 1 byte, if other trigger types are available in hardware.

JanMatCodasip: Ok, thank you for explaining it again. Understood.

src/target/riscv/riscv.c
kr-sc: I can't add this else statements because of `tools/checkpatch.sh`:

```
WARNING:UNNECESSARY_ELSE: else is not generally useful after a break or return
#192: FILE: src/target/riscv/riscv.c:975:
+		return maybe_add_trigger_t2_t6_for_bp(target, trigger, fields);
+	} else {

```

kr-sc: `assert`'s added.

src/target/riscv/riscv.c
kr-sc: There was only one general function for setting trigger for wp and bp, so this options effected on wp and bp. Later I split this function info two (one for wp and one for bp) and forgot to change help message.
Addressed

src/target/riscv/riscv.c
kr-sc: Addressed

src/target/riscv/riscv.c
kr-sc: Addressed

doc/openocd.texi
```diff
@@ -10871,6 +10871,18 @@ Control dcsr.ebreaku. When on (default), U-mode ebreak instructions trap to
...
+@end deffn
+
+@deffn {Command} {riscv napot_trigger_enable} [on|off]
+When on, allow OpenOCD to use NAPOT trigger in wp.
+@end deffn
+
+@deffn {Command} {riscv ge_lt_trigger_enable} [on|off]
+When on, allow OpenOCD to use GE/LT triggers in wp.
+@end deffn
+
```
JanMatCodasip: I have tried to make the text more clear to the reader. Let me know what you think.

```suggestion
The commands below can be used to prevent OpenOCD from using certain RISC-V trigger features, for example in cases when there are known issues in the target hardware.

@deffn {Command} {riscv eq_match_trigger_enable} [on|off]
When on (default), allow OpenOCD to use exact-match triggers in watchpoints.
@end deffn

@deffn {Command} {riscv napot_trigger_enable} [on|off]
When on (default), allow OpenOCD to use NAPOT trigger in watchpoints.
@end deffn

@deffn {Command} {riscv ge_lt_trigger_enable} [on|off]
When on (default), allow OpenOCD to use a pair of chained less-than & greater-than triggers in watchpoints.
@end deffn

```

src/target/riscv/riscv.c
```diff
@@ -4255,6 +4289,57 @@ COMMAND_HANDLER(riscv_exec_progbuf)
...
+	if (CMD_ARGC == 0) {
+		command_print(CMD, "equality match trigger enabled = %d", r->enable_equality_match_trigger);
+		return ERROR_OK;
+	} else if (CMD_ARGC != 1) {
+		LOG_ERROR("Command takes 0 or 1 parameters");
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	COMMAND_PARSE_ON_OFF(CMD_ARGV[0], r->enable_equality_match_trigger);
+	return ERROR_OK;
```
JanMatCodasip: Please, use the below structure, as already recommended in the other merge request: https://github.com/riscv/riscv-openocd/pull/918#discussion_r1332496576

```
	if (CMD_ARGC == 0) {
		/* Do something */
	} else if (CMD_ARGC == 1) {
		/* Do something else */
	} else {
		/* Report error */
	}
```

JanMatCodasip (commented):
This looks fine to me overall. I have few last suggestions.

I will not be able to make another review of this since I am out of office until Oct 9. This should not be an issue as this merge request is very close to completion and only minor items are left.

src/target/riscv/riscv.c
JanMatCodasip: > I can't add this else statements because ...

Ah, OK, understood. This is then one of the cases where checkpatch forces somewhat worse code style, IMO.

src/target/riscv/riscv.c
JanMatCodasip: Thanks.

src/target/riscv/riscv.c
JanMatCodasip: Or, maybe because of the checkpatch rules, this will be:

```
	if (CMD_ARGC == 0) {
		/* Do something */
	} else if (CMD_ARGC == 1) {
		/* Do something else */
	}

	/* Report error */
```

src/target/riscv/riscv.c
JanMatCodasip: Thanks.

doc/openocd.texi
kr-sc: Addressed

src/target/riscv/riscv.c
kr-sc: Addressed

timsifive:
I finally read this. General comments:
1. You can set custom triggers simply by writing tselect, tdata1, etc. OpenOCD won't overwrite triggers that are already set. No need to mess around with poll or dmi_read/dmi_write.
2. Work-around for buggy hardware is a good reason for configuration options, assuming they don't make the code too complex.

doc/openocd.texi
```diff
@@ -10871,6 +10871,20 @@ Control dcsr.ebreaku. When on (default), U-mode ebreak instructions trap to
 OpenOCD. When off, they generate a breakpoint exception handled internally.
 @end deffn
 
+The commands below can be used to prevent OpenOCD from using certain RISC-V trigger features, for example in cases when there are known issues in the target hardware.
```
timsifive: Can you wrap this line? No need to be strictly 80 columns, but 166 is a bit much.
(I know there are a few exceptions to this already, but going forward we're trying to keep the line lengths reasonable.)

src/target/riscv/riscv.c
timsifive: The existing set of commands are a bit of a mixture, but almost all of them include a verb. Then the fact that set_enable_virtual and set_enable_virt2phys already exists, means the most consistent naming for the new commands is set_enable_eq_match_trigger, set_enable_napot_trigger_enable, and set_enable_ge_lt_trigger.

timsifive (changes requested):
The implemented feature applies only to watchpoints, but the documentation and command names don't say anything about that.

Instead of making the command names even more cumbersome, I propose:
`riscv set_enable_eq_match_trigger [r][w][x]`

This lets a user set those trigger types for any combination of read, write, and execute. It might make the implementation a bit more straightforward even. I think you can simply modify try_setup_single_match_trigger() and try_setup_chained_match_triggers() to return error when a type is passed that doesn't fit with the r/w/x bits that are requested.

kr-sc:
> The implemented feature applies only to watchpoints, but the documentation and command names don't say anything about that.

`When on (default), allow OpenOCD to use exact-match triggers in watchpoints.` - description of `eq_match_trigger_enable` command from `doc/openocd.texi`

`When on, allow OpenOCD to use equality match trigger in wp` - help message of `eq_match_trigger_enable` command

kr-sc:
> Instead of making the command names even more cumbersome, I propose:
riscv set_enable_eq_match_trigger [r][w][x]

As I know, OpenOCD use only equality match trigger for bp. Do we really need command that disables, for example, NAPOT trigger for bp (e mode)?

kr-sc:
> This lets a user set those trigger types for any combination of read, write, and execute. It might make the implementation a bit more straightforward even. I think you can simply modify try_setup_single_match_trigger() and try_setup_chained_match_triggers() to return error when a type is passed that doesn't fit with the r/w/x bits that are requested.

We don't now type of trigger inside of `try_setup_single_match_trigger`, because we only pass `trigger_request_info` struct in it, so I think that it's better not to modify this function.

doc/openocd.texi
kr-sc: Addressed

src/target/riscv/riscv.c
kr-sc: Addressed

kr-sc:
Change output of `set_ebable_*_trigger` comamnds when it is called without args (like in https://github.com/riscv/riscv-openocd/pull/918#discussion_r1336318229)

JanMatCodasip (approved):
I have re-checked the code visually and it looks all right.

timsifive (approved):
> We don't know type of trigger inside of `try_setup_single_match_trigger`, because we only pass `trigger_request_info` struct in it, so I think that it's better not to modify this function.

The type is encoded in tdata1 of the trigger_request_info, though. It would be trivial to add a line there that checks the type and match fields, and then compares it against whatever the user configured.

Regardless. Since you've already implemented this, I guess it's good enough. In the future, can you open an issue or something when you're planning on adding new commands? That way we can have a little discussion on what commands make the most sense. I think my idea would have been just as much work to implement, and would have left us with a command that is more useful in the future, when OpenOCD implementation changes, or when we encounter hardware with different bugs.

kr-sc:
> In the future, can you open an issue or something when you're planning on adding new commands?

Yes, next time I'll do it.

src/target/riscv/riscv.c
```diff
@@ -877,52 +878,58 @@ static int maybe_add_trigger_t2_t6(struct target *target,
...
 
-		LOG_TARGET_DEBUG(target, "trying to setup LT+GE chained match trigger pair");
-		struct trigger_request_info lt_1 = {
-			.tdata1 = fields.common | fields.size.any | fields.chain.enable |
-				fields.match.lt,
-			.tdata2 = trigger->address + trigger->length,
-			.tdata1_ignore_mask = fields.tdata1_ignore_mask
-		};
-		struct trigger_request_info ge_2 = {
+		if (r->enable_ge_lt_trigger) {
```
aap-sc, line 882: @kr-sc , I'd like to request few additional improvements:

1. `enable_ge_lt_trigger` should be renamed to `wp_enable_ge_lt_trigger` (this applies to other controls as well). This is to clearly identify that these controls are for watchpoints only.
2. when `trigger->length == 0` we should allow for equality match to be used regardless of this control (otherwise we won't be able to set such wathcpoints)

aap-sc: @timsifive , @JanMatCodasip FYI

PR #918, opened 2023-09-14
openocd does not allow to query status of dcsr.ebreak{u,s,m}

kr-sc:
Extend riscv set_ebreak* commands.
Now it can be called without args to print current value.

riscv_ebreak* flags are moved to riscv_info struct.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -135,9 +135,6 @@ int riscv_command_timeout_sec = DEFAULT_COMMAND_TIMEOUT_SEC;
 int riscv_reset_timeout_sec = DEFAULT_RESET_TIMEOUT_SEC;
 
 static bool riscv_enable_virt2phys = true;
-bool riscv_ebreakm = true;
-bool riscv_ebreaks = true;
-bool riscv_ebreaku = true;
 
 bool riscv_enable_virtual;
 
@@ -3748,32 +3745,53 @@ COMMAND_HANDLER(riscv_set_enable_virt2phys)
 
 COMMAND_HANDLER(riscv_set_ebreakm)
 {
-	if (CMD_ARGC != 1) {
-		LOG_ERROR("Command takes exactly 1 parameter");
-		return ERROR_COMMAND_SYNTAX_ERROR;
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	if (CMD_ARGC == 0) {
+		command_print(CMD, "riscv_ebreakm enabled: %s", r->riscv_ebreakm ? "on" : "off");
+		return ERROR_OK;
+	} else if (CMD_ARGC == 1) {
+		COMMAND_PARSE_ON_OFF(CMD_ARGV[0], r->riscv_ebreakm);
+		return ERROR_OK;
 	}
-	COMMAND_PARSE_ON_OFF(CMD_ARGV[0], riscv_ebreakm);
-	return ERROR_OK;
+
+	LOG_ERROR("Command takes 0 or 1 parameters");
+	return ERROR_COMMAND_SYNTAX_ERROR;
 }
 
 COMMAND_HANDLER(riscv_set_ebreaks)
 {
-	if (CMD_ARGC != 1) {
-		LOG_ERROR("Command takes exactly 1 parameter");
-		return ERROR_COMMAND_SYNTAX_ERROR;
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	if (CMD_ARGC == 0) {
+		command_print(CMD, "riscv_ebreaks enabled: %s", r->riscv_ebreaks ? "on" : "off");
+		return ERROR_OK;
+	} else if (CMD_ARGC == 1) {
+		COMMAND_PARSE_ON_OFF(CMD_ARGV[0], r->riscv_ebreaks);
+		return ERROR_OK;
 	}
-	COMMAND_PARSE_ON_OFF(CMD_ARGV[0], riscv_ebreaks);
-	return ERROR_OK;
+
+	LOG_ERROR("Command takes 0 or 1 parameters");
+	return ERROR_COMMAND_SYNTAX_ERROR;
 }
 
 COMMAND_HANDLER(riscv_set_ebreaku)
 {
-	if (CMD_ARGC != 1) {
-		LOG_ERROR("Command takes exactly 1 parameter");
-		return ERROR_COMMAND_SYNTAX_ERROR;
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	if (CMD_ARGC == 0) {
+		command_print(CMD, "riscv_ebreaku enabled: %s", r->riscv_ebreaku ? "on" : "off");
+		return ERROR_OK;
+	} else if (CMD_ARGC == 1) {
+		COMMAND_PARSE_ON_OFF(CMD_ARGV[0], r->riscv_ebreaku);
+		return ERROR_OK;
 	}
-	COMMAND_PARSE_ON_OFF(CMD_ARGV[0], riscv_ebreaku);
-	return ERROR_OK;
+
+	LOG_ERROR("Command takes 0 or 1 parameters");
+	return ERROR_COMMAND_SYNTAX_ERROR;
 }
 
 COMMAND_HELPER(riscv_clear_trigger, int trigger_id, const char *name)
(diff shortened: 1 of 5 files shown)


src/target/riscv/riscv.c
```diff
@@ -4196,6 +4196,19 @@ COMMAND_HANDLER(handle_info)
...
+	riscv_print_info_line(CMD, "OpenOCD_flag_dcsr", "ebreaks", riscv_ebreaks);
+	riscv_print_info_line(CMD, "OpenOCD_flag_dcsr", "ebreaku", riscv_ebreaku);
+
+	riscv_reg_t dcsr;
+
+	const bool dcsr_read_success = riscv_get_register(target, &dcsr, GDB_REGNO_DCSR) == ERROR_OK;
+
+	riscv_print_info_line_if_available(CMD, "dcsr", "ebreakm", get_field(dcsr, CSR_DCSR_EBREAKM), dcsr_read_success);
+	riscv_print_info_line_if_available(CMD, "dcsr", "ebreaks", get_field(dcsr, CSR_DCSR_EBREAKS), dcsr_read_success);
+	riscv_print_info_line_if_available(CMD, "dcsr", "ebreaku", get_field(dcsr, CSR_DCSR_EBREAKU), dcsr_read_success);
```
JanMatCodasip: Is the purpose of these fields to tell you the current state of these `dcsr.ebreakX` bits?

Why do we need to add them to the riscv_info output? Why not to read the value of `dcsr` directly using one of the usual register read commands?

kr-sc: Firstly, I want to add current state of OpenOCD `riscv_ebreak*` flags. After that, I thought that user can be a little confused by these values (because there is an active dcsr state and the one that will be set upon next resume), so I decided to add actual `dcsr.ebreak*` values to `riscv info` output.

> Why not to read the value of dcsr directly using one of the usual register read commands?

Now it become much easier to check this values.

JanMatCodasip (dismissed):
#### (1) As for the riscv_ebreak_* OpenOCD's internal configuration flags:

I agree that there should be a way how the user can view the current settings (a "getter"). This applies to not only to commands `riscv set_ebreak*`, but in general to all configuration commands.

I am not in favor of putting it into `riscv info`. The output of `riscv info` would become a mixture of the detected hardware capabilities and current OpenOCD's settings, which are two different sets of information.

The getters had already been discussed a little here:
- https://github.com/riscv/riscv-openocd/pull/681#discussion_r808242682
- https://github.com/riscv/riscv-openocd/pull/681#discussion_r807615060

@timsifive  Since this topic has re-occurred, I recommend to discuss and agree on how the config getters should look like. I don't have strong preferences, but would like it to be:
- consistent (at least in the RISC-V target)
- intuitive and easy for the user, if possible
- done in such a way that it passes upstream review

#### (2) As for adding `dcsr.ebreak*` values to `riscv info`:

I recommend to not add any information that can be read equally easy by the existing commands (`reg`, `get_reg`) and possibly some bit-masking. 

Furthermore, items of `riscv info` would ideally be documented so that the user knows what they mean. That's another reason why to not add more items (unless the benefits of having them there are large enough).

kr-sc:
1) Remove flags values from `riscv info` output
2) Change fucntionality of `riscv set_ebreak*` commands: now it prints actual `riscv_ebreak*` flags value when it is called without args
3) Move flags values to `riscv_info` struct

src/target/riscv/riscv.c
```diff
@@ -3748,31 +3745,52 @@ COMMAND_HANDLER(riscv_set_enable_virt2phys)
 
 COMMAND_HANDLER(riscv_set_ebreakm)
 {
-	if (CMD_ARGC != 1) {
-		LOG_ERROR("Command takes exactly 1 parameter");
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	if (CMD_ARGC == 0) {
+		command_print(CMD, "riscv_ebreakm enabled = %d", r->riscv_ebreakm);
```
JanMatCodasip: Optionally - if you like:

```suggestion
		command_print(CMD, "riscv_ebreakm enabled = %s", r->riscv_ebreakm ? "yes" : "no");
```

src/target/riscv/riscv.c
```diff
@@ -4196,6 +4214,7 @@ COMMAND_HANDLER(handle_info)
 		riscv_enumerate_triggers(target) == ERROR_OK;
 	riscv_print_info_line_if_available(CMD, "hart", "trigger_count",
 				r->trigger_count, trigger_count_available);
+
```
JanMatCodasip: This whitespace change (blank line) can now be reverted.

src/target/riscv/riscv.c
```diff
@@ -3748,31 +3745,52 @@ COMMAND_HANDLER(riscv_set_enable_virt2phys)
...
+		command_print(CMD, "riscv_ebreakm enabled = %d", r->riscv_ebreakm);
+		return ERROR_OK;
+	} else if (CMD_ARGC != 1) {
+		LOG_ERROR("Command takes 0 or 1 parameters");
 		return ERROR_COMMAND_SYNTAX_ERROR;
 	}
-	COMMAND_PARSE_ON_OFF(CMD_ARGV[0], riscv_ebreakm);
+
+	COMMAND_PARSE_ON_OFF(CMD_ARGV[0], r->riscv_ebreakm);
 	return ERROR_OK;
```
JanMatCodasip: For readability, I recommend this structure. Applies to all the three commands.

```
	if (CMD_ARGC == 0) {
		command_print(CMD, "riscv_ebreakm enabled = %d", r->riscv_ebreakm);
		return ERROR_OK;
	} else if (CMD_ARGC == 1) {
		COMMAND_PARSE_ON_OFF(CMD_ARGV[0], r->riscv_ebreakm);
		return ERROR_OK;
	} else {
		LOG_ERROR("Command takes 0 or 1 parameters");
	}
```

src/target/riscv/riscv.c
```diff
@@ -3748,31 +3745,52 @@ COMMAND_HANDLER(riscv_set_enable_virt2phys)
 
 COMMAND_HANDLER(riscv_set_ebreakm)
 {
-	if (CMD_ARGC != 1) {
-		LOG_ERROR("Command takes exactly 1 parameter");
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	if (CMD_ARGC == 0) {
```
JanMatCodasip, line 3751: Please, update also the usage hint for the three commands:

```
Old: .usage = "on|off"
New: .usage = "[on|off]"
```

https://github.com/riscv/riscv-openocd/blob/caa92500be5a7bfdbc165bd9adced6dff2bcff91/src/target/riscv/riscv.c#L4474

JanMatCodasip, line 3751: Please update also the documentation for the commands in openocd.texi.

JanMatCodasip (commented):
This change looks all right to me overall - thanks. I have only few smaller suggestions.

Note that I am out of office until Oct 9 so I won't be able make another review of this change. But since it is almost complete, that should not be an issue. I'll leave the last quick check + merge to @timsifive.

src/target/riscv/riscv.c
kr-sc: I think, that now it much easier to parse output of this command than with your suggestion, so I would like to leave it as it is.

src/target/riscv/riscv.c
kr-sc: Removed. Addressed.

src/target/riscv/riscv.c
kr-sc: Addressed

src/target/riscv/riscv.c
kr-sc: Addressed

kr-sc: Addressed

src/target/riscv/riscv.c
timsifive: Let's follow existing OpenOCD commands. E.g. handle_poll_command() prints out a boolean value, and it uses:
`command_print(CMD, "background polling: %s", jtag_poll_get_enabled() ? "on" : "off");`

src/target/riscv/riscv.h
```diff
@@ -297,6 +297,10 @@ struct riscv_info {
 	yes_no_maybe_t vsew64_supported;
 
 	bool range_trigger_fallback_encountered;
+
+	bool riscv_ebreakm;
+	bool riscv_ebreaks;
+	bool riscv_ebreaku;
```
timsifive, line 303: I like this change. Can you mention it in the description of the PR?

timsifive (changes requested):

src/target/riscv/riscv.c
kr-sc: Addressed

src/target/riscv/riscv.h
kr-sc: Commit message updated. Addressed.

timsifive (approved):
Looks good. Thanks for this improvement.

PR #920, opened 2023-09-21
target/riscv: use cacheable read/write function to handle DCSR

lz-bro:
Avoid cache and memory inconsistencies on DCSR
Change-Id: c99efe3ec60daa6d7059e02f06a638ca18f08ab7

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -4844,6 +4844,18 @@ static int riscv_set_or_write_register(struct target *target,
 
 	keep_alive();
 
+	if (regid == GDB_REGNO_PC) {
+		return riscv_set_or_write_register(target, GDB_REGNO_DPC, value, write_through);
+	} else if (regid == GDB_REGNO_PRIV) {
+		riscv_reg_t dcsr;
+
+		if (riscv_get_register(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
+			return ERROR_FAIL;
+		dcsr = set_field(dcsr, CSR_DCSR_PRV, get_field(value, VIRT_PRIV_PRV));
+		dcsr = set_field(dcsr, CSR_DCSR_V, get_field(value, VIRT_PRIV_V));
+		return riscv_set_or_write_register(target, GDB_REGNO_DCSR, dcsr, write_through);
+	}
+
 	if (!target->reg_cache) {
 		assert(!target_was_examined(target));
 		LOG_TARGET_DEBUG(target,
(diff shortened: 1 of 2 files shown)


JanMatCodasip (changes requested):
Hi @lz-bro, thank you for catching the bug.

I recommend a different way of fixing it:

The functions `riscv013_set_register` and `riscv013_get_register` should remain low-level and not deal with the register cache. That means they should only call `register_read_direct()` and `register_write_direct()`, not the higher-level cache aware functions `riscv_get_register()` and `riscv_set_register()`.

My proposal is to move the `rid` checks to `riscv_get_register()`:

```c
int riscv_get_register(struct target *target, riscv_reg_t *value,
		enum gdb_regno regid)
{
	RISCV_INFO(r);
	assert(r->get_register);

	keep_alive();

	/* Handle virtual registers - moved code */

	if (rid == GDB_REGNO_PC) {
		return riscv_get_register(target, value, GDB_REGNO_DPC);
	} else if (rid == GDB_REGNO_PRIV) {
		uint64_t dcsr;
		if (riscv_get_register(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
			return ERROR_FAIL;
		/*...*/
	}
	
	/* Handle read from the actual register - existing code */
	
	if (!target->reg_cache) {
		assert(!target_was_examined(target));
		LOG_TARGET_DEBUG(target, "No cache, reading %s from target",
				gdb_regno_name(regid));
		return r->get_register(target, value, regid);
	}

	struct reg *reg = get_reg_cache_entry(target, regid);
	/* ... */
}
```

Analogously for `riscv_set_register()`.

@timsifive Does this make sense to you, too? Or am I overreaching on this one?

lz-bro:
@JanMatCodasip Thank you for your suggestion, I will modify it later

src/target/riscv/riscv.c
```diff
@@ -4822,6 +4822,19 @@ static int riscv_set_or_write_register(struct target *target,
 
 	keep_alive();
 
+	if (rid == GDB_REGNO_PC) {
+		LOG_TARGET_DEBUG(target, "writing PC to DPC: 0x%" PRIx64, value);
```
timsifive: Can you remove this debug print? That would be the other half of #915.

timsifive (changes requested):
It's not clear to me what bug this is fixing. Was there a discussion somewhere else?

Also, it's failing several of the automated build actions (which I just enabled for you). Please take a look.

src/target/riscv/riscv.c
lz-bro: 1. This bug was encountered when I used riscv-tests to test FPGA：
https://github.com/riscv-software-src/riscv-tests/blob/d4eaa5bd6674b51d3b9b24913713c4638e99cdd9/debug/gdbserver.py#L1591-L1601
When rid is priv, the cacheable read/write function is not used to handle dcsr, which may result in incorrect dcsr from cache.
2. I looked at several failed automated build operations, Check Code Style doesn't seem to be the cause for me, and I've fixed the other errors.
3. I will remove the debug print later and resubmit it.

timsifive: Thanks for explaining.

timsifive (dismissed):
Looks good. Cleans up the code and fixes a bug. :-)

JanMatCodasip (dismissed):
LGTM. Thank you.

JanMatCodasip:
@lz-bro Please note that there is a conflict that should be resolved prior to the merge.

Thanks.

lz-bro:
> @lz-bro Please note that there is a conflict that should be resolved prior to the merge.
> 
> Thanks.
I have resolved the conflict，thank you.

JanMatCodasip (approved):

timsifive (approved):

en-sc:
The move of `priv` read/write from version-specific part causes `priv.v` to be read incorrectly for 0.11 targets.
The issue is: `dcsr[5]` is `dcsr.v` for the current spec, but it is `dcsr.debugint` for 0.11
Will prepare a patch with a fix.

en-sc:
> Will prepare a patch with a fix.

https://github.com/riscv/riscv-openocd/pull/997

PR #921, opened 2023-09-21
target/riscv: support riscv repeat_read by sysbus access

lz-bro:
Fix riscv repeat_read by sysbus access
Change-Id:  a8ffda6e705cff8ff0e9bd802151694eea73d000

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3175,7 +3175,7 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
 
 	RISCV013_INFO(info);
 	target_addr_t next_address = address;
-	target_addr_t end_address = address + count * size;
+	target_addr_t end_address = address + (increment ? count : 1) * size;
 
 	while (next_address < end_address) {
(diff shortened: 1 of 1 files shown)


timsifive:
This looks right. How did you test this?

lz-bro:
> This looks right. How did you test this?
When I run riscv-tests in RepeatReadTest case:
https://github.com/riscv-software-src/riscv-tests/blob/d4eaa5bd6674b51d3b9b24913713c4638e99cdd9/debug/gdbserver.py#L915-L918
memmory access by sysbus:
`riscv set_mem_access sysbus progbuf`
openocd log show:
sysbus
sysbusfail
progbuf
There is obviously a bug in using sysbus to repeat_read memory, although it cannot access the mtime address. After fixing this bug, I simply tested it and checked it from the log.

timsifive (approved):
Looking at this closer, the existing code did the right thing, but the debug messages are wrong. E.g. `./gdbserver.py targets/RISC-V/spike64.py RepeatReadTest` tests a configuration without SBA and the test passes as expected, because sbcs.autoincrement is not set.

Still, this is worth fixing, and the existing code might be working by accident. This change makes things a bit more clear.

PR #924, opened 2023-09-27
server/gdb_server: Step unavailable targets.

timsifive:
When gdb requests to step an unavailable target, pretend the target is still running and the step hasn't yet completed. When the target becomes available, the step can complete.

Change-Id: I969ab56139f72a757552928d59edf6eabd598fa4

### DIFF
--- a/src/server/gdb_server.c
+++ b/src/server/gdb_server.c
@@ -3176,9 +3176,15 @@ static bool gdb_handle_vcont_packet(struct connection *connection, const char *p
 			return true;
 		}
 
-		retval = target_step(ct, current_pc, 0, 0);
-		if (retval == ERROR_TARGET_NOT_HALTED)
-			LOG_INFO("target %s was not halted when step was requested", target_name(ct));
+		if (ct->state == TARGET_UNAVAILABLE) {
+			LOG_TARGET_ERROR(ct, "Target is unavailable, so cannot be stepped. "
+					     "Pretending to gdb that it is running until it's available again.");
+			retval = ERROR_FAIL;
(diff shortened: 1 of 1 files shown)


src/server/gdb_server.c
```diff
@@ -3176,9 +3176,15 @@ static bool gdb_handle_vcont_packet(struct connection *connection, const char *p
 			return true;
 		}
 
-		retval = target_step(ct, current_pc, 0, 0);
-		if (retval == ERROR_TARGET_NOT_HALTED)
-			LOG_INFO("target %s was not halted when step was requested", target_name(ct));
+		if (ct->state == TARGET_UNAVAILABLE) {
+			LOG_TARGET_ERROR(ct, "Target is unavailable, so cannot be stepped. "
```
JanMatCodasip, line 3180: You may consider whether to use just LOG_TARGET_WARNING() here, but I don't have a strong opinion.

JanMatCodasip (approved):
This looks all right to me - checked visually.

JanMatCodasip:
Please consider updating the commit message (for easier reference in the future):

```
When gdb requests to step an unavailable target, report success. (...)
-->
When gdb requests to step an unavailable target, pretend the target is still running and the step hasn't yet completed. (...)
```

(or similar wording)

src/server/gdb_server.c
timsifive: I'll leave this for now. We can change it if real-world use says this is too noisy.

timsifive:
Fixed the PR description. Thanks.

PR #925, opened 2023-09-27
gdb_server,rtos: Differentiate rtos_get_gdb_reg failing and not imple…

timsifive:
…mented

If it fails, then pass that failure on. If it's simply not implemented, then we can fall through and try target_get_gdb_reg_list_noread().

This difference matters when the target representing the current hwthread is unavailable, but the target that is linked to the gdb connection is available. In that case we want the operation to return an error to gdb, instead of reading the register from the target that is available.

Change-Id: I9c84ca556f818c5580e25ab349a34a226fcf0f43

### DIFF
--- a/src/rtos/rtos.c
+++ b/src/rtos/rtos.c
@@ -591,7 +591,7 @@ int rtos_get_gdb_reg(struct connection *connection, int reg_num)
 
 		free(reg_list);
 	}
-	return ERROR_FAIL;
+	return ERROR_NOT_IMPLEMENTED;
 }
 
 /** Return a list of general registers. */
--- a/src/rtos/rtos.c
+++ b/src/rtos/rtos.c
@@ -529,69 +529,71 @@ int rtos_get_gdb_reg(struct connection *connection, int reg_num)
 {
 	struct target *target = get_target_from_connection(connection);
 	threadid_t current_threadid = target->rtos->current_threadid;
-	if ((target->rtos) && (current_threadid != -1) &&
-			(current_threadid != 0) &&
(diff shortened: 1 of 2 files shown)


JanMatCodasip:
I will be able to review this one (and also #926, #927) tomorrow morning.

src/rtos/rtos.c
```diff
@@ -591,7 +591,7 @@ int rtos_get_gdb_reg(struct connection *connection, int reg_num)
 
 		free(reg_list);
 	}
-	return ERROR_FAIL;
+	return ERROR_NOT_IMPLEMENTED;
```
JanMatCodasip: Overall, this looks all right to me. I have just two suggestions:

1) Code style: For readability, I'd recommend to use here the early-return code structure, as shown below.

2) What should be returned at the very end of the function, that is if the `reg_num` is not found in `reg_list`? I believe it should be ERROR_FAIL.

```c
int rtos_get_gdb_reg(...)
{
    if ( (!target->rtos) || 
        current_threadid == -1 ||
        current_threadid == 0 ||
        /* ... */ ) {
        /* early return */
        return ERROR_NOT_IMPLEMENTED;
    }

    /* The rest of the algorithm continues here */
    /* ... */
    /* ... */

    /* What should be returned if we reach the end? */
    LOG_ERROR("Did not find register number %d in RTOS register list", reg_num);
    return ERROR_FAIL; 
    
    /* or */

    return ERROR_NOT_IMPLEMENTED;
}
```

timsifive: Done. Thanks.

JanMatCodasip (approved):

PR #926, opened 2023-09-27
server/gdb_server: Handle events if first target is unavailable

timsifive:
When a target in an SMP group is unavailable, the gdb layer might get an event for a different target in that SMP group, but not one that is the primary target for that gdb connection. So propagate events if they're for any of the targets in the SMP group, not just if it's for the first one in that group.

Change-Id: I8d6738762acc7c0aef96f56ce2cb7f2eeb233b33

### DIFF
--- a/src/server/gdb_server.c
+++ b/src/server/gdb_server.c
@@ -166,6 +166,24 @@ struct target *get_available_target_from_connection(struct connection *connectio
 	return target;
 }
 
+/** Return true iff the given connection includes the given target. */
+static bool gdb_connection_includes_target(struct connection *connection, struct target *target)
+{
+	struct gdb_service *gdb_service = connection->service->priv;
+	struct target *service_target = gdb_service->target;
+	if (service_target->smp) {
+		struct target_list *tlist;
+		foreach_smp_target(tlist, service_target->smp_targets) {
+			struct target *t = tlist->target;
+			if (t == target)
+				return true;
+		}
+		return false;
+	}
+	/* Non-SMP target. */
+	return service_target == target;
+}
+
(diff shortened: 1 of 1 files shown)


src/server/gdb_server.c
```diff
@@ -166,6 +166,22 @@ struct target *get_available_target_from_connection(struct connection *connectio
...
+	struct target *service_target = gdb_service->target;
+	if (service_target->smp) {
+		struct target_list *tlist;
+		foreach_smp_target(tlist, service_target->smp_targets) {
+			struct target *t = tlist->target;
+			if (t == target)
+				return true;
+		}
+	}
+	return service_target == target;
```
JanMatCodasip: I'm sorry if I'm too pedantic. For SMP targets, the last line `return service_target == target` is redundant. If reached, it will be always `false` for SMP.

To make the situation clearer to the reader, you may consider this structure of code:

```suggestion
	if (service_target->smp) {
		struct target_list *tlist;
		foreach_smp_target(tlist, service_target->smp_targets) {
			struct target *t = tlist->target;
			if (t == target)
				return true;
		}
		return false;
	} else {
		/* Non-SMP target */
		return service_target == target;	
	}
```

JanMatCodasip (commented):
This looks all right to me. I have just one suggestion.

src/server/gdb_server.c
timsifive: Done. Thanks.

timsifive: (checkpatch says I'm not allowed to use the `else`)

JanMatCodasip (approved):

PR #927, opened 2023-09-27
server/gdb_server: Fake resuming unavailable targets.

timsifive:
When asked to resume an unavailable target, resume any available targets and report success.

Change-Id: Ieafc63794c1a6eba8948c0f9ce84fa74f9765041

### DIFF
--- a/src/server/gdb_server.c
+++ b/src/server/gdb_server.c
@@ -3039,7 +3039,7 @@ static bool gdb_handle_vcont_packet(struct connection *connection, const char *p
 	__attribute__((unused)) int packet_size)
 {
 	struct gdb_connection *gdb_connection = connection->priv;
-	struct target *target = get_available_target_from_connection(connection);
+	struct target *target = get_target_from_connection(connection);
 	const char *parse = packet;
 	int retval;
 
@@ -3060,6 +3060,24 @@ static bool gdb_handle_vcont_packet(struct connection *connection, const char *p
 	/* simple case, a continue packet */
 	if (parse[0] == 'c') {
 		gdb_running_type = 'c';
+
+		if (target->state == TARGET_UNAVAILABLE) {
+			struct target *available_target = get_available_target_from_connection(connection);
+			if (target == available_target) {
+				LOG_DEBUG("All targets for this gdb connection "
+						"are unavailable.  Fake to gdb that the resume "
+						"succeeded and the target is now running.");
+				gdb_connection->frontend_state = TARGET_RUNNING;
+				gdb_connection->output_flag = GDB_OUTPUT_ALL;
+				target_call_event_callbacks(target, TARGET_EVENT_GDB_START);
+				return true;
+			}
+			LOG_TARGET_DEBUG(target, "Target is unavailable. Resume %s instead.",
+					target_name(available_target));
+			/* Resume an available target. */
+			target = available_target;
+		}
+
 		LOG_DEBUG("target %s continue", target_name(target));
 		gdb_connection->output_flag = GDB_OUTPUT_ALL;
 		retval = target_resume(target, 1, 0, 0, 0);


src/server/gdb_server.c
```diff
@@ -3036,6 +3036,19 @@ static bool gdb_handle_vcont_packet(struct connection *connection, const char *p
 
 	/* simple case, a continue packet */
 	if (parse[0] == 'c') {
+		if (target->state == TARGET_UNAVAILABLE) {
+			struct target *available_target = get_available_target_from_connection(connection);
+			if (target == available_target) {
+				/* If the only target is unavailable, then just
+				 * fake to gdb that it is now running. */
```
JanMatCodasip: I'd suggest a reword - to make the comment more accurate:

```suggestion
				/* All targets for this gdb connections are unavailable --
				 * fake to gdb that the resume succeeded and the target is now running. */
```

src/server/gdb_server.c
```diff
@@ -3036,6 +3036,19 @@ static bool gdb_handle_vcont_packet(struct connection *connection, const char *p
...
+				gdb_connection->frontend_state = TARGET_RUNNING;
+				gdb_connection->output_flag = GDB_OUTPUT_ALL;
+				return true;
```
JanMatCodasip: When pretending that the resume succeeded, do we want to do also (some of) these?

- assign `gdb_running_type = 'c';` ?
- call `target_call_event_callbacks(target, TARGET_EVENT_GDB_START);` ?
- make a debug print - for easier log analysis?

src/server/gdb_server.c
```diff
@@ -3036,6 +3036,19 @@ static bool gdb_handle_vcont_packet(struct connection *connection, const char *p
...
+			}
+			/* Resume an available target. */
+			target = available_target;
```
JanMatCodasip, line 3078: Here I'd also add a debug print, something like this:

```
LOG_DEBUG("Target '%s' is unavailable. Resuming an available target '%s' instead.", target_name(target), target_name(available_target));
```

src/server/gdb_server.c
```diff
@@ -3016,7 +3016,7 @@ static bool gdb_handle_vcont_packet(struct connection *connection, const char *p
 	__attribute__((unused)) int packet_size)
```
JanMatCodasip, line 3039: It seems to me that the GDB's `c` packet would require the same treatment (`gdb_step_continue_packet()`).

Or, is GDB not using the plain `c` packet (in favor of `vCont:c`)? If that is the case, would you know if this is GDB's general behavior on all targets? Or only on multi-threaded targets?

JanMatCodasip (commented):
This looks overall fine to me - checked visually only.

I have several suggestions and questions.

src/server/gdb_server.c
timsifive: Done. Thanks.

src/server/gdb_server.c
timsifive: I've never seen gdb use a plain 'c' packet. I assume that if vCont is supported, then it always uses that.

src/server/gdb_server.c
timsifive: Yes, on all 3 points.

OpenOCD itself doesn't seem to use TARGET_EVENT_GDB_START, but it seems more consistent to report it here than to not report.

src/server/gdb_server.c
timsifive: Done.

src/server/gdb_server.c
TommyMurphyTM1234: > I've never seen gdb use a plain 'c' packet. I assume that if vCont is supported, then it always uses that.

Do you ever use, say, an Eclipse or VSCode based environment for debugging? I just wonder if that sort of setup, that usually uses the GDB/MI rather than the regular CLI to drive GDB, might cause `c` packets to be used in some cases? Wouldn't it be safer to apply the same changes to the OpenOCD `c` packet handler just in case?

src/server/gdb_server.c
JanMatCodasip: > OpenOCD itself doesn't seem to use TARGET_EVENT_GDB_START

My understanding is that this an event for which the users can write their own hooks in TCL:

```tcl
mychip.cpu configure -event gdb-start {
    # ....
}
```

https://openocd.org/doc/html/CPU-Configuration.html#Target-Events

src/server/gdb_server.c
JanMatCodasip: The current merge request dealing with `vCont` is OK with me. This was a note to keep the back of one's mind, just in case the `c` packet should ever occur and started causing troubles.

src/server/gdb_server.c
JanMatCodasip: @timsifive In the current diff, I can still see the original code. 

Can you please double-check whether your latest changes have been pushed?

timsifive: Yup. Not pushed yet. I tend to comment before I make the change, and then I needed to run tests, and then got side-tracked. But it's coming.

src/server/gdb_server.c
timsifive: Reading the code for handling 'c' packets, I'm not convinced it does anything at all. I'm not going to attempt to address that unless I see a problem report.

@TommyMurphyTM1234 I don't use Eclipse, but I do sometimes go over -d3 logs from people who do.

JanMatCodasip (approved):

PR #928, opened 2023-09-28
target/riscv: cache requests to trigger configuration

AnastasiyaChernikova:
Depending on configuration, the existing implementation of watchpoints is rather inefficient for certain scenarios. Consider HW that:

1. triggers 0-3 can be used as instruction breakpoints
2. triggers 4-7 can be used as data breakpoints (watchpoints)
3.  NAPOT triggers are not supported.

Now, consider that we have a pending watchpoint. And we perform a "step" operation. According to the current implementation:

* OpenOCD will disable watchpoints
* Perform a single-step
* Will try to restore the original watchpoints. It will need 12 attempts to find a suitable trigger: (8 attempts to try NAPOT, and another 4 to try GE+LE).

This patch introduces a dedicated cache for requests to triggers. It significantly speeds things up, since we cache failed attempts and no additional interactions with HW is necessary.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -128,6 +128,17 @@ struct trigger {
 	int unique_id;
 };
 
+struct tdata2_cache {
+	struct list_head elem_tdata2;
+	riscv_reg_t tdata2;
+};
+
+struct tdata1_cache {
+	riscv_reg_t tdata1;
+	struct list_head tdata2_cache_head;
+	struct list_head elem_tdata1;
+};
+
 /* Wall-clock timeout for a command/access. Settable via RISC-V Target commands.*/
 int riscv_command_timeout_sec = DEFAULT_COMMAND_TIMEOUT_SEC;
 
@@ -503,6 +514,25 @@ static void free_custom_register_names(struct target *target)
 	info->custom_register_names.reg_names = NULL;
 }
 
+static void free_wp_triggers_cache(struct target *target)
+{
+	RISCV_INFO(r);
+
+	for (unsigned int i = 0; i < r->trigger_count; ++i) {
+		struct tdata1_cache *elem_1, *tmp_1;
+		list_for_each_entry_safe(elem_1, tmp_1, &r->wp_triggers_negative_cache[i], elem_tdata1) {
+			struct tdata2_cache *elem_2, *tmp_2;
+			list_for_each_entry_safe(elem_2, tmp_2, &elem_1->tdata2_cache_head, elem_tdata2) {
+				list_del(&elem_2->elem_tdata2);
+				free(elem_2);
+			}
+			list_del(&elem_1->elem_tdata1);
+			free(elem_1);
+		}
+	}
+	free(r->wp_triggers_negative_cache);
+}
+
 static void riscv_deinit_target(struct target *target)
 {
 	LOG_TARGET_DEBUG(target, "riscv_deinit_target()");
@@ -519,6 +549,7 @@ static void riscv_deinit_target(struct target *target)
 		tt->deinit_target(target);
 
 	riscv_free_registers(target);
+	free_wp_triggers_cache(target);
 
 	if (!info)
 		return;
@@ -653,8 +684,8 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
 			LOG_TARGET_DEBUG(target,
 				"wrote 0x%" PRIx64 " to tdata2 but read back 0x%" PRIx64,
 				tdata2, tdata2_rb);
-
-		riscv_set_register(target, GDB_REGNO_TDATA1, 0);
+		if (riscv_set_register(target, GDB_REGNO_TDATA1, 0) != ERROR_OK)
+			return ERROR_FAIL;
 		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
 	}
 
@@ -720,6 +751,99 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
 			trig_info.tdata1, trig_info.tdata2, trig_info.tdata1_ignore_mask);
 };
 
+static struct tdata1_cache *tdata1_cache_alloc(struct list_head *tdata1_cache_head, riscv_reg_t tdata1)
+{
+	struct tdata1_cache *elem = (struct tdata1_cache *)calloc(1, sizeof(struct tdata1_cache));
+	elem->tdata1 = tdata1;
+	INIT_LIST_HEAD(&elem->tdata2_cache_head);
+	list_add_tail(&elem->elem_tdata1, tdata1_cache_head);
+	return elem;
+}
+
+static void tdata2_cache_alloc(struct list_head *tdata2_cache_head, riscv_reg_t tdata2)
+{
+	struct tdata2_cache * const elem = calloc(1, sizeof(struct tdata2_cache));
+	elem->tdata2 = tdata2;
+	list_add(&elem->elem_tdata2, tdata2_cache_head);
+}
+
+struct tdata2_cache *tdata2_cache_search(struct list_head *tdata2_cache_head, riscv_reg_t find_tdata2)
+{
+	struct tdata2_cache *elem_2;
+	list_for_each_entry(elem_2, tdata2_cache_head, elem_tdata2) {
+		if (elem_2->tdata2 == find_tdata2)
+			return elem_2;
+	}
+	return NULL;
+}
+
+struct tdata1_cache *tdata1_cache_search(struct list_head *tdata1_cache_head, riscv_reg_t find_tdata1)
+{
+	struct tdata1_cache *elem_1;
+	list_for_each_entry(elem_1, tdata1_cache_head, elem_tdata1) {
+		if (elem_1->tdata1 == find_tdata1)
+			return elem_1;
+	}
+	return NULL;
+}
+
+static void create_wp_trigger_cache(struct target *target)
+{
+	RISCV_INFO(r);
+
+	r->wp_triggers_negative_cache = (struct list_head *)calloc(r->trigger_count,
+		sizeof(struct list_head));
+	for (unsigned int i = 0; i < r->trigger_count; ++i)
+		INIT_LIST_HEAD(&r->wp_triggers_negative_cache[i]);
+}
+
+static void wp_triggers_cache_add(struct target *target, unsigned int idx, riscv_reg_t tdata1,
+	riscv_reg_t tdata2, int error_code)
+{
+	RISCV_INFO(r);
+
+	struct tdata1_cache *tdata1_cache = tdata1_cache_search(&r->wp_triggers_negative_cache[idx], tdata1);
+	if (!tdata1_cache) {
+		tdata1_cache = tdata1_cache_alloc(&r->wp_triggers_negative_cache[idx], tdata1);
+	} else {
+		struct tdata2_cache *tdata2_cache = tdata2_cache_search(&tdata1_cache->tdata2_cache_head, tdata2);
+		if (tdata2_cache) {
+			list_move(&tdata2_cache->elem_tdata2, &tdata1_cache->tdata2_cache_head);
+			return;
+		}
+	}
+	tdata2_cache_alloc(&tdata1_cache->tdata2_cache_head, tdata2);
+}
+
+static bool wp_triggers_cache_search(struct target *target, unsigned int idx,
+	riscv_reg_t tdata1, riscv_reg_t tdata2)
+{
+	RISCV_INFO(r);
+
+	struct tdata1_cache *tdata1_cache = tdata1_cache_search(&r->wp_triggers_negative_cache[idx], tdata1);
+	if (!tdata1_cache)
+		return false;
+	struct tdata2_cache *tdata2_cache = tdata2_cache_search(&tdata1_cache->tdata2_cache_head, tdata2);
+	if (!tdata2_cache)
+		return false;
+	assert(tdata1_cache->tdata1 == tdata1 && tdata2_cache->tdata2 == tdata2);
+	return true;
+}
+
+static int try_use_trigger_and_cache_result(struct target *target, unsigned int idx, riscv_reg_t tdata1,
+	riscv_reg_t tdata2, riscv_reg_t tdata1_ignore_mask)
+{
+	if (wp_triggers_cache_search(target, idx, tdata1, tdata2))
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+
+	int ret = set_trigger(target, idx, tdata1, tdata2, tdata1_ignore_mask);
+
+	/* Add these values to the cache to remember that they are not supported. */
+	if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+		wp_triggers_cache_add(target, idx, tdata1, tdata2, ret);
+	return ret;
+}
+
 static int try_setup_single_match_trigger(struct target *target,
 		struct trigger *trigger, struct trigger_request_info trig_info)
 {
@@ -735,8 +859,9 @@ static int try_setup_single_match_trigger(struct target *target,
 	for (unsigned int idx = 0;
 			find_next_free_trigger(target, trigger_type, false, &idx) == ERROR_OK;
 			++idx) {
-		ret = set_trigger(target, idx, trig_info.tdata1, trig_info.tdata2,
-				trig_info.tdata1_ignore_mask);
+		ret = try_use_trigger_and_cache_result(target, idx, trig_info.tdata1, trig_info.tdata2,
+			trig_info.tdata1_ignore_mask);
+
 		if (ret == ERROR_OK) {
 			r->trigger_unique_id[idx] = trigger->unique_id;
 			return ERROR_OK;
@@ -763,12 +888,17 @@ static int try_setup_chained_match_triggers(struct target *target,
 	for (unsigned int idx = 0;
 			find_next_free_trigger(target, trigger_type, true, &idx) == ERROR_OK;
 			++idx) {
-		ret = set_trigger(target, idx, t1.tdata1, t1.tdata2,
-				t1.tdata1_ignore_mask);
-		if (ret != ERROR_OK)
+		ret = try_use_trigger_and_cache_result(target, idx, t1.tdata1, t1.tdata2,
+			t1.tdata1_ignore_mask);
+
+		if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
 			continue;
-		ret = set_trigger(target, idx + 1, t2.tdata1, t2.tdata2,
-				t2.tdata1_ignore_mask);
+		else if (ret != ERROR_OK)
+			return ret;
+
+		ret = try_use_trigger_and_cache_result(target, idx + 1, t2.tdata1, t2.tdata2,
+			t2.tdata1_ignore_mask);
+
 		if (ret == ERROR_OK) {
 			r->trigger_unique_id[idx] = trigger->unique_id;
 			r->trigger_unique_id[idx + 1] = trigger->unique_id;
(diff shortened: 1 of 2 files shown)


timsifive (changes requested):
I love this improvement. It's going to help a lot, as you explained.

Why did you choose to use a linked list for the data structure? It looks like you're always allocating the whole thing. I think using an array would simplify the code significantly, but perhaps I'm missing some downside.

src/target/riscv/riscv.c
```diff
@@ -700,6 +734,110 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
...
+}
+
+static int try_use_trigger_and_cache_result(struct target *target, unsigned int idx, riscv_reg_t tdata1,
+	riscv_reg_t tdata2, riscv_reg_t tdata1_ignore_mask)
+{
+	if (is_use_wp_trigger_in_cache(target, idx, tdata1, tdata2))
+			return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+
+	int ret = set_trigger(target, idx, tdata1, tdata2, tdata1_ignore_mask);
+	add_use_wp_trigger_in_cache(target, idx, tdata1, tdata2, ret);
```
MarekVCodasip: Myself, I would probably not handle the return value inside the add_use_wp_trigger_in_cache(). Also, I think if it returned an error which isn't ERROR_TARGET_RESOURCE_NOT_AVAILABLE (error when setting/reading triggers), the cache should be invalidated. As we have no idea at which step the error happened.

AnastasiyaChernikova: In this case, this check is needed in order not to cache calls that have a return code of ERROR_OK

AnastasiyaChernikova:
>Why did you choose to use a linked list for the data structure? It looks like you're always allocating the whole thing. I think using an array would simplify the code significantly, but perhaps I'm missing some downside.

When caching tdata2 data, each new call is written to the top of the list (or moved if such a call has already been made), thus making it easier to find the data. And for tdata1 a list to make code uniform.

src/target/riscv/riscv.c
```diff
@@ -700,6 +734,110 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
 			trig_info.tdata1, trig_info.tdata2, trig_info.tdata1_ignore_mask);
 };
 
+static struct tdata1_cache *create_new_tdata1_cache(struct list_head *tdata1_cache_head, riscv_reg_t tdata1)
+{
+
```
JanMatCodasip: There are extra blank lines at the beginning of multiple functions - these can be removed.

```suggestion
```

Please check all the functions. I have commented on a single occurrence only but there are more of them.

src/target/riscv/riscv.c
```diff
@@ -700,6 +734,110 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
...
+	RISCV_INFO(r);
+	
+	if (error_code != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+		return;
+
+	struct tdata1_cache *tdata1_cache = find_tdata1(&(r->wp_triggers_cache[idx]), tdata1);
+	if (tdata1_cache == NULL) {
+		tdata1_cache = create_new_tdata1_cache(&(r->wp_triggers_cache[idx]), tdata1);
```
timsifive: Are you making a new cache per trigger index here? I don't understand that. There can only be one tdata1 value per trigger at a time.

src/target/riscv/riscv.c
```diff
@@ -700,6 +734,110 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
...
+	return NULL;
+}
+
+static void create_wp_trigger_cache(struct target *target)
+{
+
+	RISCV_INFO(r);
+
+	r->wp_triggers_cache = (struct list_head *)calloc(r->trigger_count,
+		sizeof(struct list_head));
```
timsifive: One cache per trigger makes sense. But then all you need is the contents of tdata1/tdata2 along with some bits to mark whether they are valid, right?

I completely do not understand why you're using lists within each wp_triggers_cache entry.

src/target/riscv/riscv.c
AnastasiyaChernikova: We can use one trigger to set wp of different types (eq, napot, ge, lt). And in this implementation, all this information is cached. This is necessary if we installed wp on a certain trigger, then deleted it and want to reinstall it.

src/target/riscv/riscv.c
AnastasiyaChernikova: Usually one tdata1 value corresponds to many tdata2 values, so this implementation seemed more optimal

timsifive: Can you give an example where you have multiple tdata2 values for a single trigger?

aap-sc: I guess this happens when a single trigger is used for different watchpoints/breakpoints . Like user set a breakpoint at some address, then disabled it and set at some other address. This will result in accumulation of different `tdata2` values.

timsifive: But why do we care to accumulate different `tdata2` values for a single trigger? I thought the point was for OpenOCD to keep track of what `tdata2` is for each trigger, so it doesn't need to read it from the target, which is slow.

en-sc: As far as I understand the point is to cache non-successful results of calls to `set_trigger` (`ERROR_TARGET_RESOURCE_NOT_AVAILABLE`). This makes an assumption that if a combination of `tdata1` and `tdata2` resulted in `ERROR_TARGET_RESOURCE_NOT_AVAILABLE` once, the result will be the same. Such structure allows to skip attempts to set a trigger.
e.g.:
* A breakpoint needs to be restored.
* The required combination of `tdata1` and `tdata2` is only supported on the last trigger.
* While the breakpoint was set for the first time, unsuccessful attempts to set it on the triggers with lower indexes were collected into the cache.
* These attempts will be skipped during restore (no communication with DM is required).

timsifive: Ah, that's the part I was missing! Thanks for explaining. I'll review again.

timsifive: And of course the original description said "It significantly speeds things up, since we cache failed attempts". Sorry for not reading more closely.

src/target/riscv/riscv.h
```diff
@@ -137,6 +137,8 @@ struct riscv_info {
 	/* The number of triggers per hart. */
 	unsigned int trigger_count;
 
+	struct list_head *wp_triggers_cache;
```
timsifive: This needs a comment explaining what the cache is for. What does it mean for a value to be in the cache?

Also mention that it's an array of caches, one for each trigger.

src/target/riscv/riscv.c
```diff
@@ -700,6 +732,99 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
 			trig_info.tdata1, trig_info.tdata2, trig_info.tdata1_ignore_mask);
 };
 
+static struct tdata1_cache *create_new_tdata1_cache(struct list_head *tdata1_cache_head, riscv_reg_t tdata1)
```
timsifive: Following the example of e.g. riscv_batch_alloc, let's call this:
```suggestion
static struct tdata1_cache *tdata1_cache_alloc(struct list_head *tdata1_cache_head, riscv_reg_t tdata1)
```

(How does this square with e.g. riscv_free_registers()? It's a bit messy and inconsistent, but at least when there's a simple struct then using _alloc() seems more consistent.)

src/target/riscv/riscv.c
```diff
@@ -700,6 +732,99 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
...
+{
+	struct tdata1_cache *elem = (struct tdata1_cache *)calloc(1, sizeof(struct tdata1_cache));
+	elem->tdata1 = tdata1;
+	INIT_LIST_HEAD(&elem->tdata2_cache_head);
+	list_add_tail(&elem->elem_tdata1, tdata1_cache_head);
+	return elem;
+}
+
+static void create_new_tdata2_cache(struct list_head *tdata2_cache_head, riscv_reg_t tdata2)
```
```suggestion
static void tdata2_cache_alloc(struct list_head *tdata2_cache_head, riscv_reg_t tdata2)
```

src/target/riscv/riscv.c
```diff
@@ -700,6 +732,99 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
...
+{
+	struct tdata2_cache * const elem = calloc(1, sizeof(struct tdata2_cache));
+	elem->tdata2 = tdata2;
+	list_add(&elem->elem_tdata2, tdata2_cache_head);
+}
+
+struct tdata2_cache *find_call_in_tdata2_cache(struct list_head *tdata2_cache_head, riscv_reg_t find_tdata2)
```
timsifive: What is "call" in this context? Isn't this simply `tdata2_cache_search`?

src/target/riscv/riscv.c
```diff
@@ -700,6 +732,99 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
...
+{
+	struct tdata2_cache *elem_2;
+	list_for_each_entry(elem_2, tdata2_cache_head, elem_tdata2) {
+		if (elem_2->tdata2 == find_tdata2)
+			return elem_2;
+	}
+	return NULL;
+}
+
+struct tdata1_cache *find_tdata1(struct list_head *tdata1_cache_head, riscv_reg_t find_tdata1)
```
```suggestion
struct tdata1_cache *tdata1_cache_search(struct list_head *tdata1_cache_head, riscv_reg_t find_tdata1)
```

src/target/riscv/riscv.c
```diff
@@ -700,6 +732,99 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
...
+		INIT_LIST_HEAD(&r->wp_triggers_cache[i]);
+}
+
+static void add_use_wp_trigger_in_cache(struct target *target, unsigned int idx, riscv_reg_t tdata1,
+	riscv_reg_t tdata2, int error_code)
+{
+	RISCV_INFO(r);
+
+	if (error_code != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+		return;
```
timsifive: This error check belongs outside this function, which makes for easier to understand code.

```suggestion
static void wp_triggers_cache_add(struct target *target, unsigned int idx, riscv_reg_t tdata1,
	riscv_reg_t tdata2)
{
	RISCV_INFO(r);

```

src/target/riscv/riscv.c
```diff
@@ -700,6 +732,99 @@ static void log_trigger_request_info(struct trigger_request_info trig_info)
...
+		if (tdata2_cache) {
+			list_move(&tdata2_cache->elem_tdata2, &tdata1_cache->tdata2_cache_head);
+			return;
+		}
+	}
+	create_new_tdata2_cache(&tdata1_cache->tdata2_cache_head, tdata2);
+}
+
+static bool is_use_wp_trigger_in_cache(struct target *target, unsigned int idx,
+	riscv_reg_t tdata1, riscv_reg_t tdata2)
```
```suggestion
static bool wp_triggers_cache_search(struct target *target, unsigned int idx,
	riscv_reg_t tdata1, riscv_reg_t tdata2)
```

src/target/riscv/riscv.c
timsifive: I agree with @MarekVCodasip that the error check should be here, not inside add_use_wp_trigger_in_cache().

It also needs a comment like "Add these values to the cache to remember that they are not supported."

src/target/riscv/riscv.c
```diff
@@ -743,12 +869,21 @@ static int try_setup_chained_match_triggers(struct target *target,
...
+		switch (ret) {
+		case ERROR_OK:
+			break;
+		case ERROR_TARGET_RESOURCE_NOT_AVAILABLE:
 			continue;
-		ret = set_trigger(target, idx + 1, t2.tdata1, t2.tdata2,
-				t2.tdata1_ignore_mask);
+		default:
+			return ret;
+		}
```
timsifive: This is correct, but not idiomatic.

I'd go with:
```
if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
   continue;
else if (ret != ERROR_OK)
   return ret;
```

timsifive (changes requested):
OK, I think I got it now. Overall looks good. I have a lot of naming/comment suggestions that would have made this code more obvious to me, but nothing really about the structure of the code.

src/target/riscv/riscv.h
JanMatCodasip: I had several ideas that basically correspond to Tim's findngs but forgot to submit my pending review comments :man_facepalming:  -- I'm sorry.

Here I'd recommend to:

- 1) stress that this is a **negative** cache directly in the name of the variable -- we cache items that are known to **not** work
- 2) add an explanatory comment

For example:

```suggestion
	/* Data structure to record known unsupported tdata1+tdata2 trigger CSR values.
	 * This is to avoid repetitive attempts to set trigger configurations that are already
	 * known to be unsupported in the HW. */
	struct list_head *wp_triggers_negative_cache;
```

or 

```suggestion
	/* Data structure to record known unsupported tdata1+tdata2 trigger CSR values.
	 * This is to avoid repetitive attempts to set trigger configurations that are already
	 * known to be unsupported in the HW. */
	struct list_head *wp_triggers_unsupported_values;
```

or similar

src/target/riscv/riscv.c
JanMatCodasip: Alternate suggestion for your consideration:

`is_known_unsupported_trigger(...)`

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.h
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
timsifive: Did you forget to push this?

AnastasiyaChernikova:
@timsifive All issues were addressed. Could you please take a look?

src/target/riscv/riscv.c
JanMatCodasip: Here I also vote to simplify the function name, as Tim suggested above: `add_use_wp_trigger_in_cache()` --> `wp_triggers_cache_add()`

JanMatCodasip (changes requested):
There seem to be two unaddressed (overlooked?) items:

- https://github.com/riscv/riscv-openocd/pull/928#discussion_r1377676673
- https://github.com/riscv/riscv-openocd/pull/928#discussion_r1374817608

Other than that, the code looks all right.

src/target/riscv/riscv.c
timsifive: Bump.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

src/target/riscv/riscv.c
AnastasiyaChernikova: Addressed.

timsifive (dismissed):

JanMatCodasip (dismissed):

AnastasiyaChernikova:
@timsifive  this commit has an unexpected (to me test failure) that I haven't observed locally. That being said - I used relatively old riscv-tests to check my changes (riscv-tests ed0a63ea4f , which is from around Jul 10, 2023). Looks like I'll need to re-test my changes with an updated riscv-tests.

aap-sc:
> @timsifive this commit has an unexpected (to me test failure) that I haven't observed locally. That being said - I used relatively old riscv-tests to check my changes (riscv-tests ed0a63ea4f , which is from around Jul 10, 2023). Looks like I'll need to re-test my changes with an updated riscv-tests.

@AnastasiyaChernikova , @timsifive, @JanMatCodasip  actually, I've tried to update `riscv-tests` version to the current TOT and I have `UnavailableHaltedTest` test failure without this MR applied. Looks like the test is broken. Could it be a known issue with the test itself?

aap-sc:
> I have `UnavailableHaltedTest` test failure without this MR applied.

and just for the reference (I think this should be mentioned). This attempt to update was in repository that has few additional patches applied on top of riscv-openocd (minor fixes that were backported from https://sourceforge.net/p/openocd/  . And I used a relatively old version of spike (around 2 months old)

TommyMurphyTM1234:
> > @timsifive this commit has an unexpected (to me test failure) that I haven't observed locally. That being said - I used relatively old riscv-tests to check my changes (riscv-tests ed0a63ea4f , which is from around Jul 10, 2023). Looks like I'll need to re-test my changes with an updated riscv-tests.
> 
> @AnastasiyaChernikova , @timsifive, @JanMatCodasip actually, I've tried to update `riscv-tests` version to the current TOT and I have `UnavailableHaltedTest` test failure without this MR applied. Looks like the test is broken. Could it be a known issue with the test itself?

If I build everything (without the changes in this PR) from scratch from the latest sources (toolchain, Spike, RISC-V OpenOCD) then the `UnavailableHaltedTest` runs succeeed.

If I copy in the modified `riscv.c` and `riscv.h` from here and rebuild OpenOCD those test runs still succeed.

* https://github.com/AnastasiyaChernikova/riscv-openocd/tree/triggers/src/target/riscv

timsifive:
UnavailableHaltedTest seems to fail intermittently in the github action. I haven't seen it fail this way when I run it locally. :-( Hopefully I'll have some time to investigate more next week.

timsifive:
I've restarted this test run several times, and it fails UnavailableHaltedTest consistently. Is it simply based on older code?
Can you rebase this off the latest `riscv` branch? That should also address the conflict that has appeared.

timsifive:
For some reason, in this branch only, UnavailableHaltedTest consistently fails. I have no idea why. It doesn't make any sense with this change and what that test does. I cannot reproduce the failure on my machine either. :-(

I'm weary merging this change, though, because I don't want all testing to suddenly start failing. Is this blocking other work for you?

aap-sc:
@timsifive 

> For some reason, in this branch only, UnavailableHaltedTest consistently fails. I have no idea why. It doesn't make any sense with this change and what that test does. I cannot reproduce the failure on my machine either. :-(

I can actually.  However, **the issue is reproducible even without this patch** on my machine.

This may be related to python version or some weird timing issues. Sometimes OpenOCD log file contain non-printable characters (like ^H and stuff). It seems that this sometimes messes up matching algorithms used in testing infrastructure. There are 2 types of issues I observe:

- `testlib.TestLibError: Timed out waiting for b'riscv\\ dmi_write\\ 0x1f\\ 0x3\\\n' ` - this I cannot reproduce reliably (yet).
- `cannot use a string pattern on a bytes-like object` these are caused that sometimes we have a mismatch between regexp and `self.log_buf` from `expect` method of `testlib.py`. Mismatch happens because sometimes these objects are `bytes` and sometimes these are strings

aap-sc:
> This may be related to python version or some weird timing issues.

I think I've found the root cause. It's a timing/parsing issue. I've filed https://github.com/riscv-software-src/riscv-tests/issues/520 against riscv-tests.

timsifive (approved):

PR #929, opened 2023-09-28
do not assume DTM version unless dtmcontrol is read successfully

aap-sc:
Change-Id: I02906dbbe04778623a51774bc21fa45b55d12881

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -273,7 +273,7 @@ static uint16_t dram_address(unsigned int index)
 		return 0x40 + index - 0x10;
 }
 
-static uint32_t dtmcontrol_scan(struct target *target, uint32_t out)
+static int dtmcontrol_scan(struct target *target, uint32_t out, uint32_t *in_ptr)
 {
 	struct scan_field field;
 	uint8_t in_value[4];
@@ -300,7 +300,9 @@ static uint32_t dtmcontrol_scan(struct target *target, uint32_t out)
 	uint32_t in = buf_get_u32(field.in_value, 0, 32);
 	LOG_DEBUG("DTMCONTROL: 0x%x -> 0x%x", out, in);
 
-	return in;
+	if (in_ptr)
+		*in_ptr = in;
+	return ERROR_OK;
 }
 
 static uint32_t idcode_scan(struct target *target)
@@ -338,7 +340,7 @@ static void increase_dbus_busy_delay(struct target *target)
 			info->dtmcontrol_idle, info->dbus_busy_delay,
 			info->interrupt_high_delay);
 
-	dtmcontrol_scan(target, DTMCONTROL_DBUS_RESET);
+	dtmcontrol_scan(target, DTMCONTROL_DBUS_RESET, NULL /* discard value */);
 }
 
 static void increase_interrupt_high_delay(struct target *target)
@@ -1459,16 +1461,17 @@ static int step(struct target *target, int current, target_addr_t address,
 static int examine(struct target *target)
 {
 	/* Don't need to select dbus, since the first thing we do is read dtmcontrol. */
+	uint32_t dtmcontrol;
+	if (dtmcontrol_scan(target, 0, &dtmcontrol) != ERROR_OK || dtmcontrol == 0) {
+		LOG_ERROR("Could not scan dtmcontrol. Check JTAG connectivity/board power.");
+		return ERROR_FAIL;
+	}
 
-	uint32_t dtmcontrol = dtmcontrol_scan(target, 0);
 	LOG_DEBUG("dtmcontrol=0x%x", dtmcontrol);
 	LOG_DEBUG("  addrbits=%d", get_field(dtmcontrol, DTMCONTROL_ADDRBITS));
 	LOG_DEBUG("  version=%d", get_field(dtmcontrol, DTMCONTROL_VERSION));
 	LOG_DEBUG("  idle=%d", get_field(dtmcontrol, DTMCONTROL_IDLE));
-	if (dtmcontrol == 0) {
-		LOG_ERROR("dtmcontrol is 0. Check JTAG connectivity/board power.");
-		return ERROR_FAIL;
-	}
+
 	if (get_field(dtmcontrol, DTMCONTROL_VERSION) != 0) {
 		LOG_ERROR("Unsupported DTM version %d. (dtmcontrol=0x%x)",
 				get_field(dtmcontrol, DTMCONTROL_VERSION), dtmcontrol);
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -411,14 +411,14 @@ static void select_dmi(struct target *target)
 	jtag_add_ir_scan(target->tap, &select_dbus, TAP_IDLE);
 }
 
-static uint32_t dtmcontrol_scan(struct target *target, uint32_t out)
+static int dtmcontrol_scan(struct target *target, uint32_t out, uint32_t *in_ptr)
 {
 	struct scan_field field;
 	uint8_t in_value[4];
 	uint8_t out_value[4] = { 0 };
 
 	if (bscan_tunnel_ir_width != 0)
-		return dtmcontrol_scan_via_bscan(target, out);
+		return dtmcontrol_scan_via_bscan(target, out, in_ptr);
 
 	buf_set_u32(out_value, 0, 32, out);
 
@@ -441,7 +441,9 @@ static uint32_t dtmcontrol_scan(struct target *target, uint32_t out)
 	uint32_t in = buf_get_u32(field.in_value, 0, 32);
 	LOG_DEBUG("DTMCS: 0x%x -> 0x%x", out, in);
 
-	return in;
+	if (in_ptr)
+		*in_ptr = in;
+	return ERROR_OK;
 }
 
 static void increase_dmi_busy_delay(struct target *target)
@@ -452,7 +454,7 @@ static void increase_dmi_busy_delay(struct target *target)
 			info->dtmcs_idle, info->dmi_busy_delay,
 			info->ac_busy_delay);
 
-	dtmcontrol_scan(target, DTM_DTMCS_DMIRESET);
+	dtmcontrol_scan(target, DTM_DTMCS_DMIRESET, NULL /* discard result */);
 }
 
 /**
@@ -597,7 +599,7 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 		} else if (status == DMI_STATUS_SUCCESS) {
 			break;
 		} else {
-			dtmcontrol_scan(target, DTM_DTMCS_DMIRESET);
+			dtmcontrol_scan(target, DTM_DTMCS_DMIRESET, NULL /* discard result */);
 			break;
 		}
 		if (time(NULL) - start > timeout_sec)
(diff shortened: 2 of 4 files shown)


aap-sc:
@timsifive I know that changes to riscv-0.11 are discouraged, however in this case I believe it's better to add this fix. There were a couple of serious bugs:

1. dtmcontrol_scan could return an error code as a scanned value.
2. in case when `dtmcontrol` was read as "0" we had an implicit fallback to v0.11 that caused register cache to be populated. This could result in a weird situations like:
```
XLEN=64, misa=0x14112f
```

src/target/riscv/riscv-011.c
```diff
@@ -273,7 +273,7 @@ static uint16_t dram_address(unsigned int index)
 		return 0x40 + index - 0x10;
 }
 
-static uint32_t dtmcontrol_scan(struct target *target, uint32_t out)
+static uint32_t dtmcontrol_scan(struct target *target, uint32_t out, uint32_t *in_ptr)
```
timsifive: Here and elsewhere, if you're now returning ERROR_ values, the type should be int.

src/target/riscv/riscv.c
```diff
@@ -411,11 +414,13 @@ static struct target_type *get_target_type(struct target *target)
...
 	switch (info->dtm_version) {
-		case 0:
+		case DTM_DTMCS_VERSION_0_11:
 			return &riscv011_target;
-		case 1:
+		case DTM_DTMCS_VERSION_1_0:
 			return &riscv013_target;
 		default:
+			// TODO: once we have proper support for non-examined targets
+			// we should have an assert here
```
timsifive: I think OpenOCD still wants /* */ comments (although the enforcement disappeared from the checkpatch script at some point).

src/target/riscv/riscv.c
```diff
@@ -1545,8 +1555,14 @@ static int riscv_examine(struct target *target)
...
 
 	RISCV_INFO(info);
-	uint32_t dtmcontrol = dtmcontrol_scan(target, 0);
+	uint32_t dtmcontrol;
+	if (dtmcontrol_scan(target, 0, &dtmcontrol) != ERROR_OK || dtmcontrol == 0) {
+		LOG_TARGET_ERROR(target, "Could not read dtmcontrol. Check JTAG connectivity/board power.");
+		return ERROR_FAIL;
+	}
 	LOG_TARGET_DEBUG(target, "dtmcontrol=0x%x", dtmcontrol);
+	// FIXME: if examine fails we may need to restore the original dtm_version
```
timsifive: That makes sense. Can you add that change? Shouldn't be much more complicated.

timsifive (changes requested):
How did you test the riscv-011 changes?

src/target/riscv/riscv-011.c
aap-sc: my bad. will fix.

src/target/riscv/riscv.c
aap-sc: will try... though I'm not sure that we need to fix it in this patch. `examine` has it's share of issues:

for example, it halts the target, but if something goes wrong - the target is not resumed. The fix for such situations may be complicated. But having said that - I'll try to address at least `dtm_version` stuff

aap-sc:
> How did you test the riscv-011 changes?

Regarding this...
-  I haven't tested on v0.11 since I don't have such HW, unfortunately. As far as v0.11 goes I've only tried to make sure that the change is reasonable by visually inspecting the code.
- In case of v0.13 there was extensive testing: the whole riscv_tests and additional tests on fpga

TommyMurphyTM1234:
> > How did you test the riscv-011 changes?
> 
> Regarding this...
> 
> * I haven't tested on v0.11 since I don't have such HW, unfortunately. As far as v0.11 goes I've only tried to make sure that the change is reasonable by visually inspecting the code.

If it helps, I can run the tests against a HiFive1 (Rev A I think) in the next couple of days?

aap-sc:
> If it helps, I can run the tests against a HiFive1 (Rev A I think) in the next couple of days?

Oh, it would be great! If you have the bandwidth - please, do. Though I'll need to address few comments by @timsifive

src/target/riscv/riscv-011.c
aap-sc: addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
aap-sc: addressed..

Apparently my whining about examine implementation is not relevant to `riscv.c`

TommyMurphyTM1234:
> > If it helps, I can run the tests against a HiFive1 (Rev A I think) in the next couple of days?
> 
> Oh, it would be great! If you have the bandwidth - please, do. Though I'll need to address few comments by @timsifive

I take it that this PR still needs some changes before it makes sense for me to test against the HiFive1?
(Just to confirm - my HiFive1 is indeed Rev A01 but I presume that that's OK?).

aap-sc:
> I take it that this PR still needs some changes before it makes sense for me to test against the HiFive1?

I've already addressed all concerns raised by @timsifive  . Maybe we should wait for Tim to confirm that this is good enough.

src/target/riscv/riscv-013.c
```diff
@@ -1845,17 +1847,19 @@ static int examine(struct target *target)
 	/* Don't need to select dbus, since the first thing we do is read dtmcontrol. */
 	LOG_TARGET_DEBUG(target, "dbgbase=0x%x", target->dbgbase);
 
-	uint32_t dtmcontrol = dtmcontrol_scan(target, 0);
+	uint32_t dtmcontrol;
+	if (dtmcontrol_scan(target, 0, &dtmcontrol) != ERROR_OK || dtmcontrol == 0) {
+		LOG_TARGET_ERROR(target, "could not scan dtmcontrol. Check JTAG connectivity/board power.");
```
JanMatCodasip: Just nit-picking: 

```suggestion
		LOG_TARGET_ERROR(target, "Could not scan dtmcontrol. Check JTAG connectivity/board power.");
```

src/target/riscv/riscv.c
```diff
@@ -389,14 +390,16 @@ static uint32_t dtmcontrol_scan(struct target *target, uint32_t out)
 
 	int retval = jtag_execute_queue();
 	if (retval != ERROR_OK) {
-		LOG_ERROR("failed jtag scan: %d", retval);
+		LOG_TARGET_ERROR(target, "failed dtmcontrol scan, error code = %d", retval);
```
```suggestion
		LOG_TARGET_ERROR(target, "dtmcontrol scan failed, error code = %d", retval);
```

src/target/riscv/riscv.c
```diff
@@ -1542,25 +1552,40 @@ static int riscv_examine(struct target *target)
...
+	}
 	LOG_TARGET_DEBUG(target, "dtmcontrol=0x%x", dtmcontrol);
 	info->dtm_version = get_field(dtmcontrol, DTMCONTROL_VERSION);
 	LOG_TARGET_DEBUG(target, "version=0x%x", info->dtm_version);
 
+	int examination_status = ERROR_FAIL;
 	struct target_type *tt = get_target_type(target);
 	if (!tt)
-		return ERROR_FAIL;
+		goto report_examination_failure;
```
JanMatCodasip: The goto label can be shorter, if you like. For example:

```suggestion
		goto examine_fail;
```

src/target/riscv/riscv.c
```diff
@@ -1542,25 +1552,40 @@ static int riscv_examine(struct target *target)
...
 
-	int result = tt->init_target(info->cmd_ctx, target);
-	if (result != ERROR_OK)
-		return result;
+	examination_status = tt->init_target(info->cmd_ctx, target);
```
JanMatCodasip: This can be shortened to:

```suggestion
	examine_status = tt->init_target(info->cmd_ctx, target);
```

It becomes easier to read - 3 syllables instead of 5.

src/target/riscv/riscv.c
```diff
@@ -1542,25 +1552,40 @@ static int riscv_examine(struct target *target)
...
+
+	examination_status = tt->examine(target);
+	if (examination_status != ERROR_OK)
+		goto report_examination_failure;
+
+	return ERROR_OK;
 
-	return tt->examine(target);
+report_examination_failure:
+	info->dtm_version = DTM_DTMCS_VERSION_UNKNOWN;
```
JanMatCodasip, line 1580: If examination of the DTM succeeds (dtmcontrol_scan()) but the later examination of the target fails, I believe we should keep the value of `info->dtm_version` because the DTM examination actually passed.

In that case, we should still be able to issue e.g. low-level DMI commands (`riscv dmi_read`, `riscv dmi_write`). I am not sure if that works today, but I believe we should not close the door on that.

JanMatCodasip (commented):
I like this fix/improvement and the increased level of robustness it brings.

I have only few minor comments, otherwise it looks fine.

src/target/riscv/riscv.c
timsifive: I would prefer to reset dtm_version if examine() fails. If examine() fails, it might be because the target is in some wonky state, held in reset, whatever, and the dtm version itself might be suspect. (Obviously this is not supposed to be that way, but hardware isn't always the way it's supposed to be, especially during initialization.) The added cost of detecting it again is very low

JanMatCodasip: Ok, good point, agreed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

JanMatCodasip (approved):
The changes look good to me. 

When testing on HiFive01 Rev A01 with [known fails](https://github.com/riscv/riscv-openocd/issues/869) skipped:

- On HiFive1.py target, the testing passed without any issues.
- On HiFive1-flash.py, there were 8 exceptions, occurring both with and without this change. I haven't investigated these further.

```
:::::::::::::::::::::::::::[ ran 72 tests in 465s ]:::::::::::::::::::::::::::
28 tests returned not_applicable
36 tests returned pass
8 tests returned exception
   DebugBreakpoint > logs/20231009-082756-HiFive1Flash-DebugBreakpoint.log
   DebugExit > logs/20231009-082844-HiFive1Flash-DebugExit.log
   DebugFunctionCall > logs/20231009-082926-HiFive1Flash-DebugFunctionCall.log
   Hwbp1 > logs/20231009-083025-HiFive1Flash-Hwbp1.log
   Hwbp2 > logs/20231009-083108-HiFive1Flash-Hwbp2.log
   Registers > logs/20231009-083240-HiFive1Flash-Registers.log
   TooManyHwbp > logs/20231009-083342-HiFive1Flash-TooManyHwbp.log
   UserInterrupt > logs/20231009-083442-HiFive1Flash-UserInterrupt.log

```

TommyMurphyTM1234:
> The changes look good to me.
> 
> When testing on HiFive01 Rev A01 with [known fails](https://github.com/riscv/riscv-openocd/issues/869) skipped:
> 
> * On HiFive1.py target, the testing passed without any issues.
> * On HiFive1-flash.py, there were 8 exceptions, occurring both with and without this change. I haven't investigated these further.

If it's of any use, I plan to look at this imminently and could also look at those additional (?) exceptions?

* https://github.com/riscv-software-src/riscv-tests/issues/488

JanMatCodasip:
> If it's of any use, I plan to look at this imminently and could also look at those additional (?) exceptions?

Any help with the failing tests is appreciated :)

timsifive (approved):

aap-sc:
Hi @timsifive , just wanted to double-check if there are any additional changes/fixes required for this PR?

PR #931, opened 2023-10-03
target: Update messages connected with `examine`

kr-sc:
Move `examine_attempted` flag to target struct to make it target specific. 

`Info` messages for retry and `Error` messages for failure added.

P.S. I have to request this patch here because of https://review.openocd.org/c/openocd/+/6964. There is no such patch in upstream openocd repo.

### DIFF
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -744,16 +744,22 @@ static int default_check_reset(struct target *target)
  * Keep in sync */
 int target_examine_one(struct target *target)
 {
+	if (target->examine_attempted)
+		LOG_TARGET_INFO(target, "Retry examination.");
+
+	LOG_TARGET_INFO(target, "Examination started.");
+
 	target_call_event_callbacks(target, TARGET_EVENT_EXAMINE_START);
 
 	int retval = target->type->examine(target);
 	if (retval != ERROR_OK) {
+		LOG_TARGET_ERROR(target, "Examination failed. examine() -> %d", retval);
 		target_reset_examined(target);
 		target_call_event_callbacks(target, TARGET_EVENT_EXAMINE_FAIL);
 		return retval;
 	}
 
-	LOG_USER("[%s] Target successfully examined.", target_name(target));
+	LOG_TARGET_INFO(target, "Target successfully examined.");
 	target_set_examined(target);
 	target_call_event_callbacks(target, TARGET_EVENT_EXAMINE_END);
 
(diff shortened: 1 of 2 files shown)


src/target/target.c
```diff
@@ -744,16 +744,22 @@ static int default_check_reset(struct target *target)
...
 	int retval = target->type->examine(target);
 	if (retval != ERROR_OK) {
+		LOG_TARGET_ERROR(target, "Examination failed. examine() -> %d", retval);
 		target_reset_examined(target);
 		target_call_event_callbacks(target, TARGET_EVENT_EXAMINE_FAIL);
 		return retval;
 	}
 
-	LOG_USER("[%s] Target successfully examined.", target_name(target));
+	LOG_TARGET_INFO(target, "Target successfully examined.");
```
timsifive, line 762: Why are you changing this from _USER to _INFO? (I don't actually know what the difference is.)

timsifive (changes requested):
Looks OK.

src/target/target.c
kr-sc: Difference that output of `LOG_USER` will be showed with any `debug_level`, `LOG_INFO` will be showed with `debug_level 2`. I added message `Examination started` using `LOG_TARGET_INFO` so I thought that is good idea to use `LOG_TARGET_INFO` for this message too,

timsifive: OK

timsifive (approved):

JanMatCodasip (approved):

en-sc:
@kr-sc , @timsifive,  unfortunately, this needs to be reverted. The implementation is buggy. 

We have `target_examine()` function:
```
    for (target = all_targets; target; target = target->next) {
        /* defer examination, but don't skip it */
        if (!target->tap->enabled) {
            jtag_register_event_callback(jtag_enable_callback,
                    target);
            continue;
        }
        if (target->defer_examine)
            continue;

        target->examine_attempted = true;
```
This function is called only once. From src/openocd.c. So, the original intention for the variable is just to mark that this function was called.  I guess that this may result in subtle errors, since the condition for:
```
    if (retval != ERROR_OK && target->examine_attempted) {                                        
        target_reset_examined(target);                                                              
        retval = target_examine_one(target);                                                        
        if (retval != ERROR_OK) {                                                                   
           LOG_TARGET_DEBUG(target, "Polling again in %dms",                                         
              target->backoff.interval);                                                              
           return retval;                                                                            
         }                                                                                           
    }
```    
Will never be true for the targets that had a deferred examine.

kr-sc:
@en-sc is right, unfortunately
mr with revert here https://github.com/riscv/riscv-openocd/pull/934

PR #934, opened 2023-10-16
Revert "target: Update messages connected with `examine`"

kr-sc:
This reverts commit a3db93b1cea56614baf04f8e88742049da9f2f1e.

Reason for revert: https://github.com/riscv/riscv-openocd/pull/931#issuecomment-1761550506

### DIFF
--- a/src/target/target.c
+++ b/src/target/target.c
@@ -744,22 +744,16 @@ static int default_check_reset(struct target *target)
  * Keep in sync */
 int target_examine_one(struct target *target)
 {
-	if (target->examine_attempted)
-		LOG_TARGET_INFO(target, "Retry examination.");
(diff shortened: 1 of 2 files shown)


timsifive (commented):
Can you update the description of this PR to explain why the change needs to be reverted?

kr-sc:
> Can you update the description of this PR to explain why the change needs to be reverted?

@en-sc explained here https://github.com/riscv/riscv-openocd/pull/931#issuecomment-1761550506

Also, @en-sc found a bug that message  `LOG_TARGET_INFO(target, "Retry examination.");`  is printed even at the first examine atttempt.

timsifive (approved):

PR #935, opened 2023-10-16
From upstream

timsifive:
Merge down up to 0384fe5d596f42388f8b84d42959d899f29388ab from upstream.

checkpatch has complaints, but I'm going to override them because they came from upstream code.


MarekVCodasip (approved):
I checked the differences, looks fine to me.

JanMatCodasip (approved):
@MarekVCodasip has done the bulk of this review and has checked the differences. 

We have also skimmed through the differences together and it looks fine to me as well.

PR #937, opened 2023-10-17
contrib: Match upstream.

timsifive:
Upstream has a checkpatch failure here. I had fixed it because I didn't know how else to properly get around it back then. Reintroduce the problem. Now this file is identical to upstream.

Change-Id: Ic03b6bb42945ddbcfd2fe12c0cab5b05eda1a50c


MarekVCodasip:
@JanMatCodasip asked me to look at it. I consider it a good thing to converge the two forks. Same as with the other patch.

MarekVCodasip (approved):

JanMatCodasip (approved):

PR #945, opened 2023-10-25
target/riscv: Fix memory access when MMU is enabled and address couldn't be translated

kr-sc:
Now:
1) If mmu is disabled, virt2phys succeeded and returns physical address 
2) If mmu is enabled, but translation fails, read/write_memory fails

Change-Id: I312309c660239014b3278cb77cadc5618de8e4de

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2095,10 +2095,10 @@ static int riscv_effective_privilege_mode(struct target *target, int *v_mode, in
 
 static int riscv_mmu(struct target *target, int *enabled)
 {
-	if (!riscv_enable_virt2phys) {
-		*enabled = 0;
+	*enabled = 0;
+
+	if (!riscv_enable_virt2phys)
 		return ERROR_OK;
-	}
 
 	/* Don't use MMU in explicit or effective M (machine) mode */
 	riscv_reg_t priv;
@@ -2129,6 +2129,7 @@ static int riscv_mmu(struct target *target, int *enabled)
 				/* In hypervisor mode regular satp translation
 				 * doesn't happen. */
 				return ERROR_OK;
+
 		}
 
 		riscv_reg_t vsatp;
@@ -2155,7 +2156,6 @@ static int riscv_mmu(struct target *target, int *enabled)
 			*enabled = 1;
 		} else {
 			LOG_TARGET_DEBUG(target, "No V-mode address translation enabled.");
-			*enabled = 0;
 		}
 
 		return ERROR_OK;
@@ -2164,7 +2164,6 @@ static int riscv_mmu(struct target *target, int *enabled)
 	/* Don't use MMU in explicit or effective M (machine) mode */
 	if (effective_mode == PRV_M) {
 		LOG_TARGET_DEBUG(target, "SATP/MMU ignored in Machine mode.");
-		*enabled = 0;
 		return ERROR_OK;
 	}
 
@@ -2172,13 +2171,11 @@ static int riscv_mmu(struct target *target, int *enabled)
 	if (riscv_get_register(target, &satp, GDB_REGNO_SATP) != ERROR_OK) {
 		LOG_TARGET_DEBUG(target, "Couldn't read SATP.");
 		/* If we can't read SATP, then there must not be an MMU. */
-		*enabled = 0;
 		return ERROR_OK;
 	}
 
 	if (get_field(satp, RISCV_SATP_MODE(xlen)) == SATP_MODE_OFF) {
 		LOG_TARGET_DEBUG(target, "MMU is disabled.");
-		*enabled = 0;
 	} else {
 		LOG_TARGET_DEBUG(target, "MMU is enabled.");
 		*enabled = 1;
@@ -2315,6 +2312,8 @@ static int riscv_virt2phys_v(struct target *target, target_addr_t virtual, targe
 			break;
 		case SATP_MODE_OFF:
 			vsatp_info = NULL;
+			LOG_TARGET_DEBUG(target, "vsatp mode is %d. No VS-stage translation. (vsatp: 0x%" PRIx64 ")",
+				vsatp_mode, vsatp);
 			break;
 		default:
 			LOG_TARGET_ERROR(target,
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv.c
```diff
@@ -2282,17 +2282,19 @@ static int riscv_address_translate(struct target *target,
 static int riscv_virt2phys_v(struct target *target, target_addr_t virtual, target_addr_t *physical)
 {
 	riscv_reg_t vsatp;
-	if (riscv_get_register(target, &vsatp, GDB_REGNO_VSATP) != ERROR_OK) {
-		LOG_TARGET_ERROR(target, "Failed to read vsatp register.");
+	int result = riscv_get_register(target, &vsatp, GDB_REGNO_VSATP);
+	if (result != ERROR_OK) {
+		LOG_TARGET_ERROR(target, "Failed to read vsatp register. Error: %d", result);
 		return ERROR_FAIL;
```
```suggestion
		return result;
```

src/target/riscv/riscv.c
```diff
@@ -2282,17 +2282,19 @@ static int riscv_address_translate(struct target *target,
...
 	unsigned int xlen = riscv_xlen(target);
 	int vsatp_mode = get_field(vsatp, RISCV_SATP_MODE(xlen));
 	LOG_TARGET_DEBUG(target, "VS-stage translation mode: %d", vsatp_mode);
 	riscv_reg_t hgatp;
-	if (riscv_get_register(target, &hgatp, GDB_REGNO_HGATP) != ERROR_OK) {
-		LOG_TARGET_ERROR(target, "Failed to read hgatp register.");
+	result = riscv_get_register(target, &hgatp, GDB_REGNO_HGATP);
+	if (result != ERROR_OK) {
+		LOG_TARGET_ERROR(target, "Failed to read hgatp register. Error: %d", result);
 		return ERROR_FAIL;
```
```suggestion
		return result;
```

src/target/riscv/riscv.c
```diff
@@ -2390,23 +2396,28 @@ static int riscv_virt2phys(struct target *target, target_addr_t virtual, target_
 	int enabled;
 	if (riscv_mmu(target, &enabled) != ERROR_OK)
 		return ERROR_FAIL;
-	if (!enabled)
-		return ERROR_FAIL;
-
+	if (!enabled) {
+		*physical = virtual;
```
timsifive, line 2395: Does this fix anything, or is this just defensive coding?

src/target/riscv/riscv.c
```diff
@@ -2456,11 +2467,14 @@ static int riscv_read_memory(struct target *target, target_addr_t address,
...
-		address = physical_addr;
+	int result = target->type->virt2phys(target, address, &physical_addr);
+	if (result != ERROR_OK) {
+		LOG_TARGET_ERROR(target, "Address translation failed.");
+		return result;
+	}
 
 	RISCV_INFO(r);
-	return r->read_memory(target, address, size, count, buffer, size);
+	return r->read_memory(target, physical_addr, size, count, buffer, size);
```
timsifive, line 2470: Wow. That's kind of a big bug. This code never worked at all?

timsifive (changes requested):
Looks good overall. Please return result where that is now easily possible.

src/target/riscv/riscv.c
aap-sc: it's kinda worked. there was re-assignment of the address variable to a translated address.

src/target/riscv/riscv.c
kr-sc: After internal discussion we decided not to print/return error code here, because it looks like useless for users.

src/target/riscv/riscv.c
kr-sc: As you can see, logic of `read/write_memory` functions  was changed (assignment to `address` was removed). So, when MMU is disabled we want to perform a memory access using untraslated address, and for this we need this change.

src/target/riscv/riscv.c
kr-sc: Yes, now we store real address in `physical_addr` and it filled by `riscv_virt2phys` even if MMU is disabled.

src/target/riscv/riscv.c
```diff
@@ -2095,6 +2095,8 @@ static int riscv_effective_privilege_mode(struct target *target, int *v_mode, in
 
 static int riscv_mmu(struct target *target, int *enabled)
 {
+	*enabled = 0;
```
JanMatCodasip, line 2098: Some `return`s in this function are preceded by explicit `*enabled = <value>;` whereas others rely on this initialization at the very top of the function.

For consistency, I would recommend to use the same style for all returns.

aap-sc: As you can see there were some execution paths that returned ERROR_OK without actually  initializing the value (like on line 21333). 

So, I guess that this is just a  defensive coding (not the real returned value). That being said - it would not hurt to add a comment to clarify the intention.

JanMatCodasip: The initialization at the very top (defensive approach) looks good to me. 

I repeat my vote to unify the rest of the function: to remove the extra `*enabled = 0` assignments. They are redundant, and present only in some of the code paths, not everywhere (an inconsistent style that I would recommend to avoid).

aap-sc: > I repeat my vote to unify the rest of the function: to remove the extra *enabled = 0 assignment

Ah.. I see. I've misinterpreted the original suggestion. Yes, it would be an improvement.

src/target/riscv/riscv.c
timsifive: That is fine.

src/target/riscv/riscv.c
timsifive: Thanks for explaining.

src/target/riscv/riscv.c
kr-sc: Same comment as above

src/target/riscv/riscv.c
kr-sc: Redundant assignments removed. Addressed.

timsifive (approved):

PR #946, opened 2023-10-25
target/riscv: update debug register printers

en-sc:
Change-Id: I069bbe069a3aaa7fd3a4f6eccde40f813db33cc9


en-sc:
This update is made to include [a bugfix](https://github.com/riscv/riscv-debug-spec/pull/908), recently merged into `riscv-debug-spec`

timsifive (approved):

PR #949, opened 2023-10-31
Remove mention of esp32c2, esp32c3 from doc

JanMatCodasip:
Targets "esp32c2" and "esp32c3" should not be mentioned in the doc under "target types" because these are not standalone OpenOCD targets.

They are merely a set of .cfg files which use the generic "riscv" target.


JanMatCodasip:
Noticed as a diff against vanilla upstream during the review of https://github.com/riscv/riscv-openocd/pull/947.

timsifive (approved):

PR #951, opened 2023-11-01
target/riscv: dump_field() shouldn't always decode

en-sc:
Sometimes, the value from of some DMI scans has no meaning (e.g. when `op` is read). Such values should not be decoded. To make the dumps more consistent, `<no decoding available>` is printed when there is no decoding for a register.

Change-Id: I415f06a5a80f2fc8fb8ab3f79132bdf0602c8ad6

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -389,8 +389,7 @@ static void dump_field(struct target *target, int idle, const struct scan_field 
 	unsigned int in_data = get_field(in, DTM_DMI_DATA);
 	unsigned int in_address = in >> DTM_DMI_ADDRESS_OFFSET;
 
-	log_printf_lf(LOG_LVL_DEBUG,
-			__FILE__, __LINE__, __PRETTY_FUNCTION__,
+	log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __func__,
 			"%db %s %08x @%02x -> %s %08x @%02x; %di",
 			field->num_bits, op_string[out_op], out_data, out_address,
 			status_string[in_op], in_data, in_address, idle);
@@ -398,13 +397,13 @@ static void dump_field(struct target *target, int idle, const struct scan_field 
 	if (out_op == DTM_DMI_OP_WRITE) {
 		char out_decoded[decode_dmi(target, NULL, out_address, out_data) + 1];
 		decode_dmi(target, out_decoded, out_address, out_data);
-		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __PRETTY_FUNCTION__,
+		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __func__,
 				"write: %s", out_decoded);
 	}
 	if (!discard_in && in_op == DTM_DMI_OP_SUCCESS) {
 		char in_decoded[decode_dmi(target, NULL, in_address, in_data) + 1];
 		decode_dmi(target, in_decoded, in_address, in_data);
-		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __PRETTY_FUNCTION__,
+		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __func__,
 				"read: %s", in_decoded);
 	}
 }
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -371,7 +371,7 @@ static unsigned int decode_dmi(struct target *target, char *text, unsigned int a
 	return decode_dm(text, address - dm->base, data);
 }
 
-static void dump_field(struct target *target, int idle, const struct scan_field *field)
+static void dump_field(struct target *target, int idle, const struct scan_field *field, bool discard_in)
 {
 	static const char * const op_string[] = {"-", "r", "w", "?"};
 	static const char * const status_string[] = {"+", "?", "F", "b"};
@@ -390,18 +390,22 @@ static void dump_field(struct target *target, int idle, const struct scan_field 
 	unsigned int in_address = in >> DTM_DMI_ADDRESS_OFFSET;
 
 	log_printf_lf(LOG_LVL_DEBUG,
-			__FILE__, __LINE__, "scan",
+			__FILE__, __LINE__, __PRETTY_FUNCTION__,
 			"%db %s %08x @%02x -> %s %08x @%02x; %di",
 			field->num_bits, op_string[out_op], out_data, out_address,
 			status_string[in_op], in_data, in_address, idle);
 
-	char out_text[decode_dmi(target, NULL, out_address, out_data) + 1];
-	unsigned int out_len = decode_dmi(target, out_text, out_address, out_data);
-	char in_text[decode_dmi(target, NULL, in_address, in_data) + 1];
-	unsigned int in_len = decode_dmi(target, in_text, in_address, in_data);
-	if (in_text[0] || out_text[0]) {
-		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, "scan", "%.*s -> %.*s",
-				out_len, out_text, in_len, in_text);
+	if (out_op == DTM_DMI_OP_WRITE) {
+		char out_decoded[decode_dmi(target, NULL, out_address, out_data) + 1];
+		decode_dmi(target, out_decoded, out_address, out_data);
+		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __PRETTY_FUNCTION__,
+				"write: %s", out_decoded);
+	}
+	if (!discard_in && in_op == DTM_DMI_OP_SUCCESS) {
+		char in_decoded[decode_dmi(target, NULL, in_address, in_data) + 1];
+		decode_dmi(target, in_decoded, in_address, in_data);
+		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __PRETTY_FUNCTION__,
+				"read: %s", in_decoded);
 	}
 }
 
@@ -542,7 +546,7 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 
 	if (address_in)
 		*address_in = buf_get_u32(in, DTM_DMI_ADDRESS_OFFSET, info->abits);
-	dump_field(target, idle_count, &field);
+	dump_field(target, idle_count, &field, /*discard_in*/ !data_in);
 	return buf_get_u32(in, DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH);
 }
 
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -215,13 +215,13 @@ static void dump_field(int idle, const struct scan_field *field)
 		unsigned int in_address = in >> DTM_DMI_ADDRESS_OFFSET;
 
 		log_printf_lf(LOG_LVL_DEBUG,
-				__FILE__, __LINE__, __PRETTY_FUNCTION__,
+				__FILE__, __LINE__, __func__,
 				"%db %s %08x @%02x -> %s %08x @%02x; %di",
 				field->num_bits, op_string[out_op], out_data, out_address,
 				status_string[in_op], in_data, in_address, idle);
 	} else {
 		log_printf_lf(LOG_LVL_DEBUG,
-				__FILE__, __LINE__, __PRETTY_FUNCTION__, "%db %s %08x @%02x -> ?; %di",
+				__FILE__, __LINE__, __func__, "%db %s %08x @%02x -> ?; %di",
 				field->num_bits, op_string[out_op], out_data, out_address, idle);
 	}
 }


en-sc:
Prior to the change:
```
Debug: 393 104 riscv-013.c:388 scan(): 40b r 00000000 @11 -> + 00000000 @16; 0i
Debug: 394 104 riscv-013.c:398 scan(): dmstatus=0 { version=none, anyrunning=0, allrunning=0, anyunavail=0, allunavail=0, anynonexistent=0, allnonexistent=0, anyresumeack=0, allresumeack=0, anyhavereset=0, allhavereset=0, impebreak=0, stickyunavail=current, ndmresetpending=false, confstrptrvalid=invalid, hasresethaltreq=0, authbusy=ready, authenticated=false, anyhalted=0, allhalted=0, } -> abstractcs=0 { datacount=0, relaxedpriv=full checks, busy=ready, progbufsize=0, cmderr=none, }
Debug: 395 104 riscv-013.c:388 scan(): 40b - 00000000 @11 -> + 00430382 @11; 0i
Debug: 396 104 riscv-013.c:398 scan(): dmstatus=0 { version=none, anyrunning=0, allrunning=0, anyunavail=0, allunavail=0, anynonexistent=0, allnonexistent=0, anyresumeack=0, allresumeack=0, anyhavereset=0, allhavereset=0, impebreak=0, stickyunavail=current, ndmresetpending=false, confstrptrvalid=invalid, hasresethaltreq=0, authbusy=ready, authenticated=false, anyhalted=0, allhalted=0, } -> dmstatus=0x430382 { version=0.13, anyrunning=0, allrunning=0, anyunavail=0, allunavail=0, anynonexistent=0, allnonexistent=0, anyresumeack=1, allresumeack=1, anyhavereset=0, allhavereset=0, impebreak=1, stickyunavail=current, ndmresetpending=false, confstrptrvalid=invalid, hasresethaltreq=0, authbusy=ready, authenticated=true, anyhalted=1, allhalted=1, }
...
Debug: 5379 820 riscv-013.c:388 scan(): 40b w 0032100c @17 -> + 00000000 @04; 1i
Debug: 5380 820 riscv-013.c:398 scan(): command=0x32100c { control=0x32100c, cmdtype=0, } ->
Debug: 5381 820 riscv-013.c:388 scan(): 40b r 00000000 @16 -> + 0032100c @17; 0i
Debug: 5382 820 riscv-013.c:398 scan(): abstractcs=0 { datacount=0, relaxedpriv=full checks, busy=ready, progbufsize=0, cmderr=none, } -> command=0x32100c { control=0x32100c, cmdtype=0, }
Debug: 5383 821 riscv-013.c:388 scan(): 40b - 00000000 @16 -> + 06000002 @16; 0i
Debug: 5384 821 riscv-013.c:398 scan(): abstractcs=0 { datacount=0, relaxedpriv=full checks, busy=ready, progbufsize=0, cmderr=none, } -> abstractcs=0x6000002 { datacount=2, relaxedpriv=full checks, busy=ready, progbufsize=6, cmderr=none, }
Debug: 5385 821 riscv-013.c:388 scan(): 40b r 00000000 @05 -> + 00000000 @16; 0i
Debug: 5386 821 riscv-013.c:398 scan():  -> abstractcs=0 { datacount=0, relaxedpriv=full checks, busy=ready, progbufsize=0, cmderr=none, }
Debug: 5387 821 riscv-013.c:388 scan(): 40b - 00000000 @05 -> + 00000000 @05; 0i
Debug: 5388 821 riscv-013.c:398 scan():  ->
Debug: 5389 821 riscv-013.c:388 scan(): 40b r 00000000 @04 -> + 00000000 @05; 0i
Debug: 5390 821 riscv-013.c:398 scan():  ->
Debug: 5391 821 riscv-013.c:388 scan(): 40b - 00000000 @04 -> + 00800000 @04; 0i
Debug: 5392 821 riscv-013.c:398 scan():  ->
```
After the change:
```
Debug: 4688 1567 riscv-013.c:415 dump_field(): 40b r 00000000 @11 -> + 00000000 @04; 103i
Debug: 4689 1567 riscv-013.c:415 dump_field(): 40b - 00000000 @11 -> + 00430382 @11; 103i
Debug: 4690 1567 riscv-013.c:436 dump_field():  -> dmstatus=0x430382 { version=0.13, anyrunning=0, allrunning=0, anyunavail=0, allunavail=0, anynonexistent=0, allnonexistent=0, anyresumeack=1, allresumeack=1, anyhavereset=0, allhavereset=0, impebreak=1, stickyunavail=current, ndmresetpending=false, confstrptrvalid=invalid, hasresethaltreq=0, authbusy=ready, authenticated=true, anyhalted=1, allhalted=1, }
...
Debug: 4947 1614 riscv-013.c:415 dump_field(): 40b w 0032100c @17 -> + 00000000 @04; 103i
Debug: 4948 1614 riscv-013.c:436 dump_field(): command=0x32100c { control=0x32100c, cmdtype=0, } ->
Debug: 4949 1615 riscv-013.c:415 dump_field(): 40b r 00000000 @16 -> + 0032100c @17; 103i
Debug: 4950 1615 riscv-013.c:415 dump_field(): 40b - 00000000 @16 -> + 06000002 @16; 103i
Debug: 4951 1615 riscv-013.c:436 dump_field():  -> abstractcs=0x6000002 { datacount=2, relaxedpriv=full checks, busy=ready, progbufsize=6, cmderr=none, }
Debug: 4952 1616 riscv-013.c:415 dump_field(): 40b r 00000000 @05 -> + 00000000 @16; 103i
Debug: 4953 1617 riscv-013.c:415 dump_field(): 40b - 00000000 @05 -> + 00000000 @05; 103i
Debug: 4954 1617 riscv-013.c:436 dump_field():  -> <no decoding available>
Debug: 4955 1617 riscv-013.c:415 dump_field(): 40b r 00000000 @04 -> + 00000000 @05; 103i
Debug: 4956 1618 riscv-013.c:415 dump_field(): 40b - 00000000 @04 -> + 00800000 @04; 103i
Debug: 4957 1618 riscv-013.c:436 dump_field():  -> data0=0x800000 { data=0x800000, }
```

src/target/riscv/riscv-013.c
```diff
@@ -336,11 +336,36 @@ static unsigned int decode_dm(char *text, unsigned int address, unsigned int dat
...
+		{DM_CONFSTRPTR2, DM_CONFSTRPTR2_ORDINAL},
+		{DM_CONFSTRPTR3, DM_CONFSTRPTR3_ORDINAL},
+		{DM_NEXTDM, DM_NEXTDM_ORDINAL},
+		{DM_DATA0, DM_DATA0_ORDINAL},
+		{DM_PROGBUF0, DM_PROGBUF0_ORDINAL},
+		{DM_AUTHDATA, DM_AUTHDATA_ORDINAL},
+		{DM_DMCS2, DM_DMCS2_ORDINAL},
+		{DM_HALTSUM0, DM_HALTSUM0_ORDINAL},
+		{DM_HALTSUM1, DM_HALTSUM1_ORDINAL},
+		{DM_HALTSUM2, DM_HALTSUM2_ORDINAL},
```
timsifive: Is this oversight from another change? This doesn't seem to have anything to do with the description of this PR. Can you update the description to mention every significant change from this PR?

en-sc: Thanks! Removed the change to `decode_dm()`. Commit description is updated to mention the addition of `<no decoding available>`.

src/target/riscv/riscv-013.c
```diff
@@ -390,18 +390,29 @@ static void dump_field(struct target *target, int idle, const struct scan_field
...
 
-	char out_text[decode_dmi(target, NULL, out_address, out_data) + 1];
-	unsigned int out_len = decode_dmi(target, out_text, out_address, out_data);
-	char in_text[decode_dmi(target, NULL, in_address, in_data) + 1];
-	unsigned int in_len = decode_dmi(target, in_text, in_address, in_data);
-	if (in_text[0] || out_text[0]) {
-		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, "scan", "%.*s -> %.*s",
-				out_len, out_text, in_len, in_text);
+	const bool decode_out = out_op == DTM_DMI_OP_WRITE;
+	char out_decoded[decode_dmi(target, NULL, out_address, out_data) + 1];
```
timsifive: Can you move this inside `if (decode_out)` to avoid unnecessary work?

src/target/riscv/riscv-013.c
```diff
@@ -390,18 +390,29 @@ static void dump_field(struct target *target, int idle, const struct scan_field
...
+	const char *out_text = "";
+	if (decode_out)
+		out_text = decode_dmi(target, out_decoded, out_address, out_data)
+			? out_decoded
+			: "<no decoding available>";
+
+	char in_decoded[decode_dmi(target, NULL, in_address, in_data) + 1];
```
timsifive: Same.

src/target/riscv/riscv-013.c
```diff
@@ -390,18 +390,29 @@ static void dump_field(struct target *target, int idle, const struct scan_field
...
-	if (in_text[0] || out_text[0]) {
-		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, "scan", "%.*s -> %.*s",
-				out_len, out_text, in_len, in_text);
+	const bool decode_out = out_op == DTM_DMI_OP_WRITE;
+	char out_decoded[decode_dmi(target, NULL, out_address, out_data) + 1];
+	const char *out_text = "";
+	if (decode_out)
+		out_text = decode_dmi(target, out_decoded, out_address, out_data)
+			? out_decoded
+			: "<no decoding available>";
```
timsifive: "no decoding available" sounds a bit odd. We could decode it if we wanted to. We (probably) have the relevant field information available to decode. How about "<value ignored>"?

timsifive (changes requested):
I like this idea to reduce debug clutter.

src/target/riscv/riscv-013.c
```diff
@@ -390,18 +390,29 @@ static void dump_field(struct target *target, int idle, const struct scan_field
...
+	char in_decoded[decode_dmi(target, NULL, in_address, in_data) + 1];
+	const char *in_text = "";
+	if (decode_in)
+		in_text = decode_dmi(target, in_decoded, in_address, in_data)
+			? in_decoded
+			: "<no decoding available>";
+
+	if (decode_out || decode_in) {
+		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __PRETTY_FUNCTION__,
+				"%s -> %s", out_text, in_text);
```
JanMatCodasip: I would recommend to print the decoded values onto separate lines. And also to indent them (e.g. by 2 spaces) so that in the log they appear visually related to the preceding raw DMI print. Something like:

```suggestion
	if (decode_out) {
		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __PRETTY_FUNCTION__,
				"  writing: %s", out_text);
		// or "out:" instead of "writing:"
	}
	if (decode_in) {
		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __PRETTY_FUNCTION__,
				"  read: %s", in_text);
		// or "in:" instead of "read:"
	}	
```

Reasons:
- The decoded text produces very long log lines, especially when both "in" and "out" would be printed.
- When only one of the directions is decoded, it would be easier to tell if it is "in" or "out" by reading the prefix than checking whether the text is on the left (or on the right) of the `->` sign.

src/target/riscv/riscv-013.c
```diff
@@ -542,7 +553,7 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 
 	if (address_in)
 		*address_in = buf_get_u32(in, DTM_DMI_ADDRESS_OFFSET, info->abits);
-	dump_field(target, idle_count, &field);
+	dump_field(target, idle_count, &field, /*decode_in*/ data_in);
```
JanMatCodasip: I believe the condition for `decode_in` should be extended like this:

- We should keep track what type of DMI operation is currently in flight for the given debug module (e.g. store it to the instance `dm013_info_t`).
- The received data is only valid (and should only be decoded) when when the current operation is DMI_OP_READ and at the same time the received `op==0` (success). 

In all other cases than the above, the received DMI data is irrelevant and should not be decoded. Decoding it would be misleading to the users who read the log.

JanMatCodasip (changes requested):

src/target/riscv/riscv-013.c
en-sc: Currently, the condition is:
The value is decoded if it is not discarded by the caller of the function.

IMHO this is preferable, since it is more robust -- if there is an error in OpenOCD and a meaningless value is used and analyzed further in the code, such value will be present in the log. Moreover, the solution you suggest requires a lot of changes -- there is a separate interface to DMI operations (`batch.h/c`), which would need to update this field in `dm013_info_t`. If someone was to implement it -- the following still should be true: `!data_in || (<last op is read> && <in_op == success>)`. 
If you wish, I can add an assertion into `dump_field()`:
```
assert(!decode_in || in_op == 0)
```

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
JanMatCodasip: Thank you for replying. 

> If you wish, I can add an assertion into dump_field(): assert(!decode_in || in_op == 0)

Please don't. This assertion would not hold true when the answer from the target is `in_op==BUSY`.

> Currently, the condition is: The value is decoded if it is not discarded by the caller of the function. IMHO this is preferable, since it is more robust -- if there is an error in OpenOCD and a meaningless value is used and analyzed further in the code, such value will be present in the log.

Whether the received data is valid (and should be interpreted, for example decoded), is precisely determined by the JTAG DTM specification. That is, by the history of `op` values exchanged between OpenOCD and the target. Not by the fact that the caller of `dmi_scan()` reads the data_in.

In fact that the caller requests the data_in:
- does not mean the dmi.data is valid, 
- does not mean the caller will make any use of it.

> Moreover, the solution you suggest requires a lot of changes

I understand that the variant I proposed is more complex to implement. It seems OK to me to skip that part for now, and (if you like) file a TODO issue for it.

For now, I believe you can at least partially improve the situation by not decoding the data when the target responds that the previous operation was unsuccessful:

```
	/* The write operation only got started if the previous operation succeeded. 
	 * Otherwise the write is ignored. */
	if (out_op == DTM_DMI_OP_WRITE && in_op == DMI_OP_SUCCESS) {
		/* ... */
	}
	/* The read data can only be valid if the previous operation succeeded. */	
	if (decode_in &&  in_op == DMI_OP_SUCCESS) {
		char in_decoded[decode_dmi(target, NULL, in_address, in_data) + 1];
		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __PRETTY_FUNCTION__, "in: %s",
				decode_dmi(target, in_decoded, in_address, in_data)
				? in_decoded
				: "");
	}
```

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

JanMatCodasip (dismissed):
This looks good to me.

As [discussed](https://github.com/riscv/riscv-openocd/pull/951#discussion_r1386089782), a more elaborate algorithm to detect validity of the incoming data could be implemented. Anyway, this change is still nice improvement over the current state.

src/target/riscv/riscv-013.c
```diff
@@ -390,18 +390,24 @@ static void dump_field(struct target *target, int idle, const struct scan_field
...
-	char out_text[decode_dmi(target, NULL, out_address, out_data) + 1];
-	unsigned int out_len = decode_dmi(target, out_text, out_address, out_data);
-	char in_text[decode_dmi(target, NULL, in_address, in_data) + 1];
-	unsigned int in_len = decode_dmi(target, in_text, in_address, in_data);
-	if (in_text[0] || out_text[0]) {
-		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, "scan", "%.*s -> %.*s",
-				out_len, out_text, in_len, in_text);
+	if (out_op == DTM_DMI_OP_WRITE) {
+		char out_decoded[decode_dmi(target, NULL, out_address, out_data) + 1];
+		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __PRETTY_FUNCTION__, "out: %s",
```
```suggestion
		log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __PRETTY_FUNCTION__, "write: %s",
```

src/target/riscv/riscv-013.c
```diff
@@ -390,18 +390,24 @@ static void dump_field(struct target *target, int idle, const struct scan_field
...
+				decode_dmi(target, out_decoded, out_address, out_data)
+				? out_decoded
+				: "");
```
timsifive: I think I was wrong with my initial feedback. This string shows up when decode_dmi() returns an empty string. In that case "<no decoding available>" is exactly the right thing to print. Sorry.

timsifive (changes requested):
Minor feedback.

src/target/riscv/riscv-013.c
en-sc: Addressed. Changed `"in"` to `"read"` as well.

src/target/riscv/riscv-013.c
en-sc: Sorry, I didn't quite understand what you mean. I have replaced `"<no decoding available>"` to `""` as you suggested. Though I decided to use the output of `decode_dmi()` directly (without the ternary operator).

timsifive (dismissed):

timsifive:
This `__PRETTY_FUNCTION__` change makes the output terrible:
```
Debug: 4020 1652 batch.c:224 void dump_field(int, const struct scan_field *)(): 40b w 5555aaaa @3c -> ?; 1i
```
I don't care to see the prototype for dump_field() all the time. Please revert that. There is really no reason to mention the function name at all. It just needs to be something concise that we can easily ID as a RISC-V JTAG scan.

en-sc:
> ```
> Debug: 4020 1652 batch.c:224 void dump_field(int, const struct scan_field *)(): 40b w 5555aaaa @3c -> ?; 1i
> ```
@timsifive, thank you for the catch! IMHO, this is quite interesting. The very line you quoted was not affected by my commit (it's a similar function for batch operations).
Have you compiled using Clang?
The thing is `__PRETTY_FUNCTION__` is a GNU extension and it is defined to be the same as `__func__` for C sources (in this usecase) but differ for C++ sources (full signature instead of just a name) by [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/Function-Names.html).
Currently Clang does support `__PRETTY_FUNCTION__`, though it uses GCC's C++ variant across C and C++.

I would like to leave log's format as consistent as possible, so I have just replaced `__PRETTY_FUNCTION__` with `__func__` in a separate commit (`__func__` is already used in `LOG_DEBUG` and so on).

timsifive:
Oh, I *am* building with clang. I switched because I was checking if something worked. I didn't realize that would cause this difference.

timsifive:
Can you rebase this on top of the current `riscv`? It makes it more clear what changes are actually new.

en-sc:
> Can you rebase this on top of the current `riscv`? It makes it more clear what changes are actually new.

Done.

timsifive (approved):

JanMatCodasip (approved):

PR #952, opened 2023-11-03
target/riscv: Stop caching writes to DPC

MarekVCodasip:
Since DPC is WARL (same rules as MEPC according to the specification), it is possible that
writes to it won't result in the exact value present. Therefore, writes to it shouldn't be cached, same as with other WARL registers.

Change-Id: Ib7a4041bc2730c3cd40d5a4db8ba52a8e8788256

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -4960,8 +4960,6 @@ static bool gdb_regno_cacheable(enum gdb_regno regno, bool is_write)
 	 * CSRs. */
 	switch (regno) {
 		case GDB_REGNO_DPC:
-			return true;
-
 		case GDB_REGNO_VSTART:
 		case GDB_REGNO_VXSAT:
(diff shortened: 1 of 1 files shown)


JanMatCodasip (approved):
We have already discussed and reviewed it with Marek internally.

timsifive (changes requested):
But we can still cache it for reads, right? So for DPC it should really return !is_write.

JanMatCodasip:
I believe that's right. And that's exactly what the code says on line 4980. https://github.com/riscv/riscv-openocd/pull/952/files#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR4980

It is just not immediately apparent due to the diff shown being too short :-)

timsifive (approved):
Thanks for explaining, @JanMatCodasip.

PR #953, opened 2023-11-03
target/riscv: gdb_regno_name takes an enum.

timsifive:
Otherwise it won't compile for me. Not sure why that doesn't affect the automated builds.

Change-Id: Ic66c743e1698c4c0772e5601723cb5c711b4fa5c

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -5539,7 +5539,7 @@ static void init_custom_csr_names(struct target *target)
 	}
 }
 
-const char *gdb_regno_name(struct target *target, unsigned int regno)
+const char *gdb_regno_name(struct target *target, enum gdb_regno regno)
 {
 	RISCV_INFO(info);
 


timsifive:
@AnastasiyaChernikova

JanMatCodasip (approved):
There was a mismatch between declaration of gdb_regno_name():

```
src/target/riscv/gdb_regs.h:123:const char *gdb_regno_name(struct target *target, enum gdb_regno regno);
```

and the definition:

```
src/target/riscv/riscv.c:5542:const char *gdb_regno_name(struct target *target, unsigned int regno)
```

So this fix is definitely correct. I assume that depending on compiler version and -Werror strictness the compiler may complain. 

Out of curiosity, I tried two GCC versions locally (8.5.0 and 12.2.1) but wasn't able to reproduce it.

timsifive:
FWIW, I'm running gcc (GCC) 13.2.1 20230801

PR #954, opened 2023-11-03
Merge commit '05ee88915520d1dd82da94a016a9374a1f3a8129' from upstream

timsifive:


timsifive (commented):
Lots of changes, but not much that could actually affect us.

Most changes that can be ignored are:
* Add angie debug device.
* PLD changes.
* JTAG driver changes, which seem like trivial error handling.
* Lots of target changes changing to LOG_TARGET* macros, many of which convert a warning to an error.
* Lots of ESP32 changes, none of which address the RISC-V core present on esp32s2 and esp32s3 (IIRC)
* Lots of `pld device` -> `pld create` in config files

timsifive:
As usual, there are minor checkpatch complaints that I will override.

aap-sc:
@timsifive , just a note. Once this commit is merged it would not hurt to address relevant TODO from this MR: https://github.com/riscv/riscv-openocd/pull/897/files

tcl/target/esp_common.cfg
```diff
@@ -77,12 +77,21 @@ proc create_esp_target { ARCH } {
 	set_esp_common_variables
 	create_esp_jtag
 	create_openocd_targets
+<<<<<<< HEAD
 	configure_openocd_events $ARCH
+=======
+	configure_openocd_events
+>>>>>>> 05ee88915520d1dd82da94a016a9374a1f3a8129
```
MarekVCodasip: There are several unresolved conflicts in this file. I believe the HEAD variant should be taken and the other one dropped.

MarekVCodasip:
We took a look with @JanMatCodasip. It all looks fine, except for the esp_common.cfg file mentioned above.

timsifive:
I did my best to resolve the remaining esp_common.cfg conflicts. I don't have a target to test on, however.
@erhankur can you confirm that the ESP config still works?

erhankur:
> I did my best to resolve the remaining esp_common.cfg conflicts. I don't have a target to test on, however. @erhankur can you confirm that the ESP config still works?

@timsifive LGTM. It works. Thanks

tcl/target/esp_common.cfg
JanMatCodasip: Resolved by Tim (below).

JanMatCodasip (approved):

PR #956, opened 2023-11-07
[do-not-merge] pipeline test

aap-sc:
Change-Id: I294ead2621b44f059b9eaf2f285b9457acdf1bad


aap-sc:
To whom it may concern. Looks like pipeline passed.

PR #958, opened 2023-11-09
target/riscv: Replace [sg]et_field macros with functions.

timsifive:
Compilers are good at optimizing, and with functions it's abundantly clear what all the types involved are. This change means we don't have to be super careful about the type of values because of what the macro might do to them that might cause overflow.

The only place where the return type matters is in printf-style functions, and I made get_value32() for those cases where a change was needed.

This should set the stage for simply copying the latest debug_defines.h from the debug spec build again.

See discussion at end of https://github.com/riscv/riscv-debug-spec/pull/902

Change-Id: I5fb19d0cfc1e20137832a7b344b05db215ce00e1

### DIFF
--- a/src/target/riscv/field_helpers.h
+++ b/src/target/riscv/field_helpers.h
@@ -1,0 +1,47 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+
+#ifndef FIELD_HELPERS_H
+#define FIELD_HELPERS_H
+
+#include <stdint.h>
+#include <assert.h>
+
+static inline uint64_t get_field(uint64_t reg, uint64_t mask)
+{
+	return (reg & mask) / (mask & ~(mask << 1));
+}
+
+static inline uint32_t get_field32(uint64_t reg, uint64_t mask)
+{
+	uint64_t value = get_field(reg, mask);
+	assert(value <= UINT32_MAX);
+	return value;
+}
+
+static inline uint64_t set_field(uint64_t reg, uint64_t mask, uint64_t val)
+{
+	/* Clear current value from field. */
+	reg &= ~mask;
+	uint64_t low_field_bit = mask & ~(mask << 1);
+	/* Assert if the value doesn't fit in the field. */
+	assert(((val * low_field_bit) & ~mask) == 0);
+	reg |= (val * low_field_bit) & mask;
+	return reg;
+}
+
+static inline uint32_t set_field32(uint32_t reg, uint32_t mask, uint32_t val)
+{
+	return (uint32_t)set_field(reg, mask, val);
+}
+
+static inline uint64_t field_value(uint64_t mask, uint64_t val)
+{
+	return set_field(0, mask, val);
+}
+
+static inline uint32_t field_value32(uint32_t mask, uint32_t val)
+{
+	return set_field32(0, mask, val);
+}
+
+#endif
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -29,6 +29,7 @@
 #include "asm.h"
 #include "batch.h"
 #include "debug_reg_printer.h"
+#include "field_helpers.h"
 
 static int riscv013_on_step_or_resume(struct target *target, bool step);
 static int riscv013_step_or_resume_current_hart(struct target *target,
@@ -84,9 +85,6 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
  * called directly by OpenOCD, which can't assume anything about what's
  * currently in IR. They should set IR to dbus explicitly.
  */
-
-#define get_field(reg, mask) (((reg) & (mask)) / ((mask) & ~((mask) << 1)))
-#define set_field(reg, mask, val) (((reg) & ~(mask)) | (((val) * ((mask) & ~((mask) << 1))) & (mask)))
 
 #define RISCV013_INFO(r) riscv013_info_t *r = get_info(target)
 
@@ -779,9 +777,9 @@ static int dmstatus_read_timeout(struct target *target, uint32_t *dmstatus,
 	int dmstatus_version = get_field(*dmstatus, DM_DMSTATUS_VERSION);
 	if (dmstatus_version != 2 && dmstatus_version != 3) {
 		LOG_ERROR("OpenOCD only supports Debug Module version 2 (0.13) and 3 (1.0), not "
-				"%d (dmstatus=0x%x). This error might be caused by a JTAG "
+				"%" PRId32 " (dmstatus=0x%" PRIx32 "). This error might be caused by a JTAG "
 				"signal issue. Try reducing the JTAG clock speed.",
-				get_field(*dmstatus, DM_DMSTATUS_VERSION), *dmstatus);
+				get_field32(*dmstatus, DM_DMSTATUS_VERSION), *dmstatus);
 	} else if (authenticated && !get_field(*dmstatus, DM_DMSTATUS_AUTHENTICATED)) {
 		LOG_ERROR("Debugger is not authenticated to target Debug Module. "
 				"(dmstatus=0x%x). Use `riscv authdata_read` and "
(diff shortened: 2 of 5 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1859,15 +1857,15 @@ static int examine(struct target *target)
...
-	LOG_TARGET_DEBUG(target, "  dmireset=%d", get_field(dtmcontrol, DTM_DTMCS_DMIRESET));
-	LOG_TARGET_DEBUG(target, "  idle=%d", get_field(dtmcontrol, DTM_DTMCS_IDLE));
-	LOG_TARGET_DEBUG(target, "  dmistat=%d", get_field(dtmcontrol, DTM_DTMCS_DMISTAT));
-	LOG_TARGET_DEBUG(target, "  abits=%d", get_field(dtmcontrol, DTM_DTMCS_ABITS));
-	LOG_TARGET_DEBUG(target, "  version=%d", get_field(dtmcontrol, DTM_DTMCS_VERSION));
+	LOG_TARGET_DEBUG(target, "  dmireset=%d", get_field32(dtmcontrol, DTM_DTMCS_DMIRESET));
+	LOG_TARGET_DEBUG(target, "  idle=%d", get_field32(dtmcontrol, DTM_DTMCS_IDLE));
+	LOG_TARGET_DEBUG(target, "  dmistat=%d", get_field32(dtmcontrol, DTM_DTMCS_DMISTAT));
+	LOG_TARGET_DEBUG(target, "  abits=%d", get_field32(dtmcontrol, DTM_DTMCS_ABITS));
+	LOG_TARGET_DEBUG(target, "  version=%d", get_field32(dtmcontrol, DTM_DTMCS_VERSION));
```
en-sc: I would suggest using `log_debug_reg(target, DTM_DTMCS_ORDINAL, dtmcontrol)` instead.

src/target/riscv/riscv-013.c
```diff
@@ -1859,15 +1857,15 @@ static int examine(struct target *target)
...
 
 	if (get_field(dtmcontrol, DTM_DTMCS_VERSION) != 1) {
 		LOG_TARGET_ERROR(target, "Unsupported DTM version %d. (dtmcontrol=0x%x)",
```
```suggestion
		LOG_TARGET_ERROR(target, "Unsupported DTM version %" PRIu32 ". (dtmcontrol=0x%" PRIx32 ")",
```

src/target/riscv/riscv-013.c
```diff
@@ -781,7 +779,7 @@ static int dmstatus_read_timeout(struct target *target, uint32_t *dmstatus,
 		LOG_ERROR("OpenOCD only supports Debug Module version 2 (0.13) and 3 (1.0), not "
 				"%d (dmstatus=0x%x). This error might be caused by a JTAG "
```
```suggestion
				"%" PRIu32 " (dmstatus=0x%" PRIx32 "). This error might be caused by a JTAG "
```

src/target/riscv/field_helpers.h
```diff
@@ -0,0 +1,34 @@
...
+	uint64_t value = get_field(reg, mask);
+	assert(value <= UINT32_MAX);
+	return value;
+}
+static inline uint64_t set_field(uint64_t reg, uint64_t mask, uint64_t val)
+{
+	/* Clear current value from field. */
+	reg &= ~mask;
+	uint64_t low_field_bit = mask & ~(mask << 1);
+	reg |= (val * low_field_bit) & mask;
```
en-sc, line 28:
```suggestion
        assert(!((val * low_field_bit) & ~mask));
	reg |= (val * low_field_bit) & mask;
```

src/target/riscv/field_helpers.h
```diff
@@ -0,0 +1,34 @@
...
+	return (reg & mask) / (mask & ~(mask << 1));
+}
+
+static inline uint32_t get_field32(uint64_t reg, uint64_t mask)
+{
+	uint64_t value = get_field(reg, mask);
+	assert(value <= UINT32_MAX);
+	return value;
+}
+static inline uint64_t set_field(uint64_t reg, uint64_t mask, uint64_t val)
```
JanMatCodasip, line 21: > Tim: [...] The only place where the return type matters is in printf-style functions, and I made get_value32() for those cases where a change was needed.

Actually, compilers or static analysis tools may complain about "implicit narrowing" when the 64-bit return value is assigned to a 32-bit value. 

For example clang-tidy has rules for it: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/narrowing-conversions.html

GCC has the`-Wconversion` option for it:
```
main.c: In function ‘main’:
main.c:6:24: warning: conversion from ‘uint64_t’ {aka ‘long unsigned int’} to ‘uint32_t’ {aka ‘unsigned int’} may change value [-Wconversion]
  uint32_t second_val = val;
```

For that reason I would recommend to create also the `set_field32()` variant. To have it ready for use in case any of those warnings pop up.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/field_helpers.h
timsifive: Looks like you want to check that the value fits in the mask. Is that right?
Done.

src/target/riscv/field_helpers.h
timsifive: I just tried a build with clang and I got no warnings. But I've added a helper for it.

src/target/riscv/field_helpers.h
en-sc: > Looks like you want to check that the value fits in the mask. Is that right?

Yes. Thanks!

en-sc (approved):

src/target/riscv/field_helpers.h
JanMatCodasip: Thanks.

src/target/riscv/field_helpers.h
```diff
@@ -0,0 +1,42 @@
...
+static inline uint32_t set_field32(uint32_t reg, uint32_t mask, uint32_t val)
+{
+	return (uint32_t)set_field(reg, mask, val);
+}
+
+static inline uint64_t field_value(uint64_t mask, uint64_t val)
+{
+	return set_field(0, mask, val);
+}
+
```
JanMatCodasip, line 41: I would recommend to add also the `field_value32()` variant. For the same reasons as `set_field32()`, as already discussed above. 

```suggestion
static inline uint64_t field_value(uint64_t mask, uint64_t val)
{
	return set_field(0, mask, val);
}

static inline uint32_t field_value32(uint32_t mask, uint32_t val)
{
	return set_field32(0, mask, val);
}
```

JanMatCodasip (dismissed):
I checked this visually and it looks fine to me. I only have one last minor suggestion.

src/target/riscv/field_helpers.h
timsifive: Done.

src/target/riscv/riscv-013.c
aap-sc: [nitpicking] we still have "%d" /  "%x" as part of this commit, while the return type is uint32_t. Probably would not hurt to replace with PRIu32 / PRIx32. This is the price we have to pay to use fixed-width types :(.

JanMatCodasip (approved):
Looks good to me. 

I agree that the occurrences of `%d` (and similar) should be replaced by `PRId32` (and similar). I don't have preference whether to do it in this merge request or in a following one.

timsifive:
Let's do it separate. More, smaller PRs are almost always preferable. Especially if the changes are individually helpful.

Unless the build as-is causes warnings on some compiler that I'm not using?

aap-sc:
> Unless the build as-is causes warnings on some compiler that I'm not using?

No, at least not in the environment we have. The comment was mostly a nitpick - it can surely be addressed later

aap-sc (approved):

PR #959, opened 2023-11-10
target/riscv: improve error handling in `write_memory_progbuf()`

en-sc:
The goal of this commit is to provide more robust error handling in `write_memory_progbuf()`. This is achieved by rewriting it in a fashion similar to `read_memory_progbuf()`.

The motivation is: some instability in `load_image` was encountered. No stable reproduction could be obtained, so the root cause was not determined. Therefore, it was decided to clean-up the code, that may be implicated in such failures.

Examples of unhanded errors in the code prior to this commit:
* Most of `dmi_write()` return values are discarded.
* If `dm_read()` on `abstractcs` failed (line 4546), `abstractauto` was not cleared.

Furthermore, the structure of the code was quite complicated, which made it hard to analyze and reason whether or not all possible failures are handled properly.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -230,3 +230,13 @@ size_t riscv_batch_available_scans(struct riscv_batch *batch)
 {
 	return batch->allocated_scans - batch->used_scans - 4;
 }
+
+bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch)
+{
+	if (!batch->used_scans)
+		return false;
+	assert(batch->last_scan == RISCV_SCAN_TYPE_NOP);
+	const struct scan_field *field = batch->fields + batch->used_scans - 1;
+	const uint64_t in = buf_get_u64(field->in_value, 0, field->num_bits);
+	return get_field(in, DTM_DMI_OP) == DTM_DMI_OP_BUSY;
+}
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3658,7 +3658,7 @@ struct memory_access_info {
  * dm_data[0:1] contains mem[address + (index_on_target - 2) * increment]
  */
 static int read_memory_progbuf_inner_on_ac_busy(struct target *target,
-		uint32_t start_index, uint32_t elements_to_read, uint32_t *elements_read,
+		uint32_t start_index, uint32_t *elements_read,
 		struct memory_access_info access)
 {
 	increase_ac_busy_delay(target);
@@ -3811,7 +3811,7 @@ static int read_memory_progbuf_inner_run_and_process_batch(struct target *target
 	case CMDERR_BUSY:
 		LOG_TARGET_DEBUG(target, "memory read resulted in busy response");
 		if (read_memory_progbuf_inner_on_ac_busy(target, start_index,
-					elements_to_read, &elements_to_extract_from_batch, access)
+					&elements_to_extract_from_batch, access)
 				!= ERROR_OK)
 			return ERROR_FAIL;
 		break;
@@ -4394,182 +4394,277 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
 	return ERROR_OK;
 }
 
+/**
+ * This function is used to start the memory-writing pipeline.
+ * As part of the process, the function writes the first item and waits for completion,
+ * so forward progress is ensured.
+ * The pipeline looks like this:
+ * debugger -> dm_data[0:1] -> s1 -> memory
+ * Prior to calling it, the program buffer should contain the appropriate
+ * program.
+ * This function sets DM_ABSTRACTAUTO_AUTOEXECDATA to trigger second stage of the
+ * pipeline (dm_data[0:1] -> s1) whenever dm_data is written.
+ */
+static int write_memory_progbuf_startup(struct target *target, target_addr_t *address_p,
+		const uint8_t *buffer, uint32_t size)
+{
+	/* TODO: There is potential to gain some performance if the operations below are
+	 * executed inside the first DMI batch (not separately). */
+	if (register_write_direct(target, GDB_REGNO_S0, *address_p) != ERROR_OK)
+		return ERROR_FAIL;
+
+	/* Write the first item to data0 [, data1] */
+	assert(size <= 8);
+	const uint64_t value = buf_get_u64(buffer, 0, 8 * size);
+	if (write_abstract_arg(target, /*index*/ 0, value, size > 4 ? 64 : 32)
+			!= ERROR_OK)
+		return ERROR_FAIL;
+
+	/* Write and execute command that moves the value from data0 [, data1]
+	 * into S1 and executes program buffer. */
+	uint32_t command = access_register_command(target,
+			GDB_REGNO_S1, riscv_xlen(target),
+			AC_ACCESS_REGISTER_POSTEXEC |
+			AC_ACCESS_REGISTER_TRANSFER |
+			AC_ACCESS_REGISTER_WRITE);
+	if (execute_abstract_command(target, command) != ERROR_OK)
+		return ERROR_FAIL;
+
+	log_memory_access64(*address_p, value, size, /*is_read*/ false);
+
+	/* The execution of the command succeeded, which means:
+	 * - write of the first item to memory succeeded
+	 * - address on the target (S0) was incremented
+	 */
+	*address_p += size;
+
+	/* TODO: Setting abstractauto.autoexecdata is not necessary for a write
+	 * of one element. */
+	return dm_write(target, DM_ABSTRACTAUTO,
+			1 << DM_ABSTRACTAUTO_AUTOEXECDATA_OFFSET);
+}
+
+/**
+ * This function reverts the changes made by `write_memory_progbuf_startup()`
+ */
+static int write_memory_progbuf_teardown(struct target *target)
+{
+	return dm_write(target, DM_ABSTRACTAUTO, 0);
+}
+
+/**
+ * This function attempts to restore the pipeline after a busy on abstract
+ * access or a DMI busy by reading the content of s0 -- the address of the
+ * failed write.
+ */
+static int write_memory_progbuf_handle_busy(struct target *target,
+		target_addr_t *address_p, uint32_t size, const uint8_t *buffer)
+{
+	riscv013_clear_abstract_error(target);
+	increase_ac_busy_delay(target);
+
+	if (write_memory_progbuf_teardown(target) != ERROR_OK)
+		return ERROR_FAIL;
+
+	target_addr_t address_on_target;
+	if (register_read_direct(target, &address_on_target, GDB_REGNO_S0) != ERROR_OK)
+		return ERROR_FAIL;
+	const uint8_t * const curr_buff = buffer + (address_on_target - *address_p);
+	LOG_TARGET_DEBUG(target, "Restarting from 0x%" TARGET_PRIxADDR, *address_p);
+	*address_p = address_on_target;
+	/* This restores the pipeline and ensures one item gets reliably written */
+	return write_memory_progbuf_startup(target, address_p, curr_buff, size);
+}
+
+/**
+ * This function fills the batch with DMI writes (but does not execute the batch).
+ * It returns the next address -- the address that will be the start of the next batch.
+ */
+static target_addr_t write_memory_progbuf_fill_batch(struct riscv_batch *batch,
+		target_addr_t start_address, target_addr_t end_address, uint32_t size,
+		const uint8_t *buffer)
+{
+	assert(size <= 8);
+	const unsigned int writes_per_element = size > 4 ? 2 : 1;
+	const size_t batch_capacity = riscv_batch_available_scans(batch) / writes_per_element;
+	/* This is safe even for the edge case when writing at the very top of
+	 * the 64-bit address space (in which case end_address overflows to 0).
+	 */
+	const target_addr_t batch_end_address = start_address +
+		MIN((target_addr_t)batch_capacity * size,
+				end_address - start_address);
+	for (target_addr_t address = start_address; address != batch_end_address;
+			address += size, buffer += size) {
+		assert(size <= 8);
+		const uint64_t value = buf_get_u64(buffer, 0, 8 * size);
+		log_memory_access64(address, value, size, /*is_read*/ false);
+		if (writes_per_element == 2)
+			riscv_batch_add_dm_write(batch, DM_DATA1,
+					(uint32_t)(value >> 32), false);
+		riscv_batch_add_dm_write(batch, DM_DATA0, (uint32_t)value, false);
+	}
+	return batch_end_address;
+}
+
+/**
+ * This function runs the batch of writes and updates address_p with the
+ * address of the next write.
+ */
+static int write_memory_progbuf_run_batch(struct target *target, struct riscv_batch *batch,
+		target_addr_t *address_p, target_addr_t end_address, uint32_t size,
+		const uint8_t *buffer)
+{
+	if (batch_run(target, batch) != ERROR_OK)
+		return ERROR_FAIL;
+
+	/* Note that if the scan resulted in a Busy DMI response, it
+	 * is this call to wait_for_idle() that will cause the dmi_busy_delay
+	 * to be incremented if necessary. */
+	uint32_t abstractcs;
+
+	if (wait_for_idle(target, &abstractcs) != ERROR_OK)
+		return ERROR_FAIL;
+	RISCV013_INFO(info);
+	info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
+
+	const bool dmi_busy_encountered = riscv_batch_dmi_busy_encountered(batch);
+	if (info->cmderr == CMDERR_NONE && !dmi_busy_encountered) {
+		LOG_TARGET_DEBUG(target, "Successfully written memory block M[0x%" TARGET_PRIxADDR
+				".. 0x%" TARGET_PRIxADDR ")", *address_p, end_address);
+		*address_p = end_address;
+		return ERROR_OK;
+	} else if (info->cmderr == CMDERR_BUSY || dmi_busy_encountered) {
+		if (info->cmderr == CMDERR_BUSY)
+			LOG_TARGET_DEBUG(target, "Encountered abstract command busy response while writing block M[0x%"
+					TARGET_PRIxADDR ".. 0x%" TARGET_PRIxADDR ")", *address_p, end_address);
+		if (dmi_busy_encountered)
+			LOG_TARGET_DEBUG(target, "Encountered DMI busy response while writing block M[0x%"
+					TARGET_PRIxADDR ".. 0x%" TARGET_PRIxADDR ")", *address_p, end_address);
+		/* TODO: If dmi busy is encountered, the address of the last
+		 * successful write can be deduced by analysing the batch.
+		 */
+		return write_memory_progbuf_handle_busy(target, address_p, size, buffer);
+	}
+	LOG_TARGET_ERROR(target, "Error when writing memory, abstractcs=0x%" PRIx32,
+			abstractcs);
+	riscv013_clear_abstract_error(target);
+	return ERROR_FAIL;
+}
+
+static int write_memory_progbuf_try_to_write(struct target *target,
+		target_addr_t *address_p, target_addr_t end_address, uint32_t size,
+		const uint8_t *buffer)
+{
+	RISCV013_INFO(info);
+	struct riscv_batch * const batch = riscv_batch_alloc(target, RISCV_BATCH_ALLOC_SIZE,
+			info->dmi_busy_delay + info->ac_busy_delay);
+	if (!batch)
+		return ERROR_FAIL;
+
+	const target_addr_t batch_end_addr = write_memory_progbuf_fill_batch(batch,
+			*address_p, end_address, size, buffer);
+
+	int result = write_memory_progbuf_run_batch(target, batch, address_p,
+			batch_end_addr, size, buffer);
+	riscv_batch_free(batch);
+	return result;
+}
+
+static int riscv_program_store_mprv(struct riscv_program *p, enum gdb_regno d,
+		enum gdb_regno b, int offset, unsigned int size, bool mprven)
+{
+	if (mprven && riscv_program_csrrsi(p, GDB_REGNO_ZERO, CSR_DCSR_MPRVEN,
+				GDB_REGNO_DCSR) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (riscv_program_store(p, d, b, offset, size) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (mprven && riscv_program_csrrci(p, GDB_REGNO_ZERO, CSR_DCSR_MPRVEN,
+				GDB_REGNO_DCSR) != ERROR_OK)
+		return ERROR_FAIL;
+
+	return ERROR_OK;
+}
+
+static int write_memory_progbuf_fill_progbuf(struct target *target,
+		uint32_t size, bool mprven)
+{
+	if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
+		return ERROR_FAIL;
+	if (riscv_save_register(target, GDB_REGNO_S1) != ERROR_OK)
+		return ERROR_FAIL;
+
+	struct riscv_program program;
+
+	riscv_program_init(&program, target);
+	if (riscv_program_store_mprv(&program, GDB_REGNO_S1, GDB_REGNO_S0, 0, size,
+				mprven) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (riscv_program_addi(&program, GDB_REGNO_S0, GDB_REGNO_S0, size) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (riscv_program_ebreak(&program) != ERROR_OK)
+		return ERROR_FAIL;
+
+	return riscv_program_write(&program);
+}
+
+static int write_memory_progbuf_inner(struct target *target, target_addr_t start_addr,
+		uint32_t size, uint32_t count, const uint8_t *buffer, bool mprven)
+{
+	if (write_memory_progbuf_fill_progbuf(target, size,
+				mprven) != ERROR_OK)
+		return ERROR_FAIL;
+
+	target_addr_t addr_on_target = start_addr;
+	if (write_memory_progbuf_startup(target, &addr_on_target, buffer, size) != ERROR_OK)
+		return ERROR_FAIL;
+
+	const target_addr_t end_addr = start_addr + (target_addr_t)size * count;
+
+	for (target_addr_t next_addr_on_target = addr_on_target; addr_on_target != end_addr;
+			addr_on_target = next_addr_on_target) {
+		const uint8_t * const curr_buff = buffer + (addr_on_target - start_addr);
+		if (write_memory_progbuf_try_to_write(target, &next_addr_on_target,
+					end_addr, size, curr_buff) != ERROR_OK) {
+			write_memory_progbuf_teardown(target);
+			return ERROR_FAIL;
+		}
+		/* write_memory_progbuf_try_to_write() ensures that at least one item
+		 * gets successfully written even when busy condition is encountered.
+		 * These assertions shuld hold when next_address_on_target overflows. */
+		assert(next_addr_on_target - addr_on_target > 0);
+		assert(next_addr_on_target - start_addr <= (target_addr_t)size * count);
+	}
+
+	return write_memory_progbuf_teardown(target);
+}
+
 static int write_memory_progbuf(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, const uint8_t *buffer)
 {
-	RISCV013_INFO(info);
-
 	if (riscv_xlen(target) < size * 8) {
-		LOG_TARGET_ERROR(target, "XLEN (%d) is too short for %d-bit memory write.",
+		LOG_TARGET_ERROR(target, "XLEN (%u) is too short for %" PRIu32 "-bit memory write.",
 				riscv_xlen(target), size * 8);
 		return ERROR_FAIL;
 	}
 
-	LOG_TARGET_DEBUG(target, "writing %d words of %d bytes to 0x%08lx", count, size, (long)address);
+	LOG_TARGET_DEBUG(target, "writing %" PRIu32 " words of %" PRIu32
+			" bytes to 0x%" TARGET_PRIxADDR, count, size, address);
 
-	select_dmi(target);
+	if (dm013_select_target(target) != ERROR_OK)
+		return ERROR_FAIL;
 
 	uint64_t mstatus = 0;
 	uint64_t mstatus_old = 0;
 	if (modify_privilege(target, &mstatus, &mstatus_old) != ERROR_OK)
 		return ERROR_FAIL;
 
-	/* s0 holds the next address to write to
-	 * s1 holds the next data value to write
-	 */
+	const bool mprven = riscv_enable_virtual && get_field(mstatus, MSTATUS_MPRV);
 
-	int result = ERROR_OK;
-	if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
-		return ERROR_FAIL;
-	if (riscv_save_register(target, GDB_REGNO_S1) != ERROR_OK)
-		return ERROR_FAIL;
-
-	/* Write the program (store, increment) */
-	struct riscv_program program;
-	riscv_program_init(&program, target);
-	if (riscv_enable_virtual && has_sufficient_progbuf(target, 5) && get_field(mstatus, MSTATUS_MPRV))
-		riscv_program_csrrsi(&program, GDB_REGNO_ZERO, CSR_DCSR_MPRVEN, GDB_REGNO_DCSR);
-
-	switch (size) {
-		case 1:
-			riscv_program_sbr(&program, GDB_REGNO_S1, GDB_REGNO_S0, 0);
-			break;
-		case 2:
-			riscv_program_shr(&program, GDB_REGNO_S1, GDB_REGNO_S0, 0);
-			break;
-		case 4:
-			riscv_program_swr(&program, GDB_REGNO_S1, GDB_REGNO_S0, 0);
-			break;
-		case 8:
-			riscv_program_sdr(&program, GDB_REGNO_S1, GDB_REGNO_S0, 0);
-			break;
-		default:
-			LOG_TARGET_ERROR(target, "Unsupported size: %d", size);
-			result = ERROR_FAIL;
-			goto error;
-	}
-
-	if (riscv_enable_virtual && has_sufficient_progbuf(target, 5) && get_field(mstatus, MSTATUS_MPRV))
-		riscv_program_csrrci(&program, GDB_REGNO_ZERO,  CSR_DCSR_MPRVEN, GDB_REGNO_DCSR);
-	riscv_program_addi(&program, GDB_REGNO_S0, GDB_REGNO_S0, size);
-
-	result = riscv_program_ebreak(&program);
-	if (result != ERROR_OK)
-		goto error;
-	riscv_program_write(&program);
-
-	riscv_addr_t cur_addr = address;
-	riscv_addr_t distance = (riscv_addr_t)count * size;
-	bool setup_needed = true;
-	LOG_TARGET_DEBUG(target, "Writing until final address 0x%016" PRIx64, cur_addr + distance);
-	while (cur_addr - address < distance) {
-		LOG_TARGET_DEBUG(target, "Transferring burst starting at address 0x%016" PRIx64,
-				cur_addr);
-
-		struct riscv_batch *batch = riscv_batch_alloc(
-				target,
-				RISCV_BATCH_ALLOC_SIZE,
-				info->dmi_busy_delay + info->ac_busy_delay);
-		if (!batch)
-			goto error;
-
-		/* To write another word, we put it in S1 and execute the program. */
-		for (riscv_addr_t offset = cur_addr - address; offset < distance; offset += size) {
-			const uint8_t *t_buffer = buffer + offset;
-
-			uint64_t value = buf_get_u64(t_buffer, 0, 8 * size);
-
-			log_memory_access64(cur_addr, value, size, false);
-			cur_addr += size;
-
-			if (setup_needed) {
-				result = register_write_direct(target, GDB_REGNO_S0,
-						address + offset);
-				if (result != ERROR_OK) {
-					riscv_batch_free(batch);
-					goto error;
-				}
-
-				/* Write value. */
-				if (size > 4)
-					dm_write(target, DM_DATA1, value >> 32);
-				dm_write(target, DM_DATA0, value);
-
-				/* Write and execute command that moves value into S1 and
-				 * executes program buffer. */
-				uint32_t command = access_register_command(target,
-						GDB_REGNO_S1, riscv_xlen(target),
-						AC_ACCESS_REGISTER_POSTEXEC |
-						AC_ACCESS_REGISTER_TRANSFER |
-						AC_ACCESS_REGISTER_WRITE);
-				result = execute_abstract_command(target, command);
-				if (result != ERROR_OK) {
-					riscv_batch_free(batch);
-					goto error;
-				}
-
-				/* Turn on autoexec */
-				dm_write(target, DM_ABSTRACTAUTO,
-						1 << DM_ABSTRACTAUTO_AUTOEXECDATA_OFFSET);
-
-				setup_needed = false;
-			} else {
-				/* Note that data1 "might not be preserved after
-				 * an abstract command is executed," so this
-				 * can't be optimized by only writing data1 when
-				 * it has changed. */
-				if (size > 4)
-					riscv_batch_add_dm_write(batch, DM_DATA1, value >> 32, false);
-				riscv_batch_add_dm_write(batch, DM_DATA0, value, false);
-				if (riscv_batch_full(batch))
-					break;
-			}
-		}
-
-		result = batch_run(target, batch);
-		riscv_batch_free(batch);
-		if (result != ERROR_OK)
-			goto error;
-
-		/* Note that if the scan resulted in a Busy DMI response, it
-		 * is this read to abstractcs that will cause the dmi_busy_delay
-		 * to be incremented if necessary. */
-
-		uint32_t abstractcs;
-		bool dmi_busy_encountered;
-		result = dm_op(target, &abstractcs, &dmi_busy_encountered,
-				DMI_OP_READ, DM_ABSTRACTCS, 0, false, true);
-		if (result != ERROR_OK)
-			goto error;
-		while (get_field(abstractcs, DM_ABSTRACTCS_BUSY))
-			if (dm_read(target, &abstractcs, DM_ABSTRACTCS) != ERROR_OK)
-				return ERROR_FAIL;
-		info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
-		if (info->cmderr == CMDERR_NONE && !dmi_busy_encountered) {
-			LOG_TARGET_DEBUG(target, "Successful (partial?) memory write");
-		} else if (info->cmderr == CMDERR_BUSY || dmi_busy_encountered) {
-			if (info->cmderr == CMDERR_BUSY)
-				LOG_TARGET_DEBUG(target, "Memory write resulted in abstract command busy response.");
-			else if (dmi_busy_encountered)
-				LOG_TARGET_DEBUG(target, "Memory write resulted in DMI busy response.");
-			riscv013_clear_abstract_error(target);
-			increase_ac_busy_delay(target);
-
-			dm_write(target, DM_ABSTRACTAUTO, 0);
-			result = register_read_direct(target, &cur_addr, GDB_REGNO_S0);
-			if (result != ERROR_OK)
-				goto error;
-			setup_needed = true;
-		} else {
-			LOG_TARGET_ERROR(target, "Error when writing memory, abstractcs=0x%08lx", (long)abstractcs);
-			riscv013_clear_abstract_error(target);
-			result = ERROR_FAIL;
-			goto error;
-		}
-	}
-
-error:
-	dm_write(target, DM_ABSTRACTAUTO, 0);
+	int result = write_memory_progbuf_inner(target, address, size, count, buffer, mprven);
 
 	/* Restore MSTATUS */
 	if (mstatus != mstatus_old)
--- a/src/target/riscv/batch.h
+++ b/src/target/riscv/batch.h
@@ -75,4 +75,7 @@ void riscv_batch_add_nop(struct riscv_batch *batch);
 /* Returns the number of available scans. */
 size_t riscv_batch_available_scans(struct riscv_batch *batch);
 
+/* Return true iff the last scan in the batch returned DMI_OP_BUSY. */
+bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch);
+
 #endif
--- a/src/target/riscv/program.c
+++ b/src/target/riscv/program.c
@@ -99,6 +99,23 @@ int riscv_program_sbr(struct riscv_program *p, enum gdb_regno d, enum gdb_regno 
 	return riscv_program_insert(p, sb(d, b, offset));
 }
 
+int riscv_program_store(struct riscv_program *p, enum gdb_regno d, enum gdb_regno b, int offset,
+		unsigned int size)
+{
+	switch (size) {
+	case 1:
+		return riscv_program_sbr(p, d, b, offset);
+	case 2:
+		return riscv_program_shr(p, d, b, offset);
+	case 4:
+		return riscv_program_swr(p, d, b, offset);
+	case 8:
+		return riscv_program_sdr(p, d, b, offset);
+	}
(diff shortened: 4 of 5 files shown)


timsifive (changes requested):
Can you update the description so it says how error handling is improved? What case works now that didn't work before?

timsifive:
I see you've added some comments to the code. I appreciate those, but I'd still like to see a description on the PR about what problem this is solving, and roughly how it does it. That helps me as a reviewer right now, and it also helps anyone later looking at this change in the history.

en-sc:
> Can you update the description so it says how error handling is improved? What case works now that didn't work before?

Done.

src/target/riscv/batch.c
```diff
@@ -230,3 +230,13 @@ size_t riscv_batch_available_scans(struct riscv_batch *batch)
 {
 	return batch->allocated_scans - batch->used_scans - 4;
 }
+
+bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch)
```
timsifive, line 234:
```suggestion
/* Return true iff the last scan in the batch returned DMI_OP_BUSY. */
bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch)
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
 	return ERROR_OK;
 }
 
-static int write_memory_progbuf(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, const uint8_t *buffer)
+static uint64_t get_from_buffer_and_log_write(target_addr_t address,
```
timsifive: How about `log_memory_access64_buf`? Probably best to add is_read here so the function is more general and matches `log_memory_access64` more closely.

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+	return value;
+}
 
-	if (riscv_xlen(target) < size * 8) {
-		LOG_TARGET_ERROR(target, "XLEN (%d) is too short for %d-bit memory write.",
-				riscv_xlen(target), size * 8);
-		return ERROR_FAIL;
-	}
+static int write_word_to_dm_data_regs(struct target *target, target_addr_t address,
+		const uint8_t *buffer, uint32_t size)
```
timsifive: `write_abstract_arg_buf`?

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+ * program.
+ * This function sets DM_ABSTRACTAUTO_AUTOEXECDATA to trigger second stage of the
+ * pipeline (dm_data[0:1] -> s1) whenever dm_data is written.
+ */
+static int write_memory_progbuf_inner_startup(struct target *target, target_addr_t *address_p,
+		const uint8_t *buffer, uint32_t size)
+{
+	/* FIXME: The same result can be achieved by starting the batch with
+	 * corresponding DMI operations.
+	 */
```
timsifive: Why is that a FIXME?

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+			GDB_REGNO_S1, riscv_xlen(target),
+			AC_ACCESS_REGISTER_POSTEXEC |
+			AC_ACCESS_REGISTER_TRANSFER |
+			AC_ACCESS_REGISTER_WRITE);
+	if (execute_abstract_command(target, command) != ERROR_OK)
 		return ERROR_FAIL;
 
-	/* s0 holds the next address to write to
-	 * s1 holds the next data value to write
+	/* Address on the target was incremented during the execution.
```
```suggestion
	/* If the write succeeded, address on the target was incremented during the execution.
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+	return dm_write(target, DM_ABSTRACTAUTO,
+			1 << DM_ABSTRACTAUTO_AUTOEXECDATA_OFFSET);
+}
+
+/**
+ * This function attempts to restore the pipeline after a busy on abstract
+ * access or a DMI busy by reading the content of s0 -- the address of the
+ * failed write.
+ */
+static int write_memory_progbuf_inner_on_busy(struct target *target,
```
```suggestion
static int write_memory_progbuf_on_busy(struct target *target,
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+}
 
-	/* Write the program (store, increment) */
-	struct riscv_program program;
-	riscv_program_init(&program, target);
-	if (riscv_enable_virtual && has_sufficient_progbuf(target, 5) && get_field(mstatus, MSTATUS_MPRV))
-		riscv_program_csrrsi(&program, GDB_REGNO_ZERO, CSR_DCSR_MPRVEN, GDB_REGNO_DCSR);
+/**
+ * This function fills the batch with writes and returns the address on target
+ * of the next ellement, if the batch is executed fully.
```
```suggestion
 * of the next element, if the batch is executed fully.
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+ */
+static target_addr_t write_memory_progbuf_inner_fill_batch(struct riscv_batch *batch,
```
```suggestion
static target_addr_t write_memory_progbuf_fill_batch(struct riscv_batch *batch,
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+/**
+ * This function is used to start the memory-writing pipeline.
+ * The pipeline looks like this:
+ * debugger -> dm_data[0:1] -> s1 -> memory
+ * Prior to calling it, the program buffer should contain the appropriate
+ * program.
+ * This function sets DM_ABSTRACTAUTO_AUTOEXECDATA to trigger second stage of the
+ * pipeline (dm_data[0:1] -> s1) whenever dm_data is written.
+ */
+static int write_memory_progbuf_inner_startup(struct target *target, target_addr_t *address_p,
```
```suggestion
static int write_memory_progbuf_startup(struct target *target, target_addr_t *address_p,
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
-	dm_write(target, DM_ABSTRACTAUTO, 0);
+	return dm_write(target, DM_ABSTRACTAUTO, 0);
+}
+
+static int write_memory_progbuf(struct target *target, target_addr_t address,
+		uint32_t size, uint32_t count, const uint8_t *buffer)
+{
+	if (riscv_xlen(target) < size * 8) {
+		LOG_TARGET_ERROR(target, "XLEN (%u) is too short for %" PRIu64 "-bit memory write.",
+				riscv_xlen(target), (uint64_t)size * 8);
```
JanMatCodasip: Detail: I believe it is not necessary to cast the `size` to uint64_t.

```suggestion
	assert(size == 1 || size == 2 || size == 4 || size == 8);
	if (riscv_xlen(target) < size * 8) {
		LOG_TARGET_ERROR(target, "XLEN (%u) is too short for %" PRIu32 "-bit memory write.",
				riscv_xlen(target), size * 8);
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+		return ERROR_FAIL;
+
+	for (target_addr_t end_address_on_target = address + (target_addr_t)size * count,
+			next_address_on_target = address_on_target;
+			address_on_target != end_address_on_target;
+			address_on_target = next_address_on_target) {
+		const uint8_t * const curr_buff = buffer + (address_on_target - address);
+		if (write_memory_progbuf_inner_try_to_write(target, &next_address_on_target,
+					end_address_on_target, size, curr_buff) != ERROR_OK) {
 			dm_write(target, DM_ABSTRACTAUTO, 0);
```
JanMatCodasip: As a reader of the code, I find it a bit strange that this function did not set the `abstractauto` register but is clearing it.

Preferably:
- 1) Either the same function that sets some state also clears it after itself.
- 2) Or there would be a pair of functions - one to sets things up, the other one to clean it up afterwards.

Here I suggest the option 2) - create a function called `write_memory_progbuf_inner_teardown()` (or a similar name) that cleans up after its counterpart `write_memory_progbuf_inner_startup()`.

src/target/riscv/riscv-013.c
JanMatCodasip: Do you mean that here is some room to optimize the code for performance by using a batch?

```suggestion
	/* TODO: There is potential to gain some performance if the operations below are 
	 * executed as a single DMI batch.
	 */
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+ * This function fills the batch with writes and returns the address on target
+ * of the next ellement, if the batch is executed fully.
+ */
+static target_addr_t write_memory_progbuf_inner_fill_batch(struct riscv_batch *batch,
+		target_addr_t start_address, target_addr_t end_address, uint32_t size,
+		const uint8_t *buffer)
+{
+	assert(size <= 8);
+	const uint32_t reads_per_element = size > 4 ? 2 : 1;
+	const uint32_t batch_capacity = riscv_batch_available_scans(batch) / reads_per_element;
```
JanMatCodasip: There seems to be a typo: `reads_per_element` --> `writes_per_element`

Here I believe you can stick to generic data types (`unsigned int`). No need to use fixed-width types.

```suggestion
	const unsigned int writes_per_element = (size > 4) ? 2 : 1;
	const unsigned int batch_capacity = riscv_batch_available_scans(batch) / writes_per_element;
```

src/target/riscv/riscv-013.c
JanMatCodasip: The comment does not correspond to what the function is doing.

```suggestion
 * This function fills the batch with DMI writes (but does not execute the batch).
 * It returns the next address -- the address that will be the start of the next batch.
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+		return ERROR_OK;
+	} else if (info->cmderr == CMDERR_BUSY || dmi_busy_encountered) {
+		if (info->cmderr == CMDERR_BUSY)
+			LOG_TARGET_DEBUG(target, "Memory write resulted in abstract command busy response.");
+		if (dmi_busy_encountered)
+			LOG_TARGET_DEBUG(target, "Memory write resulted in DMI busy response.");
+		/* TODO: If dmi busy is encountered, the address of the last
+		 * successful write can be deduced by analysing the batch.
+		 */
+		return write_memory_progbuf_inner_on_busy(target, address_p, size, buffer);
```
JanMatCodasip: This can be given a more descriptive name:

```suggestion
		return write_memory_progbuf_inner_handle_busy(target, address_p, size, buffer);
```

or 

```suggestion
		return write_memory_progbuf_inner_get_failed_addr(target, address_p, size, buffer);
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+	RISCV013_INFO(info);
+	info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
 
-		struct riscv_batch *batch = riscv_batch_alloc(
-				target,
-				RISCV_BATCH_ALLOC_SIZE,
-				info->dmi_busy_delay + info->ac_busy_delay);
-		if (!batch)
-			goto error;
+	bool dmi_busy_encountered = riscv_batch_dmi_busy_encountered(batch);
```
JanMatCodasip: If you like, this can be const.

```suggestion
	const bool dmi_busy_encountered = riscv_batch_dmi_busy_encountered(batch);
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+	if (info->cmderr == CMDERR_NONE && !dmi_busy_encountered) {
+		LOG_TARGET_DEBUG(target, "Successful memory write M[0x%" TARGET_PRIxADDR
+				", 0x%" TARGET_PRIxADDR ")", *address_p, end_address);
```
JanMatCodasip: Here you can make it more clear to the log reader that a whole block has been successfully written:
- by saying it is a block
- by printing the range of addresses separated by `..` as opposed to `,`

```suggestion
		LOG_TARGET_DEBUG(target, "Successfully written memory block M[0x%" TARGET_PRIxADDR
				".. 0x%" TARGET_PRIxADDR ")", *address_p, end_address);
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
-			if (setup_needed) {
-				result = register_write_direct(target, GDB_REGNO_S0,
-						address + offset);
-				if (result != ERROR_OK) {
-					riscv_batch_free(batch);
-					goto error;
-				}
+/**
+ * This function assumes that currently, the address of the next store (value
+ * of s1) is the same as address argument.
```
JanMatCodasip: You can drop the word "currently" - the meaning stays and the sentence becomes clearer.

```suggestion
 * This function assumes that the address of the next store (value
 * of s1) is the same as address argument.
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
-			break;
-		default:
-			LOG_TARGET_ERROR(target, "Unsupported size: %d", size);
-			result = ERROR_FAIL;
-			goto error;
+	for (target_addr_t address = start_address;
+			address - start_address < distance;
+			address += size, buffer += size) {
+		uint64_t value = get_from_buffer_and_log_write(address, buffer, size);
+		if (size > 4)
```
JanMatCodasip: Really minor detail. If you like:

```suggestion
		if (writes_per_element == 2)
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+	if (info->cmderr == CMDERR_NONE && !dmi_busy_encountered) {
+		LOG_TARGET_DEBUG(target, "Successful memory write M[0x%" TARGET_PRIxADDR
+				", 0x%" TARGET_PRIxADDR ")", *address_p, end_address);
+		*address_p = end_address;
+		return ERROR_OK;
+	} else if (info->cmderr == CMDERR_BUSY || dmi_busy_encountered) {
+		if (info->cmderr == CMDERR_BUSY)
+			LOG_TARGET_DEBUG(target, "Memory write resulted in abstract command busy response.");
+		if (dmi_busy_encountered)
+			LOG_TARGET_DEBUG(target, "Memory write resulted in DMI busy response.");
```
JanMatCodasip: Optionally, if you like, you may include the memory address range also in these messages (to help with log analysis).

```
Encountered abstract command busy response while writing block M[addr_start .. addr_end]

Encountered DMI busy response while writing block M[addr_start .. addr_end]
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+
+	uint64_t mstatus = 0;
+	uint64_t mstatus_old = 0;
+	if (modify_privilege(target, &mstatus, &mstatus_old) != ERROR_OK)
+		return ERROR_FAIL;
+
+	const bool mprven = riscv_enable_virtual && get_field(mstatus, MSTATUS_MPRV);
+	/* TODO: Here it is possible to handle writes of a single element
+	 * separately, without saving the s1 register (see read_memory_progbuf()).
+	 */
```
JanMatCodasip: Is this comment up to date, please?

I don't see `s1` being saved here. The comment also mentions read_memory_progbuf(), which looks like a copy-paste typo.

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+		return ERROR_FAIL;
+
+	target_addr_t address_on_target = address;
+	if (write_memory_progbuf_inner_startup(target, &address_on_target, buffer, size) != ERROR_OK)
+		return ERROR_FAIL;
+
+	for (target_addr_t end_address_on_target = address + (target_addr_t)size * count,
+			next_address_on_target = address_on_target;
+			address_on_target != end_address_on_target;
+			address_on_target = next_address_on_target) {
```
JanMatCodasip: These statements inside the `for` loops are bit complex. I'd propose to take the variable definitions out.

Please, rename the variables so that the names are shorter and more descriptive. For example:

```
end_address_on_target --> end_addr  (and should be const)
address_on_target --> batch_start_addr
next_address_on_target  --> next_batch_start_addr
```

In fact, this is a `do - while` loop masked as `for` loop. The assignment `next_address_on_target = address_on_target` is a "hack" to make sure the first iteration is always executed, but the assigned value is never used in the algorithm, and is over-written on line 4646.

For that reason, please, change it to `do - while`:

```suggestion
	target_addr_t batch_start_addr = address;
	const target_addr_t end_addr = address + (target_addr_t)size * count;
	target_addr_t next_batch_start_addr;
	do {
		/* ... */
		
		batch_start_addr = next_batch_start_addr;
	} while (batch_start_addr != end_addr);
```

JanMatCodasip (changes requested):
I have managed to review approx. 70 % of the code, and will have to return to this to complete the review next week.

I am sending the comments I have so far.

Overall, I like this rewrite very much as it improves the code, which will make troubleshooting and future code maintenance easier.

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,299 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+}
 
-	if (riscv_enable_virtual && has_sufficient_progbuf(target, 5) && get_field(mstatus, MSTATUS_MPRV))
-		riscv_program_csrrci(&program, GDB_REGNO_ZERO,  CSR_DCSR_MPRVEN, GDB_REGNO_DCSR);
-	riscv_program_addi(&program, GDB_REGNO_S0, GDB_REGNO_S0, size);
+/**
+ * This function runs the batch of writes and updates address_p with the
+ * address of the next write.
+ */
+static int write_memory_progbuf_inner_run_batch(struct target *target, struct riscv_batch *batch,
```
```suggestion
static int write_memory_progbuf_run_batch(struct target *target, struct riscv_batch *batch,
```

timsifive (changes requested):
Adding in my partial review as well. I'll get back to this more later.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Great suggestion! Addressed.

src/target/riscv/riscv-013.c
en-sc: Addressed. The wording is much better now. Thanks!

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Addressed. Thank you!

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: It is up to date, I've just made a mistake. Thanks for the catch!

src/target/riscv/batch.c
en-sc: Wouldn't it be better to add the comment in the `batch.h` instead of `batch.c`?

src/target/riscv/riscv-013.c
en-sc: Addressed. Changed the comment as @JanMatCodasip suggested [here](https://github.com/riscv/riscv-openocd/pull/959#discussion_r1395235504).

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Addressed. Changed the comment as @JanMatCodasip suggested [here](https://github.com/riscv/riscv-openocd/pull/959#discussion_r1395247401).

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: I have shortened the names.

As for other suggestions, you are mistaken. `&next_address_on_target` is passed to `write_memory_progbuf_try_to_write()` as `address_p` -- in/out parameter. It should be initialized in order to calculate how long the batch should be in `_fill_batch()`.

Moreover, this is not a `do - while` loop, since it is possible for this loop to have zero iterations -- one element is already written in `_startup()`.

I would also like to leave `_on_target` part in the names, since these values represent current value of `s0` register.

src/target/riscv/riscv-013.c
en-sc: I would like to leave this name as-is.
This function returns a value from the buffer, so naming it `log_...` would be misleading IMHO.

Adding the `is_read` argument does not seem practical to me, since when memory is read, the buffer is written. This function, on the other hand, reads the buffer.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+ */
+static target_addr_t write_memory_progbuf_fill_batch(struct riscv_batch *batch,
+		target_addr_t start_address, target_addr_t end_address, uint32_t size,
+		const uint8_t *buffer)
+{
+	assert(size <= 8);
+	const unsigned int writes_per_element = size > 4 ? 2 : 1;
+	const size_t batch_capacity = riscv_batch_available_scans(batch) / writes_per_element;
+	const target_addr_t distance = MIN((target_addr_t)batch_capacity * size,
+			end_address - start_address);
```
JanMatCodasip: Detail: Instead of `distance`, please use a more fitting name like `num_writes` or similar.

```suggestion
	const target_addr_t num_writes = MIN((target_addr_t)batch_capacity * size,
			end_address - start_address);
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
-			riscv_program_sdr(&program, GDB_REGNO_S1, GDB_REGNO_S0, 0);
-			break;
-		default:
-			LOG_TARGET_ERROR(target, "Unsupported size: %d", size);
-			result = ERROR_FAIL;
-			goto error;
+	for (target_addr_t address = start_address;
+			address - start_address < distance;
+			address += size, buffer += size) {
+		uint64_t value = get_from_buffer_and_log_write(address, buffer, size);
```
JanMatCodasip: An optional detail:

```suggestion
		const uint64_t value = get_from_buffer_and_log_write(address, buffer, size);
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+		if (writes_per_element == 2)
+			riscv_batch_add_dm_write(batch, DM_DATA1, value >> 32, false);
+		riscv_batch_add_dm_write(batch, DM_DATA0, value, false);
```
JanMatCodasip: Please, be explicit here about the conversion from uint64_t to uint32_t -- add the cast operator.

It may not be apparent to the reader that the conversion is happening unless they remember what the data type of the 3rd argument of riscv_batch_add_dm_write() is.

Furthermore, compilers may warn about implicit narrowing conversions (for example [-Wconversion](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wconversion) in GCC).

```suggestion
			riscv_batch_add_dm_write(batch, DM_DATA1, (uint32_t)(value >> 32), false);
		riscv_batch_add_dm_write(batch, DM_DATA0, (uint32_t)value, false);
```

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+ * Prior to calling it, the program buffer should contain the appropriate
+ * program.
+ * This function sets DM_ABSTRACTAUTO_AUTOEXECDATA to trigger second stage of the
+ * pipeline (dm_data[0:1] -> s1) whenever dm_data is written.
+ */
+static int write_memory_progbuf_startup(struct target *target, target_addr_t *address_p,
+		const uint8_t *buffer, uint32_t size)
+{
+	/* TODO: There is potential to gain some performance if the operations below are
+	 * executed as a single DMI batch.
```
JanMatCodasip: @en-sc Now that I read it again, is this what you actually meant?

```suggestion
	/* TODO: There is potential to gain some performance if the operations below are
	 * executed inside the first DMI batch (not separately).
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+	 */
+	if (register_write_direct(target, GDB_REGNO_S0, *address_p) != ERROR_OK)
+		return ERROR_FAIL;
 
-	select_dmi(target);
+	if (write_abstract_arg_buf(target, *address_p, buffer, size) != ERROR_OK)
```
```suggestion
	/* Write the first item to data0 [, data1] */
	if (write_abstract_arg_buf(target, *address_p, buffer, size) != ERROR_OK)
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+		return ERROR_FAIL;
 
-	uint64_t mstatus = 0;
-	uint64_t mstatus_old = 0;
-	if (modify_privilege(target, &mstatus, &mstatus_old) != ERROR_OK)
+	/* Write and execute command that moves value into S1 and
```
```suggestion
	/* Write and execute command that moves the value from data0 [, data1] into S1 and
```

src/target/riscv/riscv-013.c
JanMatCodasip: Please, consider to not add the memory logging inside this function. It is very unclear from the function name that the logging took place inside.

I would recommend to make the log print as a separate call.

JanMatCodasip: This function write_abstract_arg_buf() is used only in a single place. I would recommend to change it like this:

Option 1:

Use the existing functions buf_get_u64() and write_abstract_arg() directly, and drop the write_abstract_arg_buf() function.

Option 2:

Keep the function write_abstract_arg_buf() but extend it by by adding the "index" attribute so that it becomes really reusable, same as the existing write_abstract_arg() function. Also move the function definition so that they are right next to each other in the code.

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
 
-static int write_memory_progbuf(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, const uint8_t *buffer)
+static uint64_t get_from_buffer_and_log_write(target_addr_t address,
+		const uint8_t *buffer, uint32_t size)
 {
-	RISCV013_INFO(info);
+	assert(size <= 8);
+	uint64_t value = buf_get_u64(buffer, 0, 8 * size);
+	log_memory_access64(address, value, size, /*is_read*/ false);
```
JanMatCodasip: As already mentioned in another comment, I recommend to not combine the reading from the buffer and making a log print about an operation into a single function.

Please consider making them two separate calls since they are logically two separate operations.

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+{
+	assert(size <= 8);
+	const uint64_t value = get_from_buffer_and_log_write(address, buffer, size);
+	return write_abstract_arg(target, /*index*/ 0, value, size > 4 ? 64 : 32);
+}
 
-	LOG_TARGET_DEBUG(target, "writing %d words of %d bytes to 0x%08lx", count, size, (long)address);
+/**
+ * This function is used to start the memory-writing pipeline.
+ * The pipeline looks like this:
```
```suggestion
 * This function is used to start the memory-writing pipeline.
 * As part of the process, the function writes the first item and waits for completion,
 * so forward progress is ensured.
 * The pipeline looks like this:
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+			AC_ACCESS_REGISTER_POSTEXEC |
+			AC_ACCESS_REGISTER_TRANSFER |
+			AC_ACCESS_REGISTER_WRITE);
+	if (execute_abstract_command(target, command) != ERROR_OK)
 		return ERROR_FAIL;
 
-	/* s0 holds the next address to write to
-	 * s1 holds the next data value to write
+	/* If the write succeeded, address on the target was incremented
+	 * during the execution.
```
JanMatCodasip: Sorry to nitpick on comments, but since the algorithm is complex I believe the clarity of the comments is very important.

```suggestion
	/* The execution of the command succeeded, which means:
	 * - write of the first item to memory succeeded
	 * - address on the target (S0) was incremented
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
-		riscv_batch_free(batch);
-		if (result != ERROR_OK)
-			goto error;
+	return ERROR_OK;
+}
 
-		/* Note that if the scan resulted in a Busy DMI response, it
-		 * is this read to abstractcs that will cause the dmi_busy_delay
-		 * to be incremented if necessary. */
+static int write_memory_progbuf_fill_progbuf(struct target *target,
```
JanMatCodasip, line 4590: I would recommend to split this function to two, so that the saving of the registers is not hidden under the "fill_progbuf" name.

- `write_memory_progbuf_backup_registers()`
- `write_memory_progbuf_fill_progbuf()`

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
-	 * s1 holds the next data value to write
+	/* If the write succeeded, address on the target was incremented
+	 * during the execution.
 	 */
+	*address_p += size;
 
-	int result = ERROR_OK;
-	if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
+	return dm_write(target, DM_ABSTRACTAUTO,
+			1 << DM_ABSTRACTAUTO_AUTOEXECDATA_OFFSET);
```
JanMatCodasip, line 4444: If the number of items to write (`count`) is 1, it is not necessary to set and later clear abstractauto[0].

@en-sc @timsifive Please consider whether you are OK with the current state or would like to optimize it.

If you decide to leave it as is, @en-sc, please add a TODO, explaining that this could be improved.

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+{
+	if (batch_run(target, batch) != ERROR_OK)
+		return ERROR_FAIL;
 
-	result = riscv_program_ebreak(&program);
-	if (result != ERROR_OK)
-		goto error;
-	riscv_program_write(&program);
+	/* Note that if the scan resulted in a Busy DMI response, it
+	 * is this read to abstractcs that will cause the dmi_busy_delay
```
```suggestion
	 * is this call to wait_for_idle() that will cause the dmi_busy_delay
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+
+	for (target_addr_t next_addr_on_target = addr_on_target; addr_on_target != end_addr;
+			addr_on_target = next_addr_on_target) {
+		const uint8_t * const curr_buff = buffer + (addr_on_target - address);
+		if (write_memory_progbuf_try_to_write(target, &next_addr_on_target,
+					end_addr, size, curr_buff) != ERROR_OK) {
+			write_memory_progbuf_teardown(target);
+			return ERROR_FAIL;
 		}
+		if (next_addr_on_target == addr_on_target) {
```
JanMatCodasip: Can this condition ever be true? I think it cannot.

That's because the call graph is:

```
write_memory_progbuf_try_to_write() 
   |
   v
write_memory_progbuf_run_batch() 
   |
   |  if (DMI busy) or (absractcs.busy)
   v
write_memory_progbuf_handle_busy()
   |
   v
write_memory_progbuf_startup()
```

And `write_memory_progbuf_startup()` ensures at least one item gets reliably written, so forward progress is guaranteed already.

So it seems this `if` block can be dropped and replaced by:

```
/* write_memory_progbuf_try_to_write() ensures that at least one item 
 * gets successfully written even when busy condition is encountered. */
assert(next_addr_on_target > addr_on_target);
```

Once done, it won't be necessary to have variables `addr_on_target` and `next_addr_on_target`. Only one will suffice, making the code even more simple to read and reason about.

The function `write_memory_progbuf_ensure_forward_progress()` can then be dropped, too.

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+	if (riscv_program_addi(&program, GDB_REGNO_S0, GDB_REGNO_S0, size) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (riscv_program_ebreak(&program) != ERROR_OK)
+		return ERROR_FAIL;
+
+	return riscv_program_write(&program);
+}
+
+static int write_memory_progbuf_inner(struct target *target, target_addr_t address,
```
JanMatCodasip: Renaming `address` to `start_addr` will help readability (because there are multiple other `*addr*` variables in the function already):

```suggestion
static int write_memory_progbuf_inner(struct target *target, target_addr_t start_addr,
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,309 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
 		return ERROR_FAIL;
-	if (riscv_save_register(target, GDB_REGNO_S1) != ERROR_OK)
+
+	target_addr_t address_on_target;
+	if (register_read_direct(target, &address_on_target, GDB_REGNO_S0) != ERROR_OK)
 		return ERROR_FAIL;
+	const uint8_t * const curr_buff = buffer + (address_on_target - *address_p);
+	LOG_TARGET_DEBUG(target, "Restarting from 0x%" TARGET_PRIxADDR, *address_p);
+	*address_p = address_on_target;
+	return write_memory_progbuf_startup(target, address_p, curr_buff, size);
```
JanMatCodasip, line 4476:
```suggestion
	/* This restores the pipeline and ensures one item gets reliably written */
	return write_memory_progbuf_startup(target, address_p, curr_buff, size);
```

src/target/riscv/riscv-013.c
JanMatCodasip: Yes, I was mistaken in the two above points, sorry. Thank you for explanation.

Some of the other review suggestions should lead to further simplification of this function. So in the end, this function should become easier to understand for future readers.

JanMatCodasip (changes requested):
Thank you for the changes. 

I have completed my review and don't expect to have any more (major) comments.

src/target/riscv/riscv-013.c
en-sc: I would like to leave it as-is.
The motivation is -- the registers clobbered by a program is a property of a program, so it would be less clear whether the set of saved registers is correct.

src/target/riscv/riscv-013.c
en-sc: The intention is to enforce and ensure that each write is logged properly. Since the approach to writing it to the target is different in `_fill_batch()`  and `_startup()`, the logging is done on reads from buffer. If you insist, I will separate these calls, though IMHO this will make it harder to guarantee correctness of the logs such way.

src/target/riscv/riscv-013.c
en-sc: I think it would be better to implement this optimization in a separate patch. Added a TODO.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Partially addressed. Please, see [this comment for details](https://github.com/riscv/riscv-openocd/pull/959#discussion_r1408827978).

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Yes. Thanks! Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: I think `distance` is quite fitting in this case. It refers to the actual difference between pointers (in bytes), not the number of elements to write nor the number of `dmi` writes to perform.

src/target/riscv/riscv-013.c
en-sc: Thanks for the catch! Addressed.

en-sc: Though I left `next_addr_on_target`. Otherwise it is impossible to write the assertions.

src/target/riscv/riscv-013.c
en-sc: Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,280 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+			addr_on_target = next_addr_on_target) {
+		const uint8_t * const curr_buff = buffer + (addr_on_target - start_addr);
+		if (write_memory_progbuf_try_to_write(target, &next_addr_on_target,
+					end_addr, size, curr_buff) != ERROR_OK) {
+			write_memory_progbuf_teardown(target);
+			return ERROR_FAIL;
 		}
+		/* write_memory_progbuf_try_to_write() ensures that at least one item
+		 * gets successfully written even when busy condition is encountered. */
+		assert(next_addr_on_target - addr_on_target > 0);
```
JanMatCodasip, line 4638: Could you simply write it like this?

```suggestion
		assert(next_addr_on_target > addr_on_target);
```

Or have you chosen the more complex expression for a specific reason?

src/target/riscv/riscv-013.c
JanMatCodasip: Thank you. Sure, feel free to leave the `next_addr_on_target` in the code.

src/target/riscv/riscv-013.c
JanMatCodasip: Sounds fine to me.

src/target/riscv/batch.c
JanMatCodasip: I can see the comment has been moved to batch.h, so this one can be perhaps considered resolved.

src/target/riscv/riscv-013.c
JanMatCodasip: Ah, I overlooked the the fact that this is the size of the batch in bytes. Thanks for explaining.

In that case, you could name it `bytes_to_write` (or similar) to help the reader. Or if you decide to keep `distance`, that's also OK to me.

src/target/riscv/riscv-013.c
JanMatCodasip: Good point, let's leave those operations together in one function, then.

src/target/riscv/riscv-013.c
JanMatCodasip: Thank you for explaining your rationale. I've continued the discussion in [the other](https://github.com/riscv/riscv-openocd/pull/959/files#r1394771633), tightly coupled thread.

src/target/riscv/riscv-013.c
JanMatCodasip: I would recommend to take one of these options:

(A) Keep the function as is and keep also the (somewhat cumbersome) name - since it is actually saying what it is doing.

(B) Split the function to the two calls - obtaining data from buffer & logging. This logically separates the operations and avoids the need to pick one name for the mix. I believe these operations are logically separate and only artificially grouped into one function, albeit with [good intention](https://github.com/riscv/riscv-openocd/pull/959/files#r1408827978).

I'll leave the rest of the discussion up to you :)

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,280 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
-	if (riscv_save_register(target, GDB_REGNO_S0) != ERROR_OK)
+	/* TODO: Setting abstractauto.autoexecdata is not necessary for a write
+	 * of one element. */
+	return dm_write(target, DM_ABSTRACTAUTO,
+			1 << DM_ABSTRACTAUTO_AUTOEXECDATA_OFFSET);
+}
+
+/**
+ * This function is used to revert the changes made by
+ * `write_memory_progbuf_startup()`
```
JanMatCodasip: Detail: Simplification of the sentence

```suggestion
 * This function is reverts the changes made by
 * `write_memory_progbuf_startup()`
```

src/target/riscv/riscv-013.c
en-sc: > Or have you chosen the more complex expression for a specific reason?

Yes. Consider writing on address `(target_addr_t)(-1) - 1` of two bytes. In this case the `next_addr_on_target` will be zero. This will result in the assertion failing.

JanMatCodasip: Thanks for explaining. Could you please add a comment about that, e.g. `/* This expression is safe when next_address_on_target overflows */` (or similar).

JanMatCodasip (commented):
Thank you for addressing the review suggestions so far. 

I have left several last comments, otherwise it looks all right to me - very close to ready.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Addressed.

en-sc:
@timsifive, can you please take a look once again?

src/target/riscv/riscv-013.c
timsifive: (C) Do we even need this function? If we split the functionality (which seems reasonable), then a user needs to call 2 functions, but this whole thing is already just 2 function calls.

src/target/riscv/riscv-013.c
timsifive: I agree that `distance` is not immediately clear. Instead, make an `end_address` variable. That should be trivial in the places you use `distance`, and it's immediately obvious what it means.

timsifive (dismissed):
I really appreciate all the function comments you wrote. They really help understand what is going on.

This code is definitely more readable than what it was before. Thanks.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,274 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+static target_addr_t write_memory_progbuf_fill_batch(struct riscv_batch *batch,
+		target_addr_t start_address, target_addr_t end_address, uint32_t size,
+		const uint8_t *buffer)
+{
+	assert(size <= 8);
+	const unsigned int writes_per_element = size > 4 ? 2 : 1;
+	const size_t batch_capacity = riscv_batch_available_scans(batch) / writes_per_element;
+	/* With the possibility of uint64_t overflow in mind: */
+	end_address = start_address + MIN((target_addr_t)batch_capacity * size,
+			end_address - start_address);
```
```suggestion
	end_address = MIN(start_address + (target_addr_t)batch_capacity * size,
			end_address);
```

src/target/riscv/riscv-013.c
```diff
@@ -4394,182 +4394,274 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+ * It returns the next address -- the address that will be the start of the next batch.
+ */
+static target_addr_t write_memory_progbuf_fill_batch(struct riscv_batch *batch,
+		target_addr_t start_address, target_addr_t end_address, uint32_t size,
+		const uint8_t *buffer)
+{
+	assert(size <= 8);
+	const unsigned int writes_per_element = size > 4 ? 2 : 1;
+	const size_t batch_capacity = riscv_batch_available_scans(batch) / writes_per_element;
+	/* With the possibility of uint64_t overflow in mind: */
```
timsifive: I don't understand this comment, and I don't think we should try to support memory accesses where the instruction wraps. It might work, but there's no guarantee of it. If that is important to support, we must have a test in riscv-tests to cover that case.

You could assert that end_address > start_address after computing the new end_address (and possibly add an explicit check higher up that returns an error, maybe in a different PR).

timsifive (changes requested):

src/target/riscv/riscv-013.c
JanMatCodasip: Optional, detail: 

On top of what Tim suggests: I'd create new `batch_end_address` variable for it instead of re-using the end_address parameter. 

```suggestion
	const target_addr_t batch_end_address = start_address + MIN((target_addr_t)batch_capacity * size,
			end_address - start_address);
```

src/target/riscv/riscv-013.c
JanMatCodasip: IIUC, the comment refers to the edge case when the data block to write ends at the very top end of the 64-bit address space. So the data block is still in one piece (won't wrap around), but `end_address` already overflows to 0. I think it has been [mentioned here](). @en-sc, have I understood it correctly?

If so:
- You can perhaps rephrase the comment to `/* This is safe even for the edge case when the data block ends at the very top of the 64-bit address space (in which case end_address overflows to 0). */` (or similar).
- Optional: It would be also nice to have a test for this edge case.

I agree that reading/writing blocks that would wrap around should be forbidden, unless properly tested. This can be handled in a separate merge request.

src/target/riscv/riscv-013.c
JanMatCodasip: Optional: A separate variable can be created for the result (to avoid re-using and modifying the end_address input argument).

E.g.: `const target_addr_t batch_end_address = ...;`

I prefer to not change the value of input arguments inside functions as it improves readability. But I understand it is more a matter of personal taste (and [debate](https://softwareengineering.stackexchange.com/questions/369449/is-there-a-reason-to-not-modify-values-of-parameters-passed-by-value)).

JanMatCodasip (commented):
I have only left few last comments in the open threads, otherwise I do not have any more suggestions and all the rest looks all right to me.

Thank you.

src/target/riscv/riscv-013.c
en-sc: @JanMatCodasip, thank you for the explanation! You are exactly right. I will adjust the comment.

On a side note, this is a quote from the unprivileged spec:
>The memory address space is
circular, so that the byte at address 2^XLEN −1 is adjacent to the byte at address zero. Accordingly,
memory address computations done by the hardware ignore overflow and instead wrap around
modulo 2^XLEN .

Doesn't this mean that the 'wrapping' memory access support is desirable? Though I don't see any practical application for this.

src/target/riscv/riscv-013.c
en-sc: > A separate variable can be created for the result (to avoid re-using and modifying the end_address input argument).

Addressed.

src/target/riscv/riscv-013.c
timsifive: It might be desirable, but with the current implementation I don't trust that it works. If we want to support it in OpenOCD, we have to have tests for it because it's a corner case that would be easy to break.

Seems like work for a separate PR, if we think it's important.

timsifive (approved):

JanMatCodasip (approved):
Thank you.

PR #961, opened 2023-11-10
target/riscv: clarify usage of `coreid`

en-sc:
By definition in `target/target.h`, `coreid` is not a unique identifier of a target -- it can be the same for targets on different TAPs.

Change-Id: Ifce78da55fffe28dd8b6b06ecae7d8c4e305c0a2


en-sc:
Related PR in mainline OpenOCD: https://review.openocd.org/c/openocd/+/7997

timsifive (approved):
Fine by me. I don't think coreid is very useful now that we have the target name in all these messages.

MarekVCodasip (approved):

PR #962, opened 2023-11-13
repair git2cl fetch url

bendlas:
seems like upstream for git2cl is gone: https://git.savannah.nongnu.org/git/git2cl
replace submodule url with one from github


timsifive:
https://git.savannah.nongnu.org/git/git2cl.git/ exists, though.

bendlas:
That's so weird, my github action fetching this fails: https://github.com/bendlas/nur-packages/actions/runs/6855003000/job/18639133787#step:7:641

bendlas:
yeah, I think that was due to some specific config on git. on latest runner version it works ...

PR #964, opened 2023-11-16
introduce execution status for riscv_program

aap-sc:
This commit addresses https://github.com/riscv/riscv-openocd/issues/933

Change-Id: I318d5fd6af3a17d1c1099401427ca80ffbec0007

### DIFF
--- a/src/target/riscv/program.c
+++ b/src/target/riscv/program.c
@@ -11,6 +11,7 @@
 #include "helper/log.h"
 
 #include "asm.h"
+#include "debug_defines.h"
 #include "encoding.h"
 
 /* Program interface. */
@@ -26,6 +27,7 @@ int riscv_program_init(struct riscv_program *p, struct target *target)
 	for (size_t i = 0; i < RISCV_MAX_DEBUG_BUFFER_SIZE; ++i)
 		p->debug_buffer[i] = -1;
 
+	p->execution_result = RISCV_DBGBUF_EXEC_RESULT_NOT_EXECUTED;
 	return ERROR_OK;
 }
 
@@ -46,6 +48,7 @@ int riscv_program_exec(struct riscv_program *p, struct target *t)
 {
 	keep_alive();
 
+	p->execution_result = RISCV_DBGBUF_EXEC_RESULT_UNKNOWN_ERROR;
 	riscv_reg_t saved_registers[GDB_REGNO_XPR31 + 1];
 	for (size_t i = GDB_REGNO_ZERO + 1; i <= GDB_REGNO_XPR31; ++i) {
 		if (p->writes_xreg[i]) {
@@ -67,10 +70,16 @@ int riscv_program_exec(struct riscv_program *p, struct target *t)
 	if (riscv_program_write(p) != ERROR_OK)
 		return ERROR_FAIL;
 
-	if (riscv_execute_debug_buffer(t) != ERROR_OK) {
+	uint32_t cmderr;
+	if (riscv_execute_debug_buffer(t, &cmderr) != ERROR_OK) {
+		p->execution_result = (cmderr == DM_ABSTRACTCS_CMDERR_EXCEPTION)
+			? RISCV_DBGBUF_EXEC_RESULT_EXCEPTION
+			: RISCV_DBGBUF_EXEC_RESULT_UNKNOWN_ERROR;
+		/* TODO: what happens if we fail here, but need to restore registers? */
 		LOG_TARGET_DEBUG(t, "Unable to execute program %p", p);
 		return ERROR_FAIL;
 	}
+	p->execution_result = RISCV_DBGBUF_EXEC_RESULT_SUCCESS;
 
 	for (size_t i = GDB_REGNO_ZERO; i <= GDB_REGNO_XPR31; ++i)
 		if (p->writes_xreg[i])
--- a/src/target/riscv/program.c
+++ b/src/target/riscv/program.c
@@ -24,20 +24,19 @@ int riscv_program_init(struct riscv_program *p, struct target *target)
 	for (size_t i = 0; i < RISCV_REGISTER_COUNT; ++i)
 		p->writes_xreg[i] = 0;
 
-	for (size_t i = 0; i < RISCV_MAX_DEBUG_BUFFER_SIZE; ++i)
-		p->debug_buffer[i] = -1;
+	for (size_t i = 0; i < RISCV_MAX_PROGBUF_SIZE; ++i)
+		p->progbuf[i] = -1;
 
-	p->execution_result = RISCV_DBGBUF_EXEC_RESULT_NOT_EXECUTED;
+	p->execution_result = RISCV_PROGBUF_EXEC_RESULT_NOT_EXECUTED;
 	return ERROR_OK;
 }
 
 int riscv_program_write(struct riscv_program *program)
 {
 	for (unsigned i = 0; i < program->instruction_count; ++i) {
-		LOG_TARGET_DEBUG(program->target, "debug_buffer[%02x] = DASM(0x%08x)",
-				i, program->debug_buffer[i]);
-		if (riscv_write_debug_buffer(program->target, i,
-					program->debug_buffer[i]) != ERROR_OK)
+		LOG_TARGET_DEBUG(program->target, "progbuf[%02x] = DASM(0x%08x)",
+				i, program->progbuf[i]);
+		if (riscv_write_progbuf(program->target, i, program->progbuf[i]) != ERROR_OK)
 			return ERROR_FAIL;
 	}
 	return ERROR_OK;
@@ -48,7 +47,7 @@ int riscv_program_exec(struct riscv_program *p, struct target *t)
 {
 	keep_alive();
 
-	p->execution_result = RISCV_DBGBUF_EXEC_RESULT_UNKNOWN_ERROR;
+	p->execution_result = RISCV_PROGBUF_EXEC_RESULT_UNKNOWN;
 	riscv_reg_t saved_registers[GDB_REGNO_XPR31 + 1];
 	for (size_t i = GDB_REGNO_ZERO + 1; i <= GDB_REGNO_XPR31; ++i) {
 		if (p->writes_xreg[i]) {
@@ -61,9 +60,9 @@ int riscv_program_exec(struct riscv_program *p, struct target *t)
 
 	if (riscv_program_ebreak(p) != ERROR_OK) {
 		LOG_TARGET_ERROR(t, "Unable to insert ebreak into program buffer");
-		for (size_t i = 0; i < riscv_debug_buffer_size(p->target); ++i)
+		for (size_t i = 0; i < riscv_progbuf_size(p->target); ++i)
 			LOG_TARGET_ERROR(t, "ram[%02x]: DASM(0x%08" PRIx32 ") [0x%08" PRIx32 "]",
-					(int)i, p->debug_buffer[i], p->debug_buffer[i]);
+					(int)i, p->progbuf[i], p->progbuf[i]);
 		return ERROR_FAIL;
 	}
 
@@ -71,15 +70,15 @@ int riscv_program_exec(struct riscv_program *p, struct target *t)
 		return ERROR_FAIL;
 
 	uint32_t cmderr;
-	if (riscv_execute_debug_buffer(t, &cmderr) != ERROR_OK) {
+	if (riscv_execute_progbuf(t, &cmderr) != ERROR_OK) {
 		p->execution_result = (cmderr == DM_ABSTRACTCS_CMDERR_EXCEPTION)
-			? RISCV_DBGBUF_EXEC_RESULT_EXCEPTION
-			: RISCV_DBGBUF_EXEC_RESULT_UNKNOWN_ERROR;
+			? RISCV_PROGBUF_EXEC_RESULT_EXCEPTION
+			: RISCV_PROGBUF_EXEC_RESULT_UNKNOWN_ERROR;
 		/* TODO: what happens if we fail here, but need to restore registers? */
 		LOG_TARGET_DEBUG(t, "Unable to execute program %p", p);
 		return ERROR_FAIL;
 	}
-	p->execution_result = RISCV_DBGBUF_EXEC_RESULT_SUCCESS;
+	p->execution_result = RISCV_PROGBUF_EXEC_RESULT_SUCCESS;
 
 	for (size_t i = GDB_REGNO_ZERO; i <= GDB_REGNO_XPR31; ++i)
 		if (p->writes_xreg[i])
@@ -200,7 +199,7 @@ int riscv_program_ebreak(struct riscv_program *p)
 {
 	struct target *target = p->target;
 	RISCV_INFO(r);
-	if (p->instruction_count == riscv_debug_buffer_size(p->target) &&
+	if (p->instruction_count == riscv_progbuf_size(p->target) &&
 			r->impebreak) {
 		return ERROR_OK;
 	}
@@ -214,14 +213,14 @@ int riscv_program_addi(struct riscv_program *p, enum gdb_regno d, enum gdb_regno
 
 int riscv_program_insert(struct riscv_program *p, riscv_insn_t i)
 {
-	if (p->instruction_count >= riscv_debug_buffer_size(p->target)) {
+	if (p->instruction_count >= riscv_progbuf_size(p->target)) {
 		LOG_TARGET_ERROR(p->target, "Unable to insert program into progbuf, "
 			"capacity would be exceeded (progbufsize=%d).",
-			(int)riscv_debug_buffer_size(p->target));
+			(int)riscv_progbuf_size(p->target));
 		return ERROR_FAIL;
 	}
 
-	p->debug_buffer[p->instruction_count] = i;
+	p->progbuf[p->instruction_count] = i;
 	p->instruction_count++;
 	return ERROR_OK;
 }
--- a/src/target/riscv/program.h
+++ b/src/target/riscv/program.h
@@ -5,16 +5,17 @@
 
 #include "riscv.h"
 
-#define RISCV_MAX_DEBUG_BUFFER_SIZE 32
+#define RISCV_MAX_PROGBUF_SIZE 32
 #define RISCV_REGISTER_COUNT 32
 #define RISCV_DSCRATCH_COUNT 2
 
 typedef enum {
-	RISCV_DBGBUF_EXEC_RESULT_NOT_EXECUTED,
-	RISCV_DBGBUF_EXEC_RESULT_EXCEPTION,
-	RISCV_DBGBUF_EXEC_RESULT_UNKNOWN_ERROR,
-	RISCV_DBGBUF_EXEC_RESULT_SUCCESS
-} riscv_dbgbuf_exec_status_t;
+	RISCV_PROGBUF_EXEC_RESULT_NOT_EXECUTED,
+	RISCV_PROGBUF_EXEC_RESULT_UNKNOWN,
+	RISCV_PROGBUF_EXEC_RESULT_EXCEPTION,
+	RISCV_PROGBUF_EXEC_RESULT_UNKNOWN_ERROR,
+	RISCV_PROGBUF_EXEC_RESULT_SUCCESS
+} riscv_progbuf_exec_result_t;
 
 /* The various RISC-V debug specifications all revolve around setting up
  * program buffers and executing them on the target.  This structure contains a
@@ -22,7 +23,7 @@ typedef enum {
 struct riscv_program {
 	struct target *target;
 
-	uint32_t debug_buffer[RISCV_MAX_DEBUG_BUFFER_SIZE];
+	uint32_t progbuf[RISCV_MAX_PROGBUF_SIZE];
 
 	/* Number of 32-bit instructions in the program. */
 	size_t instruction_count;
@@ -36,7 +37,7 @@ struct riscv_program {
 
 	/* execution result of the program */
 	/* TODO: remove this field. We should make it a parameter to riscv_program_exec */
-	riscv_dbgbuf_exec_status_t execution_result;
+	riscv_progbuf_exec_result_t execution_result;
 };
 
 /* Initializes a program with the header. */
--- a/src/target/riscv/program.h
+++ b/src/target/riscv/program.h
@@ -9,6 +9,13 @@
 #define RISCV_REGISTER_COUNT 32
 #define RISCV_DSCRATCH_COUNT 2
 
+typedef enum {
+	RISCV_DBGBUF_EXEC_RESULT_NOT_EXECUTED,
+	RISCV_DBGBUF_EXEC_RESULT_EXCEPTION,
+	RISCV_DBGBUF_EXEC_RESULT_UNKNOWN_ERROR,
+	RISCV_DBGBUF_EXEC_RESULT_SUCCESS
+} riscv_dbgbuf_exec_status_t;
+
 /* The various RISC-V debug specifications all revolve around setting up
  * program buffers and executing them on the target.  This structure contains a
  * single program, which can then be executed on targets.  */
@@ -26,6 +33,10 @@ struct riscv_program {
 
 	/* XLEN on the target. */
 	int target_xlen;
+
+	/* execution result of the program */
+	/* TODO: remove this field. We should make it a parameter to riscv_program_exec */
+	riscv_dbgbuf_exec_status_t execution_result;
 };
 
 /* Initializes a program with the header. */
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -54,7 +54,7 @@ static int riscv013_write_debug_buffer(struct target *target, unsigned int index
 static riscv_insn_t riscv013_read_debug_buffer(struct target *target, unsigned int
 		index);
 static int riscv013_invalidate_cached_debug_buffer(struct target *target);
-static int riscv013_execute_debug_buffer(struct target *target);
+static int riscv013_execute_debug_buffer(struct target *target, uint32_t *cmderr);
 static void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a, uint64_t d);
 static void riscv013_fill_dmi_read_u64(struct target *target, char *buf, int a);
 static int riscv013_dmi_write_u64_bits(struct target *target);
@@ -109,12 +109,13 @@ typedef enum slot {
 
 /*** Debug Bus registers. ***/
 
-#define CMDERR_NONE				0
-#define CMDERR_BUSY				1
-#define CMDERR_NOT_SUPPORTED	2
-#define CMDERR_EXCEPTION		3
-#define CMDERR_HALT_RESUME		4
-#define CMDERR_OTHER			7
+/* TODO: CMDERR_* defines can removed */
+#define CMDERR_NONE			DM_ABSTRACTCS_CMDERR_NONE
+#define CMDERR_BUSY			DM_ABSTRACTCS_CMDERR_BUSY
+#define CMDERR_NOT_SUPPORTED		DM_ABSTRACTCS_CMDERR_NOT_SUPPORTED
+#define CMDERR_EXCEPTION		DM_ABSTRACTCS_CMDERR_EXCEPTION
+#define CMDERR_HALT_RESUME		DM_ABSTRACTCS_CMDERR_HALT_RESUME
+#define CMDERR_OTHER			DM_ABSTRACTCS_CMDERR_OTHER
 
 #define HART_INDEX_MULTIPLE	-1
 #define HART_INDEX_UNKNOWN	-2
@@ -189,11 +190,6 @@ typedef struct {
 	bool abstract_write_fpr_supported;
 
 	yes_no_maybe_t has_aampostincrement;
-
-	/* When a function returns some error due to a failure indicated by the
-	 * target in cmderr, the caller can look here to see what that error was.
-	 * (Compare with errno.) */
-	uint8_t cmderr;
 
 	/* Some fields from hartinfo. */
 	uint8_t datasize;
@@ -838,25 +834,32 @@ static uint32_t __attribute__((unused)) abstract_register_size(unsigned width)
 
 static int wait_for_idle(struct target *target, uint32_t *abstractcs)
 {
-	RISCV013_INFO(info);
+	assert(target);
+	assert(abstractcs);
+
 	time_t start = time(NULL);
-	while (1) {
-		if (dm_read(target, abstractcs, DM_ABSTRACTCS) != ERROR_OK)
+	do {
+		if (dm_read(target, abstractcs, DM_ABSTRACTCS) != ERROR_OK) {
+			/* We couldn't read abstractcs. For safety, overwrite the output value to
+			 * prevent the caller working with a stale value of abstractcs. */
+			*abstractcs = 0;
+			LOG_TARGET_ERROR(target,
+				"potentially unrecoverable error detected - could not read abstractcs");
 			return ERROR_FAIL;
+		}
 
 		if (get_field(*abstractcs, DM_ABSTRACTCS_BUSY) == 0)
 			return ERROR_OK;
 
-		if (time(NULL) - start > riscv_command_timeout_sec) {
-			info->cmderr = get_field(*abstractcs, DM_ABSTRACTCS_CMDERR);
+	} while ((time(NULL) - start) < riscv_command_timeout_sec);
 
-			LOG_TARGET_ERROR(target, "Timed out after %ds waiting for busy to go low (abstractcs=0x%x). "
-					"Increase the timeout with riscv set_command_timeout_sec.",
-					riscv_command_timeout_sec,
-					*abstractcs);
-			return ERROR_FAIL;
-		}
-	}
+	LOG_TARGET_ERROR(target,
+		"Timed out after %ds waiting for busy to go low (abstractcs=0x%" PRIx32 "). "
+		"Increase the timeout with riscv set_command_timeout_sec.",
+		riscv_command_timeout_sec,
+		*abstractcs);
+
+	return ERROR_TIMEOUT_REACHED;
 }
 
 static int dm013_select_target(struct target *target)
@@ -865,9 +868,11 @@ static int dm013_select_target(struct target *target)
 	return dm013_select_hart(target, info->index);
 }
 
-static int execute_abstract_command(struct target *target, uint32_t command)
+static int execute_abstract_command(struct target *target, uint32_t command,
+		uint32_t *cmderr)
 {
-	RISCV013_INFO(info);
+	assert(cmderr);
+	*cmderr = CMDERR_NONE;
 	if (debug_level >= LOG_LVL_DEBUG) {
 		switch (get_field(command, DM_COMMAND_CMDTYPE)) {
 			case 0:
@@ -879,17 +884,27 @@ static int execute_abstract_command(struct target *target, uint32_t command)
 		}
 	}
 
-	if (dm_write_exec(target, DM_COMMAND, command, false) != ERROR_OK)
+	if (dm_write_exec(target, DM_COMMAND, command, false /* ensure success */) != ERROR_OK)
 		return ERROR_FAIL;
 
-	uint32_t abstractcs = 0;
-	int result = wait_for_idle(target, &abstractcs);
-
-	info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
-	if (info->cmderr != 0 || result != ERROR_OK) {
-		LOG_TARGET_DEBUG(target, "command 0x%x failed; abstractcs=0x%x", command, abstractcs);
-		/* Clear the error. */
-		dm_write(target, DM_ABSTRACTCS, DM_ABSTRACTCS_CMDERR);
+	uint32_t abstractcs;
+	int wait_result = wait_for_idle(target, &abstractcs);
+	if (wait_result != ERROR_OK) {
+		/* TODO: can we recover from this? */
+		if (wait_result == ERROR_TIMEOUT_REACHED)
+			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (timeout)", command);
+		else
+			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (unknown fatal error %d)", command, wait_result);
+		return wait_result;
+	}
+	*cmderr = get_field32(abstractcs, DM_ABSTRACTCS_CMDERR);
+	if (*cmderr != CMDERR_NONE) {
+		LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed; abstractcs=0x%" PRIx32,
+			command, abstractcs);
+		/* Attempt to clear the error. */
+		/* TODO: can we add a more substantial recovery if the clear operation fails ? */
+		if (dm_write(target, DM_ABSTRACTCS, DM_ABSTRACTCS_CMDERR) != ERROR_OK)
+			LOG_TARGET_ERROR(target, "could not clear abstractcs error");
 		return ERROR_FAIL;
 	}
 
@@ -997,9 +1012,10 @@ static int register_read_abstract_with_size(struct target *target,
 	uint32_t command = access_register_command(target, number, size,
 			AC_ACCESS_REGISTER_TRANSFER);
 
-	int result = execute_abstract_command(target, command);
+	uint32_t cmderr;
+	int result = execute_abstract_command(target, command, &cmderr);
 	if (result != ERROR_OK) {
-		if (info->cmderr == CMDERR_NOT_SUPPORTED) {
+		if (cmderr == CMDERR_NOT_SUPPORTED) {
 			if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31) {
 				info->abstract_read_fpr_supported = false;
 				LOG_TARGET_INFO(target, "Disabling abstract command reads from FPRs.");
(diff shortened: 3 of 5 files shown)


aap-sc:
Note to reviewers:

- this commit has lot's of TODO statements. I plan to work on most of these in a subsequent commits:
1. moving ` exec_status` to the output parameter of `riscv_program_exec` is the next thing I plan to implement if this PR is approved. I wanted to produce a minimal diff without causing too much disturbance across the codebase.
2. error-handling adjustment may take a while. But I hope that these comments only illustrate existing defects without introducing new ones.

src/target/riscv/riscv-013.c
```diff
@@ -857,9 +859,11 @@ static int dm013_select_target(struct target *target)
 	return dm013_select_hart(target, info->index);
 }
 
-static int execute_abstract_command(struct target *target, uint32_t command)
+static int execute_abstract_command(struct target *target, uint32_t command,
+		unsigned int *p_abstractcs_err)
```
JanMatCodasip: Can this be called just `cmderr`?

OpenOCD also do not follow the convention to have `p_` prefixes on pointers.

```suggestion
		unsigned int *cmderr)
```

src/target/riscv/riscv-013.c
```diff
@@ -53,7 +53,8 @@ static int riscv013_write_debug_buffer(struct target *target, unsigned int index
 static riscv_insn_t riscv013_read_debug_buffer(struct target *target, unsigned int
 		index);
 static int riscv013_invalidate_cached_debug_buffer(struct target *target);
-static int riscv013_execute_debug_buffer(struct target *target);
+static int riscv013_execute_debug_buffer(struct target *target,
+		riscv_dbgbuf_exec_status_t *exec_status);
```
JanMatCodasip: I have not yet done proper review and will need to return to this later.

As an early feedback, please call this output value "execution result", not "execution status". This applies to all occurrences, not just this place.

I would also prefer to get the raw `cmderr` as the result, and not have a new enumeration for it. Same as you have it in `execute_abstract_command()`.

JanMatCodasip (changes requested):

src/target/riscv/riscv-013.c
aap-sc: > I would also prefer to get the raw cmderr as the result, and not have a new enumeration for it. Same as you have it in execute_abstract_command().

Unless there is a strong pushback - I would prefer to leave this part as is. The reason is that it renders existing API inconsistent.

JanMatCodasip: Which API do you refer to, please? Do you mean the `write_debug_buffer` function pointer in `struct riscv_info`?

If so, then I wouldn't worry about passing the cmderr, for these reasons:
- There is no debug buffer in 0.11 target support at all.
- This API based on function pointers to 0.11 and 0.13 functions I would like to eventually remove and rewrite by direct function calls. That's because the API complicates reading of the the code, static analysis of the code and brings no actual benefit.

aap-sc: > Which API do you refer to, please? Do you mean the write_debug_buffer function pointer in struct riscv_info?

No. I believe I've misunderstood your original comment. I though that you've proposed to return cmderr instead of a generic `int` error code. Looks like  you just don't want `cmderr` to be translated to `riscv_dbgbuf_exec_status_t`. I haven't thought about this possibility - will take a look.

JanMatCodasip: >  Looks like you just don't want cmderr to be translated to riscv_dbgbuf_exec_status_t

Yes, that was my idea.

I believe the low-level program buffer execution function should provide the raw error code (cmderr). If the caller wishes to translate the error code, that is OK, but the raw value should remain accessible.

src/target/riscv/riscv-013.c
aap-sc: addressed. replaced with `cmderr` in this function and `abstractcs_err` elsewhere.

src/target/riscv/riscv-013.c
aap-sc: addressed * . There is a caveat, though:

1. low-level `execute_debug_buffer` returns raw `abstractcs_err` as requested.
2. ` riscv_program_exec` still "returns" translated `abstractcs_err` (that is the new enumeration). The reason is that it can fail during the preparation stage, while the respected abstract command is not ready to be executed. This corresponds to ` RISCV_DBGBUF_EXEC_RESULT_NOT_EXECUTED` enumeration value. I believe that it would be beneficial to clearly identify at what stage  `riscv_program_exec` fail.

I haven't squashed the respected changes to simplify review. Will squash once (if) changes are approved.

src/target/riscv/program.h
```diff
@@ -26,6 +33,10 @@ struct riscv_program {
 
 	/* XLEN on the target. */
 	int target_xlen;
+
+	/* execution result of the program */
+	/* TODO: remove this field. We should make it a parameter to riscv_program_exec */
```
timsifive, line 39: Can you clarify when that should happen?

src/target/riscv/riscv-013.c
```diff
@@ -831,25 +828,31 @@ static uint32_t __attribute__((unused)) abstract_register_size(unsigned width)
...
-	RISCV013_INFO(info);
+	assert(target);
+	assert(abstractcs);
+
 	time_t start = time(NULL);
-	while (1) {
-		if (dm_read(target, abstractcs, DM_ABSTRACTCS) != ERROR_OK)
+	do {
+		if (dm_read(target, abstractcs, DM_ABSTRACTCS) != ERROR_OK) {
+			/* NOTE: this value is meaninless, overwrite to avoid stale value */
```
```suggestion
			/* NOTE: this value is meaningless, overwrite to avoid stale value */
```

src/target/riscv/riscv-013.c
```diff
@@ -872,17 +877,27 @@ static int execute_abstract_command(struct target *target, uint32_t command)
...
-		/* Clear the error. */
-		dm_write(target, DM_ABSTRACTCS, DM_ABSTRACTCS_CMDERR);
+	uint32_t abstractcs;
+	int wait_result = wait_for_idle(target, &abstractcs);
+	if (wait_result != ERROR_OK) {
+		/* TODO: can we recover from this? */
+		if (wait_result == ERROR_TIMEOUT_REACHED)
+			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (timeout)", command);
+		else
+			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (unknown fatal error)", command);
```
```suggestion
			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (unknown fatal error %d)", command, wait_result);
```

aap-sc:
@JanMatCodasip , @timsifive could you take a look once again?

src/target/riscv/program.c
```diff
@@ -67,11 +70,18 @@ int riscv_program_exec(struct riscv_program *p, struct target *t)
 	if (riscv_program_write(p) != ERROR_OK)
 		return ERROR_FAIL;
 
-	if (riscv_execute_debug_buffer(t) != ERROR_OK) {
+	unsigned int abstractcs_err;
+	if (riscv_execute_debug_buffer(t, &abstractcs_err) != ERROR_OK) {
+		p->execution_result = (abstractcs_err == DM_ABSTRACTCS_CMDERR_EXCEPTION)
+			? RISCV_DBGBUF_EXEC_RESULT_EXCEPTION
+			: RISCV_DBGBUF_EXEC_RESULT_UNKNOWN;
```
timsifive: UNKNOWN doesn't feel right here. We don't know what happened, but it was definitely an error. UNKNOWN_ERROR seems more descriptive.

src/target/riscv/riscv-013.c
```diff
@@ -2884,31 +2901,42 @@ static int execute_fence(struct target *target)
...
 		riscv_program_init(&program, target);
-		riscv_program_fence_rw_rw(&program);
 		riscv_program_fence_i(&program);
-		result = riscv_program_exec(&program, target);
-		if (result != ERROR_OK)
-			LOG_TARGET_DEBUG(target, "Unable to execute pre-fence");
+		riscv_program_fence_rw_rw(&program);
+		if (riscv_program_exec(&program, target) != ERROR_OK) {
+			if (program.execution_result != RISCV_DBGBUF_EXEC_RESULT_EXCEPTION) {
+				LOG_TARGET_ERROR(target, "Unexpected error during fence execution");
```
timsifive: Can you add a comment that explains exceptions are OK, because (we assume) it means these particular fence instructions aren't implemented?

timsifive (changes requested):
Looks good overall.

src/target/riscv/program.h
aap-sc: as soon as this and https://github.com/riscv/riscv-openocd/pull/959 are merged. I plan to create a subsequent commit.

aap-sc: that is there is no technical reason preventing us from doing this right now. It's just I want to reduce the diff

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/program.c
aap-sc: addressed

src/target/riscv/riscv-013.c
aap-sc: I've added the comment, though not in this exact place. Just added detailed information regarding RISCV_DBGBUF_EXEC_RESULT_EXCEPTION to execute_fence

src/target/riscv/program.c
```diff
@@ -46,6 +48,7 @@ int riscv_program_exec(struct riscv_program *p, struct target *t)
 {
 	keep_alive();
 
+	p->execution_result = RISCV_DBGBUF_EXEC_RESULT_UNKNOWN_ERROR;
```
timsifive: Is it worth having RISCV_DBGBUF_EXEC_RESULT_UNKNOWN_ERROR and RISCV_DBGBUF_EXEC_RESULT_UNKNOWN? Using the second one here. Might help things be more clear when debugging.

timsifive (changes requested):
One last comment.

src/target/riscv/riscv-013.c
```diff
@@ -4568,14 +4608,16 @@ static int write_memory_progbuf(struct target *target, target_addr_t address,
...
-		if (info->cmderr == CMDERR_NONE && !dmi_busy_encountered) {
+
+		unsigned int cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
+		if (cmderr == CMDERR_NONE && !dmi_busy_encountered) {
 			LOG_TARGET_DEBUG(target, "Successful (partial?) memory write");
-		} else if (info->cmderr == CMDERR_BUSY || dmi_busy_encountered) {
-			if (info->cmderr == CMDERR_BUSY)
+		} else if (cmderr == CMDERR_BUSY || dmi_busy_encountered) {
+			if (cmderr == CMDERR_BUSY)
 				LOG_TARGET_DEBUG(target, "Memory write resulted in abstract command busy response.");
```
JanMatCodasip: @aap-sc, @en-sc The change of this part of the code will conflict with https://github.com/riscv/riscv-openocd/pull/959.

Which of these two pull requests would you like to have merged first? The other one would need to be rebased after that.

src/target/riscv/riscv-013.c
```diff
@@ -831,25 +828,31 @@ static uint32_t __attribute__((unused)) abstract_register_size(unsigned width)
...
+				"potentially unrecoverable error detected - could not read abstractcs");
 			return ERROR_FAIL;
+		}
 
 		if (get_field(*abstractcs, DM_ABSTRACTCS_BUSY) == 0)
 			return ERROR_OK;
 
-		if (time(NULL) - start > riscv_command_timeout_sec) {
-			info->cmderr = get_field(*abstractcs, DM_ABSTRACTCS_CMDERR);
+	} while (time(NULL) - start < riscv_command_timeout_sec);
```
JanMatCodasip: Just a detail that improves readability a little:

```suggestion
	} while ((time(NULL) - start) < riscv_command_timeout_sec);
```

src/target/riscv/riscv-013.c
```diff
@@ -990,9 +1005,10 @@ static int register_read_abstract_with_size(struct target *target,
 	uint32_t command = access_register_command(target, number, size,
 			AC_ACCESS_REGISTER_TRANSFER);
 
-	int result = execute_abstract_command(target, command);
+	unsigned int abstractcs_err;
+	int result = execute_abstract_command(target, command, &abstractcs_err);
```
JanMatCodasip: The third parameter of execute_abstract_command() is called `cmderr`, so I'd recommend to name the argument also `cmderr` (in order to not introduce different names for the same thing).

```suggestion
	unsigned int cmderr;
	int result = execute_abstract_command(target, command, &cmderr);
```

src/target/riscv/riscv-013.c
```diff
@@ -1038,9 +1054,10 @@ static int register_write_abstract(struct target *target, enum gdb_regno number,
 	if (write_abstract_arg(target, 0, value, size) != ERROR_OK)
 		return ERROR_FAIL;
 
-	int result = execute_abstract_command(target, command);
+	unsigned int abstractcs_err;
+	int result = execute_abstract_command(target, command, &abstractcs_err);
```
JanMatCodasip: Same as above.

There are more occurrences throughout this file. If you agree with the suggestion, please take a look at all of them. I have not commented on the other cases to keep the number of comments low.

```suggestion
	unsigned int cmderr;
	int result = execute_abstract_command(target, command, &cmderr);
```

src/target/riscv/program.h
```diff
@@ -9,6 +9,13 @@
 #define RISCV_REGISTER_COUNT 32
 #define RISCV_DSCRATCH_COUNT 2
 
+typedef enum {
+	RISCV_DBGBUF_EXEC_RESULT_NOT_EXECUTED,
+	RISCV_DBGBUF_EXEC_RESULT_EXCEPTION,
+	RISCV_DBGBUF_EXEC_RESULT_UNKNOWN_ERROR,
+	RISCV_DBGBUF_EXEC_RESULT_SUCCESS
+} riscv_dbgbuf_exec_status_t;
```
JanMatCodasip: 1) Please, rename the type name `riscv_dbgbuf_exec_status_t` to `riscv_dbgbuf_exec_result_t` so that the naming is consistent.

2) Optionally: Consider renaming `dbgbuf` to `progbuf`, which is a more commonly used abbreviation (but yes, there are exceptions like `execute_debug_buffer`).

src/target/riscv/riscv-013.c
```diff
@@ -3563,8 +3602,11 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
 		}
 
 		/* Execute the command */
-		result = execute_abstract_command(target, command);
+		unsigned int abstractcs_err;
+		result = execute_abstract_command(target, command, &abstractcs_err);
 
+		/* TODO: we need to modify error handling here. We can make decisions only
+		 * if abstractcs_err indicates an exception */
```
JanMatCodasip: I believe we should check here for `cmderr==2 (unsupported)`, rather than `cmderr==3 (exception)`. Do you agree?

```suggestion
		/* TODO: we need to modify error handling here. We can make decisions only
		 * if cmderr indicates "unsupported" */
```

src/target/riscv/riscv-013.c
```diff
@@ -3629,8 +3670,11 @@ static int read_memory_progbuf_inner_startup(struct target *target,
 	const uint32_t startup_command = access_register_command(target,
 			GDB_REGNO_S1, riscv_xlen(target),
 			AC_ACCESS_REGISTER_TRANSFER | AC_ACCESS_REGISTER_POSTEXEC);
-	if (execute_abstract_command(target, startup_command) != ERROR_OK)
+	unsigned int abstractcs_err;
+	if (execute_abstract_command(target, startup_command, &abstractcs_err) != ERROR_OK)
 		return ERROR_FAIL;
+	/* TODO: we need to modify error handling here. We can make decisions only
+	 * if abstractcs_err indicates an exception */
```
JanMatCodasip: Do we really need to make different decisions here based on the value of cmderr?

I believe it is enough to detect the failure here, and we don't need to examine the exact value in cmderr.

src/target/riscv/riscv-013.c
```diff
@@ -872,17 +877,27 @@ static int execute_abstract_command(struct target *target, uint32_t command)
...
+	int wait_result = wait_for_idle(target, &abstractcs);
+	if (wait_result != ERROR_OK) {
+		/* TODO: can we recover from this? */
+		if (wait_result == ERROR_TIMEOUT_REACHED)
+			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (timeout)", command);
+		else
+			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (unknown fatal error %d)", command, wait_result);
+		return wait_result;
+	}
+	*cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
```
JanMatCodasip: get_field() returns uint64_t but cmderr is of type unsigned int*. Compilers may therefore warn about narrowing conversions. For that reason, please add an implicit cast.

```suggestion
	*cmderr = (unsigned int)get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
```

src/target/riscv/riscv-013.c
```diff
@@ -872,17 +877,27 @@ static int execute_abstract_command(struct target *target, uint32_t command)
...
+	if (*cmderr != CMDERR_NONE) {
+		LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed; abstractcs=0x%" PRIx32 "",
+			command, abstractcs);
+		/* Attempt to clear the error. */
+		/* TODO: can we add a more substantial recovery if the clear operation fails ? */
+		if (dm_write(target, DM_ABSTRACTCS, DM_ABSTRACTCS_CMDERR) != ERROR_OK)
+			LOG_TARGET_ERROR(target, "could not clear abstractcs error");
```
JanMatCodasip, line 907: I would recommend to remove this TODO.

If a simple write to a debug module register fails (in this case write to abstractcs), then the target is seriously broken already. I'm afraid there is nothing more we could do, apart from printing an error. 

```suggestion
		if (dm_write(target, DM_ABSTRACTCS, DM_ABSTRACTCS_CMDERR) != ERROR_OK)
			LOG_TARGET_ERROR(target, "could not clear abstractcs error");
```

src/target/riscv/riscv-013.c
```diff
@@ -4103,7 +4141,8 @@ static int read_memory_progbuf_inner_one(struct target *target,
 	uint32_t command = access_register_command(target, GDB_REGNO_S1,
 			riscv_xlen(target), AC_ACCESS_REGISTER_WRITE |
 			AC_ACCESS_REGISTER_TRANSFER | AC_ACCESS_REGISTER_POSTEXEC);
-	if (execute_abstract_command(target, command) != ERROR_OK)
+	unsigned int abstractcs_err;
+	if (execute_abstract_command(target, command, &abstractcs_err) != ERROR_OK)
```
JanMatCodasip: The cmderr is not taken into account here. 

Would it be possible to make the third argument optional? If the user does not care about the exact cmderr value, they can pass NULL.

src/target/riscv/riscv-013.c
```diff
@@ -4529,7 +4568,8 @@ static int write_memory_progbuf(struct target *target, target_addr_t address,
 						AC_ACCESS_REGISTER_POSTEXEC |
 						AC_ACCESS_REGISTER_TRANSFER |
 						AC_ACCESS_REGISTER_WRITE);
-				result = execute_abstract_command(target, command);
+				unsigned int abstractcs_err;
+				result = execute_abstract_command(target, command, &abstractcs_err);
```
JanMatCodasip: Same as above - the cmderr is not used here. Another place where the caller could simply pass NULL as the third argument.

src/target/riscv/riscv-013.c
```diff
@@ -4992,15 +5034,17 @@ static int riscv013_invalidate_cached_debug_buffer(struct target *target)
 	return ERROR_OK;
 }
 
-static int riscv013_execute_debug_buffer(struct target *target)
+static int riscv013_execute_debug_buffer(struct target *target,
+		unsigned int *abstractcs_err)
 {
+	assert(abstractcs_err);
```
JanMatCodasip: I'd recommend to move the assert() to the inner function execute_abstract_command(), that is, the place where the pointer is actually assigned to.

src/target/riscv/riscv-013.c
```diff
@@ -4992,15 +5034,17 @@ static int riscv013_invalidate_cached_debug_buffer(struct target *target)
 	return ERROR_OK;
 }
 
-static int riscv013_execute_debug_buffer(struct target *target)
+static int riscv013_execute_debug_buffer(struct target *target,
+		unsigned int *abstractcs_err)
```
JanMatCodasip: Please consider calling the third argument `cmderr`, same as it is in `execute_abstract_command()`.

src/target/riscv/riscv.c
```diff
@@ -5360,10 +5360,11 @@ riscv_insn_t riscv_read_debug_buffer(struct target *target, int index)
 	return r->read_debug_buffer(target, index);
 }
 
-int riscv_execute_debug_buffer(struct target *target)
+int riscv_execute_debug_buffer(struct target *target,
+	unsigned int *abstractcs_err)
```
JanMatCodasip: Same as above - consider calling the last parameter `cmderr` so that the names are unified.

```suggestion
int riscv_execute_debug_buffer(struct target *target,
	unsigned int *cmderr)
```

src/target/riscv/riscv-013.c
JanMatCodasip: There is an identical comment also in `read_memory_abstract()`.

src/target/riscv/riscv-013.c
```diff
@@ -3647,20 +3691,18 @@ static int read_memory_progbuf_inner_startup(struct target *target,
 		goto clear_abstractauto_and_fail;
 
 	uint32_t abstractcs;
-
 	if (wait_for_idle(target, &abstractcs) != ERROR_OK)
 		goto clear_abstractauto_and_fail;
 
-	info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
-	switch (info->cmderr) {
+	uint32_t cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
```
JanMatCodasip: In some places, cmderr has type of uint32_t and in others it is unsigned int.

Please, could you unify it.

Also when assigning to uint32_t, consider using get_field32() so that we avoid potential compiler warnings about narrowing conversions.

Please, take a look at all the uses of get_field(), touched in this pull request. I haven't commented on each of them individually.

JanMatCodasip (changes requested):
Thank you for the changes, it looks fine overall.  I have completed my review and don't expect to have any more major comments than what I have posted.

src/target/riscv/riscv-013.c
aap-sc: I would prefer if this change was merged first. https://github.com/riscv/riscv-openocd/pull/959 - is not that big we can wait.

aap-sc: ouch. 

> I would prefer if this change was merged first. https://github.com/riscv/riscv-openocd/pull/959 - is not that big we can wait.

it's vice versa. I mean: let's merge https://github.com/riscv/riscv-openocd/pull/959 . This can wait.

src/target/riscv/program.h
aap-sc: will take a look, thanks

src/target/riscv/riscv-013.c
aap-sc: > Please, could you unify it.

I think I've  tried to do that, but the final scope of changes was quite large. I'll take a look once again, but **if** the number of changes is large, I would prefer to address it in a subsequent commit.

JanMatCodasip: If you could double-check the lines with get_field(), affected by this merge request, that would be nice.

As for the other uses of get_field(), I agree that these can (and should) be handled in a separate pull request.

src/target/riscv/riscv-013.c
JanMatCodasip: Correction: If the execute_abstract_command() is updated to accept NULL as the third argument, then the assert should of course be dropped (not moved anywhere else).

src/target/riscv/riscv-013.c
aap-sc: > I believe we should check here for cmderr==2 (unsupported), rather than cmderr==3 (exception). Do you agree?

 we should check for both, actually. The intention was to check that some error code is set . That is cmderr is not NONE (happens in case of timeouts)

src/target/riscv/riscv-013.c
aap-sc: > I believe it is enough to detect the failure here, and we don't need to examine the exact value in cmderr.

That's why we have a TODO here. I'm not sure. For example - what if we detect a timeout? Ideally, we should find a way to restore the operation.

src/target/riscv/riscv-013.c
aap-sc: > I'm afraid there is nothing more we could do, apart from printing an error. 

The TODO is to investigate if we have any options to recover somehow. For example @en-sc had a suggestion (during an internal review) to just drop a marker that target target was examined to enforce debugger to restore the state or give a user a reasonable feedback. **maybe** we could also reset the DM. I mean there may be some crazy variant that will make a recovery possible.

src/target/riscv/riscv-013.c
aap-sc: > Would it be possible to make the third argument optional? If the user does not care about the exact cmderr value, they can pass NULL.

I'm reluctant to do this. I want to enforce programmer  to always think about error detection. We should not have situations where we don't care about errors.  We already had way too many commits to correct the situation (and more to come...)

src/target/riscv/riscv-013.c
aap-sc: double checked. This was the only place - I'll adjust.

src/target/riscv/program.c
aap-sc: yep. I think it would not hurt. will address.

aap-sc: addressed

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/program.h
aap-sc: addressed.

> nut yes, there are exceptions like execute_debug_buffer

not anymore :). decided to rename all "*debug_buffer*" stuff to "*progbuf*"

src/target/riscv/riscv-013.c
aap-sc: yep. addressed

src/target/riscv/riscv-013.c
aap-sc: same point  is above. I wold leave it for now.

NOTE: I plan to address this TODO (we have an internal task to tackle this). These comments will be either removed or we'll propose something

src/target/riscv/riscv-013.c
aap-sc: addressed (assert was removed/moved)

src/target/riscv/riscv-013.c

src/target/riscv/riscv.c

src/target/riscv/riscv-013.c

src/target/riscv/program.h
```diff
@@ -5,17 +5,25 @@
...
 #define RISCV_DSCRATCH_COUNT 2
 
+typedef enum {
+	RISCV_DBGBUF_EXEC_RESULT_NOT_EXECUTED,
+	RISCV_DBGBUF_EXEC_RESULT_UNKNOWN,
+	RISCV_DBGBUF_EXEC_RESULT_EXCEPTION,
+	RISCV_DBGBUF_EXEC_RESULT_UNKNOWN_ERROR,
+	RISCV_DBGBUF_EXEC_RESULT_SUCCESS
+} riscv_progbuf_exec_result_t;
+
```
JanMatCodasip: Thank you for unifying the names (debug_buffer -> progbuf) in many places in the code.

One last nitpick - `RISCV_DBGBUF_*` items in this enum can be renamed to `RISCV_PROGBUF_*`, too.

```suggestion
typedef enum {
	RISCV_PROGBUF_EXEC_RESULT_NOT_EXECUTED,
	RISCV_PROGBUF_EXEC_RESULT_UNKNOWN,
	RISCV_PROGBUF_EXEC_RESULT_EXCEPTION,
	RISCV_PROGBUF_EXEC_RESULT_UNKNOWN_ERROR,
	RISCV_PROGBUF_EXEC_RESULT_SUCCESS
} riscv_progbuf_exec_result_t;

```

src/target/riscv/riscv-013.c
JanMatCodasip: Thank you for explanation.

I believe it is a more general question about what OpenOCD should do (if anything at all) when it encounters an evidently broken target.

An argument could be made to just report this error to the user, and let the user deal with it (which would be my preference in such situations).

Anyway, let's leave the TODO in place as you originally suggested.

src/target/riscv/program.h
JanMatCodasip: Thanks.

src/target/riscv/riscv-013.c
```diff
@@ -831,25 +828,31 @@ static uint32_t __attribute__((unused)) abstract_register_size(unsigned width)
...
-	RISCV013_INFO(info);
+	assert(target);
+	assert(abstractcs);
+
 	time_t start = time(NULL);
-	while (1) {
-		if (dm_read(target, abstractcs, DM_ABSTRACTCS) != ERROR_OK)
+	do {
+		if (dm_read(target, abstractcs, DM_ABSTRACTCS) != ERROR_OK) {
+			/* NOTE: this value is meaningless, overwrite to avoid stale value */
```
JanMatCodasip: Did I please get correctly the meaning of this comment?

```suggestion
			/* We couldn't read abstractcs. For safety, overwrite the output value to prevent the caller
			 * working with a stale value of abstractcs. */
```

src/target/riscv/program.c
```diff
@@ -58,20 +60,27 @@ int riscv_program_exec(struct riscv_program *p, struct target *t)
...
+		/* TODO: what happens if we fail here, but need to restore registers? */
 		LOG_TARGET_DEBUG(t, "Unable to execute program %p", p);
 		return ERROR_FAIL;
 	}
+	p->execution_result = RISCV_DBGBUF_EXEC_RESULT_SUCCESS;
 
+	/* TODO: what happens if we can not restore context ? */
 	for (size_t i = GDB_REGNO_ZERO; i <= GDB_REGNO_XPR31; ++i)
 		if (p->writes_xreg[i])
 			riscv_set_register(t, i, saved_registers[i]);
```
JanMatCodasip: If I read the code correctly, this register save/restore logic inside riscv_program_write() is not used anymore (dead code). No part of code assigns `writes_xreg[]` to any other value than 0.

To save registers prior to progbuf execution, a different strategy is used now -- explicit calls to `riscv_save_register()`.

For that reason, I'd recommend to not add TODOs here. 

Instead, please consider removing all these:
- both the loops on lines 51-59 and 84-86 in riscv_program_exec()
- writes_xreg array in struct riscv_program
- saved_registers array in riscv_program_exec()

src/target/riscv/riscv-013.c
JanMatCodasip: I fully agree that errors should always be taken into account. That however does not mean that the caller of `execute_abstract_command()` must always introspect the detailed cause of the error (cmderr) :-)

After thinking about it more, I believe it would be good to have 4 distinct return codes for execute_abstract_command():

- (1) `ERROR_OK` - Command succeeded. In this case the caller needs not look at cmderr (because cmderr is guaranteed to be 0).
- (2) `ERROR_CMD_FAILED` (or similar name) - Command was executed but failed due to cmderr!=0. The caller can look at cmderr to determine the actual failure reason, if it wishes, but it does not need to do so. All depends on the caller's situation, and how detailed information about the error it needs to have.
- (3) `ERROR_FAIL` - During the execution of the command, the DMI read or write operation failed. The target (or the connection to it) is seriously broken. The value in cmderr is irrelevant in this case.
- (4) `ERROR_TIMEOUT_REACHED` - Timeout occurred. The value in cmderr is not relevant in this case, either.

At the moment, cases (2) and (3) is merged into one ERROR_FAIL return code. It would be good to split it because those are two different outcomes. Please, let me know if you have similar or different opinion. But no need to make the changes directly in this merge request. It would be good to agree on the strategy.

Based on the above, I still believe the third argument of execute_abstract_command() should be made optional. The caller should detect the overall status (1) - (4) through the return code. And only in case of (2), the cmderr can be introspected, but it is not mandatory to do so in all cases.

src/target/riscv/riscv-013.c
JanMatCodasip: I have continued the discussion here: https://github.com/riscv/riscv-openocd/pull/964#discussion_r1422155725

Please let me know if you share the opinion or see it differently.

src/target/riscv/riscv-013.c
JanMatCodasip: I have continued the discussion here: https://github.com/riscv/riscv-openocd/pull/964#discussion_r1422155725

Please let me know if you share the opinion or see it differently.

src/target/riscv/program.h
aap-sc: strange, I though I did that ...

src/target/riscv/riscv-013.c
aap-sc: yep

src/target/riscv/program.c
aap-sc: > Instead, please consider removing all these:
> 
> both the loops on lines 51-59 and 84-86 in riscv_program_exec()
> writes_xreg array in struct riscv_program
> saved_registers array in riscv_program_exec()

If no objections I would adjust this code in a subsequent commit (if @en-sc haven't done this already). Spoiler: there is a **huge** (the largest yet) patch incoming by @en-sc that refactores work with registers.

aap-sc:
> That however does not mean that the caller of execute_abstract_command() must always introspect the detailed cause of the error (cmderr) :-)

> It would be good to agree on the strategy.

If there are lot's of cases when we don't want to introspect the detailed cause, then I agree that it may be better to return generic error-codes with `cmderr` being optional. Though I still need to do some investigation before forming a definite opinion regarding `must always introspect the detailed cause` part.

src/target/riscv/program.h
timsifive: Random comment: It is much easier to review changes if you make a separate PR for renaming a bunch of stuff without changing any functionality.

src/target/riscv/program.c
JanMatCodasip: Understood. Sure, this can be done in a separate pull request.

src/target/riscv/riscv-013.c
JanMatCodasip: @aap-sc, https://github.com/riscv/riscv-openocd/pull/959 has just been merged. 

So this pull request can now be rebased.

JanMatCodasip (commented):
From my point of view:

- 1) There is one last open thread (https://github.com/riscv/riscv-openocd/pull/964#discussion_r1422021336) with cosmetic rename stuff.
- 2) This pull request should be rebased (and conflicts solved).

After that I consider it ready for merge. Thanks.

src/target/riscv/program.h
aap-sc: > Random comment: It is much easier to review changes if you make a separate PR for renaming a bunch of stuff without changing any functionality.

I moved these renames to a separate change set specifically to simplify the review. If this is still inconvenient, I can factor-out this in a separate MR.

aap-sc: @timsifive I've splitted the whole change set into 2 distinct commits: one with functional changes the other is with the rename of `dbgbuf->progbuf`. I've tested them separately.  If necessary, I can move the second changeset to a separate MR.

@JanMatCodasip : addressed.

aap-sc:
rebased and addressed all comments. @timsifive , @JanMatCodasip could you please take a look?

src/target/riscv/riscv-013.c
```diff
@@ -879,17 +885,27 @@ static int execute_abstract_command(struct target *target, uint32_t command)
...
+		/* TODO: can we recover from this? */
+		if (wait_result == ERROR_TIMEOUT_REACHED)
+			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (timeout)", command);
+		else
+			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (unknown fatal error %d)", command, wait_result);
+		return wait_result;
+	}
+	*cmderr = (unsigned int)get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
+	if (*cmderr != CMDERR_NONE) {
+		LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed; abstractcs=0x%" PRIx32 "",
```
timsifive: Remove that last `""`

src/target/riscv/riscv-013.c
```diff
@@ -879,17 +885,27 @@ static int execute_abstract_command(struct target *target, uint32_t command)
...
+	int wait_result = wait_for_idle(target, &abstractcs);
+	if (wait_result != ERROR_OK) {
+		/* TODO: can we recover from this? */
+		if (wait_result == ERROR_TIMEOUT_REACHED)
+			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (timeout)", command);
+		else
+			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (unknown fatal error %d)", command, wait_result);
+		return wait_result;
+	}
+	*cmderr = (unsigned int)get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
```
timsifive: If you make `cmd_err` a uint32_t, and call get_field32, then you don't need to do a cast here.

src/target/riscv/riscv-013.c
```diff
@@ -2891,31 +2909,50 @@ static int execute_fence(struct target *target)
...
 		riscv_program_init(&program, target);
-		riscv_program_fence_rw_rw(&program);
 		riscv_program_fence_i(&program);
-		result = riscv_program_exec(&program, target);
-		if (result != ERROR_OK)
-			LOG_TARGET_DEBUG(target, "Unable to execute pre-fence");
+		riscv_program_fence_rw_rw(&program);
+		if (riscv_program_exec(&program, target) != ERROR_OK) {
+			if (program.execution_result != RISCV_DBGBUF_EXEC_RESULT_EXCEPTION) {
+				LOG_TARGET_ERROR(target, "Unexpected error during fence execution");
```
timsifive, line 2932: There should be a mechanism to only print this error (and the similar ones below) once. If a target doesn't implement some kinds of `fence` then it's good to let the user know, but this is not something they can do anything about. Probably should be a warning, not an error, too.

src/target/riscv/riscv-013.c
```diff
@@ -3654,20 +3699,18 @@ static int read_memory_progbuf_inner_startup(struct target *target,
 		goto clear_abstractauto_and_fail;
 
 	uint32_t abstractcs;
-
 	if (wait_for_idle(target, &abstractcs) != ERROR_OK)
 		goto clear_abstractauto_and_fail;
 
-	info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
-	switch (info->cmderr) {
+	cmderr = (unsigned int)get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
```
timsifive: Same with the cast.

src/target/riscv/riscv-013.c
```diff
@@ -3823,20 +3866,17 @@ static int read_memory_progbuf_inner_run_and_process_batch(struct target *target
...
 	uint32_t abstractcs;
-
 	if (wait_for_idle(target, &abstractcs) != ERROR_OK)
 		return ERROR_FAIL;
 
 	uint32_t elements_to_extract_from_batch;
 
-	info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
-	switch (info->cmderr) {
+	unsigned int cmderr = (unsigned int)get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
```
timsifive: Same with the cast.

src/target/riscv/riscv-013.c
```diff
@@ -4558,17 +4600,16 @@ static int write_memory_progbuf_run_batch(struct target *target, struct riscv_ba
 
 	if (wait_for_idle(target, &abstractcs) != ERROR_OK)
 		return ERROR_FAIL;
-	RISCV013_INFO(info);
-	info->cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
 
+	unsigned int cmderr = get_field(abstractcs, DM_ABSTRACTCS_CMDERR);
```
timsifive: I guess the cast is optional?

timsifive (changes requested):

src/target/riscv/program.h
timsifive: Thanks.

src/target/riscv/riscv-013.c
aap-sc: > I guess the cast is optional?

Regarding the cast. There was a discussion in this review to introduce one https://github.com/riscv/riscv-openocd/pull/964#discussion_r1417006866 (since compilers could complain - I tend to agree with this point). Though, likely, it's better to use `get_field32` as you've suggested. Will take a look.

src/target/riscv/riscv-013.c
aap-sc: But in this case the condition is `!= RISCV_DBGBUF_EXEC_RESULT_EXCEPTION` . This should never happen if target behave correctly. Meaning, if target does not implement ifence, it should result in illegal instruction - we just do `LOG_TARGET_DEBUG` in that case . This LOG_TARGET_ERROR is an indication that something is seriously wrong

src/target/riscv/riscv-013.c
aap-sc: here and around . addressed . replaced `unsigned int` to `uint32_t` . Currently this replacement is in a separate changeset (to simplify review). Will squash later

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
timsifive: Oh, I read too quickly and didn't notice `!=`.

timsifive (dismissed):

aap-sc:
Note to reviewers: the latest update is just a squash of `unsigned int ` -> `uint32_t` replacing. No new changes introduced.

JanMatCodasip (dismissed):
Looks good, thank you.

aap-sc:
@timsifive , @JanMatCodasip there was a small merge conflict with recently committed https://github.com/riscv/riscv-openocd/pull/978 (in riscv013_clear_abstract_error) . I've resolved it. Can we merge this one?

aap-sc:
@timsifive, @JanMatCodasip gentle ping. Also @en-sc could you please take a look?

en-sc (approved):
LGTM

JanMatCodasip (approved):
Looks good to me, thank you.

PR #967, opened 2023-11-21
Merge up to 4b1ea8511a7da9d7201df40302e3341c6e97ffdd from upstream

timsifive:
There was a messy conflict in breakpoints.c. @kr-sc, can you take a look? It's a combination of your changes upstream and my changes here that caused the conflict.

As usual, upstream changes have checkpatch issues which we'll ignore.


src/target/breakpoints.c
```diff
@@ -293,36 +293,49 @@ static void breakpoint_free(struct target *data_target, struct target *breakpoin
 	}
 
 	if (!breakpoint)
-		return;
+		return ERROR_OK;
```
timsifive:
```suggestion
		return ERROR_BREAKPOINT_NOT_FOUND;
```

src/target/breakpoints.c
```diff
@@ -568,38 +616,62 @@ static int watchpoint_remove_internal(struct target *target, target_addr_t addre
...
+		if (retval != ERROR_WATCHPOINT_NOT_FOUND) {
+			num_found_watchpoints++;
+
+			if (retval != ERROR_OK)
+				LOG_TARGET_ERROR(target, "failed to remove watchpoint at address" TARGET_ADDR_FMT, address);
+		}
 	}
+
+	if (num_found_watchpoints == 0)
+		LOG_TARGET_ERROR(target, "no watchpoint at address " TARGET_ADDR_FMT " found", address);
```
kr-sc, line 643: We should add `return ERROR_WATCHPOINT_NOT_FOUND;` here because if wp is missing at address we will get `ERROR_OK` (on smp targets). Fixed in OS here: https://review.openocd.org/c/openocd/+/7910

kr-sc:
@timsifive, also, I think that you should also add this one https://review.openocd.org/c/openocd/+/7940.
It fixes endless loop in case of clearing target from wp/bp.

JanMatCodasip:
I have started checking the changes and am halfway through.

I expect to finish it early next week, and will also look at the suggestions posted so far.

src/target/breakpoints.c
timsifive: Good catch. Done.

timsifive:
> @timsifive, also, I think that you should also add this one https://review.openocd.org/c/openocd/+/7940.

I cherry-picked that one.

src/target/breakpoints.c
```diff
@@ -550,11 +576,19 @@ static void watchpoint_free(struct target *target, struct watchpoint *watchpoint
 	}
 
 	if (!watchpoint)
-		return;
+		return ERROR_WATCHPOINT_NOT_FOUND;
```
JanMatCodasip, line 579: I believe this `if` is redundant - the condition can't ever be true. That's because both the callers of `watchpoint_free()` already ensure that the argument is a valid watchpoint of the given target. 

An assert should be used here instead.

I will submit a change to the vanilla upstream by the end of the week. No need to change anything in this merge request.

src/target/breakpoints.c
```diff
@@ -293,36 +293,51 @@ static void breakpoint_free(struct target *data_target, struct target *breakpoin
 	}
 
 	if (!breakpoint)
-		return;
+		return ERROR_BREAKPOINT_NOT_FOUND;
```
JanMatCodasip, line 296: This is the same case as in watchpoint_free(). The condition of the `if` can't be true, and so the `if` should be replaced by an assert. I will submit a short patch to the upstream shortly. No need to change anything in this merge request.

src/target/breakpoints.c
JanMatCodasip: This `if` block is dead code; discussed here: https://github.com/riscv/riscv-openocd/pull/967#discussion_r1408102525

JanMatCodasip (approved):
I have finished my (visual) review and all looks well, including breakpoints.c after cherry-picking the two upstream changes suggested by @kr-sc (thanks).

The two cosmetic suggestion (removal of pointless `if`) block I will submit to openocd.org separately.

Thank you.

PR #971, opened 2023-12-01
break from long loops on shutdown request

en-sc:
In loops that typically take longer time to complete, check if there is a pending shutdown request. If so, terminate the loop.  This allows to respond to a signal requesting a shutdown during some loops which do not return control to main OpenOCD loop.

### DIFF
--- a/src/target/riscv/program.c
+++ b/src/target/riscv/program.c
@@ -47,6 +47,9 @@ int riscv_program_exec(struct riscv_program *p, struct target *t)
 {
 	keep_alive();
 
+	if (openocd_is_shutdown_pending())
+		return ERROR_SERVER_INTERRUPTED;
+
 	p->execution_result = RISCV_PROGBUF_EXEC_RESULT_UNKNOWN;
 	riscv_reg_t saved_registers[GDB_REGNO_XPR31 + 1];
 	for (size_t i = GDB_REGNO_ZERO + 1; i <= GDB_REGNO_XPR31; ++i) {
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -595,6 +595,9 @@ static int dmi_op_timeout(struct target *target, uint32_t *data_in,
 
 	keep_alive();
 
+	if (openocd_is_shutdown_pending())
+		return ERROR_SERVER_INTERRUPTED;
+
 	time_t start = time(NULL);
 	/* This first loop performs the request.  Note that if for some reason this
 	 * stays busy, it is actually due to the previous access. */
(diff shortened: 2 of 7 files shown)


en-sc:
Related patch to mainline OpenOCD: https://review.openocd.org/c/openocd/+/8032

JanMatCodasip (changes requested):
Thank you for this improvement.

As already noted in the upstream review (https://review.openocd.org/c/openocd/+/8032):

- Please create a new error code to denote interrupted operation, e.g. ERROR_INTERRUPTED or ERROR_SHUTDOWN
- Is the new function `openocd_shutdown_by_signal()` needed? If so, please give it a better name.

src/target/riscv/program.c
```diff
@@ -45,6 +45,8 @@ int riscv_program_write(struct riscv_program *program)
 int riscv_program_exec(struct riscv_program *p, struct target *t)
 {
 	keep_alive();
+	if (openocd_is_shutdown_pending())
+		return ERROR_SERVER_REMOTE_CLOSED;
```
JanMatCodasip: In total, I can see three cases in this merge request where the error code is different than `ERROR_SERVER_INTERRUPTED`. Are these intentional or just typos?

```suggestion
	if (openocd_is_shutdown_pending())
		return ERROR_SERVER_INTERRUPTED;
```

src/target/riscv/riscv-013.c
```diff
@@ -3193,6 +3195,9 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
 						return ERROR_FAIL;
 					}
 					keep_alive();
+					if (openocd_is_shutdown_pending())
+						return ERROR_FAIL;
```
JanMatCodasip: This is the 2nd case.

src/target/riscv/riscv.c
```diff
@@ -5237,6 +5239,8 @@ int riscv_get_register(struct target *target, riscv_reg_t *value,
 	assert(r->get_register);
 
 	keep_alive();
+	if (openocd_is_shutdown_pending())
+		return ERROR_FAIL;
```
JanMatCodasip: 3rd case.

src/target/riscv/program.c
en-sc: These are just typos. Thanks for the catch! Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv.c

en-sc:
@JanMatCodasip, can you please take a look at this one. The only difference with recently merged to mainline https://review.openocd.org/gitweb?p=openocd.git;a=commitdiff;h=2e920a212fbe2de705811d547c169c1ae1611a02 is RISC-V specific.

timsifive (dismissed):

JanMatCodasip (dismissed):
Looks good, thank you.

JanMatCodasip:
@en-sc There is now a little conflict in `riscv_program_exec()` after merging another merge request. Could you please resolve it. Thank you.

en-sc:
@JanMatCodasip, I've resolved the merge conflict you mentioned. Please, take a look.

JanMatCodasip (approved):

JanMatCodasip:
> I've resolved the merge conflict you mentioned.

Thank you. If nobody comes with any other feedback, let's merge tomorrow.

Generally, I would recommend to have a cool-down period 1-2 days for approved merge requests prior to the actual merge -- to give anyone from the community some time window to post additional comments, if desired so.

PR #972, opened 2023-12-01
target/riscv: report helpfull location during register decode

en-sc:
`LOG_TARGET_DEBUG()` reports file, line and function name at the call
site. This information is not helpfull if it always points to the same
location inside `log_debug_reg()`.

Additionally, avoid using VLA in `log_debug_reg()`, since OpenOCD style guide(`doc/manual/style.txt`) prohibits it:

> - use malloc() to create dynamic arrays. Do @b not use @c alloca
> or variable length arrays on the stack. non-MMU hosts(uClinux) and
> pthreads require modest and predictable stack usage.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -302,9 +302,14 @@ static void log_debug_reg(struct target *target, enum riscv_debug_reg_ordinal re
 	if (debug_level < LOG_LVL_DEBUG)
 		return;
 	const riscv_debug_reg_ctx_t context = get_riscv_debug_reg_ctx(target);
-	char buf[riscv_debug_reg_to_s(NULL, reg, context, value) + 1];
+	char * const buf = malloc(riscv_debug_reg_to_s(NULL, reg, context, value) + 1);
+	if (!buf) {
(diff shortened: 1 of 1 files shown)


timsifive (approved):
Looks good. Can you share a sample output line emitted by the new code?

en-sc:
Prior to the change:
```
Debug: 82 2 riscv-013.c:307 log_debug_reg(): [riscv.cpu] dtmcs=0x61 { version=1.0, dmistat=0, idle=0, dmireset=0, dtmhardreset=0, errinfo=not implemented, abits=6, }
```
After:
```
Debug: 82 3 riscv-013.c:1870 examine(): [riscv.cpu] dtmcs=0x61 { version=1.0, dmistat=0, idle=0, dmireset=0, dtmhardreset=0, errinfo=not implemented, abits=6, }
```
This PR does not update `debug_reg_printer.[ch]`, so the actual decoding string is flawed. Please, see [this comment](https://github.com/riscv/riscv-openocd/pull/973#issuecomment-1842467719).

JanMatCodasip (approved):

PR #973, opened 2023-12-04
Update riscv/debug_defines (to sync with riscv-debug-spec commit 40b9a05)

kr-sc:
Change-Id: Ie969866d1de83360a5f45e96e22108b58b8aa02f


timsifive (dismissed):
Looks OK. Just to check, can you show some debug output that uses the new code?

en-sc:
I would suggest updating to a more recent version to include these changes: https://github.com/riscv/riscv-debug-spec/pull/908, https://github.com/riscv/riscv-debug-spec/pull/909. Please, note that `debug_reg_printer.[ch]` should also be updated if the changes are included.

kr-sc:
> Looks OK. Just to check, can you show some debug output that uses the new code?

Before:
```
log_debug_reg(): [riscv.cpu] dtmcs=0x61 { version=1.0, dmistat=0, idle=0, dmireset=0, dtmhardreset=0, errinfo=not implemented, abits=6, }
```

After:
```
log_debug_reg(): [riscv.cpu] dtmcs=0x61 {version=1_0 errinfo=not_implemented abits=6}
```

timsifive:
I made https://github.com/riscv/riscv-debug-spec/pull/922 to address the checkpatch issue. When that merges, please update this PR.

timsifive (approved):

src/target/riscv/debug_reg_printer.h
```diff
@@ -2,6 +2,12 @@
 
 #include "debug_defines.h"
 
+enum riscv_debug_reg_show {
```
JanMatCodasip, line 5: Please, could you give a descriptive name to this enum, e.g. `display_mode` (or similar), rather than `show`.

```suggestion
enum riscv_debug_reg_display_mode {
```

src/target/riscv/debug_reg_printer.c
```diff
@@ -62,22 +62,35 @@ static uint64_t riscv_debug_reg_field_value(riscv_debug_reg_field_info_t field,
 
 static unsigned int riscv_debug_reg_fields_to_s(char *buf, unsigned int offset,
 	struct riscv_debug_reg_field_list_t (*get_next)(riscv_debug_reg_ctx_t contex),
-	riscv_debug_reg_ctx_t context, uint64_t value)
+	riscv_debug_reg_ctx_t context, uint64_t value,
+	enum riscv_debug_reg_show show)
```
JanMatCodasip, line 66:
```suggestion
	enum riscv_debug_reg_display_mode display_mode)
```

src/target/riscv/debug_reg_printer.c
```diff
@@ -62,22 +62,35 @@ static uint64_t riscv_debug_reg_field_value(riscv_debug_reg_field_info_t field,
...
+		if (show == RISCV_DEBUG_REG_SHOW_ALL ||
+				(show == RISCV_DEBUG_REG_HIDE_UNNAMED_0 &&
+					(field_value != 0 ||
+						(list.field.values && list.field.values[0]))) ||
+				(show == RISCV_DEBUG_REG_HIDE_ALL_0 && field_value != 0)) {
+			curr += get_len_or_sprintf(buf, curr, separator);
+			curr += riscv_debug_reg_field_to_s(buf, curr, list.field, context,
+							field_value);
+			separator = " ";
+		}
```
JanMatCodasip, line 85: The condition of the "if" is complex, which makes it hard to understand and modify. Please, could you simplify this whole part of the code - for example as shown below.

```suggestion

		uint64_t field_value = riscv_debug_reg_field_value(list.field, value);

		const bool isZero = (field_value == 0);
		const bool isUnnamed = (!list.field.values) || (!list.field.values[0]);		
		
		if (display_mode == RISCV_DEBUG_REG_HIDE_UNNAMED_0 && isUnnamed && isZero)
			// Don't display this field
			continue;
			
		if (display_mode == RISCV_DEBUG_REG_HIDE_ALL_0 && isZero)
			// Don't display this field
			continue;			
		
		// Show this field
		curr += get_len_or_sprintf(buf, curr, separator);
		curr += riscv_debug_reg_field_to_s(buf, curr, list.field, context,
							field_value);
		separator = " ";		
		
```

JanMatCodasip (changes requested):
I have few suggestions related to code structure or naming. Otherwise this looks good.

kr-sc:
@JanMatCodasip I think that we need patch to debus spec with your suggestions. What if we merge this patch and make a subsequent update with your suggestions to debug spec and openocd later?

JanMatCodasip:
> kr-sc: I think that we need patch to debug spec with your suggestions.

Ah, when looking at the changes, I missed the fact that the files `debug_reg_printer.*` are auto-generated from the debug spec repository. Thank you for the reminder.

> kr-sc: What if we merge this patch and make a subsequent update with your suggestions to debug spec and openocd later?

Sure, I agree, let's merge this now as is.

If you are happy with the suggested changes, I believe it would be good to apply them to the riscv-debug-spec repo.

JanMatCodasip (approved):

src/target/riscv/debug_reg_printer.h
JanMatCodasip: Correction per [kr-sc's comment](https://github.com/riscv/riscv-openocd/pull/973#issuecomment-1849825669): This change should go to the debug spec repository instead (riscv-debug-spec).

src/target/riscv/debug_reg_printer.c
JanMatCodasip: Same as above: This change should go to the debug spec repository instead (riscv-debug-spec).

src/target/riscv/debug_reg_printer.c
JanMatCodasip: Same as above: This change should go to the debug spec repository instead (riscv-debug-spec).

PR #978, opened 2023-12-13
Clean up clang static analyzer complaints.

timsifive:
I don't think there are any real bugs here, but at least this gives us a clean slate moving forward.

Change-Id: I29c6c398c28dfe580f9a2deb3bdbcfc491a2ceb6

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -425,6 +425,8 @@ static dbus_status_t dbus_scan(struct target *target, uint16_t *address_in,
 		.out_value = out,
 		.in_value = in
 	};
+	if (address_in)
+		*address_in = 0;
 
 	if (info->addrbits == 0) {
 		LOG_TARGET_ERROR(target, "Can't access DMI because addrbits=0.");
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -34,7 +34,7 @@
 static int riscv013_on_step_or_resume(struct target *target, bool step);
 static int riscv013_step_or_resume_current_hart(struct target *target,
 		bool step);
-static void riscv013_clear_abstract_error(struct target *target);
+static int riscv013_clear_abstract_error(struct target *target);
 
 /* Implementations of the functions in struct riscv_info. */
 static int riscv013_get_register(struct target *target,
@@ -907,12 +907,12 @@ static riscv_reg_t read_abstract_arg(struct target *target, unsigned index,
 			LOG_TARGET_ERROR(target, "Unsupported size: %d bits", size_bits);
 			return ~0;
 		case 64:
-			dm_read(target, &v, DM_DATA0 + offset + 1);
-			value |= ((uint64_t) v) << 32;
+			if (dm_read(target, &v, DM_DATA0 + offset + 1) == ERROR_OK)
+				value |= ((uint64_t)v) << 32;
 			/* falls through */
 		case 32:
-			dm_read(target, &v, DM_DATA0 + offset);
-			value |= v;
+			if (dm_read(target, &v, DM_DATA0 + offset) == ERROR_OK)
+				value |= v;
 	}
 	return value;
 }
@@ -1181,9 +1181,14 @@ static int prep_for_register_access(struct target *target,
 {
 	assert(orig_mstatus);
 
-	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
+	if (!is_fpu_reg(regno) && !is_vector_reg(regno)) {
+		/* If we don't assign orig_mstatus, clang static analysis
+		 * complains when this value is passed to
+		 * cleanup_after_register_access(). */
+		*orig_mstatus = 0;
 		/* No special preparation needed */
 		return ERROR_OK;
+	}
 
 	LOG_TARGET_DEBUG(target, "Preparing mstatus to access %s",
 			gdb_regno_name(target, regno));
@@ -2999,12 +3004,12 @@ static target_addr_t sb_read_address(struct target *target)
 	target_addr_t address = 0;
 	uint32_t v;
 	if (sbasize > 32) {
-		dm_read(target, &v, DM_SBADDRESS1);
-		address |= v;
+		if (dm_read(target, &v, DM_SBADDRESS1) == ERROR_OK)
+			address |= v;
 		address <<= 32;
 	}
-	dm_read(target, &v, DM_SBADDRESS0);
-	address |= v;
+	if (dm_read(target, &v, DM_SBADDRESS0) == ERROR_OK)
+		address |= v;
 	return address;
 }
 
@@ -5076,8 +5081,10 @@ static int riscv013_write_debug_buffer(struct target *target, unsigned int index
 static riscv_insn_t riscv013_read_debug_buffer(struct target *target, unsigned int index)
 {
 	uint32_t value;
-	dm_read(target, &value, DM_PROGBUF0 + index);
-	return value;
+	if (dm_read(target, &value, DM_PROGBUF0 + index) == ERROR_OK)
+		return value;
+	else
+		return 0;
 }
 
 static int riscv013_invalidate_cached_debug_buffer(struct target *target)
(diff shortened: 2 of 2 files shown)


src/target/riscv/riscv-011.c
```diff
@@ -425,6 +425,8 @@ static dbus_status_t dbus_scan(struct target *target, uint16_t *address_in,
 		.out_value = out,
 		.in_value = in
 	};
+	if (address_in)
+		*address_in = 0;
```
en-sc, line 429: This seems a bit strange. After the change, `address_in` and `data_in` are handled differently. What is the issue you are fixing here? I would suggest adjusting the call-site to eliminate the warning, if possible.

src/target/riscv/riscv-013.c
```diff
@@ -907,12 +907,12 @@ static riscv_reg_t read_abstract_arg(struct target *target, unsigned index,
...
-			dm_read(target, &v, DM_DATA0 + offset + 1);
-			value |= ((uint64_t) v) << 32;
+			if (dm_read(target, &v, DM_DATA0 + offset + 1) == ERROR_OK)
+				value |= ((uint64_t)v) << 32;
 			/* falls through */
 		case 32:
-			dm_read(target, &v, DM_DATA0 + offset);
-			value |= v;
+			if (dm_read(target, &v, DM_DATA0 + offset) == ERROR_OK)
+				value |= v;
```
en-sc, line 915: I don't think this is the way to fix the issue. It will not be possible to distinguish between reading a zero and reading nothing due to an error. I would suggest changing the signature of the function:
```
static int read_abstract_arg(..., riscv_reg_t *value)
```
Although this would create a much bigger diff -- a separate patch perhaps would be a better solution.

If you wish to leave this as-is for now, please, add a `FIXME` comment.

src/target/riscv/riscv-013.c
```diff
@@ -1181,9 +1181,11 @@ static int prep_for_register_access(struct target *target,
 {
 	assert(orig_mstatus);
 
-	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
+	if (!is_fpu_reg(regno) && !is_vector_reg(regno)) {
+		*orig_mstatus = 0;
```
en-sc, line 1188: Please, add a comment, explaining that this assignment is needed to silence the static analyzer warning. Also, if you like, I would suggest to move the initialization to the point of declaration. IMHO, this would make it clearer (if not FPU nor vector register, do nothing in `prep_...()` and in `cleanup_...()`)

src/target/riscv/riscv-013.c
```diff
@@ -2999,12 +3001,12 @@ static target_addr_t sb_read_address(struct target *target)
...
-		dm_read(target, &v, DM_SBADDRESS1);
-		address |= v;
+		if (dm_read(target, &v, DM_SBADDRESS1) == ERROR_OK)
+			address |= v;
 		address <<= 32;
 	}
-	dm_read(target, &v, DM_SBADDRESS0);
-	address |= v;
+	if (dm_read(target, &v, DM_SBADDRESS0) == ERROR_OK)
+		address |= v;
```
en-sc, line 3012: I have the same concern as with `read_abstract_arg()`.

src/target/riscv/riscv-013.c
```diff
@@ -5076,8 +5078,10 @@ static int riscv013_write_debug_buffer(struct target *target, unsigned int index
 static riscv_insn_t riscv013_read_debug_buffer(struct target *target, unsigned int index)
 {
 	uint32_t value;
-	dm_read(target, &value, DM_PROGBUF0 + index);
-	return value;
+	if (dm_read(target, &value, DM_PROGBUF0 + index) == ERROR_OK)
+		return value;
+	else
+		return 0;
```
en-sc, line 5087: I have the same concern as with `read_abstract_arg()`. Also, wouldn't initializing the `value` to zero fix the warning?

src/target/riscv/riscv-013.c
```diff
@@ -5248,5 +5254,7 @@ static void riscv013_clear_abstract_error(struct target *target)
 		}
 	}
 	/* Clear the error status. */
-	dm_write(target, DM_ABSTRACTCS, DM_ABSTRACTCS_CMDERR);
+	if (dm_write(target, DM_ABSTRACTCS, DM_ABSTRACTCS_CMDERR) != ERROR_OK)
+		return ERROR_FAIL;
+	return ERROR_OK;
```
en-sc: Consider the following:
* The loop exits due to timeout.
* `dm_write` succeds.
* `ERROR_OK` is returned, although `abstractcs.cmderr` may be still set.

I would recommend replacing the loop with a call to `wait_for_idle`, as suggested [here](https://github.com/riscv/riscv-openocd/pull/964/files#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63R5277), like so:
```
int res = wait_for_idle(...);
if (dm_write(target, DM_ABSTRACTCS, DM_ABSTRACTCS_CMDERR) != ERROR_OK)
  res = ERROR_FAIL;
return res;
```

en-sc (commented):
Thank you for the patch! The cleanup is great, though I have a couple concerns.

en-sc:
Also, can you please tell how did you run the analyzer? Is it possible to integrate the check into CI?

timsifive:
> Also, can you please tell how did you run the analyzer? Is it possible to integrate the check into CI?

The HACKING file describes how to run the analyzer. I assume it's possible to integrate the check, but I'm not very familiar with these tools.

src/target/riscv/riscv-011.c
timsifive: There's a caller which prints out address_in in an error message, and the compiler complains that the value is uninitialized when printed. Since this is riscv-011 which no one is testing, I prefer to keep the change as small as possible.

src/target/riscv/riscv-013.c
timsifive: I agree it would be better to pass up the error code, but I'm just trying to fix the compiler warnings, not everything that is wrong with the code. I filed #981 to track this.

src/target/riscv/riscv-013.c
timsifive: Done.

src/target/riscv/riscv-013.c
timsifive: Added to the same issue.

src/target/riscv/riscv-013.c
timsifive: Yes, initializing to 0 would fix the warning, but I think this makes it more clear what is going on. It also highlights that we're ignoring the error better.

src/target/riscv/riscv-013.c
timsifive: Done. Thanks.

src/target/riscv/riscv-013.c
```diff
@@ -1181,9 +1181,14 @@ static int prep_for_register_access(struct target *target,
 {
 	assert(orig_mstatus);
 
-	if (!is_fpu_reg(regno) && !is_vector_reg(regno))
+	if (!is_fpu_reg(regno) && !is_vector_reg(regno)) {
+		/* If we don't assign orig_mstatus, clang static analyses
```
en-sc, line 1185:
```suggestion
		/* If we don't assign orig_mstatus, clang static analysis
```

```suggestion
		/* If we don't assign orig_mstatus, clang static analysis
```

en-sc (dismissed):
LGTM, aside from one typo.

src/target/riscv/riscv-013.c
timsifive: Fixed. Thanks.

en-sc (approved):

src/target/riscv/riscv-013.c
JanMatCodasip: I also agree with @en-sc that this should (eventually) be fixed differently - by passing the error code up.

Thank you for filling the #981 to address this.

JanMatCodasip (approved):
@timsifive Thank you for the patch. This is a nice step forward but eventually we need to continue by what you mentioned in #981. 

Curious, how do you launch the clang static analyzer (command-line vs. IDE integration) and what version have you used?

timsifive:
> Curious, how do you launch the clang static analyzer (command-line vs. IDE integration) and what version have you used?

I followed these instructions: https://github.com/riscv/riscv-openocd/blob/fb28d7e6fdf3e3b4a44ab11a736c3451bd709249/HACKING#L51

Using clang version 16.0.6

PR #989, opened 2024-01-10
Merge up to adcc8ef87bc1ed47c95f1f2d23072b2b916e1555 from upstream.

en-sc:


en-sc:
`checkpatch` fails due to macro argument reuse:
```
CHECK:MACRO_ARG_REUSE: Macro argument reuse 'reg' - possible side-effects?
#3675: FILE: src/target/mips32.h:692:
+#define MIPS32_JRHB(isa, reg)		(isa ? MMIPS32_JRHB(reg) : MIPS32_ISA_JRHB(reg))
```
This is actually a bug in `checkpatch`. Ternary operator evaluates only one of the arguments (after evaluating the condition), so the `reg` argument is used once (https://en.cppreference.com/w/c/language/operator_other#Conditional_operator).

MarekVCodasip (approved):
Looks good to me aside from the remote-bitbang protocol. It seems that there is a conflict between SWD write 0 0 and sleep for 1 microsecond. I already sent an email to the upstream mailing list about it.

It should not affect us as RISC-V does not support SWD.

JanMatCodasip (commented):
@MarekVCodasip, thank you for the review.

@en-sc, I'd recommend to merge it on Tuesday if no one else requests any further changes by then (so that anyone who might be interested to take a look still has some time to do it).

PR #991, opened 2024-01-10
target/riscv: fix addressing in `dm_read`/`dm_write`

en-sc:
There was an error in `dm_read`/`dm_write`: DMI address was checked against DM registers disregarding DM base address.

To solve the issue `dmi_address()` function was introduced.

Change-Id: Ia3be619417b5f5b53db5dfe302db05170d6787c9

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -691,6 +691,16 @@ static int dmi_write_exec(struct target *target, uint32_t address,
 	return dmi_op(target, NULL, NULL, DMI_OP_WRITE, address, value, true, ensure_success);
 }
 
+static uint32_t riscv013_get_dmi_address(const struct target *target, uint32_t address)
+{
+	assert(target);
+	uint32_t base = 0;
+	RISCV013_INFO(info);
+	if (info && info->dm)
+		base = info->dm->base;
+	return address + base;
+}
+
 static int dm_op_timeout(struct target *target, uint32_t *data_in,
 		bool *dmi_busy_encountered, int op, uint32_t address,
 		uint32_t data_out, int timeout_sec, bool exec, bool ensure_success)
@@ -2769,8 +2779,7 @@ static int init_target(struct command_context *cmd_ctx,
 	generic_info->authdata_write = &riscv013_authdata_write;
 	generic_info->dmi_read = &dmi_read;
 	generic_info->dmi_write = &dmi_write;
-	generic_info->dm_read = &dm_read;
-	generic_info->dm_write = &dm_write;
+	generic_info->get_dmi_address = &riscv013_get_dmi_address;
 	generic_info->read_memory = read_memory;
 	generic_info->hart_count = &riscv013_hart_count;
 	generic_info->data_bits = &riscv013_data_bits;
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3690,143 +3690,138 @@ COMMAND_HANDLER(riscv_authdata_write)
 	return r->authdata_write(target, value, index);
 }
 
-COMMAND_HANDLER(riscv_dmi_read)
+static uint32_t riscv_get_dmi_address(const struct target *target, uint32_t dm_address)
 {
-	if (CMD_ARGC != 1) {
-		LOG_ERROR("Command takes 1 parameter");
-		return ERROR_COMMAND_SYNTAX_ERROR;
-	}
+	assert(target);
+	RISCV_INFO(r);
+	if (!r || !r->get_dmi_address)
+		return dm_address;
+	return r->get_dmi_address(target, dm_address);
+}
 
-	struct target *target = get_current_target(CMD_CTX);
+static int riscv_dmi_read(struct target *target, uint32_t *value, uint32_t address)
+{
 	if (!target) {
 		LOG_ERROR("target is NULL!");
 		return ERROR_FAIL;
 	}
-
 	RISCV_INFO(r);
 	if (!r) {
 		LOG_TARGET_ERROR(target, "riscv_info is NULL!");
 		return ERROR_FAIL;
 	}
-
-	if (r->dmi_read) {
-		uint32_t address, value;
-		COMMAND_PARSE_NUMBER(u32, CMD_ARGV[0], address);
-		if (r->dmi_read(target, &value, address) != ERROR_OK)
-			return ERROR_FAIL;
-		command_print(CMD, "0x%" PRIx32, value);
-	} else {
-		LOG_TARGET_ERROR(target, "dmi_read is not implemented for this target.");
+	if (!r->dmi_read) {
+		LOG_TARGET_ERROR(target, "dmi_read is not implemented.");
 		return ERROR_FAIL;
 	}
-	return ERROR_OK;
+	return r->dmi_read(target, value, address);
 }
 
-COMMAND_HANDLER(riscv_dmi_write)
+static int riscv_dmi_write(struct target *target, uint32_t dmi_address, uint32_t value)
 {
-	if (CMD_ARGC != 2) {
-		LOG_ERROR("Command takes exactly 2 arguments");
-		return ERROR_COMMAND_SYNTAX_ERROR;
-	}
-
-	struct target *target = get_current_target(CMD_CTX);
-	RISCV_INFO(r);
-
-	uint32_t address, value;
-	COMMAND_PARSE_NUMBER(u32, CMD_ARGV[0], address);
-	COMMAND_PARSE_NUMBER(u32, CMD_ARGV[1], value);
-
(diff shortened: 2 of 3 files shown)


src/target/riscv/riscv.h
```diff
@@ -242,8 +242,11 @@ struct riscv_info {
 	int (*dmi_read)(struct target *target, uint32_t *value, uint32_t address);
 	int (*dmi_write)(struct target *target, uint32_t address, uint32_t value);
 
-	int (*dm_read)(struct target *target, uint32_t *value, uint32_t address);
-	int (*dm_write)(struct target *target, uint32_t address, uint32_t value);
+	/* Get the DMI address of target's DM's register.
+	 * The function should return the passed address
+	 * if the target is not assigned a DM yet.
+	 */
+	uint32_t (*dmi_address)(const struct target *target, uint32_t address);
```
JanMatCodasip: Optional details:

This could be called `get_dmi_address`.

The second argument could be called `dm_address`.

```suggestion
	uint32_t (*get_dmi_address)(const struct target *target, uint32_t dm_address);
```

src/target/riscv/riscv.c
```diff
@@ -4653,28 +4645,28 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 	},
 	{
 		.name = "dmi_read",
-		.handler = riscv_dmi_read,
+		.handler = handle_riscv_dmi_read,
 		.mode = COMMAND_ANY,
 		.usage = "address",
 		.help = "Perform a 32-bit DMI read at address, returning the value."
```
JanMatCodasip: Optional detail:

While making changes here, you can also improve the help messages, if you like.

```suggestion
		.help = "Read and return 32-bit value from the given address on the RISC-V DMI bus."
```

src/target/riscv/riscv.c
```diff
@@ -4653,28 +4645,28 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
 	},
 	{
 		.name = "dmi_write",
-		.handler = riscv_dmi_write,
+		.handler = handle_riscv_dmi_write,
 		.mode = COMMAND_ANY,
 		.usage = "address value",
 		.help = "Perform a 32-bit DMI write of value at address."
```
```suggestion
		.help = "Write a 32-bit value to the given address on the RISC-V DMI bus."
```

src/target/riscv/riscv.c
```diff
@@ -4653,28 +4645,28 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
 	},
 	{
 		.name = "dm_read",
-		.handler = riscv_dm_read,
+		.handler = handle_riscv_dm_read,
 		.mode = COMMAND_ANY,
 		.usage = "reg_address",
 		.help = "Perform a 32-bit read from DM register at reg_address, returning the value."
```
```suggestion
		.help = "Read and return 32-bit value from a debug module's register at reg_address."
```

src/target/riscv/riscv.c
```diff
@@ -4653,28 +4645,28 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
 	},
 	{
 		.name = "dm_write",
-		.handler = riscv_dm_write,
+		.handler = handle_riscv_dm_write,
 		.mode = COMMAND_ANY,
 		.usage = "reg_address value",
 		.help = "Write a 32-bit value to the DM register at reg_address."
```
```suggestion
		.help = "Write a 32-bit value to the debug module's register at reg_address."
```

src/target/riscv/riscv-013.c
```diff
@@ -691,6 +691,16 @@ static int dmi_write_exec(struct target *target, uint32_t address,
 	return dmi_op(target, NULL, NULL, DMI_OP_WRITE, address, value, true, ensure_success);
 }
 
+static uint32_t riscv_013_dmi_address(const struct target *target, uint32_t address)
```

```suggestion
static uint32_t riscv_013_get_dmi_address(const struct target *target, uint32_t dm_address)
```

src/target/riscv/riscv.c
```diff
@@ -3690,143 +3690,135 @@ COMMAND_HANDLER(riscv_authdata_write)
...
+	if (progbuf_touched || dm_deactivated) {
+		if (r->invalidate_cached_progbuf) {
+			/* Here the return value of invalidate_cached_progbuf()
+			 * is ignored. It is okay to do so for now, since the
+			 * only case an error is returned is a failure to
+			 * assign a DM to the target, which would have already
+			 * caused an error during dmi_write().
+			 * FIXME: invalidate_cached_progbuf() should be void.
+			 */
+			r->invalidate_cached_progbuf(target);
```
JanMatCodasip: Fixing of this part was the main motivation for this pull request, am I right?

Side comment: I have been considering dropping the progbuf cache completely in future - for reliability and robustness. It could be replaced by populating the program buffer by a single batch of DMI transactions, leading to high performance and no issues with maintaining the cache (and invalidating it).

src/target/riscv/riscv.h
```diff
@@ -242,8 +242,11 @@ struct riscv_info {
 	int (*dmi_read)(struct target *target, uint32_t *value, uint32_t address);
 	int (*dmi_write)(struct target *target, uint32_t address, uint32_t value);
 
-	int (*dm_read)(struct target *target, uint32_t *value, uint32_t address);
-	int (*dm_write)(struct target *target, uint32_t address, uint32_t value);
+	/* Get the DMI address of target's DM's register.
+	 * The function should return the passed address
+	 * if the target is not assigned a DM yet.
```
JanMatCodasip, line 247: I am a bit uncertain about this part. I'd prefer to raise an error - to make sure the function cannot be used until the target is assigned a DM.

But this can be handled later (if and when).

JanMatCodasip (dismissed):
LGTM, thank you for the fix. 

I only have few comments related to details.

src/target/riscv/riscv.h
en-sc: I agree with you. This change is a bit questionable.
The reason for the current solution is `get_dm()` function. Currently `get_dm()` function:
* Returns `NULL` if a DM was not allocated/selected prior and allocation failed (out of memory).
* Returns non-examined DM (the corresponding structure is just initialized, but is yet to be filled with correct values) if none of the targets, corresponding to this DM were examined.
* Returns a properly examined DM otherwise.

The second case is quite worrying. I would like to make this function a pure accessor and search for DM / allocate it explicitly in the `examine()`. I'm preparing the corresponding patch. This would make it easier to reason about current state of a DM and allow to replace this weird behavior of `get_dmi_address()` with an assertion.

src/target/riscv/riscv.h

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv.c
en-sc: Yes. You are right.

> I have been considering dropping the progbuf cache completely in future

I have conflicted feelings about your suggestion. Using batched operations will significantly speed-up writing program buffer, though still there is a benefit in implementing program buffer cache.
Consider reading a bunch of FPRs/CSRs via program buffer. The first instruction changes, but the second one is always `ebreak`. Program buffer cache would save a scan on each read even in the case of batched operations.

src/target/riscv/riscv.h
JanMatCodasip: > I would like to make this function a pure accessor and search for DM / allocate it explicitly in the examine(). I'm preparing the corresponding patch.

That sounds like good solution.

src/target/riscv/riscv.c
JanMatCodasip: >  [...] there is a benefit in implementing program buffer cache. Consider reading a bunch of FPRs/CSRs via program buffer. The first instruction changes, but the second one is always ebreak. Program buffer cache would save a scan on each read even in the case of batched operations

I assume that when the batches are used, the performance benefit would be small (and will later collect measurements to confirm).

I'd prefer robustness and reliability (even at minor performance cost). Progbuf cache in my opinion works against the reliability. 

For example:
- Program buffer may be (per the RISC-V spec) memory-mapped, and the contents of the memory can change by a program writing to it (accidentally or on purpose).
- Program buffer contents may change by direct DMI access, and our code only catches a subset of the cases.
- Program buffer contents might change by other DMI bus master accessing the registers of the debug module, without OpenOCD knowing.

In other words, I believe caches are good when they are managed reliably (flushed and invalidated as needed). When there are known gaps, I'd prefer to not use them at all.

JanMatCodasip: In any case, this was only a side-discussion. No need to change anything in this merge request. :-)

Marking the thread as resolved.

JanMatCodasip (approved):

PR #992, opened 2024-01-10
target/riscv: remove `riscv_hart_count()`

en-sc:
The motivalion for the change:
* `riscv_hart_count()` is used only once to print the value into the log
  during exmination.
* The returned value is a bit confusing: it's not the total number of
targets on the TAP. It is the number of targets accessable through the
same DM. So the name of the function is misleading.
* This value is already reported on `-d3` level.

So the function seems redundant and can be safely removed.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2150,8 +2150,7 @@ static int examine(struct target *target)
 	/* Some regression suites rely on seeing 'Examined RISC-V core' to know
 	 * when they can connect with gdb/telnet.
 	 * We will need to update those suites if we want to change that text. */
-	LOG_TARGET_INFO(target, "Examined RISC-V core; found %d harts",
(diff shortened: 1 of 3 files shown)


JanMatCodasip (approved):
This looks all right, thank you for the cleanup.

Let's merge Thursday morning (unless there are more comments).

erhankur:
I am cherry-picking commits periodically to our fork. It would be helpful to have a brief explanation in all commits about the reason for the change.

en-sc:
> I am cherry-picking commits periodically to our fork. It would be helpful to have a brief explanation in all commits about the reason for the change.

Addressed. @erhankur, please, take a look.

erhankur:
> > I am cherry-picking commits periodically to our fork. It would be helpful to have a brief explanation in all commits about the reason for the change.
> 
> Addressed. @erhankur, please, take a look.

Thank you @en-sc

PR #995, opened 2024-01-16
target/riscv: cleanup `get_riscv_debug_reg_ctx()`

en-sc:
This commit makes the function safe to use throughout the lifetime of a target.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -280,13 +280,17 @@ static dm013_info_t *get_dm(struct target *target)
 	return dm;
 }
 
-static riscv_debug_reg_ctx_t get_riscv_debug_reg_ctx(struct target *target)
+static riscv_debug_reg_ctx_t get_riscv_debug_reg_ctx(const struct target *target)
 {
-	RISCV_INFO(r);
+	if (!target_was_examined(target)) {
+		const riscv_debug_reg_ctx_t default_context = {0};
+		return default_context;
+	}
+
 	RISCV013_INFO(info);
 	const riscv_debug_reg_ctx_t context = {
-		.XLEN = { .value = r->xlen, .is_set = true },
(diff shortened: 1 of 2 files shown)


en-sc:
Note: the change to `target/target.c` is a part of https://review.openocd.org/c/openocd/+/8092

src/target/riscv/riscv-013.c
```diff
@@ -280,13 +280,17 @@ static dm013_info_t *get_dm(struct target *target)
 	return dm;
 }
 
-static riscv_debug_reg_ctx_t get_riscv_debug_reg_ctx(struct target *target)
+static riscv_debug_reg_ctx_t get_riscv_debug_reg_ctx(const struct target *target)
 {
-	RISCV_INFO(r);
+	if (!target_was_examined(target)) {
+		const riscv_debug_reg_ctx_t default_context = {0};
```
MarekVCodasip, line 286: Thanks for the patch.
I do feel like there should be a warning or at least a debug print about trying to get the register context before examination.

MarekVCodasip:
Otherwise, the change looks fine to me.

src/target/riscv/riscv-013.c
en-sc: Thank you for the review!
I would like to disagree.
The value of zero here means that the corresponding `is_set` bit is not set (since the values are not examined yet).
If a register decoding requires a value which is not set from the context, it would fail with an assertion.
An example of such an assertion can be seen here: https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/debug_defines.c#L178

MarekVCodasip: Alright then. Thanks for the clearup.

MarekVCodasip (approved):

JanMatCodasip:

PR #997, opened 2024-01-17
target/riscv: move read redirection for `priv` to `riscv-013.c`

en-sc:
The reason for the change is a conflict: `dcsr[5]` is `dcsr.v` in current spec, but it is `dcsr.debugint` in 0.11. This causes `priv` register to be read incorrectly.

Change-Id: If2d8fdcd8536afa4c7149c453101b00ce0df1ce0

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -4851,6 +4851,19 @@ struct target_type riscv013_target = {
 static int riscv013_get_register(struct target *target,
 		riscv_reg_t *value, enum gdb_regno rid)
 {
+	/* It would be beneficial to move this redirection to the
+	 * version-independent section, but there is a conflict:
+	 * `dcsr[5]` is `dcsr.v` in current spec, but it is `dcsr.debugint` in 0.11.
+	 */
+	if (rid == GDB_REGNO_PRIV) {
+		uint64_t dcsr;
+		if (riscv_get_register(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
+			return ERROR_FAIL;
+		*value = set_field(0, VIRT_PRIV_V, get_field(dcsr, CSR_DCSR_V));
+		*value = set_field(*value, VIRT_PRIV_PRV, get_field(dcsr, CSR_DCSR_PRV));
+		return ERROR_OK;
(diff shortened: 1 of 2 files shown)


en-sc:
@lz-bro, FYI.

TommyMurphyTM1234:
Confirmed as working with SiFive HiFive1 Rev A01:

* https://github.com/riscv/riscv-openocd/issues/996#issuecomment-1896039265

JanMatCodasip:
FYI, I will probably be able to review this on Monday.

src/target/riscv/riscv-013.c
```diff
@@ -4851,6 +4851,19 @@ struct target_type riscv013_target = {
 static int riscv013_get_register(struct target *target,
 		riscv_reg_t *value, enum gdb_regno rid)
 {
+	/* This would be nice to move this redirection, to version-independent part,
```
MarekVCodasip: I would reword this:
```suggestion
	/* It would be beneficial to move this redirection to the version-independent section,
```

MarekVCodasip (changes requested):
Otherwise, it looks fine to me.

src/target/riscv/riscv-013.c

MarekVCodasip: Thanks.

MarekVCodasip (approved):

JanMatCodasip:

JanMatCodasip:
> TommyMurphyTM1234: Confirmed as working with SiFive HiFive1 Rev A01: [...]

Thanks for checking it. :+1:

lz-bro:
Need to move write register for priv  to riscv-013.c?

en-sc:
> Need to move write register for priv to riscv-013.c?

@iz-bro, this is a valid concern. However, I've decided not to move the write in this commit.
The reasoning is:
* This is a bug-fix. Bug is related to reading the register, not writing it.
* Assuming the other code is correct, only a user-requested write can cause an issue with writing `dcsr[5]` via writing `priv` on 0.11.

I'm currently working on a patch that significantly changes register accesses. I will post a PR changing the code in question soon. So, I would like not to change current version too much to avoid doing extra work).

Issue #196, opened 2018-01-26
investigate output of `poll`


timsifive:
This is verbose on ARM. http://openocd.org/doc/html/Server-Configuration.html says eg.:
```
> poll
background polling: on
target state: halted
target halted in ARM state due to debug-request, \
               current mode: Supervisor
cpsr: 0x800000d3 pc: 0x11081bfc
MMU: disabled, D-Cache: disabled, I-Cache: enabled
>
```

On RISC-V we don't get that output:
```
> poll
background polling: on
TAP: riscv.cpu (enabled)
> status
```

timsifive:
I think the ARM example is old. Current mainline OpenOCD displays the `TAP: ...` line instead of `target state: ...` and that seems to have been the case since 2008. Then the remaining output comes from the target's `arch_state()` which is supposed to "Issue USER() w/architecture specific status." We could print out which hardware trigger was halted, but aside from that I can't imagine what we'd want to display.

TommyMurphyTM1234:
No - a build of the very latest openocd from https://github.com/riscv/riscv-openocd gives the following output for a Cortex-M3 target poll:

```
Open On-Chip Debugger
> poll
background polling: on
TAP: M2S090.cpu (enabled)
> reset init
target halted due to debug-request, current mode: Thread
xPSR: 00000000 pc: 00000000 msp: 00000000
lr (/32): 0x00000000
xPSR (/32): 0x01000000
> poll
background polling: on
TAP: M2S090.cpu (enabled)
target halted due to debug-request, current mode: Thread
xPSR: 0x01000000 pc: 00000000 msp: 00000000
> resume
> poll
background polling: on
TAP: M2S090.cpu (enabled)
>                            
```   

Note also that the documentation (http://openocd.org/doc/pdf/openocd.pdf) says this about the poll command (with no "on" or "off" argument):

> Poll the current target for its current state. (Also, see [target curstate], page 68.)
> If that target is in debug mode, architecture specific information about the current
> state is printed. An optional parameter allows background polling to be enabled and
> disabled.

So I believe that the RISC-V implementation of poll should act similarly and in accordance with the documentation. At the very least the poll command should state that the target is in "debug halt" when that is the case?

en-sc:
The issue is addressed by #1002.

PR #1002, opened 2024-01-23
target/riscv: report info about target during `poll`

en-sc:
Addresses issue #196.

Change-Id: I71146c7bc769cb9727e57da33e9f514eedef9ce4

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -2712,9 +2712,17 @@ static int riscv_get_gdb_reg_list(struct target *target,
 
 static int riscv_arch_state(struct target *target)
 {
+	assert(target->state == TARGET_HALTED);
+	const bool semihosting_active = target->semihosting &&
+		target->semihosting->is_active;
+	LOG_USER("%s halted due to %s.%s",
+			target_name(target),
+			debug_reason_name(target),
+			semihosting_active ? " Semihosting is active." : "");
 	struct target_type *tt = get_target_type(target);
 	if (!tt)
 		return ERROR_FAIL;
+	assert(tt->arch_state);
 	return tt->arch_state(target);
 }
 


src/target/riscv/riscv.c
```diff
@@ -2712,6 +2712,13 @@ static int riscv_get_gdb_reg_list(struct target *target,
 
 static int riscv_arch_state(struct target *target)
 {
+	assert(target->state == TARGET_HALTED);
+	LOG_USER("%s halted due to %s.%s",
+			target_name(target),
+			debug_reason_name(target),
+			(target->semihosting && target->semihosting->is_active) ?
+			" Semihosting is active." :
+			"");
```
JanMatCodasip: Minor detail for readability of the LOG_USER call. (Optional - if you like.)

```suggestion
	const bool semihosting_active = (target->semihosting && target->semihosting->is_active);
	LOG_USER("%s halted due to %s.%s",
			target_name(target),
			debug_reason_name(target),
			semihosting_active ? " Semihosting is active." : "");
```

src/target/riscv/riscv.c
```diff
@@ -2712,6 +2712,13 @@ static int riscv_get_gdb_reg_list(struct target *target,
...
 	struct target_type *tt = get_target_type(target);
 	if (!tt)
 		return ERROR_FAIL;
```
JanMatCodasip, line 2724: Another optional detail: Since the function pointers in `riscv_target`, `riscv011_target`, `riscv013_target` and in `arch_info` (and maybe others) are becoming not easy to orient in, I'd use the opportuninty and add here an assert for additional safety. 

```suggestion
		return ERROR_FAIL;
	assert(tt->arch_state);
```

JanMatCodasip (dismissed):
Looks good, thanks.

I have just two small and optional suggestions.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: Addressed.

JanMatCodasip (approved):

JanMatCodasip:

Issue #1005, opened 2024-01-25
Flush dirty registers on shutdown is broken


fborisovskii:
Hello,

After merging to commit
https://github.com/riscv/riscv-openocd/commit/8dbb1250f587484774752e8610f04a98157ca10a
I've got Error in openocd log after sending monitor shutdown from gbd session.

```
Debug: 17077 1640 gdb_server.c:438 gdb_log_outgoing_packet(): [riscv.cpu0] {1} sending packet: $OK#9a
Debug: 17078 1641 gdb_server.c:420 gdb_log_incoming_packet(): [riscv.cpu0] {1} received packet: qRcmd,73687574646f776e
Debug: 17079 1641 gdb_server.c:2820 gdb_query_packet(): qRcmd: shutdown
Debug: 17080 1641 command.c:154 script_debug(): command - shutdown
User : 17081 1641 server.c:758 handle_shutdown_command(): shutdown command invoked
Debug: 17082 1641 gdb_server.c:438 gdb_log_outgoing_packet(): [riscv.cpu0] {1} sending packet: $O73687574646f776e20636f6d6d616e6420696e766f6b65640a#87
Debug: 17083 1641 gdb_server.c:438 gdb_log_outgoing_packet(): [riscv.cpu0] {1} sending packet: $EA8#be
Debug: 17084 1641 gdb_server.c:1154 gdb_connection_closed(): {1} GDB Close, Target: riscv.cpu0, state: halted, gdb_actual_connections=0
Debug: 17085 1641 target.c:1790 target_call_event_callbacks(): target event 8 (gdb-end) for core riscv.cpu0
Debug: 17086 1641 target.c:1790 target_call_event_callbacks(): target event 23 (gdb-detach) for core riscv.cpu0
Debug: 17087 1641 breakpoints.c:319 breakpoint_remove_all_internal(): [riscv.cpu0] Delete all breakpoints
Debug: 17088 1641 riscv.c:535 riscv_deinit_target(): [riscv.cpu0] riscv_deinit_target()
Debug: 17089 1641 riscv.c:1788 riscv_flush_registers(): [riscv.cpu0] Flushing register cache
Debug: 17090 1641 riscv.c:1800 riscv_flush_registers(): [riscv.cpu0] fp is dirty; write back 0x80004054
Debug: 17091 1641 riscv-013.c:4874 riscv013_set_register(): [riscv.cpu0] writing 0x80004054 to register fp
[...]

PR #1006, opened 2024-01-26
Revert "break from long loops on shutdown request"

en-sc:
This reverts commits 2e920a212fbe2de705811d547c169c1ae1611a02 and 8dbb1250f587484774752e8610f04a98157ca10a.

The reason is, after `openocd_is_shutdown_pending()` becomes true, arbitrary command may be executed:
* In `target_destroy()` and the corresponding `target->type->deinit_target()`.
* In user-specifyed `pre_shutdown_commands` list.

Change-Id: Icd00d1d954cd45e255880a6f76c3a74c098d6a17

### DIFF
--- a/src/server/server.h
+++ b/src/server/server.h
@@ -118,6 +118,5 @@ COMMAND_HELPER(server_port_command, unsigned short *out);
 
 #define ERROR_SERVER_REMOTE_CLOSED		(-400)
 #define ERROR_CONNECTION_REJECTED		(-401)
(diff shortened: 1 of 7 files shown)


en-sc:
This addresses #1005.

JanMatCodasip:
Thank you for preparing the revert. I'd recommend to merge this on Monday if there is no other feedback by then.

JanMatCodasip (approved):

MarekVCodasip (approved):
Makes sense, thank for preparing the revert.

Issue #1010, opened 2024-01-30
Halt groups may cause debug reason to be reported incorrectly


en-sc:
This is part of a log of `Sv39Test` on `spike64-2-hwthread` platform from [riscv-tests/debug](https://github.com/riscv-software-src/riscv-tests/tree/master/debug) (a patched version is used to report target state in `LOG_TARGET_*`):
```
Debug: 32052 11108 riscv.c:2988 riscv_poll_hart(): [riscv.cpu0]<running> polling, target->state=1
...
Debug: 32058 11109 riscv-013.c:407 dump_field(): read: dmstatus=0x430c82 {version=0_13 anyrunning=1 allrunning=1 anyresumeack=1 allresumeack=1 impebreak=1 authenticated=true}
Debug: 32059 11109 riscv.c:2988 riscv_poll_hart(): [riscv.cpu1]<running> polling, target->state=1
...
Debug: 32065 11111 riscv-013.c:407 dump_field(): read: dmstatus=0x430382 {version=0_13 anyresumeack=1 allresumeack=1 impebreak=1 authenticated=true anyhalted=1 allhalted=1}
Debug: 32066 11111 riscv.c:3040 riscv_poll_hart(): [riscv.cpu1]<running>   triggered a halt; previous_target_state=1
Debug: 32067 11111 riscv-013.c:1706 register_read_direct(): [riscv.cpu1]<halted> Reading dcsr
...
Debug: 32155 11131 riscv-013.c:1725 register_read_direct(): [riscv.cpu1]<halted> dcsr = 0x4000b183
Debug: 32156 11131 riscv-013.c:5080 riscv013_halt_reason(): [riscv.cpu1]<halted> dcsr.cause: 0x6
Debug: 32157 11131 riscv.c:1844 set_debug_reason(): [riscv.cpu1]<halted> debug_reason=0
Debug: 32158 11131 riscv.c:3221 riscv_openocd_poll(): [riscv.cpu1]<halted> should_remain_halted=0, should_resume=0
Debug: 32159 11131 riscv.c:3235 riscv_openocd_poll(): [riscv.cpu1]<halted> halt all; halted=1
Debug: 32160 11131 riscv.c:1921 riscv_halt(): [riscv.cpu1]<halted> halting all harts
Debug: 32161 11131 riscv.c:1853 halt_prep(): [riscv.cpu0]<running> prep hart, debug_reason=5
[...]

PR #1011, opened 2024-02-02
target/riscv: set `state` and `debug_reason` in `riscv_halt_go_all_harts()`

en-sc:
If targets are in a halt group, and a target in the group reaches a breakpoint, the target's state was able to remain `TARGET_RUNNING`.

Addresses issue #1010

Change-Id: I734bc6da71d289c4d05b417c8bf67a7d1a56574f

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1874,6 +1874,12 @@ static int riscv_halt_go_all_harts(struct target *target)
 		return ERROR_FAIL;
 	if (state == RISCV_STATE_HALTED) {
 		LOG_TARGET_DEBUG(target, "Hart is already halted.");
+		if (target->state != TARGET_HALTED) {
+			target->state = TARGET_HALTED;
+			enum riscv_halt_reason halt_reason = riscv_halt_reason(target);
+			if (set_debug_reason(target, halt_reason) != ERROR_OK)
+				return ERROR_FAIL;
+		}
 	} else {
 		if (r->halt_go(target) != ERROR_OK)
 			return ERROR_FAIL;


src/target/riscv/riscv.c
```diff
@@ -1874,6 +1874,12 @@ static int riscv_halt_go_all_harts(struct target *target)
 		return ERROR_FAIL;
 	if (state == RISCV_STATE_HALTED) {
 		LOG_TARGET_DEBUG(target, "Hart is already halted.");
+		if (target->state != TARGET_HALTED) {
+			target->state = TARGET_HALTED;
+			enum riscv_halt_reason halt_reason = riscv_halt_reason(target);
+			if (set_debug_reason(target, halt_reason) != ERROR_OK)
+				return ERROR_FAIL;
```
aap-sc, line 1881: `return ERROR_FAIL`. 

I'm not sure that return of ERROR_FAIL is the right way. Logically we should just set halt_reason to unknown/undefined and report LOG_TARGET_WARNING/LOG_TARGET_ERROR

en-sc: 1. `ERROR_FAIL` is not possible here currently, since:
    * [`set_debug_reason()` returns `ERROR_FAIL` iff `halt_reason` is `RISCV_HALT_ERROR`](https://github.com/riscv/riscv-openocd/blob/24d71d7a729c9d2b6c31bbc6f5fac304d74e8291/src/target/riscv/riscv.c#L1842)
    * [`riscv013_get_hart_state()`](https://github.com/riscv/riscv-openocd/blob/24d71d7a729c9d2b6c31bbc6f5fac304d74e8291/src/target/riscv/riscv-013.c#L2684) does not return `RISCV_HALT_ERROR`
   * 0.11 targets do not implement `r->get_hart_state()`
2. This is consistent with the code around ([`halt_prep()`](https://github.com/riscv/riscv-openocd/pull/1011/files#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR1849) / [`halt_go()`](https://github.com/riscv/riscv-openocd/pull/1011/files#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR1893))

aap-sc: It may be the case that there is a misunderstanding. I probably selected a wrong line of code. Consider this:

1. We want to halt all relevant targets.  This is the primary goal.
2. For a particular hart - we already know that it is halted. So our primary goal is complete.
3. We also want to figure out debug reason. We do this by calling `riscv013_halt_reason`. This procedure can fail.
4. I don't think that we need to abort the whole process in this case. We can just notify user that we could not determine the reason for the halt.

aap-sc: Re-read your comments. Looks like currently we have the exact behavior I requested, actually. `riscv_halt_reason` returns `RISCV_HALT_UNKNOWN` to indicate that something went terribly wrong. `RISCV_HALT_UNKNOWN` is translated to `DBG_REASON_UNDEFINED` which is exactly what I've requested.

It seems that `set_debug_reason` returning an error code is a dead code. While somewhat confusing, this is outside of the scope of this commit. So the comment is resolved. Thanks.

aap-sc:
LGTM

JanMatCodasip (approved):
LGTM, thank you.

Let's merge on Monday (unless there are more comments).

PR #1014, opened 2024-02-06
Fixes and cleanup in riscv batch and related functions

JanMatCodasip:
Fixes:

- Data types of address & data parameters in `riscv_batch_add_*()` and `riscv*_fill_dm*()` changed to uint64_t and uint32_t.
- Fixed the comparison in `riscv_batch_full()`.
- Fixed assertions in `riscv_batch_get_dmi_read_op()` and `riscv_batch_get_dmi_read_data()`.

Cleanup:

- Simplified calloc() fail handling in riscv_batch_alloc().
- Added explicit NULL assignments in riscv_batch_alloc() for clarity and readability. Don't rely on calloc().
- Removed suffix `_u64` from `riscv_*_fill_dm*()` since it does not have any meaning.
- Renamed `*dmi_write_u64_bits()` to `*get_dmi_scan_length()` which better describes its purpose.

Change-Id: Id70e5968528d64b2ee5476f1c00e08459a1e291d

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -13,7 +13,8 @@
 #define DMI_SCAN_MAX_BIT_LENGTH (DTM_DMI_MAX_ADDRESS_LENGTH + DTM_DMI_DATA_LENGTH + DTM_DMI_OP_LENGTH)
 #define DMI_SCAN_BUF_SIZE (DIV_ROUND_UP(DMI_SCAN_MAX_BIT_LENGTH, 8))
 
-#define BATCH_RESERVED_SCANS 4
+/* Reserve extra room in the batch (needed for the last NOP operation) */
+#define BATCH_RESERVED_SCANS 1
 
 static void dump_field(int idle, const struct scan_field *field);
 
@@ -37,6 +38,9 @@ struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans, size_
 	out->bscan_ctxt = NULL;
 	out->read_keys = NULL;
 
+	/* FIXME: There is potential for memory usage reduction. We could allocate
+	 * smaller buffers than DMI_SCAN_BUF_SIZE (that is, buffers that correspond to
+	 * the real DR scan length on the given target) */
 	out->data_out = malloc(sizeof(*out->data_out) * scans * DMI_SCAN_BUF_SIZE);
 	if (!out->data_out) {
 		LOG_ERROR("Failed to allocate data_out in RISC-V batch.");
@@ -84,7 +88,7 @@ void riscv_batch_free(struct riscv_batch *batch)
 
 bool riscv_batch_full(struct riscv_batch *batch)
 {
-	return batch->used_scans >= (batch->allocated_scans - BATCH_RESERVED_SCANS);
+	return riscv_batch_available_scans(batch) == 0;
 }
 
 int riscv_batch_run(struct riscv_batch *batch)
@@ -98,7 +102,7 @@ int riscv_batch_run(struct riscv_batch *batch)
 
 	for (size_t i = 0; i < batch->used_scans; ++i) {
 		if (bscan_tunnel_ir_width != 0)
-			riscv_add_bscan_tunneled_scan(batch->target, batch->fields+i, batch->bscan_ctxt+i);
+			riscv_add_bscan_tunneled_scan(batch->target, batch->fields + i, batch->bscan_ctxt + i);
 		else
 			jtag_add_dr_scan(batch->target->tap, 1, batch->fields + i, TAP_IDLE);
 
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -13,58 +13,62 @@
 #define DMI_SCAN_MAX_BIT_LENGTH (DTM_DMI_MAX_ADDRESS_LENGTH + DTM_DMI_DATA_LENGTH + DTM_DMI_OP_LENGTH)
 #define DMI_SCAN_BUF_SIZE (DIV_ROUND_UP(DMI_SCAN_MAX_BIT_LENGTH, 8))
 
+#define BATCH_RESERVED_SCANS 4
+
 static void dump_field(int idle, const struct scan_field *field);
 
 struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans, size_t idle)
 {
-	scans += 4;
+	scans += BATCH_RESERVED_SCANS;
 	struct riscv_batch *out = calloc(1, sizeof(*out));
-	if (!out)
-		goto error0;
+	if (!out) {
+		LOG_ERROR("Failed to allocate struct riscv_batch");
+		return NULL;
+	}
+
 	out->target = target;
 	out->allocated_scans = scans;
 	out->idle_count = idle;
-	out->data_out = malloc(sizeof(*out->data_out) * (scans) * DMI_SCAN_BUF_SIZE);
+	out->last_scan = RISCV_SCAN_TYPE_INVALID;
+
+	out->data_out = NULL;
+	out->data_in = NULL;
+	out->fields = NULL;
+	out->bscan_ctxt = NULL;
+	out->read_keys = NULL;
+
+	out->data_out = malloc(sizeof(*out->data_out) * scans * DMI_SCAN_BUF_SIZE);
 	if (!out->data_out) {
 		LOG_ERROR("Failed to allocate data_out in RISC-V batch.");
-		goto error1;
+		goto alloc_error;
 	};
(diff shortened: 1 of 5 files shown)


src/target/riscv/batch.c
```diff
@@ -80,7 +84,7 @@ void riscv_batch_free(struct riscv_batch *batch)
 
 bool riscv_batch_full(struct riscv_batch *batch)
 {
-	return batch->used_scans > (batch->allocated_scans - 4);
+	return batch->used_scans >= (batch->allocated_scans - BATCH_RESERVED_SCANS);
```
en-sc: I would suggest reusing `riscv_batch_available_scans()` here:
```suggestion
	return riscv_batch_available_scans(batch) != 0;
```

src/target/riscv/batch.c
```diff
@@ -13,58 +13,62 @@
 #define DMI_SCAN_MAX_BIT_LENGTH (DTM_DMI_MAX_ADDRESS_LENGTH + DTM_DMI_DATA_LENGTH + DTM_DMI_OP_LENGTH)
 #define DMI_SCAN_BUF_SIZE (DIV_ROUND_UP(DMI_SCAN_MAX_BIT_LENGTH, 8))
 
+#define BATCH_RESERVED_SCANS 4
```
en-sc: Wouldn't the value of 1 be sufficient?
As far as I understand this is only needed to always be able to insert a `nop` before running the batch.

src/target/riscv/batch.c
```diff
@@ -13,58 +13,62 @@
...
-	out->data_out = malloc(sizeof(*out->data_out) * (scans) * DMI_SCAN_BUF_SIZE);
+	out->last_scan = RISCV_SCAN_TYPE_INVALID;
+
+	out->data_out = NULL;
+	out->data_in = NULL;
+	out->fields = NULL;
+	out->bscan_ctxt = NULL;
+	out->read_keys = NULL;
+
+	out->data_out = malloc(sizeof(*out->data_out) * scans * DMI_SCAN_BUF_SIZE);
```
en-sc, line 44: I'm not sure about using `DMI_SCAN_BUF_SIZE` here. IMHO, using `DIV_ROUND_UP(riscv_get_dmi_scan_length(target), 8)` would result in a significant reduction in memory usage, since `abits` can be as low as 6, but it is assumed to be 64 here.

src/target/riscv/riscv-013.c
```diff
@@ -55,13 +55,13 @@ static riscv_insn_t riscv013_read_progbuf(struct target *target, unsigned int
...
-static void riscv013_fill_dmi_write_u64(struct target *target, char *buf, int a, uint64_t d);
-static void riscv013_fill_dmi_read_u64(struct target *target, char *buf, int a);
-static int riscv013_dmi_write_u64_bits(struct target *target);
-static void riscv013_fill_dmi_nop_u64(struct target *target, char *buf);
-static void riscv013_fill_dm_write_u64(struct target *target, char *buf, int a, uint64_t d);
-static void riscv013_fill_dm_read_u64(struct target *target, char *buf, int a);
-static void riscv013_fill_dm_nop_u64(struct target *target, char *buf);
+static void riscv013_fill_dmi_write(struct target *target, char *buf, uint64_t a, uint32_t d);
+static void riscv013_fill_dmi_read(struct target *target, char *buf, uint64_t a);
+static void riscv013_fill_dmi_nop(struct target *target, char *buf);
```
en-sc, line 60: I think these functions can be inlined into corresponding `riscv013_fill_dm_*()`.
As an alternative, `riscv013_fill_dmi_*` functions can be exposed instead of `riscv013_fill_dm_*` and the address translation can be done in `riscv_fill_dm_*` wrappers via `riscv_get_dmi_address()` (defined in `riscv.c`).
IMHO the second option is preferable, since it results in a more uniform interface (the same approach is implemented for non-batched accesses (there are `dmi_read`/`dmi_write` in `struct riscv_info`, but no `dm_read`/`dm_write`))

en-sc (commented):
Thank you for the change! The cleanup is great and can be merged as-is. However, I have a few suggestions.

src/target/riscv/batch.c
JanMatCodasip: Thanks for the suggestion - done.

(I think there's a typo and you meant `riscv_batch_available_scans(batch) == 0`, correct? ;-))

src/target/riscv/batch.c
JanMatCodasip: As far as I understand, we really need `4` reserved scans:

- one for the final NOP
- three for the optional bscan bridge tunneling (see here: https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/batch.c#L97)

I have added a comment to make it clearer to the readers.

src/target/riscv/batch.c
JanMatCodasip: Good catch, I agree. There is potential for saving some memory.

If you don't mind, I'd prefer to make this change separately, and it may take me some time to get to that. So for now I have added at least a `FIXME` comment next to the allocation to make it clear that there is a room for improvement.

src/target/riscv/batch.c
en-sc: > (I think there's a typo and you meant `riscv_batch_available_scans(batch) == 0`, correct? ;-))

Oh... Yes, sorry)

src/target/riscv/riscv-013.c
JanMatCodasip: I also agree that this can be improved.

As for the callbacks (function pointers) in `struct target_type riscv011_target`, `struct target_type riscv013_target` and `struct riscv_info`, I would like to clean it up and get rid of the callbacks completely in near term. Instead, I'd prefer to make direct calls to the `riscv011_*()` and `riscv013_*()` functions in the whole codebase.

For that reason, if you don't mind, I'd like to only clean up the data types and function names, but retain the structure of the function calls for now. The structural changes I would leave for a separate merge request.

src/target/riscv/batch.c
en-sc: As far as I understand these three additional scans are needed for each tunneled scans and are allocated in the `bscan_ctxt` array. It seems to me [`riscv_add_bscan_tunneled_scan()`](https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv.c#L6322) uses one element of `fields` array and one element of `bscan_ctxt` array, which itself contains extra fields (https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv.h#L327).
Am I missing something?

src/target/riscv/batch.c
en-sc: IMHO, that's even better, since the change in question is out of scope of the commit.

src/target/riscv/riscv-013.c
en-sc: Yes, this is out of scope.

src/target/riscv/batch.c
JanMatCodasip: I could have missed the `bscan_ctxt`. I will double-check it again.

JanMatCodasip: Hi @en-sc, you're right, the number of the reserved scans should be `1`, and thanks for checking it after me. I mis-read the code in `riscv_batch_run()`. 

I have updated this PR.

src/target/riscv/batch.c
```diff
@@ -125,33 +133,33 @@ int riscv_batch_run(struct riscv_batch *batch)
 	return ERROR_OK;
 }
 
-void riscv_batch_add_dm_write(struct riscv_batch *batch, unsigned int address, uint64_t data,
+void riscv_batch_add_dm_write(struct riscv_batch *batch, uint64_t address, uint32_t data,
```
en-sc, line 136: As stated in the spec [3.1 Debug Module Interface (DMI)]:
> The DMI uses between 7 and 32 address bits.

So maybe it is better to use `uint32_t` for the address?

en-sc (approved):
The change is great. I have only minor question, though, IMHO, it does not block the patch from being merged.

src/target/riscv/batch.c
JanMatCodasip: @en-sc Thanks for pointing that out. I missed that the debug spec restricts the DMI address size to 32 bits in the section 3.1 (as you have quoted). 

My reason for uint64_t was the dtmcs.abits field which may contain value up to 63 (but I missed the above restriction).

I doubt that there is any real-world hardware with abits > 32 (what would be the reason for such a large address space for debug?). So we can restrict this, but at the same time we should add a check for abits value, read from the target. I'd prefer to do this in a separate merge request.

JanMatCodasip: I have filed an issue for it: https://github.com/riscv-collab/riscv-openocd/issues/1019

PR #1015, opened 2024-02-11
Fix error on exit with uninitialized target

tom-van:
riscv_deinit_target() tries to get_target_type() even if the examination failed or was not called at all. dtm_version is not known and an annoying error is printed:

  Error: [riscv.cpu.0] Unsupported DTM version: -1
  Error: [riscv.cpu.0] Could not identify target type.

Avoid calling get_target_type() if dtm_version is
DTM_DTMCS_VERSION_UNKNOWN.

Change-Id: I48239b1b11561357e845f3a6cb3a8e3b19e35715

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -535,9 +535,12 @@ static void riscv_deinit_target(struct target *target)
 	LOG_TARGET_DEBUG(target, "riscv_deinit_target()");
 
 	struct riscv_info *info = target->arch_info;
-	struct target_type *tt = get_target_type(target);
-	if (!tt)
-		LOG_TARGET_ERROR(target, "Could not identify target type.");
+	struct target_type *tt = NULL;
+	if (info->dtm_version != DTM_DTMCS_VERSION_UNKNOWN) {
+		tt = get_target_type(target);
+		if (!tt)
+			LOG_TARGET_ERROR(target, "Could not identify target type.");
+	}
 
 	if (riscv_flush_registers(target) != ERROR_OK)
 		LOG_TARGET_ERROR(target, "Failed to flush registers. Ignoring this error.");


src/target/riscv/riscv.c
```diff
@@ -535,9 +535,12 @@ static void riscv_deinit_target(struct target *target)
 	LOG_TARGET_DEBUG(target, "riscv_deinit_target()");
 
 	struct riscv_info *info = target->arch_info;
-	struct target_type *tt = get_target_type(target);
-	if (!tt)
-		LOG_TARGET_ERROR(target, "Could not identify target type.");
+	struct target_type *tt = NULL;
+	if (info->dtm_version != DTM_DTMCS_VERSION_UNKNOWN) {
```
en-sc, line 539: This is not particularly safe. Please, consider a target that was not properly initialized. `target->arch_info` can be still zero or point to uninitialized memory.
The condition should at least be modified to:
```
if (info && info->dtm_version != DTM_DTMCS_VERSION_UNKNOWN) {
```
However, under such condition `get_target_type()` will never print an error, and IMHO, if a target was never initialized but `riscv_deinit_target()` is called -- an error should be printed.

In general, I would like to leave the messaging here.
In case `examine()` failed, the state of the target is not the same as if the target was just initialized. Moreover, it is not documented either. This is an issue. I am preparing some patches aimed at addressing this. BTW, your patch with `riscv013_dm_free()` is a great help in this regard, thanks!

If you still find the message too annoying -- please, consider changing it to `LOG_TARGET_DEBUG()`.

en-sc (changes requested):
Thank you for the patch! I also find these messages annoying, but there is an underlying issue -- a proper deinitialization can not be guaranteed in such cases for now. So, IMHO, a message should be printed here.

src/target/riscv/riscv.c
tom-van: > The condition should at least be modified to:

> `if (info && info->dtm_version != DTM_DTMCS_VERSION_UNKNOWN) {`

Good point, will fix if we go on with this patch.

> In case examine() failed, the state of the target is not the same as if the target was just initialized. Moreover, it is not documented either. This is an issue. I am preparing some patches aimed at addressing this.

Oh I think I see at least one issue:
If tt->init_target() succeds
https://github.com/riscv/riscv-openocd/blob/5d4fa0001e2f6bd23caf10fd02f980053adb2b3b/src/target/riscv/riscv.c#L1740
and the following version specific examination fails
https://github.com/riscv/riscv-openocd/blob/5d4fa0001e2f6bd23caf10fd02f980053adb2b3b/src/target/riscv/riscv.c#L1744-L1746
dtm_version gets reset to unknown
https://github.com/riscv/riscv-openocd/blob/5d4fa0001e2f6bd23caf10fd02f980053adb2b3b/src/target/riscv/riscv.c#L1750-L1752
and later
https://github.com/riscv/riscv-openocd/blob/5d4fa0001e2f6bd23caf10fd02f980053adb2b3b/src/target/riscv/riscv.c#L533
will not recognize target type and therefore will not call
https://github.com/riscv/riscv-openocd/blob/5d4fa0001e2f6bd23caf10fd02f980053adb2b3b/src/target/riscv/riscv.c#L545-L546

Hmm, much worse than an annoying message. Giving up for now

JanMatCodasip:
Hi @tom-van, 

do you intend to revive this PR? If so, please submit it as a patch to the upstream OpenOCD - maybe you already did?

In any case, consider closing this PR. I would like to archive this riscv-openocd fork in near term, to not create confusion for newcomers about where to submit their RISC-V related changes.

PR #1016, opened 2024-02-11
Free dm and target_list structures

tom-van:
Fix memory leak on exit.

Change-Id: I6a89ac0c93b11da35b90eec3abcc5b6fd5d1be68

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -280,6 +280,29 @@ static dm013_info_t *get_dm(struct target *target)
 	return dm;
 }
 
+static void riscv013_dm_free(struct target *target)
+{
+	RISCV013_INFO(info);
+	dm013_info_t *dm = info->dm;
+	if (!dm)
+		return;
+
+	target_list_t *target_entry;
+	list_for_each_entry(target_entry, &dm->target_list, list) {
+		if (target_entry->target == target) {
+			list_del(&target_entry->list);
+			free(target_entry);
+			break;
+		}
+	}
+
+	if (list_empty(&dm->target_list)) {
+		list_del(&dm->list);
+		free(dm);
+	}
+	info->dm = NULL;
+}
+
 static riscv_debug_reg_ctx_t get_riscv_debug_reg_ctx(const struct target *target)
 {
 	if (!target_was_examined(target)) {
@@ -1863,6 +1886,8 @@ static void deinit_target(struct target *target)
 	struct riscv_info *info = target->arch_info;
 	if (!info)
 		return;
+
+	riscv013_dm_free(target);
 
 	free(info->version_specific);
 	/* TODO: free register arch_info */


src/target/riscv/riscv-013.c
```diff
@@ -280,6 +280,29 @@ static dm013_info_t *get_dm(struct target *target)
 	return dm;
 }
 
+static void riscv013_dm_free(struct target *target)
```
en-sc, line 283:
```suggestion
static void riscv013_dm_free(const struct target *target)
```

en-sc (changes requested):
Thank you for the patch! I have only a minor nitpick.

src/target/riscv/riscv-013.c
```diff
@@ -280,6 +280,29 @@ static dm013_info_t *get_dm(struct target *target)
...
+			free(target_entry);
+			break;
+		}
+	}
+
+	if (list_empty(&dm->target_list)) {
+		list_del(&dm->list);
+		free(dm);
+	}
+	info->dm = NULL;
```
tom-van, line 303: Here we change riscv013_info_t structure which is tightly coupled with the target

src/target/riscv/riscv-013.c
tom-van: Evgeniy, you're right that struct target is not changed by dm_free, but on the other hand it does change riscv013_info_t.
I would consider the const attribute as misleading, wouldn't you?

en-sc: Yes, it is a bit misleading in that regard.
However, I would argue that this highlights that the function does not change the state of the architecture-independent part (`struct target`) and only operates on riscv013-specific part (`riscv013_info_t`), opposed to e.g. `examine()` or `deassert_reset()`.

tom-van: Sorry, I would regard such construct as `const` attribute misuse (and will refuse such change in the upstream OpenOCD repo). Let's take `const` seriously and use it only when neither main `struct target` nor any of private sub-containers are changed. If the similar code were written in C++, the compiler would not allow to pass a non-const pointer to a subclass inside of `const` method. C does not prevent us to do so, but we should not abuse that C imperfection.

MarekVCodasip: I don't feel strongly about it, but I would prefer this to be non-const, for the reasons @tom-van outlined.

MarekVCodasip (approved):
Thanks for the fix. LGTM

src/target/riscv/riscv-013.c
en-sc: First of all, I do agree this is a question of taste. So I'm approving your PR.

However, I would still like to have an argument on the topic if you have the time.
> If the similar code were written in C++, the compiler would not allow to pass a non-const pointer to a subclass inside of `const` method. C does not prevent us to do so, but we should not abuse that C imperfection.

I don't quite understand, what do you mean by a subclass? Are you referring to a derived class not being able to change the fields of a base class if the method is marked const?

The thing is, I think passing the target via a const pointer models here a protected  inheritance. You can observe the fields of the base class (this would require public getters in the base class in case of C++), but you are not able to change them, since they are protected.

In general, I find it confusing when a concept is assigned an implicit meaning. Here is a pointer to a structure, but one of it's members is considered tightly coupled to it, so another set of rules is applied to it (here regarding constness). I find such rules hard to navigate.

en-sc (approved):
LGTM

src/target/riscv/riscv-013.c
tom-van: Thanks for explanation.

> The thing is, I think passing the target via a const pointer models here a protected inheritance.

Never seen it from this point of view. I admit it makes some sense to me. But the 'protected' inheritance model should
not allow to pass const pointer from the caller. It would look like
```
fn(struct target *t)
{
  const struct target *ct = t;
  // no more use of t
}
```
Anyway the 'protected' inheritance model is contra-intuitive. Let's simply use `const` pointer parameter if and only if the whole hierarchy under the pointer is not written by the function and all its callees - i.e. pure getters/observers.

en-sc:
I would suggest merging this on Thursday, unless any other feedback/discussion arises.

JanMatCodasip:
> I would suggest merging this on Thursday, unless any other feedback/discussion arises.

Sure, please go ahead :)

PR #1018, opened 2024-02-15
Merge up to efdd5e0 from upstream

en-sc:
There is an ongoing discussion on https://review.openocd.org/c/openocd/+/8124 regarding 0d3d4c981ac77b600ce95c9ea6f1cdb280127342, but AFAIU it seems that the patch does not break anything.


MarekVCodasip (approved):
LGTM

PR #1020, opened 2024-02-21
target/riscv: Implement ability to re-examine target

kr-sc:
Add a new OpenOCD command `re_examine` to enforce re-examination of target. Some of existing hardware have a writable misa.V bit. It means that it can be enabled by software. We need an opportunity to enforce `examine` procedure in OpenOCD to query the capabilities of the target.

Also this patch includes commit with fix of a little issue connected with clearing triggers cache.

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -11515,6 +11515,13 @@ riscv exec_progbuf 0x0330000f 0x0000100f
 riscv exec_progbuf 0x94a20405
 @end example
 
+@subsection RISC-V Syntacore hardware-specific Commands
+
+@deffn {Command} {riscv re_examine}
+Enforce (re)examination of target. This command requires that all watchpoints and
+breakpoints to be deleted before use.
+@end deffn
+
 @section ARC Architecture
 @cindex ARC
 
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -4560,6 +4560,28 @@ COMMAND_HANDLER(riscv_set_enable_ge_lt_trigger)
 	return ERROR_COMMAND_SYNTAX_ERROR;
 }
 
+COMMAND_HANDLER(handle_re_examine_target)
+{
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	if (target->watchpoints || target->breakpoints) {
+		LOG_TARGET_ERROR(target, "Please, remove all breakpoints and watchpoints.");
+		return ERROR_FAIL;
+	}
+
+	if (riscv_flush_registers(target) != ERROR_OK) {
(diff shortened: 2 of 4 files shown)


doc/openocd.texi
```diff
@@ -11515,6 +11515,13 @@ riscv exec_progbuf 0x0330000f 0x0000100f
 riscv exec_progbuf 0x94a20405
 @end example
 
+@subsection RISC-V Syntacore hardware-specific Commands
```
MarekVCodasip, line 11518: I would not call this a Syntacore hardware-specific command.

Spike also has the ability to change at runtime which extensions are implemented. The RISC-V specification does allow the ISA to change at runtime if the hardware supports it.

src/target/target.h
```diff
@@ -450,6 +450,15 @@ static inline void target_set_examined(struct target *target)
...
 
+/**
+ * Reset the @c examined flag for the given target.
+ * Pure paranoia -- targets are zeroed on allocation.
+ */
+static inline void target_reset_examined(struct target *target)
+{
+	target->examined = false;
+}
+
```
MarekVCodasip, line 461: I am not sure if touching target.h/c is the best approach.

Myself, I would open a patch on the upsteam OpenOCD changing the visibility of this function.

For now, I would set the `target->examined = false` directly in the handle_re_examine_target() and add a // TODO to change it later.

src/target/riscv/riscv.c
```diff
@@ -4556,6 +4560,28 @@ COMMAND_HANDLER(riscv_set_enable_ge_lt_trigger)
 	return ERROR_COMMAND_SYNTAX_ERROR;
 }
 
+COMMAND_HANDLER(handle_re_examine_target)
+{
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	if (target->watchpoints || target->breakpoints) {
+		LOG_TARGET_ERROR(target, "Please, remove all breakpoints and watchpoints.");
```
MarekVCodasip, line 4569: I would change the message, something akin to: "Re-examination requires the removal of all breakpoints and watchpoints."

MarekVCodasip (changes requested):
Thanks for the patch. I have found a few things.

I would like others to also take a look, I have a feeling that re-examination needs more discussion.

doc/openocd.texi
aap-sc: > I would not call this a Syntacore hardware-specific command.

This is a mistake on our side. This code was in Syntacore codebase for a while and we just missed that this part had to be re-written during internal review.

aap-sc:
NOTE: @MarekVCodasip Kirill will be out for a while... This patch won't get any updates because of this for some time. Likely, we'll refactor the  patch a little bit and re-upload it again. In any case - thanks for review, the suggested changes will be integrated.

PR #1023, opened 2024-02-29
target/riscv: check `abstractcs.busy`

en-sc:
According to the RISC-V Debug Spec (1.0.0-rc1)[3.7 Abstract Commands]:
> While an abstract command is executing (busy in abstractcs is high), a
debugger must not change hartsel, and must not write 1 to haltreq,
resumereq, ackhavereset, setresethaltreq, or clrresethaltreq.

This patch ensures OpenOCD does not violate this rule.

The patch is separated into three commits.
1. target/riscv: cache `abstractcs.busy` in `dm013_info_t` --  provides facilities to track the state of `abstractcs.busy`
2. target/riscv: introduce `examine_dm()` function -- ensures the rule is followed during examination.
3. target/riscv: check `abstractcs.busy` -- takes care of other cases.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -127,6 +127,8 @@ typedef struct {
 	uint32_t base;
 	/* The number of harts connected to this DM. */
 	int hart_count;
+	/* Indicates we already examined this DM, so don't need to do it again. */
+	bool was_examined;
 	/* Indicates we already reset this DM, so don't need to do it again. */
 	bool was_reset;
 	/* Targets that are connected to this DM. */
@@ -201,9 +203,6 @@ typedef struct {
 	uint8_t datasize;
 	uint8_t dataaccess;
 	int16_t dataaddr;
-
-	/* The width of the hartsel field. */
-	unsigned hartsellen;
 
 	/* DM that provides access to this target. */
 	dm013_info_t *dm;
@@ -1933,6 +1932,126 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
 	return ERROR_OK;
 }
 
+static int examine_dm(struct target *target)
+{
+	dm013_info_t *dm = get_dm(target);
+	if (!dm)
+		return ERROR_FAIL;
+	if (dm->was_examined)
+		return ERROR_OK;
+
+	int result = ERROR_FAIL;
+
+	uint32_t dmcontrol;
+	if (!dm->was_reset) {
+		/* First, the Debug Module is reset. However,
+		 * `dmcontrol.hartsel` should be read first, in order not to
+		 * change it when requesting the reset, since changing it
+		 * without checking that `abstractcs.busy` is low is
+		 * prohibited.
+		 */
+		result = dm_read(target, &dmcontrol, DM_DMCONTROL);
+		if (result != ERROR_OK)
+			return result;
+		/* Initiate the reset (`dmcontrol.dmactive == 0`) leaving
+		 * `dmcontrol.hartsel` the same.
+		 */
+		dmcontrol = (dmcontrol & DM_DMCONTROL_HARTSELLO) |
+			(dmcontrol & DM_DMCONTROL_HARTSELHI);
+		result = dm_write(target, DM_DMCONTROL, dmcontrol);
+		if (result != ERROR_OK)
+			return result;
+		/* FIXME: We should poll dmcontrol until dmactive becomes 0
+		 * See https://github.com/riscv/riscv-debug-spec/pull/566
+		 */
+	} else {
+		/* The DM was already reset when examining a different hart.
+		 * No need to reset it again. But for safety, assume that an abstract
+		 * command might be in progress at the moment.
+		 */
+		dm->abstract_cmd_maybe_busy = true;
+	}
+
+	dm->current_hartid = HART_INDEX_UNKNOWN;
+
+	result = dm_write(target, DM_DMCONTROL, DM_DMCONTROL_HARTSELLO |
+			DM_DMCONTROL_HARTSELHI | DM_DMCONTROL_DMACTIVE |
+			DM_DMCONTROL_HASEL);
+	if (result != ERROR_OK)
+		return result;
+
+
+	result = dm_read(target, &dmcontrol, DM_DMCONTROL);
+	if (result != ERROR_OK)
+		return result;
+
+	/* FIXME: We should poll for dmactive==1 as the debug module
+	 * may need some time to actually activate.
+	 * See https://github.com/riscv/riscv-debug-spec/pull/566
+	 */
+	if (!get_field(dmcontrol, DM_DMCONTROL_DMACTIVE)) {
+		LOG_TARGET_ERROR(target, "Debug Module did not become active.");
+		LOG_DEBUG_REG(target, DM_DMCONTROL, dmcontrol);
+		return ERROR_FAIL;
+	}
+
+	/* The DM has been reset and has successfully came out of the reset (dmactive=1):
+	 * - either the reset has been performed during this call to examine_dm() (above);
+	 * - or the reset had already happened in an earlier call of examine_dm() when
+	 *   examining a different hart.
+	 */
+	dm->was_reset = true;
+
+	dm->hasel_supported = get_field(dmcontrol, DM_DMCONTROL_HASEL);
+
+	uint32_t hartsel =
+		(get_field(dmcontrol, DM_DMCONTROL_HARTSELHI) <<
+		 DM_DMCONTROL_HARTSELLO_LENGTH) |
+		get_field(dmcontrol, DM_DMCONTROL_HARTSELLO);
+
+	/* Before doing anything else we must first enumerate the harts. */
+	const int max_hart_count = MIN(RISCV_MAX_HARTS, hartsel + 1);
+	if (dm->hart_count < 0) {
+		for (int i = 0; i < max_hart_count; ++i) {
+			/* TODO: This is extremely similar to
+			 * riscv013_get_hart_state().
+			 * It would be best to reuse the code.
+			 */
+			result = dm013_select_hart(target, i);
+			if (result != ERROR_OK)
+				return result;
+
+			uint32_t s;
+			result = dmstatus_read(target, &s, /*authenticated*/ true);
+			if (result != ERROR_OK)
+				return result;
+
+			if (get_field(s, DM_DMSTATUS_ANYNONEXISTENT))
+				break;
+
+			dm->hart_count = i + 1;
+
+			if (get_field(s, DM_DMSTATUS_ANYHAVERESET)) {
+				dmcontrol = DM_DMCONTROL_DMACTIVE | DM_DMCONTROL_ACKHAVERESET;
+				/* TODO: Check `abstractcs.busy` here. */
+				dmcontrol = set_dmcontrol_hartsel(dmcontrol, i);
+				result = dm_write(target, DM_DMCONTROL, dmcontrol);
+				if (result != ERROR_OK)
+					return result;
+			}
+		}
+		LOG_TARGET_DEBUG(target, "Detected %d harts.", dm->hart_count);
+	}
+
+	if (dm->hart_count <= 0) {
+		LOG_TARGET_ERROR(target, "No harts found!");
+		return ERROR_FAIL;
+	}
+
+	dm->was_examined = true;
+	return ERROR_OK;
+}
+
 static int examine(struct target *target)
 {
 	/* We reset target state in case if something goes wrong during examine:
@@ -1969,38 +2088,17 @@ static int examine(struct target *target)
 		return ERROR_FAIL;
 	}
 
-	/* Reset the Debug Module. */
-	dm013_info_t *dm = get_dm(target);
-	if (!dm)
-		return ERROR_FAIL;
-	if (!dm->was_reset) {
-		dm_write(target, DM_DMCONTROL, 0);
-		dm_write(target, DM_DMCONTROL, DM_DMCONTROL_DMACTIVE);
-		dm->was_reset = true;
-	}
+	int result = examine_dm(target);
+	if (result != ERROR_OK)
+		return result;
+
+	result = dm013_select_target(target);
+	if (result != ERROR_OK)
+		return result;
+
 	/* We're here because we're uncertain about the state of the target. That
 	 * includes our progbuf cache. */
 	riscv013_invalidate_cached_progbuf(target);
-
-	dm_write(target, DM_DMCONTROL, DM_DMCONTROL_HARTSELLO |
-			DM_DMCONTROL_HARTSELHI | DM_DMCONTROL_DMACTIVE |
-			DM_DMCONTROL_HASEL);
-	dm->current_hartid = HART_INDEX_UNKNOWN;
-	uint32_t dmcontrol;
-	if (dm_read(target, &dmcontrol, DM_DMCONTROL) != ERROR_OK)
-		return ERROR_FAIL;
-	/* Ensure the HART_INDEX_UNKNOWN is flushed out */
-	if (dm013_select_hart(target, 0) != ERROR_OK)
-		return ERROR_FAIL;
-
-
-	if (!get_field(dmcontrol, DM_DMCONTROL_DMACTIVE)) {
-		LOG_TARGET_ERROR(target, "Debug Module did not become active. dmcontrol=0x%x",
-				dmcontrol);
-		return ERROR_FAIL;
-	}
-
-	dm->hasel_supported = get_field(dmcontrol, DM_DMCONTROL_HASEL);
 
 	uint32_t dmstatus;
 	if (dmstatus_read(target, &dmstatus, false) != ERROR_OK)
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1911,6 +1927,135 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
 	return ERROR_OK;
 }
 
+static int check_ac_busy_clear(struct target *target)
```
MarekVCodasip: Please, consider changing the name of this function. Something akin to `wait_for_idle_if_needed()` as it is an extension of wait_for_idle()

src/target/riscv/riscv-013.c
```diff
@@ -140,6 +142,10 @@ typedef struct {
 	/* The program buffer stores executable code. 0 is an illegal instruction,
 	 * so we use 0 to mean the cached value is invalid. */
 	uint32_t progbuf_cache[16];
+
+	/* Some operations are illegal when an abstract command is running.
+	 * The field is used to track this. */
+	bool ac_busy;
```
MarekVCodasip: Similarly to the later function check_ac_busy_clear(), please consider renaming this field to `abtractcs_busy` or `abstract_cmd_busy` or similar.

MarekVCodasip (commented):
Thanks for the patch.

As for the code, I have minor comments about naming.

As for the whole concept of this change:
  - In function examine_dm() we believe with @JanMatCodasip that the check of busy is warranted.
  - In the other functions it looks to us that `assert(!ac_busy)` would be more appropriate. Because the preceding command should always clean up after itself and leave `abstractcs.busy == 0`. Let us know what do you think and what was the rationale behind these checks to many places in the code. We don't mind keeping these as is, but still believe asserts are more appropriate in most places.

Thank you.

JanMatCodasip:
> In the other functions it looks to us that assert(!ac_busy) would be more appropriate. 

Or, possibly `LOG_ERROR("BUG: Unexpected abstractcs.busy ...");` could be used if asserts seem too harsh. 

Still, it seems to me that OpenOCD code itself should ensure that DM is left in a clean state (including abstractcs.busy==0) when any command completes. That means should be possible to start a new command without checking if busy==0. _(An exception from this rule could be malfunctioning hardware in which abstractcs.busy remains stuck at 1 for some reason, and I am afraid we cannot do much about that in OpenOCD in general.)_ 

If there is a command that can leave the DM in an unclean state, then that place should be fixed. Safety checks at the beginning of other commands seem like a good idea, but please preferably in the form of assertions or LOG_ERROR("BUG: ..."), making it very clear to the reader that these are safety checks that guard against cases that ideally should never happen.

In any case, thank you for the time invested into the patch. For example the refactor of examine is beneficial in its own right :+1:

en-sc:
> * In the other functions it looks to us that `assert(!ac_busy)` would be more appropriate. Because the preceding command should always clean up after itself and leave `abstractcs.busy == 0`.

Currently, in general, working with abstract commands is something along the lines of:
* Issue a command (there are quite a few ways to do this).
* Call `wait_for_idle()`.

However, [the loop in `wait_for_idle()` is time-bound](https://github.com/riscv-collab/riscv-openocd/pull/1023/files#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63R915), so it can not be guaranteed that after the call to  `wait_for_idle()` `abstractcs.busy` is clear.

So, let's consider the following OpenOCD session:
* User writes memory -> the write takes a long time -> `wait_for_idle()` times out -> an `ERROR_TIMEOUT_REACHED` is returned.
* However, the OpenOCD does not exit if a command returns an error.
* OpenOCD polls targets -> `hartsel` is changed while `abstractcs.busy` is high -> the rule is broken by the debugger, not by the HW.

I see a couple of options how to fix this:
1. Remove the time bound on the loop in `wait_for_idle()` -- This is bad, since there will be now way to stop OpenOCD in such loop (apart from SIGKILL).
2. Insert a check at the start of all high-level interface functions -- I have decided against it, since there are quite a of these functions and there is no guarantee that this will be sufficient (one lower level function may set `abstractcs.busy` and a fallback may not check that `abstractcs.busy` is clear).
3. Insert the checks directly in front of prohibited operations -- This approach was selected.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -140,6 +142,10 @@ typedef struct {
 	/* The program buffer stores executable code. 0 is an illegal instruction,
 	 * so we use 0 to mean the cached value is invalid. */
 	uint32_t progbuf_cache[16];
+
+	/* Some operations are illegal when an abstract command is running.
+	 * The field is used to track this. */
+	bool abtractcs_busy;
```
JanMatCodasip: I have tried to expand the comment, to make it clearer to the reader what the purpose of this flag is. Please let me know if you like it.

Also a typo in the identifier was fixed: `abtractcs_busy` --> `abstractcs_busy`

Or even better, the item can be renamed to signify that the abstract command **may** be busy, e.g. `abstract_cmd_maybe_busy` (or similar).

```suggestion
	/* Some operations are illegal when an abstract command is running.
	 * The field is used to track whether the last command timed out, and abstractcs.busy
	 * may have remained set. In that case we may need to re-check the busy state 
	 * at the beginning of the next command. */
	bool abstract_cmd_maybe_busy;
```

src/target/riscv/riscv-013.c
```diff
@@ -874,6 +879,12 @@ static int wait_for_idle(struct target *target, uint32_t *abstractcs)
 	assert(target);
 	assert(abstractcs);
 
+	dm013_info_t *dm = get_dm(target);
+	if (!dm) {
+		*abstractcs = 0;
+		return ERROR_FAIL;
+	}
```
JanMatCodasip: Can this situation happen in real life, or would that be a programmer error to call wait_for_idle() before the target is assigned to a DM?

If it is the latter, I'd recommend to add `LOG_ERROR("BUG: ...)` or use `assert(dm)` instead.

```suggestion
	if (!dm) {
		*abstractcs = 0;
		LOG_ERROR("BUG: Target %s is not assigned to any RISC-V debug module", target_name(target));
		return ERROR_FAIL;
	}
```

src/target/riscv/riscv-013.c
```diff
@@ -896,6 +908,10 @@ static int wait_for_idle(struct target *target, uint32_t *abstractcs)
 		riscv_command_timeout_sec,
 		*abstractcs);
 
+	if (!dm->abtractcs_busy)
+		LOG_TARGET_ERROR(target, "BUG: dm->abtractcs_busy was not set in time!");
+	dm->abtractcs_busy = true;
+
```
JanMatCodasip: It looks to me that this safety check can be moved to the top of the function. I believe it is the caller's responsibility to set `dm->abstractcs_busy = true`, so we can move this check to the top.

Is that correct understanding? If so, I'd prefer to move this block.

Maybe the error message can also be updated to make it even clearer (if you like):

```suggestion
	if (!dm->abtractcs_busy)
		LOG_TARGET_ERROR(target, "BUG: dm->abtractcs_busy was not set at the start of abstract command");
	dm->abtractcs_busy = true;

```

src/target/riscv/riscv-013.c
```diff
@@ -1911,6 +1927,135 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
...
+static int wait_for_idle_if_needed(struct target *target)
+{
+	dm013_info_t *dm = get_dm(target);
+	if (!dm)
+		return ERROR_FAIL;
+	if (!dm->abtractcs_busy)
+		return ERROR_OK;
+
+	uint32_t abstractcs;
+	int result = wait_for_idle(target, &abstractcs);
```
JanMatCodasip: Putting comments here may help the reader understand the broader context of this function:

```suggestion
	if (!dm->abtractcs_busy)
		/* The previous abstract command ended correctly
		 * and busy was cleared. No need to do anything. */
		return ERROR_OK;

	/* The previous abstract command timed out and abstractcs.busy 
	 * may have remained set. Wait for it to get cleared. */
	uint32_t abstractcs;
	int result = wait_for_idle(target, &abstractcs);
```

src/target/riscv/riscv-013.c
```diff
@@ -1911,6 +1927,135 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
...
+		 */
+		dmcontrol = (dmcontrol & DM_DMCONTROL_HARTSELLO) |
+			(dmcontrol & DM_DMCONTROL_HARTSELHI);
+		result = dm_write(target, DM_DMCONTROL, dmcontrol);
+		if (result != ERROR_OK)
+			return result;
+	} else {
+		/* If the DM was not reset, an abstract command can be executed
+		 * at the moment.
+		 */
```
JanMatCodasip: I have tried to update the wording for clarity

```suggestion
		/* If the DM was already reset (when examining a different hart).
		 * No need to reset it again. But for safety, assume that an abstract 
		 * command might be in progress at the moment.
		 */
```

src/target/riscv/riscv-013.c
```diff
@@ -1911,6 +1927,135 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
...
+		return result;
+
+	if (!get_field(dmcontrol, DM_DMCONTROL_DMACTIVE)) {
+		LOG_TARGET_ERROR(target, "Debug Module did not become active.");
+		LOG_DEBUG_REG(target, DM_DMCONTROL, dmcontrol);
+		return ERROR_FAIL;
+	}
+
+	/* Note, that this does not mean the DM was just reset. */
+	dm->was_reset = true;
```
JanMatCodasip: This comment was a bit hard to grasp to me. 

Instead, I'd recommend to move the `was_reset = true` to line 1974 where the reset actually takes place, and the comment can be removed.

src/target/riscv/riscv-013.c
```diff
@@ -1911,6 +1927,135 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
...
+			DM_DMCONTROL_HASEL);
+	if (result != ERROR_OK)
+		return result;
+
+
+	result = dm_read(target, &dmcontrol, DM_DMCONTROL);
+	if (result != ERROR_OK)
+		return result;
+
+	if (!get_field(dmcontrol, DM_DMCONTROL_DMACTIVE)) {
```
JanMatCodasip, line 2012: This would be an unrelated improvement (for a separate merge request), so I'd recommend to put here at least a FIXME comment.

```suggestion
	/* FIXME: We should poll for dmactive==1 as the debug module 
	 * may need some time to actually activate.
	 */
	if (!get_field(dmcontrol, DM_DMCONTROL_DMACTIVE)) {
```

src/target/riscv/riscv-013.c
```diff
@@ -1911,6 +1927,135 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
...
+		if (result != ERROR_OK)
+			return result;
+		/* Initiate the reset (`dmcontrol.dmactive == 0`) leaving
+		 * `dmcontrol.hartsel` the same.
+		 */
+		dmcontrol = (dmcontrol & DM_DMCONTROL_HARTSELLO) |
+			(dmcontrol & DM_DMCONTROL_HARTSELHI);
+		result = dm_write(target, DM_DMCONTROL, dmcontrol);
+		if (result != ERROR_OK)
+			return result;
```
JanMatCodasip, line 1983:
```suggestion
		result = dm_write(target, DM_DMCONTROL, dmcontrol);
		if (result != ERROR_OK)
			return result;
		/* FIXME: We should poll dmcontrol until dmactive becomes 0 */			
```

JanMatCodasip:
@en-sc:

> Currently, in general, working with abstract commands is something along the lines of ...
> 
> However, [the loop in `wait_for_idle()` is time-bound](https://github.com/riscv-collab/riscv-openocd/pull/1023/files#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63R915), so ...
> 
> ...

Thank you for the explanation. It really helped me to understand the idea behind the proposed changes.

> I see a couple of options how to fix this:
> 
>     1. ...
> 
>     2. ...
> 
>     3. Insert the checks directly in front of prohibited operations -- This approach was selected.

For completeness, I can imagine two more options: 
- Leave the current state (the user has been warned about strange behavior of the HW and can restart the session);
- Un-examine the targets and DM when busy timeout is encountered, preventing further operations until restarted and re-examined.

But anyway, I believe your chosen solution (point 3 above) is fine and I am in favor of it. I have posted several comments with the aim to make the intention behind the changes more clear to the reader of the code. Please let me know if you are OK with the suggestions. I don't expect to have any more review comments than what I have posted so far.

Thank you.

src/target/riscv/riscv-013.c
en-sc: Thanks for the suggestion. This is much better wording!
Addressed.
Although I have changed the comment a bit:
`at the beginning of the next command` -> `before executing these operations` (I think it's a bit more specific).

src/target/riscv/riscv-013.c
en-sc: Addressed. Used `LOG_ERROR("BUG:...)".

I have decided against an `assert()` since `wait_for_idle()` is quite often called in error handling, so it would be easy to miss something like:

```
int foo()
{
  if (bar() != ERROR_OK) /* bar() executes a command after calling get_dm() -> returns ERROR_FAIL if get_dm() returned NULL */
    goto error;
    ...
 error:
    wait_for_idle();
    return ERROR_FAIL;
}
```

In general, I don't like the fact that `get_dm()` allocates a new DM and can return NULL. I would like for allocation of a DM to be an explicit step in the examination process and for `get_dm()` to be a simple field accessor.

AFAIU, the only thing preventing such approach is the authentication mechanism -- the only way to authenticate a DM is to attempt an examination (to know the version of the spec used) and then to re-attempt it after issuing appropriate `riscv auth write/read` commands (here a DM is needed on a not-yet-examined target).

I think this mechanism should be re-imagined.

I will create a patch with a proposal, though I'm not sure when exactly :).

src/target/riscv/riscv-013.c
en-sc: I would like to leave it as-is.

It is convenient to use `wait_for_idle()` as a default error handler in case there was a call to a subroutine which may execute an abstract command but may fail before executing the command.

I am not ready to pledge that that there are no such uses currently. 

IMHO, it is desirable for `wait_for_idle()` to be a NOP (apart from one read of `abstractcs`) in case there was no abstract command execution to begin with.

src/target/riscv/riscv-013.c
en-sc: Great clarification! Addressed.

src/target/riscv/riscv-013.c
en-sc: Thanks! Addressed (the wording is a bit different, please check).

src/target/riscv/riscv-013.c
en-sc: I would like to leave it as-is.
On the line you mentioned reset of the DM is initiated.
However, we can not be sure the DM successfully came out of the reset.
It can be established only after `dmcontrol.dmactive` was observed in a value read from the DM, which is exactly what happens here.

src/target/riscv/riscv-013.c
en-sc: Great catch! Addressed.

src/target/riscv/riscv-013.c
en-sc: Great catch! Addressed.

src/target/riscv/riscv-013.c
JanMatCodasip: Ok, understood.

In that case, how about updating the comment as follows, to make it clearer what the `was_reset = true` actually means:

```suggestion
	/* The DM has been reset and has successfully come out of the reset (dmactive=1):
	 * - either the reset has been performed by this function (above);
	 * - or the reset had already happened in an earlier call of examine_dm() when
	 *   examining a different hart.
	 */
	dm->was_reset = true;
```

src/target/riscv/riscv-013.c
JanMatCodasip: Ok, understood, thank you for explanation.

Still, could we still change the wording a bit to make it clearer what has happened? For example as shown below:

```c
	if (!dm->abstract_cmd_maybe_busy)
		LOG_TARGET_ERROR(target, "BUG: dm->abstract_cmd_maybe_busy had not been set when starting an abstract command");
	dm->abstract_cmd_maybe_busy = true;
```

JanMatCodasip (dismissed):
This looks good to me, thank you, approved.

There are only two last threads open that deal with less important items (log prints and comments).

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -1911,6 +1932,151 @@ static int set_group(struct target *target, bool *supported, unsigned int group,
...
+	 * See https://github.com/riscv/riscv-debug-spec/pull/566
+	 */
+	if (!get_field(dmcontrol, DM_DMCONTROL_DMACTIVE)) {
+		LOG_TARGET_ERROR(target, "Debug Module did not become active.");
+		LOG_DEBUG_REG(target, DM_DMCONTROL, dmcontrol);
+		return ERROR_FAIL;
+	}
+
+	/* The DM has been reset and has successfully came out of the reset (dmactive=1):
+	 * - either the reset has been performed during this call to examkine_dm() (above);
```
JanMatCodasip: Just a little typo:

```suggestion
	 * - either the reset has been performed during this call to examine_dm() (above);
```

JanMatCodasip (dismissed):

src/target/riscv/riscv-013.c
en-sc: Sorry. Addressed.

JanMatCodasip (approved):

PR #1025, opened 2024-03-01
target/riscv: decode DMI scans in batch access

en-sc:
This allows to merge the implementation in `batch.c` with the one in `riscv-013.c`.

Change-Id: Ic3821a9ce2d75a7c6e618074679595ddefb14cfc

### DIFF
--- a/src/target/riscv/batch.h
+++ b/src/target/riscv/batch.h
@@ -78,4 +78,10 @@ size_t riscv_batch_available_scans(struct riscv_batch *batch);
 /* Return true iff the last scan in the batch returned DMI_OP_BUSY. */
 bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch);
 
+/* TODO: The function is defined in `riscv-013.c`. This is done to reduce the
+ * diff of the commit. The intention is to move the function definition to
+ * a separate module (e.g. `riscv013-jtag-dtm.c/h`) in another commit. */
+void riscv_decode_dmi_scan(const struct target *target, int idle, const struct scan_field *field,
+		bool discard_in);
+
 #endif
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -363,10 +363,10 @@ static uint32_t set_dmcontrol_hartsel(uint32_t initial, int hart_index)
 	return initial;
 }
 
-static unsigned int decode_dm(char *text, unsigned int address, unsigned int data)
+static unsigned int decode_dmi(const struct target *target, char *text, uint32_t address, uint32_t data)
 {
 	static const struct {
-		unsigned int address;
+		uint32_t address;
 		enum riscv_debug_reg_ordinal ordinal;
 	} description[] = {
 		{DM_DMCONTROL, DM_DMCONTROL_ORDINAL},
@@ -377,7 +377,7 @@ static unsigned int decode_dm(char *text, unsigned int address, unsigned int dat
 	};
 
 	for (unsigned i = 0; i < ARRAY_SIZE(description); i++) {
-		if (description[i].address == address) {
+		if (riscv_get_dmi_address(target, description[i].address) == address) {
 			const riscv_debug_reg_ctx_t context = {
 				.XLEN = { .value = 0, .is_set = false },
 				.DXLEN = { .value = 0, .is_set = false },
@@ -392,19 +392,7 @@ static unsigned int decode_dm(char *text, unsigned int address, unsigned int dat
 	return 0;
 }
 
-static unsigned int decode_dmi(struct target *target, char *text, unsigned int address,
-		unsigned int data)
-{
-	dm013_info_t *dm = get_dm(target);
-	if (!dm) {
-		if (text)
-			text[0] = '\0';
-		return 0;
-	}
-	return decode_dm(text, address - dm->base, data);
-}
-
-static void dump_field(struct target *target, int idle, const struct scan_field *field, bool discard_in)
+void riscv_decode_dmi_scan(const struct target *target, int idle, const struct scan_field *field, bool discard_in)
 {
 	static const char * const op_string[] = {"-", "r", "w", "?"};
 	static const char * const status_string[] = {"+", "?", "F", "b"};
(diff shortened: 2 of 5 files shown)


src/target/riscv/batch.h
```diff
@@ -78,4 +78,9 @@ size_t riscv_batch_available_scans(struct riscv_batch *batch);
 /* Return true iff the last scan in the batch returned DMI_OP_BUSY. */
 bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch);
 
+/* TODO: The function is defined in `riscv-013.c`. This is done to reduce the
+ * diff of the commit. The intention is to move the function defenition to
```
JanMatCodasip: Just a little typo:

```suggestion
 * diff of the commit. The intention is to move the function definition to
```

src/target/riscv/batch.h
```diff
@@ -78,4 +78,9 @@ size_t riscv_batch_available_scans(struct riscv_batch *batch);
...
+ * `batch.c` in a separate commit. */
```
JanMatCodasip: > The intention is to move the function definition to batch.c

I would prefer to only keep batch-related code in `batch.c`. Since the code to decode JTAG transactions is generic and not strictly related to batches, putting it to batch.c does not seem to be correct.

Instead, it could go to a new module, e.g. `riscv013-jtag-dtm.c/h`, `riscv013-logging.c/h`, `riscv013-utils.c/h`, depending on your preference.

src/target/riscv/riscv-013.c
```diff
@@ -387,52 +387,47 @@ static unsigned int decode_dm(char *text, unsigned int address, unsigned int dat
...
-	if (!dm) {
-		if (text)
-			text[0] = '\0';
-		return 0;
-	}
-	return decode_dm(text, address - dm->base, data);
-}
-
-static void dump_field(struct target *target, int idle, const struct scan_field *field, bool discard_in)
+void riscv_dump_field(const struct target *target, int idle, const struct scan_field *field, bool discard_in)
```
JanMatCodasip: While making changes here, I'd recommend to rename the function to `riscv_decode_jtag_scan()`, `riscv_log_jtag_scan()` or similar. The word `field` is probably not be very clear to readers.

src/target/riscv/riscv-013.c
```diff
@@ -387,52 +387,47 @@ static unsigned int decode_dm(char *text, unsigned int address, unsigned int dat
...
+		LOG_DEBUG("%db %s %08x @%02x -> %s %08x @%02x; %di",
+				field->num_bits, op_string[out_op], out_data, out_address,
+				status_string[in_op], in_data, in_address, idle);
+
+		if (!discard_in && in_op == DTM_DMI_OP_SUCCESS) {
+			char in_decoded[decode_dmi(target, NULL, in_address, in_data) + 1];
+			decode_dmi(target, in_decoded, in_address, in_data);
+			/* TODO: The decoded value is based on the assumtion
+			 * that DMI address was not changed between operations.
+			 * This is true for Spike, but may differ on HW.*/
```
JanMatCodasip: Can you please check if I understood the comment correctly.

If that is correct understanding, I would recommend to extend the comment, something along these lines:

```suggestion
			/* FIXME: OpenOCD should keep track of the last DMI address from which
			 * we have initiated a DMI read so that the obtained data can always be
			 * correctly decoded.
			 * The current code assumes that the hardware provides the read address 
			 * in the dmi.address field when returning the dmi.data. That is however not
			 * required by the spec, and therefore not guaranteed. Spike ISA simulator 
			 * behaves like that but other RISC-V debug implementations may not. 
			 */
```

Alternatively, an issue can be opened with the verbose description, and a only a short FIXME comment + the issue link can be put into the code.

JanMatCodasip (changes requested):

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/batch.h

src/target/riscv/batch.h

src/target/riscv/riscv-013.c
```diff
@@ -392,52 +392,50 @@ static unsigned int decode_dm(char *text, unsigned int address, unsigned int dat
...
-	log_printf_lf(LOG_LVL_DEBUG, __FILE__, __LINE__, __func__,
-			"%db %s %08x @%02x -> %s %08x @%02x; %di",
-			field->num_bits, op_string[out_op], out_data, out_address,
-			status_string[in_op], in_data, in_address, idle);
-
+	assert(field->out_value);
+	const uint64_t out = buf_get_u64(field->out_value, 0, field->num_bits);
+	const unsigned int out_op = get_field(out, DTM_DMI_OP);
+	const unsigned int out_data = get_field(out, DTM_DMI_DATA);
+	const unsigned int out_address = out >> DTM_DMI_ADDRESS_OFFSET;
```
JanMatCodasip: Just a detail that I noticed. For "data" and "address", it would be more suitable to use a fixed-size integer. The data is always exactly 32 bits per the spec, and address is up to 32 bits long.

(It can also be done in a follow-up commit, if you'd prefer that.)

```suggestion
	const uint32_t out_data = get_field32(out, DTM_DMI_DATA);
	const uint32_t out_address = (uint32_t)(out >> DTM_DMI_ADDRESS_OFFSET);
```

src/target/riscv/riscv-013.c
```diff
@@ -392,52 +392,50 @@ static unsigned int decode_dm(char *text, unsigned int address, unsigned int dat
...
+
+	if (field->in_value) {
+		const uint64_t in = buf_get_u64(field->in_value, 0, field->num_bits);
+		const unsigned int in_op = get_field(in, DTM_DMI_OP);
+		const unsigned int in_data = get_field(in, DTM_DMI_DATA);
+		const unsigned int in_address = in >> DTM_DMI_ADDRESS_OFFSET;
```
JanMatCodasip: (same as in previous comment)

```suggestion
		const uint32_t in_data = get_field32(in, DTM_DMI_DATA);
		const uint32_t in_address = (uint32_t)(in >> DTM_DMI_ADDRESS_OFFSET);
```

JanMatCodasip (dismissed):
Thank you, LGTM.

(The last two small comments about data types can be resolved either now or in a follow-up PR, depending on your preference.)

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

en-sc:
@JanMatCodasip, please, take a look again. I would like to merge this one.

JanMatCodasip (approved):
Thank you, LGTM.

Issue #333, opened 2018-11-19
progbufsize of 1 is not supported


atthecodeface:
For memory read and write, at least, a progbufsize of 1 is not supported.

I have patches to enable this, but execute_fence fails too I believe, and I have not patched that.
I cannot provide a pull request as is right now - I know the patch needs coding style changes and I probably have not cleaned it yet. But this is a place-holder.

PR #1028, opened 2024-03-07
target/riscv: reset delays during batch scans

en-sc:
Prior to the commit, it was impossible to get a `dmi.busy` responce on a scan in a batch that was not the first scan in the batch.

Change-Id: Ib0714ecaf7d2e11878140d16d9aa6152ff20f1e9

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -89,7 +89,8 @@ bool riscv_batch_full(struct riscv_batch *batch)
 	return riscv_batch_available_scans(batch) == 0;
 }
 
-int riscv_batch_run(struct riscv_batch *batch)
+int riscv_batch_run(struct riscv_batch *batch, bool resets_delays,
+		size_t reset_delays_after)
 {
 	if (batch->used_scans == 0) {
 		LOG_TARGET_DEBUG(batch->target, "Ignoring empty batch.");
@@ -104,7 +105,9 @@ int riscv_batch_run(struct riscv_batch *batch)
 		else
 			jtag_add_dr_scan(batch->target->tap, 1, batch->fields + i, TAP_IDLE);
 
-		if (batch->idle_count > 0)
+		const bool delays_were_reset = resets_delays
+			&& (i >= reset_delays_after);
+		if (batch->idle_count > 0 && !delays_were_reset)
 			jtag_add_runtest(batch->idle_count, TAP_IDLE);
 	}
 
--- a/src/target/riscv/batch.h
+++ b/src/target/riscv/batch.h
@@ -55,8 +55,15 @@ void riscv_batch_free(struct riscv_batch *batch);
 /* Checks to see if this batch is full. */
 bool riscv_batch_full(struct riscv_batch *batch);
 
-/* Executes this scan batch. */
-int riscv_batch_run(struct riscv_batch *batch);
+/* Executes this batch of JTAG DTM DMI scans.
+ *
+ * If resets_delays is true, the algorithm will stop inserting idle cycles
+ * (JTAG Run-Test Idle) after "reset_delays_after" number of scans is
+ * performed.  This is useful for stress-testing of RISC-V algorithms in
+ * OpenOCD that are based on batches.
+ */
+int riscv_batch_run(struct riscv_batch *batch, bool resets_delays,
+		size_t reset_delays_after);
 
 /* Adds a DM register write to this batch. */
 void riscv_batch_add_dm_write(struct riscv_batch *batch, uint64_t address, uint32_t data,
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -496,6 +496,24 @@ static void increase_dmi_busy_delay(struct target *target)
 	dtmcontrol_scan(target, DTM_DTMCS_DMIRESET, NULL /* discard result */);
 }
 
+static void decrement_reset_delays_counter(struct target *target, size_t finished_scans)
+{
+	RISCV_INFO(r);
+	if (r->reset_delays_wait < 0) {
+		assert(r->reset_delays_wait == -1);
+		return;
+	}
+	if ((size_t)r->reset_delays_wait >= finished_scans) {
+		r->reset_delays_wait -= finished_scans;
+		return;
+	}
+	r->reset_delays_wait = -1;
+	LOG_TARGET_DEBUG(target,
+			"resetting learned delays (reset_delays_wait counter expired)");
+	RISCV013_INFO(info);
+	info->dmi_busy_delay = 0;
+	info->ac_busy_delay = 0;
+}
 /**
  * exec: If this is set, assume the scan results in an execution, so more
  * run-test/idle cycles may be required.
@@ -505,7 +523,6 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 		bool exec)
 {
 	riscv013_info_t *info = get_info(target);
-	RISCV_INFO(r);
 	unsigned num_bits = info->abits + DTM_DMI_OP_LENGTH + DTM_DMI_DATA_LENGTH;
 	size_t num_bytes = (num_bits + 7) / 8;
 	uint8_t in[num_bytes];
@@ -517,14 +534,7 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 	};
 	riscv_bscan_tunneled_scan_context_t bscan_ctxt;
 
-	if (r->reset_delays_wait >= 0) {
-		r->reset_delays_wait--;
-		if (r->reset_delays_wait < 0) {
-			LOG_TARGET_DEBUG(target, "reset_delays_wait done");
-			info->dmi_busy_delay = 0;
-			info->ac_busy_delay = 0;
-		}
-	}
+	decrement_reset_delays_counter(target, 1);
 
 	memset(in, 0, num_bytes);
 	memset(out, 0, num_bytes);
@@ -2617,19 +2627,16 @@ static int sb_write_address(struct target *target, target_addr_t address,
 		(uint32_t)address, false, ensure_success);
 }
 
-static int batch_run(const struct target *target, struct riscv_batch *batch)
+static int batch_run(struct target *target, struct riscv_batch *batch)
 {
-	RISCV013_INFO(info);
 	RISCV_INFO(r);
-	if (r->reset_delays_wait >= 0) {
-		r->reset_delays_wait -= batch->used_scans;
-		if (r->reset_delays_wait <= 0) {
-			batch->idle_count = 0;
-			info->dmi_busy_delay = 0;
-			info->ac_busy_delay = 0;
-		}
-	}
-	return riscv_batch_run(batch);
+	const int result = riscv_batch_run(batch, /*resets_delays*/  r->reset_delays_wait >= 0,
+			r->reset_delays_wait);
+	/* TODO: `finished_scans` should be the number of scans that have
+	 * finished, not the number of scans scheduled. */
+	const size_t finished_scans = batch->used_scans;
+	decrement_reset_delays_counter(target, finished_scans);
+	return result;
 }
 
 static int sba_supports_access(struct target *target, unsigned int size_bytes)
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -450,6 +450,7 @@ static int riscv_init_target(struct command_context *cmd_ctx,
 	LOG_TARGET_DEBUG(target, "riscv_init_target()");
 	RISCV_INFO(info);
 	info->cmd_ctx = cmd_ctx;
+	info->reset_delays_wait = -1;
 
 	select_dtmcontrol.num_bits = target->tap->ir_length;
 	select_dbus.num_bits = target->tap->ir_length;
@@ -3850,10 +3851,8 @@ COMMAND_HANDLER(riscv_reset_delays)
 {
 	int wait = 0;
 
-	if (CMD_ARGC > 1) {
-		LOG_ERROR("Command takes at most one argument");
+	if (CMD_ARGC > 1)
 		return ERROR_COMMAND_SYNTAX_ERROR;
-	}
 
 	if (CMD_ARGC == 1)
 		COMMAND_PARSE_NUMBER(int, CMD_ARGV[0], wait);
@@ -6455,7 +6454,7 @@ int riscv_init_registers(struct target *target)
 	return ERROR_OK;
 }
 
-void riscv_add_bscan_tunneled_scan(struct target *target, struct scan_field *field,
+void riscv_add_bscan_tunneled_scan(struct target *target, const struct scan_field *field,
 					riscv_bscan_tunneled_scan_context_t *ctxt)
 {
 	jtag_add_ir_scan(target->tap, &select_user4, TAP_IDLE);
--- a/src/target/riscv/riscv.h
+++ b/src/target/riscv/riscv.h
@@ -454,7 +454,7 @@ void riscv_semihosting_init(struct target *target);
 
 enum semihosting_result riscv_semihosting(struct target *target, int *retval);
 
-void riscv_add_bscan_tunneled_scan(struct target *target, struct scan_field *field,
+void riscv_add_bscan_tunneled_scan(struct target *target, const struct scan_field *field,
 		riscv_bscan_tunneled_scan_context_t *ctxt);
 
 int riscv_read_by_any_size(struct target *target, target_addr_t address, uint32_t size, uint8_t *buffer);


JanMatCodasip:
@en-sc Please, could you explain in bit more detail what issue this PR is trying to address (e.g. in the commit message). After taking an initial quick look at the code, I was not able to grasp what the issue actually is.

My understanding of the batch code is that:
- after each scan, the corresponding amount of idle cycles is really inserted: [batch.c, line 110](https://github.com/riscv-collab/riscv-openocd/blob/riscv/src/target/riscv/batch.c#L110)
- if the DMI busy situation occurs at any point in the batch (regardless whether it is the first scan or any subsequent), the busy status is sticky, so the rest of the batch will also be marked as busy, therefore it is enough to just check the status of the last operation in the batch - [batch.c, line 248](https://github.com/riscv-collab/riscv-openocd/blob/riscv/src/target/riscv/batch.c#L248)

What am I missing? :)

en-sc:
@JanMatCodasip, you are right, the description is not sufficient.

This commit is related to testing how OpenOCD responds to `dmi.busy`.

Consider testing on Spike (e.g. `riscv-tests/debug` testsuite). Spike returns `dmi.busy` if there were less then a given number of RTI cycles (`required_rti_cycles`) between DR_UPDATE and DR_CAPTURE:
https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/jtag_dtm.cc#L145
https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/jtag_dtm.cc#L202
`required_rti_cycles` gets it's value from `--dmi-rti` CLI argument and is constant throughout the run.

OpenOCD learns this required number of RTI cycles by starting with zero and increasing it if `dmi.busy` is encountered. So the required number of RTI cycles is learned during the first DMI access in the `examine()`.

To induce `dmi.busy` on demand `riscv reset_delays <x>` command is provided. This command initializes `riscv_info::reset_delays_wait` counter to the provided `<x>` value. The counter is decreased before a DMI access and when it reaches zero the learned value of RTI cycles required is reset, so the DMI access results in `dmi.busy`.

E.g. `riscv-tests/debug` testsuite issues `riscv reset_delays <x>` before most operations:
https://github.com/riscv-software-src/riscv-tests/blob/fd01e46314636666b5bed1f08465f734751a4cda/debug/testlib.py#L837

Now consider running a batch of accesses.
Before the change all the accesses in the batch had the same number of RIT cycles in between them.
So either:
* Number of accesses in the batch was greater then the value of `riscv_info::reset_delays_wait` counter and there was no `dmi.busy` throughout the batch.
* Number of accesses in the batch was less or equal then the value of `riscv_info::reset_delays_wait` counter and the first access of the batch resulted in `dmi.busy`.

Therefore it was impossible to encounter `dmi.busy` on any scan of the batch except the first one.

en-sc:
One can argue that the better solution would be to change Spike's `required_rti_cycles` value by exposing it through some other custom DM/DMI register. I have avoided this approach because:
1. This will take more time (changing both Spike and `riscv-tests/debug` testsuite).
2. In this commit I have implemented the mechanism that allows to change the number of RTI cycles used after scans in a batch. I will need this mechanism to implement more efficient register read/writes via abstract commands:

e.g. 32-bit read now involves executing the following DMI operations:
* write command
* read abstractcs
* nop
* read data0
* nop

We don't need all these NOPs (apart from the last one),  since error statuses are sticky. The easiest way to eliminate them is to use a batch. However, after writing abstract command we need more RTI cycles, so that command finishes and `abstractcs.busy` is low by the time it is read.
(I have started implementing this in #1033)

en-sc:
> ... could you explain in bit more detail what issue this PR is trying to address (e.g. in the commit message).

Commit message is changed.

src/target/riscv/batch.c
```diff
@@ -18,7 +18,13 @@
```
JanMatCodasip: My recommendation would be to implement the reset of the DMI scan delay in the batch via simpler, more straightforward code:

- Since there is currently at most one reset of `idle_count` in a given batch, I'd recommend to remove the the list, and instead use just a plain counter do determine at which point in the batch to reset .
- If you prefer to still have the possibility to specify different idle count for each scan in the batch, please allocate an integer value directly for each scan instead of a list: `out->idle_counts = malloc(sizeof(unsigned) * scans);`. The extra bit of used memory is, in my opinion, very good price to pay for code simplicity.
- Please, decrement the `r->reset_delays_wait` directly in the loop in `jtag_fill_queue()` as the scans are added into the jtag queue; instead of doing it in another module in function `maybe_add_delays_reset()`.

src/target/riscv/batch.c
```diff
@@ -249,3 +295,37 @@ bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch)
...
+{
+	assert(!list_empty(&batch->idle_counts));
+
+	idle_count_info_t *entry;
+	list_for_each_entry(entry, &batch->idle_counts, list)
+		if (entry->until_scan > scan_idx)
+			break;
+	assert(!list_entry_is_head(entry, &batch->idle_counts, list));
+	return entry;
+}
```
JanMatCodasip: As noted above, this function becomes redundant when the list is removed.

src/target/riscv/batch.c
```diff
@@ -249,3 +295,37 @@ bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch)
...
+	list_add_tail(&new_entry->list, &old_entry->list);
+	assert(new_entry->until_scan < old_entry->until_scan);
+	/* new entry now defines the range until the scan (non-inclusive) */
+	new_entry->idle_count = old_entry->idle_count;
+	/* old entry now defines the range from the scan (inclusive) */
+	old_entry->idle_count = new_idle;
+	LOG_DEBUG("Will use idle == %zu from scan %zu until scan %zu.", old_entry->idle_count,
+			new_entry->until_scan, old_entry->until_scan);
+	return ERROR_OK;
+}
```
JanMatCodasip: As also noted above, the function jtag_fill_queue can look directly at the value `r->reset_delays_wait` (and decrement it with each added scan), therefore this function would not be needed.

src/target/riscv/riscv-013.c
```diff
@@ -2514,18 +2514,32 @@ static int sb_write_address(struct target *target, target_addr_t address,
...
+		return ERROR_OK;
 	}
+	const int result = riscv_batch_change_idle_used_from_scan(batch, 0, r->reset_delays_wait);
+	r->reset_delays_wait = -1;
+	LOG_TARGET_DEBUG(target, "reset_delays_wait done");
+	RISCV013_INFO(info);
+	info->dmi_busy_delay = 0;
+	info->ac_busy_delay = 0;
+	return result;
+}
```
JanMatCodasip: Per the previous comments: The function `riscv_batch_run()` (and `jtag_fill_queue()`) can work directly with the value r->reset_delays_wait, which makes the function `maybe_add_delays_reset()` unnecessary.

JanMatCodasip:
@en-sc, thank you for explaining the reasons for this change in detail. Now it is very clear to me.

As for using Spike ISA sim for testing of OpenOCD, I agree that Spike will not help us cover all the various edge cases of RISC-V debug protocol (like DMI busy, abstract command busy or system bus busy). As an example, you can take a look at this merge request in which configurable system bus delay was proposed to Spike but only accepted in a reduced form: https://github.com/riscv-software-src/riscv-isa-sim/pull/1513

(Side note: My vision would be to implement a simple simulator of RISC-V debug module + core, much simpler than spike, just a bare minimum, which would however allow to trigger the debug edge cases that are impossible to trigger via a general-purpose simulator like Spike. It is of course unclear whether there will be ever a capacity to do it :/)

For the current pull request, I have posted suggestions how to implement the same thing this using simpler code. Please if you can look at it.

JanMatCodasip (changes requested):
1. For my overall comment, please see above: https://github.com/riscv-collab/riscv-openocd/pull/1028#issuecomment-2049110776

2. In the individual review suggestions, I have proposed how to implement the same capability using simpler code. Please let me know if you like it.

src/target/riscv/batch.c
en-sc: > * Since there is currently at most one reset of `idle_count` in a given batch, I'd recommend to remove the the list, and instead use just a plain counter do determine at which point in the batch to reset .

I will need the ability to change `idle_count` used by a particular scan in my further commits.

> * If you prefer to still have the possibility to specify different idle count for each scan in the batch, please allocate an integer value directly for each scan instead of a list: `out->idle_counts = malloc(sizeof(unsigned) * scans);`. The extra bit of used memory is, in my opinion, very good price to pay for code simplicity.

The main motivation for using a list is not to save memory.

My idea for how operations should be performed in general is the following:
* An high-level interface operation (e.g. reading memory) is split into lower level functions (e.g. read memory via program buffer).
* At the lowest level, each function follows the structure bellow:
1. Fill the batch with appropriate operations.
2. Execute a batch. If `dmi.busy` is encountered -- increase the delay and repeat from the point of failure (with possibility of timeout). No need to refill the batch. (This loop is implemented in #1033 https://github.com/riscv-collab/riscv-openocd/pull/1033/files#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63R2571)
3. Based on the results of the execution, return appropriate error code.

This approach allows to unify handling of `dmi.busy` and make it more robust.

However, the increase of delays proves to be difficult in case of non-homogeneous batch, if an array is used.
Consider a batch scans.
All but the last one should use 2 RTI cycle (`idle_count`), the last should use 1, since it's expected to take less time (e.g. in the beginning each scan start an execution of an abstract command, the last just reads DM register).
`dmi.busy` is encountered on the first scan. We need to increase the `idle_counts` (e.g. by 1) for the scans with the same `idle_count`, since they will return busy also.
So we now use `idle_count` of 3 for the following scans of the same type (executes a command). But what about the last scan with `idle_count` of  1?
I see the following options:
1. Increase all `idle_count`s in the batch by the same value -- drawback: reset (`idle_count == 0`) stops being a reset.
2. Increase all the following scans which had the same `idle_count` -- drawback: this can make it so after the increase the new value became equal with the value that was used by the following scans. So from this point on these scans are treated the same, though they were not the same initially.

I think these drawbacks are quite subtle and the resulting behavior is not intuitive.

I agree, the use of a list makes code more complex. But these drawbacks are eliminated (I can explain how, but I think this comment is already quite long :)).

> * Please, decrement the `r->reset_delays_wait` directly in the loop in `jtag_fill_queue()` as the scans are added into the jtag queue; instead of doing it in another module in function `maybe_add_delays_reset()`.

My goal is to create a general interface to manipulating `idle_count`s used in the batch. I don't think the functionality of 
`reset_delays_wait` should be treated as a special case.

src/target/riscv/batch.c
en-sc: I've replied in [this comment](https://github.com/riscv-collab/riscv-openocd/pull/1028#discussion_r1560895289).

src/target/riscv/batch.c
en-sc: I've replied in [this comment](https://github.com/riscv-collab/riscv-openocd/pull/1028#discussion_r1560895289).

src/target/riscv/riscv-013.c
en-sc: I've replied in [this comment](https://github.com/riscv-collab/riscv-openocd/pull/1028#discussion_r1560895289).

src/target/riscv/batch.c
JanMatCodasip: @en-sc Thank you for explaining your changes even more.

In another words, you are proposing to create sub-batches within a batch in order to be able to manipulate idle counts for a subset of the scans in the batch at once. 

I am sorry but I must respectfully push back on this change.

This would bring a new additional level of complexity into the batch processing. Considering that we do not have an easy way how to test this part of OpenOCD's due to Spike's limitations (as discussed above [1](https://github.com/riscv-collab/riscv-openocd/pull/1028#issuecomment-2046976554), [2](https://github.com/riscv-collab/riscv-openocd/pull/1028#issuecomment-2049110776)), I would like to absolutely avoid making this part of OpenOCD more complex when the same functionality can be delivered in a simpler, more direct way, without creating sub-structures inside the batches. Please see my proposal below.

Also the fact that long explanation was needed, IMO, underlines the principal complexity of the proposed change.

Proposal of a simpler solution:

- For each scan in a batch, store the type of the scan: 
  - The types would be:
    - Simple DMI read or write(analogous to `dmi_read()` or `dmi_write()`) 
    - DMI read or write that triggers abstract command execution (analogous to `dmi_read_exec()` or `dmi_write_exec()`)
    - DMI read or write that triggers a system bus transaction
- Each scan uses the idle count that is valid at the time of starting of the batch
- `riscv_batch_run()` would read the current idle values directly from `r->dmi_busy_delay`, `r->bus_master_write_delay`, etc. That is, single source of authority.
- Re-execution of the batch is possible.
- When re-executing a batch, up to date idle counts would always be used.
- Parameter `idle_count` in `riscv_batch_alloc()` would be removed - no longer needed.
- Reset of learned idle values would be handled by decrementing `r->reset_delays_wait` in fill_jtag_queue().

This proposal covers all the use cases discussed above:

- Different idle values for different scans in the batch
- Possibility to re-execute an existing batch - with the same or different idle values
- Reset of learned delays during batch execution

Please, let me know what you think.

en-sc: @JanMatCodasip, thank you for the suggestion!

My proposal is more complicated than the one you suggest. I would like to try to explain my motivation for this.

> This would bring a new additional level of complexity into the batch processing.

This is true, however, I believe the changes to the processing itself (running the batch) are minimal (just iterate over list entries before iterating over batch elements).

> Considering that we do not have an easy way how to test this part of OpenOCD's due to Spike's limitations (as discussed above [1](https://github.com/riscv-collab/riscv-openocd/pull/1028#issuecomment-2046976554), [2](https://github.com/riscv-collab/riscv-openocd/pull/1028#issuecomment-2049110776)).

This is precisely why I have posted this patch first. With the ability to reset delays on any scan in the batch tests from `riscv-tests/debug` do test this code.

> Also the fact that long explanation was needed, IMO, underlines the principal complexity of the proposed change.

I agree, the implementation is somewhat complex, but the interface is simple. AFAIU, `riscv_batch_change_idle_used_from_scan()` is the one function that could have been simpler. Though it is quite short.

Regarding your proposal:

* I would like for the batch to be decoupled from target. IMHO, the reference to `struct target` should be replaced by a reference to `struct dtm *` (to be implemented). This is more straightforward, since the batches access DM, not the target itself. Reference to target will complicated accessing DM registers via a batch (which I believe should be the only access method) when switching from one target to another or when the target is not yet examined.
* I don't think the set of delay types you propose will be enough. What about Quick Access? What if a target is expected to change state in between the operations (e.g. resuming)?

**TLDR**: I believe the interface you propose will prove to narrow to be used in all the cases, so it will not be possible to use throughout the codebase.

However, I'm obviously biased and, if you insist, I will implement your suggestion instead.

JanMatCodasip: @en-sc Thank you for taking time to explain your ideas further.

> I don't think the set of delay types you propose will be enough. What about Quick Access? What if a target is expected to change state in between the operations (e.g. resuming)?

Agreed. The set of delays in my comment is just an example. For each type of operation that can take a different amount of time (sysbus read, sysbus write, quick access, ...), OpenOCD should keep a separate delay value.

> I would like for the batch to be decoupled from target. IMHO, the reference to struct target should be replaced by a reference to struct dtm * (to be implemented). This is more straightforward, since the batches access DM, not the target itself.

I am neutral to decoupling the target and the batch (I am not against it and at the same time not too eager). 

Below is my updated proposal that can work when batches are decoupled from `target`: The currently valid delays can be passed to riscv_batch_run():

```c
struct riscv_learned_delays
{
	unsigned int dmi_busy_delay;
	unsigned int ac_busy_delay;
	/* ... add other delays: system bus, quick access ... */

	/* when decremented from 1 to 0, the learned delays above are cleared to 0 */
	int reset_delays_wait;
};

struct riscv_info {
	/* ... */
	struct riscv_learned_delays delays;
	/* ... */
};	

// fill batch:
riscv_batch_add_dm_write(batch, DM_SBCS, sbcs_value, DELAY_BASIC_DMI);
riscv_batch_add_dm_write(batch, DM_SBADDRESS0, addr, DELAY_SYSBUS_READ);
// ...

// execute batch:
riscv_batch_run(batch, /* struct riscv_learned_delays* */ &delays);

```

> ... I agree, the implementation is somewhat complex, but the interface is simple.  ...

I am sorry to disagree with this statement. The batch interface would become more difficult to use: the batch user would need to keep track of sub-groups in the batch so that the delays can be properly updated when re-executing the batch.

Furthermore, the subgroups in the batch do not scale well if different types of delays are mixed inside the batch. Imagine a developer wishes to implement writing to memory using progbufsize=1 via a batch (a real-word scenario, https://github.com/riscv-collab/riscv-openocd/issues/333). The resulting batch would look like this in RV32:

```
write to DATA0 (short delay)
write to COMMAND (write address to a GPR, medium delay)
write to DATA0 (short delay)
write to COMMAND (write data to a GPR + execute program buffer, long delay)
<repeat n times>
```

Due to different delays for each of the command, there would be too many sub-groups. The caller would need to go through all the sub-groups and update the delays before re-executing the batch.

Logically, the delay for a DMI operation depends on the nature of the operation itself: on what types of operations get triggered by the DMI scan. That's why I strongly believe that the delays should be assigned to each operation, and subgroups go against this principle (demonstrated on the example with progbufsize=1 above).

> I believe the interface you propose will prove to narrow to be used in all the cases, so it will not be possible to use throughout the codebase.

Please, let me know what concrete use cases have not yet been addressed, so that we can together find a way how to address them using simple, straightforward code. Thank you.

---
---

Off-topic, unrelated to this PR:

> ... via a batch (which I believe should be the only access method) ...

Do I understand it correctly that your future aim would be to use batch access for all DMI accesses? If so, then that sounds a bit too radical to me. In my opinion, both the APIs should remain in place since each of them supports different use cases:

- `dmi_read()`, `dmi_write()`, `...`: For algorithms where performance is not crucial, but where robust execution of DMI operations is desired. DMI busy states are handled gracefully without the need to care about it in the user code. Therefore, the user algorithms can remain very simple (easier to read and maintain, less bug-prone, ...).
- `riscv_batch_*`: For algorithms where performance is important. The user code must be ready for handling errors occurring in the middle of the batch, and know how to properly recover from them. Therefore, the user-written algorithms using batches will generally lead to more complex code, which is a tradeoff for the improved performance.

If you'd like to make larger changes to these two fundamental APIs at some future point, please, it would be good to discuss it first.

en-sc: > Off-topic, unrelated to this PR:
> 
> > ... via a batch (which I believe should be the only access method) ...
> 
> Do I understand it correctly that your future aim would be to use batch access for all DMI accesses? If so, then that sounds a bit too radical to me. In my opinion, both the APIs should remain in place since each of them supports different use cases:
> 
> * `dmi_read()`, `dmi_write()`, `...`: For algorithms where performance is not crucial, but where robust execution of DMI operations is desired. DMI busy states are handled gracefully without the need to care about it in the user code. Therefore, the user algorithms can remain very simple (easier to read and maintain, less bug-prone, ...).
> * `riscv_batch_*`: For algorithms where performance is important. The user code must be ready for handling errors occurring in the middle of the batch, and know how to properly recover from them. Therefore, the user-written algorithms using batches will generally lead to more complex code, which is a tradeoff for the improved performance.
> 
> If you'd like to make larger changes to these two fundamental APIs at some future point, please, it would be good to discuss it first.

The thing is, `dmi_read()`, `dmi_write()`, `...` utilize VLA to store fields. This is against the guidelines (from `doc/manual/style.txt`)
```
- use malloc() to create dynamic arrays. Do @b not use @c alloca
or variable length arrays on the stack. non-MMU hosts(uClinux) and
pthreads require modest and predictable stack usage.
```
So these should be eventually replaced by a `malloc()/calloc()`. And after this change is done it would be just easier to use appropriate batch of two accesses in `dmi_read()`, `dmi_write()`, `...`. (no change to API).

en-sc: I have simplified the commit. I will implement scan types as you suggested in subsequent commits. Please, take a look.

src/target/riscv/batch.c
```diff
@@ -106,7 +107,9 @@ int riscv_batch_run(struct riscv_batch *batch)
 		else
 			jtag_add_dr_scan(batch->target->tap, 1, batch->fields + i, TAP_IDLE);
 
-		if (batch->idle_count > 0)
+		const bool this_scan_resets_delays = resets_delays
+			&& (i >= reset_delays_after_n_scans);
+		if (batch->idle_count > 0 && !this_scan_resets_delays)
```
JanMatCodasip: For clarity, I'd recommend to rename `this_scan_resets_delays` to `delays_were_reset` (or `delays_are_reset` or similar), which is more descriptive. 

The reason is that this flag becomes `true` for not for just a single scan, but for all scans from some point onwards.

```suggestion
		const bool delays_were_reset = resets_delays
			&& (i >= reset_delays_after_n_scans);
		if (batch->idle_count > 0 && !delays_were_reset)
```

src/target/riscv/riscv-013.c
```diff
@@ -498,6 +498,23 @@ static void increase_dmi_busy_delay(struct target *target)
...
+	if (r->reset_delays_wait < 0) {
+		assert(r->reset_delays_wait == -1);
+		return;
+	}
+	if ((size_t)r->reset_delays_wait >= finished_scans) {
+		r->reset_delays_wait -= finished_scans;
+		return;
+	}
+	r->reset_delays_wait = -1;
+	LOG_TARGET_DEBUG(target, "reset_delays_wait done");
```
JanMatCodasip: If you'd like, the log message can be made more clear:

```suggestion
	LOG_TARGET_DEBUG(target, "resetting learned delays (reset_delays_wait counter expired)");
```

src/target/riscv/riscv-013.c
```diff
@@ -498,6 +498,23 @@ static void increase_dmi_busy_delay(struct target *target)
 	dtmcontrol_scan(target, DTM_DTMCS_DMIRESET, NULL /* discard result */);
 }
 
+static void decrease_reset_delays_wait(struct target *target, size_t finished_scans)
```
JanMatCodasip: Please, help me find a more descriptive name for this function. How about `decrement_reset_delays_counter()`, `handle_reset_delays_wait_counter()`, `reset_learned_delays_if_needed()` or similar.

src/target/riscv/batch.c
JanMatCodasip: > I have simplified the commit. I will implement scan types as you suggested in subsequent commits. Please, take a look.

Thank you. I have posted just few minor coding suggestions below.

> The thing is, dmi_read(), dmi_write(), ... utilize VLA to store fields. (...) 
> (...) And after this change is done it would be just easier to use appropriate batch of two accesses in dmi_read(), dmi_write(), .... (no change to API).

Thank you for explanation. I have not yet studied this part of the code in detail (especially the use of VLAs). So I do not yet have clear view whether it would be beneficial to re-implement dmi_read(), dmi_write() et. al. on top of batches, or whether to choose a different approach.

src/target/riscv/batch.h
```diff
@@ -55,8 +55,10 @@ void riscv_batch_free(struct riscv_batch *batch);
 /* Checks to see if this batch is full. */
 bool riscv_batch_full(struct riscv_batch *batch);
 
-/* Executes this scan batch. */
-int riscv_batch_run(struct riscv_batch *batch);
+/* Executes this scan batch. If resets_delays is true, after
+ * reset_delays_after_n_scans does not insert RTI cycles between scans. */
```
JanMatCodasip: I have tried to make the comment more clear:

```suggestion
/* Executes this batch of JTAG DTM DMI scans. 
 *
 * If resets_delays is true, the algorithm will stop inserting idle cycles (JTAG Run-Test Idle)
 * after "reset_delays_after" number of scans is performed. This is useful for stress-testing 
 * of RISC-V algorithms in OpenOCD that are based on batches.
 */
```

src/target/riscv/batch.h
```diff
@@ -55,8 +55,10 @@ void riscv_batch_free(struct riscv_batch *batch);
...
+int riscv_batch_run(struct riscv_batch *batch, bool resets_delays,
+		size_t reset_delays_after_n_scans);
```
JanMatCodasip: I believe the argument can be renamed to just `reset_delays_after`.

```suggestion
		size_t reset_delays_after);
```

JanMatCodasip (commented):
I am sending few smaller coding suggestions, and don't expect to have any more comments.

Overall the code looks good, thank you.

src/target/riscv/batch.h

src/target/riscv/batch.h
en-sc: Thanks! Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/batch.c

JanMatCodasip (approved):
LGTM, thank you!

Issue #688, opened 2022-04-07
Implement watchpoint hit detection for RVC load & store instructions


JanMatCodasip:
(Note to self:)

The RISC-V target in OpenOCD currently implements watchpoint detection by disassembling the current instruction. The feature was implemented in https://github.com/riscv/riscv-openocd/pull/291. See function `riscv_hit_watchpoint()` in [riscv.c](https://github.com/riscv/riscv-openocd/blob/riscv/src/target/riscv/riscv.c).

The mechanism currently recognizes only RVI instructions but RVC (compressed instruction set) is not supported and should be added to the existing implementation.

JanMatCodasip:
The issue is being addressed by https://github.com/riscv-collab/riscv-openocd/pull/1029/ --> closing it.

PR #291, opened 2018-08-23
Handle hardware watchpoints hit by RV32 loads and stores

craigblackmore:
GDB does not report hardware watchpoints correctly when they are hit because OpenOCD does not give enough information. GDB should report the hit variable and its old and new value, but currently we just get a SIGTRAP (see examples below).

Here is an example of the failure from the GDB test suite:
```
(gdb) break main
Breakpoint 1 at 0x40400072: file /home/craig/work/projects/sifive/testing-public/gdb/gdb/testsuite/gdb.base/watchpoints.c, line 33.
(gdb) target remote :3333
Remote debugging using :3333
warning: Target-supplied registers are not supported by the current architecture
0x4040088a in ?? ()
(gdb) monitor reset halt
JTAG tap: riscv.cpu tap/device found: 0x20000001 (mfg: 0x000 (<invalid>), part: 0x0000, ver: 0x2)
(gdb) monitor flash protect 0 64 last off
cleared protection for sectors 64 through 255 on flash bank 0
(gdb) load 
Loading section .init, size 0x6c lma 0x40400000
Loading section .text, size 0x650 lma 0x4040006c
Loading section .rodata, size 0x54 lma 0x404006bc
Loading section .eh_frame, size 0x2c lma 0x40400710
Loading section .data, size 0x444 lma 0x4040073c
Start address 0x40400000, load size 2944
Transfer rate: 4 KB/sec, 588 bytes/write.
(gdb) continue
Continuing.
Note: automatically using hardware breakpoints for read-only addresses.

Breakpoint 1, main () at /home/craig/work/projects/sifive/testing-public/gdb/gdb/testsuite/gdb.base/watchpoints.c:33
33	  for (count = 0; count < 4; count++) {
(gdb) watch ival1
Hardware watchpoint 2: ival1
(gdb) watch ival3
Hardware watchpoint 3: ival3
(gdb) cont
Continuing.

Program received signal SIGTRAP, Trace/breakpoint trap.
0x40400088 in main () at /home/craig/work/projects/sifive/testing-public/gdb/gdb/testsuite/gdb.base/watchpoints.c:34
34	    ival1 = count; ival2 = count;
(gdb) FAIL: gdb.base/watchpoints.exp: watchpoint hit, first time
cont
Continuing.

Program received signal SIGTRAP, Trace/breakpoint trap.
0x404000a4 in main () at /home/craig/work/projects/sifive/testing-public/gdb/gdb/testsuite/gdb.base/watchpoints.c:35
[...]

PR #1029, opened 2024-03-10
Add functions to decode RVC load and store instructions for watchpoints

MrAlexei:
This commit adds support for RVC (compressed) load and store instructions.

https://github.com/riscv-collab/riscv-openocd/issues/688
https://github.com/riscv-collab/riscv-openocd/pull/291

> For GDB to fully support hardware watchpoints, OpenOCD needs to tell GDB which data address has been hit. OpenOCD relies on a target-specific hit_watchpoint function to do this. If GDB is not given the address, it will not print the hit variable name or its old and new value.
> 
> There does not seem to be a way for the hardware to tell us which trigger
> was hit (0.13 introduced the 'hit bit' but this is optional). Alternatively,
> we can decode the instruction at dpc and find out which memory address it accesses.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1616,6 +1616,440 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
 	return ERROR_OK;
 }
 
+/**
+ * These functions are needed to extract individual bits (for offset)
+ * from the instruction
+ */
+// c.lwsp rd_n0 c_uimm8sphi c_uimm8splo - offset[5] offset[4:2|7:6]
+static uint16_t get_offset_clwsp(riscv_insn_t instruction)
+{
+	uint16_t offset_4to2and7to6_bits =
+		get_field32(instruction, INSN_FIELD_C_UIMM8SPLO);
+	uint16_t offset_4to2_bits = offset_4to2and7to6_bits >> 2;
+	uint16_t offset_7to6_bits = offset_4to2and7to6_bits & 0x3;
+	uint16_t offset_5_bit = get_field32(instruction, INSN_FIELD_C_UIMM8SPHI);
+	return (offset_4to2_bits << 2) + (offset_5_bit << 5)
+		   + (offset_7to6_bits << 6);
+}
+
+// c.ldsp rd_n0 c_uimm9sphi c_uimm9splo - offset[5] offset[4:3|8:6]
+static uint16_t get_offset_cldsp(riscv_insn_t instruction)
+{
+	uint16_t offset_4to3and8to6_bits =
+		get_field32(instruction, INSN_FIELD_C_UIMM9SPLO);
+	uint16_t offset_4to3_bits = offset_4to3and8to6_bits >> 3;
+	uint16_t offset_8to6_bits = offset_4to3and8to6_bits & 0x7;
+	uint16_t offset_5_bit = get_field32(instruction, INSN_FIELD_C_UIMM9SPHI);
+	return (offset_4to3_bits << 3) + (offset_5_bit << 5)
+		   + (offset_8to6_bits << 6);
+}
+
+// c.swsp c_rs2 c_uimm8sp_s - offset[5:2|7:6]
+static uint16_t get_offset_cswsp(riscv_insn_t instruction)
+{
+	uint16_t offset_5to2and7to6_bits =
+		get_field32(instruction, INSN_FIELD_C_UIMM8SP_S);
+	uint16_t offset_5to2_bits = offset_5to2and7to6_bits >> 2;
+	uint16_t offset_7to6_bits = offset_5to2and7to6_bits & 0x3;
+	return (offset_5to2_bits << 2) + (offset_7to6_bits << 6);
+}
+
+// c.sdsp c_rs2 c_uimm9sp_s - offset[5:3|8:6]
+static uint16_t get_offset_csdsp(riscv_insn_t instruction)
+{
+	uint16_t offset_5to3and8to6_bits =
+		get_field32(instruction, INSN_FIELD_C_UIMM9SP_S);
+	uint16_t offset_5to3_bits = offset_5to3and8to6_bits >> 3;
+	uint16_t offset_8to6_bits = offset_5to3and8to6_bits & 0x7;
+	return (offset_5to3_bits << 3) + (offset_8to6_bits << 6);
+}
+
+// c.lw rd_p rs1_p c_uimm7lo c_uimm7hi - offset[2|6] offset[5:3]
+static uint16_t get_offset_clw(riscv_insn_t instruction)
+{
+	uint16_t offset_2and6_bits = get_field32(instruction, INSN_FIELD_C_UIMM7LO);
+	uint16_t offset_2_bit = offset_2and6_bits >> 1;
+	uint16_t offset_6_bit = offset_2and6_bits & 0x1;
+	uint16_t offset_5to3_bits = get_field32(instruction, INSN_FIELD_C_UIMM7HI);
+	return (offset_2_bit << 2) + (offset_5to3_bits << 3) + (offset_6_bit << 6);
+}
+
+// c.ld rd_p rs1_p c_uimm8lo c_uimm8hi - offset[7:6] offset[5:3]
+static uint16_t get_offset_cld(riscv_insn_t instruction)
+{
+	uint16_t offset_7to6_bits = get_field32(instruction, INSN_FIELD_C_UIMM8LO);
+	uint16_t offset_5to3_bits = get_field32(instruction, INSN_FIELD_C_UIMM8HI);
+	return (offset_5to3_bits << 3) + (offset_7to6_bits << 6);
+}
+
+// c.lq rd_p rs1_p c_uimm9lo c_uimm9hi - offset[7:6] offset[5|4|8]
+static uint16_t get_offset_clq(riscv_insn_t instruction)
+{
+	uint16_t offset_7to6_bits = get_field32(instruction, INSN_FIELD_C_UIMM9LO);
+	uint16_t offset_5to4and8_bits =
+		get_field32(instruction, INSN_FIELD_C_UIMM9HI);
+	uint16_t offset_5to4_bits = offset_5to4and8_bits >> 1;
+	uint16_t offset_8_bit = offset_5to4and8_bits & 0x1;
+	return (offset_5to4_bits << 4) + (offset_7to6_bits << 6)
+		   + (offset_8_bit << 8);
+}
+
+// c.lqsp rd_n0 c_uimm10sphi c_uimm10splo - offset[5] offset[4|9:6]
+static uint16_t get_offset_clqsp(riscv_insn_t instruction)
+{
+	uint16_t offset_4and9to6_bits =
+		get_field32(instruction, INSN_FIELD_C_UIMM10SPLO);
+	uint16_t offset_4_bit = offset_4and9to6_bits >> 4;
+	uint16_t offset_9to6_bits = offset_4and9to6_bits & 0xf;
+	uint16_t offset_5_bit = get_field32(instruction, INSN_FIELD_C_UIMM10SPHI);
+	return (offset_4_bit << 4) + (offset_5_bit << 5) + (offset_9to6_bits << 6);
+}
+
+// c.sqsp c_rs2 c_uimm10sp_s - offset[5:4|9:6]
+static uint16_t get_offset_csqsp(riscv_insn_t instruction)
+{
+	uint16_t offset_5to4and9to6_bits =
+		get_field32(instruction, INSN_FIELD_C_UIMM10SP_S);
+	uint16_t offset_5to4_biits = offset_5to4and9to6_bits >> 4;
+	uint16_t offset_9to6_bits = offset_5to4and9to6_bits & 0xf;
+	return (offset_5to4_biits << 4) + (offset_9to6_bits << 6);
+}
+
+/**
+ * Decode rs1' register num for RVC.
+ * See "Table: Registers specified by the three-bit rs1′, rs2′, and rd′ fields
+ * of the CIW, CL, CS, CA, and CB formats" in "The RISC-V Instruction Set Manual
+ * Volume I: Unprivileged ISA".
+ * */
+static uint32_t get_rs1_c(riscv_insn_t instruction)
+{
+	return GDB_REGNO_S0 + get_field32(instruction, INSN_FIELD_C_SREG1);
+}
+
+static uint32_t get_opcode(const riscv_insn_t instruction)
+{
+	// opcode is first 7 bits of the instruction
+	uint32_t opcode = instruction & INSN_FIELD_OPCODE;
+	if ((instruction & 0x03) < 0x03) { // opcode size RVC
+		// RVC MASK_C = 0xe003 for load/store instructions
+		opcode = instruction & MASK_C_LD;
+	}
+	return opcode;
+}
+
+static int get_loadstore_membase_regno(struct target *target,
+		const riscv_insn_t instruction, int *regid)
+{
+	uint32_t opcode = get_opcode(instruction);
+	int rs;
+
+	switch (opcode) {
+	case MATCH_LB:
+	case MATCH_FLH & ~INSN_FIELD_FUNCT3:
+	case MATCH_SB:
+	case MATCH_FSH & ~INSN_FIELD_FUNCT3:
+		rs = get_field32(instruction, INSN_FIELD_RS1);
+		break;
+
+	case MATCH_C_LWSP:
+	case MATCH_C_LDSP: // if xlen >= 64 or MATCH_C_FLWSP:
+	case MATCH_C_FLDSP: // or MATCH_C_LQSP if xlen == 128
+	case MATCH_C_SWSP:
+	case MATCH_C_SDSP: // if xlen >= 64 or MATCH_C_FSWSP:
+	case MATCH_C_FSDSP: // or MATCH_C_SQSP if xlen == 128
+		rs = GDB_REGNO_SP;
+		break;
+
+	case MATCH_C_LW:
+	case MATCH_C_FLW: // or MATCH_C_LD if xlen >= 64
+	case MATCH_C_FLD: // or MATCH_C_LQ if xlen == 128
+	case MATCH_C_SW:
+	case MATCH_C_FSW: // or MATCH_C_SD if xlen >= 64
+	case MATCH_C_FSD: // or MATCH_C_SQ if xlen == 128
+		rs = get_rs1_c(instruction);
+		break;
+
+	default:
+		LOG_TARGET_DEBUG(target, "0x%" PRIx32 " is not a RV32I or \"C\" load or"
+						 " store", instruction);
+		return ERROR_FAIL;
+	}
+	*regid = rs;
+	return ERROR_OK;
+}
+
+static int get_loadstore_memoffset(struct target *target,
+		const riscv_insn_t instruction, int16_t *memoffset)
+{
+	uint32_t opcode = get_opcode(instruction);
+	int16_t offset;
+
+	switch (opcode) {
+	case MATCH_LB:
+	case MATCH_FLH & ~INSN_FIELD_FUNCT3:
+	case MATCH_SB:
+	case MATCH_FSH & ~INSN_FIELD_FUNCT3:
+		if (opcode == MATCH_SB || opcode == (MATCH_FSH & ~INSN_FIELD_FUNCT3)) {
+			offset = get_field32(instruction, INSN_FIELD_IMM12LO) |
+				  (get_field32(instruction, INSN_FIELD_IMM12HI) << 5);
+		} else if (opcode == MATCH_LB ||
+				   opcode == (MATCH_FLH & ~INSN_FIELD_FUNCT3)) {
+			offset = get_field32(instruction, INSN_FIELD_IMM12);
+		} else {
+			assert(false);
+		}
+		/* sign extend 12-bit imm to 16-bits */
+		if (offset & (1 << 11))
+			offset |= 0xf000;
+		break;
+
+	case MATCH_C_LWSP:
+		offset = get_offset_clwsp(instruction);
+		break;
+
+	case MATCH_C_LDSP: // if xlen >= 64 or MATCH_C_FLWSP:
+		if (riscv_xlen(target) > 32) { // MATCH_C_LDSP
+			offset = get_offset_cldsp(instruction);
+		} else { // MATCH_C_FLWSP
+			offset = get_offset_clwsp(instruction);
+		}
+		break;
+
+	case MATCH_C_FLDSP: // or MATCH_C_LQSP if xlen == 128
+		if (riscv_xlen(target) == 128) { // MATCH_C_LQSP
+			offset = get_offset_clqsp(instruction);
+		} else { // MATCH_C_FLDSP
+			offset = get_offset_cldsp(instruction);
+		}
+		break;
+
+	case MATCH_C_SWSP:
+		offset = get_offset_cswsp(instruction);
+		break;
+
+	case MATCH_C_SDSP: // if xlen >= 64 or MATCH_C_FSWSP:
+		if (riscv_xlen(target) > 32) { // MATCH_C_SDSP
+			offset = get_offset_csdsp(instruction);
+		} else { // MATCH_C_FSWSP
+			offset = get_offset_cswsp(instruction);
+		}
+		break;
+
+	case MATCH_C_FSDSP: // or MATCH_C_SQSP if xlen == 128
+		if (riscv_xlen(target) == 128) { // MATCH_C_SQSP
+			offset = get_offset_csqsp(instruction);
+		} else { // MATCH_C_FSDSP
+			offset = get_offset_csdsp(instruction); // same as C.SDSP
+		}
+		break;
+
+	case MATCH_C_LW:
+		offset = get_offset_clw(instruction);
+		break;
+
+	case MATCH_C_FLW: // or MATCH_C_LD if xlen >= 64
+		if (riscv_xlen(target) > 32) { // MATCH_C_LD
+			offset = get_offset_cld(instruction);
+		} else { // MATCH_C_FLW
+			offset = get_offset_clw(instruction); // same as C.FLW
+		}
+		break;
+
+	case MATCH_C_FLD: // or MATCH_C_LQ if xlen == 128
+		if (riscv_xlen(target) == 128) { // MATCH_C_LQ
+			offset = get_offset_clq(instruction);
+		} else { // MATCH_C_FLD
+			offset = get_offset_cld(instruction); // same as C.LD
+		}
+		break;
+
+	case MATCH_C_SW:
+		offset = get_offset_clw(instruction); // same as C.LW
+		break;
+
+	case MATCH_C_FSW: // or MATCH_C_SD if xlen >= 64
+		if (riscv_xlen(target) > 32) { // MATCH_C_SD
+			offset = get_offset_cld(instruction); // same as C.LD
+		} else { // MATCH_C_FSW
+			offset = get_offset_clw(instruction); // same as C.LW
+		}
+		break;
+
+	case MATCH_C_FSD: // or MATCH_C_SQ if xlen == 128
+		if (riscv_xlen(target) == 128) { // MATCH_C_SQ
+			offset = get_offset_clq(instruction); // same as C.LQ
+		} else { // MATCH_C_FSD
+			offset = get_offset_cld(instruction); // same as C.LD
+		}
+		break;
+
+	default:
+		LOG_TARGET_DEBUG(target, "0x%" PRIx32 " is not a RV32I or \"C\" load or"
+						 " store", instruction);
+		return ERROR_FAIL;
+	}
+	*memoffset = offset;
+	return ERROR_OK;
+}
+
+static int verify_loadstore(struct target *target,
+		const riscv_insn_t instruction, bool *is_read)
+{
+	uint32_t opcode = get_opcode(instruction);
+	bool misa_f = riscv_supports_extension(target, 'F');
+	bool misa_d = riscv_supports_extension(target, 'D');
+	enum watchpoint_rw rw;
+
+	switch (opcode) {
+	case MATCH_LB:
+	case MATCH_FLH & ~INSN_FIELD_FUNCT3:
+		rw = WPT_READ;
+		break;
+
+	case MATCH_SB:
+	case MATCH_FSH & ~INSN_FIELD_FUNCT3:
+		rw = WPT_WRITE;
+		break;
+
+	case MATCH_C_LWSP:
+		if (get_field32(instruction, INSN_FIELD_RD) == 0) {
+			LOG_TARGET_DEBUG(target,
+				"The code points with rd = x0 are reserved for C.LWSP");
+			return ERROR_FAIL;
+		}
+		rw = WPT_READ;
+		break;
+
+	case MATCH_C_LDSP: // if xlen >= 64 or MATCH_C_FLWSP:
+		if (riscv_xlen(target) > 32) { // MATCH_C_LDSP
+			if (get_field32(instruction, INSN_FIELD_RD) == 0) {
+				LOG_TARGET_DEBUG(target,
+					"The code points with rd = x0 are reserved for C.LDSP");
+				return ERROR_FAIL;
+			}
+		} else { // MATCH_C_FLWSP
+			if (!misa_f) {
+				LOG_TARGET_DEBUG(target, "Matched C.FLWSP but target doesn\'t "
+								 "have the \"F\" extension");
+				return ERROR_FAIL;
+			}
+		}
+		rw = WPT_READ;
+		break;
+
+	case MATCH_C_FLDSP: // or MATCH_C_LQSP if xlen == 128
+		if (riscv_xlen(target) == 128) { // MATCH_C_LQSP
+			if (get_field32(instruction, INSN_FIELD_RD) == 0) {
+				LOG_TARGET_DEBUG(target,
+					"The code points with rd = x0 are reserved for C.LQSP");
+				return ERROR_FAIL;
+			}
+		} else { // MATCH_C_FLDSP
+			if (!misa_d) {
+				LOG_TARGET_DEBUG(target, "Matched C.FLDSP but target doesn\'t "
+								 "have the \"D\" extension");
+				return ERROR_FAIL;
+			}
+		}
+		rw = WPT_READ;
+		break;
+
+	case MATCH_C_SWSP:
+		rw = WPT_WRITE;
+		break;
+
+	case MATCH_C_SDSP: // if xlen >= 64 or MATCH_C_FSWSP:
+		if (riscv_xlen(target) == 32) { // MATCH_C_FSWSP
+			if (!misa_f) {
+				LOG_TARGET_DEBUG(target, "Matched C.FSWSP but target doesn\'t "
+								 "have the \"F\" extension");
+				return ERROR_FAIL;
+			}
+		}
+		rw = WPT_WRITE;
+		break;
+
+	case MATCH_C_FSDSP: // or MATCH_C_SQSP if xlen == 128
+		if (riscv_xlen(target) != 128) { // MATCH_C_SQSP
+			if (!misa_d) {
+				LOG_TARGET_DEBUG(target, "Matched C.FSDSP but target doesn\'t "
+								 "have the \"D\" extension");
+				return ERROR_FAIL;
+			}
+		}
+		rw = WPT_WRITE;
+		break;
+
+	case MATCH_C_LW:
+		rw = WPT_READ;
+		break;
+
+	case MATCH_C_FLW: // or MATCH_C_LD if xlen >= 64
+		if (riscv_xlen(target) == 32) { // MATCH_C_FLW
+			if (!misa_f) {
+				LOG_TARGET_DEBUG(target, "Matched C.FLW but target doesn\'t "
+								 "have the \"F\" extension");
+				return ERROR_FAIL;
+			}
+		}
+		rw = WPT_READ;
+		break;
+
+	case MATCH_C_FLD: // or MATCH_C_LQ if xlen == 128
+		if (riscv_xlen(target) != 128) { // MATCH_C_FLD
+			if (!misa_d) {
+				LOG_TARGET_DEBUG(target, "Matched C.FLD but target doesn\'t "
+								 "have the \"D\" extension");
+				return ERROR_FAIL;
+			}
+		}
+		rw = WPT_READ;
+		break;
+
+	case MATCH_C_SW:
+		rw = WPT_WRITE;
+		break;
+
+	case MATCH_C_FSW: // or MATCH_C_SD if xlen >= 64
+		if (riscv_xlen(target) == 32) { // MATCH_C_FSW
+			if (!misa_f) {
+				LOG_TARGET_DEBUG(target, "Matched C.FSW but target doesn\'t "
+								 "have the \"F\" extension");
+				return ERROR_FAIL;
+			}
+		}
+		rw = WPT_WRITE;
+		break;
+
+	case MATCH_C_FSD: // or MATCH_C_SQ if xlen == 128
+		if (riscv_xlen(target) != 128) { // MATCH_C_FSD
+			if (!misa_d) {
+				LOG_TARGET_DEBUG(target, "Matched C.FSD but target doesn\'t "
+								 "have the \"D\" extension");
+				return ERROR_FAIL;
+			}
+		}
+		rw = WPT_WRITE;
+		break;
+
+	default:
+		LOG_TARGET_DEBUG(target, "0x%" PRIx32 " is not a RV32I or \"C\" load or"
+						 " store", instruction);
+		return ERROR_FAIL;
+	}
+
+	if (rw == WPT_WRITE) {
+		*is_read = false;
+		LOG_TARGET_DEBUG(target, "0x%" PRIx32 " is store instruction",
+						 instruction);
+	} else {
+		*is_read = true;
+		LOG_TARGET_DEBUG(target, "0x%" PRIx32 " is load instruction",
+						 instruction);
+	}
+	return ERROR_OK;
+}
+
 /* Sets *hit_watchpoint to the first watchpoint identified as causing the
  * current halt.
  *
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,282 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	int16_t imm;
+	target_addr_t mem_addr = 0;
+	bool misa_f = riscv_supports_extension(target, 'F');
+	bool misa_d = riscv_supports_extension(target, 'D');
+
+	switch (opcode) {
+	case MATCH_LB: // and MATCH_SB:
+		rs1 = (instruction & 0xf8000) >> 15;
+		riscv_get_register(target, &mem_addr, rs1);
+		if (opcode == MATCH_SB) {
```
en-sc: I do not understand how this works. `opcode == MATCH_LB` in this `case` block. How is it possible for it to equal to both `MATCH_SB` and `MATCH_LB` at the same time?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,282 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+		opcode = instruction & 0xe003;
+	}
+	uint32_t rs1;
+	int16_t imm;
+	target_addr_t mem_addr = 0;
+	bool misa_f = riscv_supports_extension(target, 'F');
+	bool misa_d = riscv_supports_extension(target, 'D');
+
+	switch (opcode) {
+	case MATCH_LB: // and MATCH_SB:
```
```suggestion
	case MATCH_LB:
	/* fallthrough */
	case MATCH_SB:
```
Is this the intention?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,282 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+		/* sign extend 12-bit imm to 16-bits */
+		if (imm & (1 << 11))
+			imm |= 0xf000;
+		mem_addr += imm;
+		LOG_TARGET_DEBUG(target, "Memory address=0x%" PRIx64, mem_addr);
+		break;
+	case MATCH_C_LWSP:
+		if (get_bits_field(instruction, 7, 5) == 0) // the code points with rd=x0 are reserved
+			return ERROR_FAIL;
+		riscv_get_register(target, &mem_addr, GDB_REGNO_SP);
```
en-sc: `riscv_get_register()` may fail. Please, handle it's error code (return value) here and around.

src/target/riscv/riscv.c
```diff
@@ -1643,52 +1919,33 @@ static int riscv_hit_watchpoint(struct target *target, struct watchpoint **hit_w
...
 		return ERROR_FAIL;
-	}
 
 	struct watchpoint *wp = target->watchpoints;
 	while (wp) {
-		/*TODO support length/mask */
-		if (wp->address == mem_addr) {
+		/*TODO support mask */
+		if (mem_addr_wp >= wp->address &&
+			mem_addr_wp <= (wp->address + wp->length)) {
```
en-sc: 1. AFAIU, the check on the upper value should be non-inclusive:
```suggestion
			mem_addr_wp < (wp->address + wp->length)) {
```
2. This will not be sufficient. Consider a WP on address `0x11` with a length of `0x1`. Memory access on the address `0x10` can trigger the WP, if it is at least two bytes wide (e.g. `lh`). I think this is somewhat out of scope of this commit, so you may leave it as is, but please, leave a comment here describing the issue.

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,282 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	}
+	uint32_t rs1;
+	int16_t imm;
+	target_addr_t mem_addr = 0;
+	bool misa_f = riscv_supports_extension(target, 'F');
+	bool misa_d = riscv_supports_extension(target, 'D');
+
+	switch (opcode) {
+	case MATCH_LB: // and MATCH_SB:
+		rs1 = (instruction & 0xf8000) >> 15;
```
```suggestion
		rs1 = (instruction & INSN_FIELD_RS1) >> 15;
```
Where possible, please, use constants form `encoding.h`.

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,282 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
 	return ERROR_OK;
 }
 
+static inline uint32_t get_bits_field(riscv_insn_t b, uint32_t lo, uint32_t len) {
```
en-sc: Please, consider using `get_field32()` from `field_helpers.h` with the appropriate mask from `encoding.h` instead.

en-sc (changes requested):
Thank you for the contribution!  Overall, the change is great.
However, I do have a few concerns and suggestions.

src/target/riscv/riscv.c
MrAlexei: Yes. I fixed it in the commit https://github.com/riscv-collab/riscv-openocd/pull/1029/commits/fd7c9b21a2725f47c5912dbc00711b3a67b9c205

src/target/riscv/riscv.c
MrAlexei: Fixed https://github.com/riscv-collab/riscv-openocd/pull/1029/commits/88a3a29606f0ef72a39d774d77ee61d8ed7c4ad6

src/target/riscv/riscv.c
MrAlexei: This function is needed to extract individual bits (for imm) that do not have a mask in the file encoding.h from the instruction.
Example for C.LW:
Now
```
static inline int16_t get_imm_clw(riscv_insn_t instruction) {
	return (get_bits_field(instruction, 6, 1) << 2)
	       + (get_bits_field(instruction, 10, 3) << 3)
		   + (get_bits_field(instruction, 5, 1) << 6);
}
```
Or if use func get_field32()
```
static inline int16_t get_imm_clw(riscv_insn_t instruction) {
	uint32_t imm2i6 = get_field32(instruction, INSN_FIELD_C_UIMM2); // similar mask but this is imm offset[2|6]
	return ((imm2i6 & 0x2) << 2)
		   | (get_field32(instruction, INSN_FIELD_C_UIMM7HI) << 3) // similar mask but this is imm offset[5:3]
		   | ((imm2i6 & 0x1) << 6);
}
```

TommyMurphyTM1234: > This function is needed to extract individual bits (for imm) that do not have a mask in the file encoding.h from the instruction. Example for C.LW: Now
> 
> ```
> static inline int16_t get_imm_clw(riscv_insn_t instruction) {
> 	return (get_bits_field(instruction, 6, 1) << 2)
> 	       + (get_bits_field(instruction, 10, 3) << 3)
> 		   + (get_bits_field(instruction, 5, 1) << 6);
> }
> ```
> 
> Or if use func get_field32()
> 
> ```
> static inline int16_t get_imm_clw(riscv_insn_t instruction) {
> 	uint32_t imm2i6 = get_field32(instruction, INSN_FIELD_C_UIMM2); // similar mask but this is imm offset[2|6]
> 	return ((imm2i6 & 0x2) << 2)
> 		   | (get_field32(instruction, INSN_FIELD_C_UIMM7HI) << 3) // similar mask but this is imm offset[5:3]
> 		   | ((imm2i6 & 0x1) << 6);
> }
> ```

Does that perhaps suggest that `encoding.h` might need updating rather than creating additional "local" utility functions?

Even if it doesn't, it probably makes sense to use descriptive `#define`s rather than "magic numbers" for the mask/shift values in order to aid readability?

src/target/riscv/riscv.c
MrAlexei: Fixed

src/target/riscv/riscv.c
MrAlexei: > Does that perhaps suggest that `encoding.h` might need updating rather than creating additional "local" utility functions?
> 
> Even if it doesn't, it probably makes sense to use descriptive `#define`s rather than "magic numbers" for the mask/shift values in order to aid readability?

I understand, I'll look at how these fields are described in the riscv-isa-manual itself and try to change my code.

TommyMurphyTM1234: Just to note, if changes to `encoding.h` are merited then I think that these would probably need to be done upstream where this file is mastered and auto-generated.

* https://github.com/riscv-collab/riscv-openocd/blob/ca7d88252633a5d452ee92e6f8f3eb20cadfb633/src/target/riscv/encoding.h#L5-L8

MrAlexei: I want to do this.
Does this bits magic look better?
```
// from riscv-isa-manual
// c.lw rd_p rs1_p c_uimm7lo c_uimm7hi     1..0=0 15..13=2
// latex_mapping['c_uimm7lo'] = 'uimm[2$\\vert$6]'
// latex_mapping['c_uimm7hi'] = 'uimm[5:3]'
//
// from scr1
// system verilog imm = {25'd0, instr[5], instr[12:10], instr[6], 2'b00};
static inline uint16_t get_imm_clw(riscv_insn_t instruction) {
    uint32_t uimm_2and6_bits = get_field32(instruction, INSN_FIELD_C_UIMM7LO);
    uint32_t uimm_3to5_bits = get_field32(instruction, INSN_FIELD_C_UIMM7HI);
    return ((uimm_2and6_bits & 0x2) << 1) | (uimm_3to5_bits << 3)
           | ((uimm_2and6_bits & 0x1) << 6);
}
```

MrAlexei:
Converted the pull request to a draft.
It is necessary to add processing for Store/Load from extensions "F" and "D"

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,357 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	int16_t imm;
+	target_addr_t mem_addr = 0;
+	bool misa_f = riscv_supports_extension(target, 'F');
+	bool misa_d = riscv_supports_extension(target, 'D');
+
+	switch (opcode) {
+	case MATCH_LB:
+		/* fallthrough */
+	case MATCH_FLH & ~INSN_FIELD_FUNCT3:
+		/* fallthrough */
```
en-sc: I'm sorry, this is my mistake. `/* fallthrough */` is not necessary here.
You can leave it if you like, but I would suggest to drop it.

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,357 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+			imm = get_imm_clwsp(instruction);
+			mem_addr += imm;
+			LOG_TARGET_DEBUG(target, "C.FLWSP memory address=0x%" PRIx64,
+							 mem_addr);
+		}
+		break;
+
+	case MATCH_C_FLDSP: // or MATCH_C_LQSP if RV128C
+		if (riscv_xlen(target) == 128) { // MATCH_C_LQSP
+			if (get_field32(instruction, INSN_FIELD_RD) == 0) // the code points with rd=x0 are reserved
```
en-sc: Here and around: please, consider moving comments, describing why the instruction is not load/store to a LOG_TARGET_DEBUG. This will make logs much clearer.
```suggestion
			if (get_field32(instruction, INSN_FIELD_RD) == 0)
			     LOG_TARGET_DEBUG(target, "The code points with rd=x0 are reserved in case of MATCH_C_LQSP.");
```

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,357 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+			return ERROR_FAIL;
+		}
+		/* sign extend 12-bit imm to 16-bits */
+		if (imm & (1 << 11))
+			imm |= 0xf000;
+		mem_addr += imm;
+		LOG_TARGET_DEBUG(target, "Memory address=0x%" PRIx64, mem_addr);
+		break;
+
+	case MATCH_C_LWSP:
```
en-sc, line 1806: [Suggestion]
AFAIU, in each case in this switch the following is done:
1. Number of the register, which contains the base of memory address is deduced from `instruction` and opcode.
2. The register is read.
3. Memory offset is deduced from the `instruction`
4. Memory address is base plus offset.

So, please, consider moving the logic/communication with target out of the switch.
What I mean is something along the lines of:
```
unsigned int rs = get_storeload_membase_regno(instruction);  /* Inside the function there is a `switch (opcode)`*/
if (riscv_get_register(target, &mem_addr, rs) != ERROR_OK)
    return ERROR_FAIL;
target_addr_t offset = get_storeload_memoffset(instruction); /* Inside the function there is a `switch (opcode)`*/
if (verify_storload(target, opcode, rs, offset) != ERROR_OK) /* This function checks that the deuced values are coherent (like 'the code points with rd=x0 are reserved' for MATCH_C_LDSP)*/
    return ERROR_FAIL;
```

IMHO, this will make the code a bit easier to read.

src/target/riscv/riscv.c
```diff
@@ -1636,59 +1987,41 @@ static int riscv_hit_watchpoint(struct target *target, struct watchpoint **hit_w
...
 		return ERROR_FAIL;
-	}
 
 	struct watchpoint *wp = target->watchpoints;
 	while (wp) {
-		/*TODO support length/mask */
-		if (wp->address == mem_addr) {
+		/*TODO support mask */
+		if (mem_addr_wp >= wp->address &&
+			mem_addr_wp < (wp->address + wp->length)) {
```
en-sc: There it would be nice to also check is the instruction is load/store and that the watchpoint is on read/write/access.
However, this is a bit out of scope of this commit, but, please, add a TODO comment.

en-sc (commented):
The patch is great! I have only a few suggestions

src/target/riscv/riscv.c
MrAlexei: > There it would be nice to also check is the instruction is load/store and that the watchpoint is on read/write/access.

I think that the Debug module can detect a memory access operation using the trigger module and generate the correct breakpoints. With this patch, gdb correctly determines what happens in the watchpoint when the hit bit is not used in the debug module.


> However, this is a bit out of scope of this commit, but, please, add a TODO comment.

Ok

src/target/riscv/riscv.c
MrAlexei: Fixed

src/target/riscv/riscv.c
MrAlexei: Added

src/target/riscv/riscv.c
MrAlexei: Added

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,441 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	case MATCH_SB:
+	case MATCH_FSH & ~INSN_FIELD_FUNCT3:
+		if (opcode == MATCH_SB || opcode == (MATCH_FSH & ~INSN_FIELD_FUNCT3)) {
+			imm = get_field32(instruction, INSN_FIELD_IMM12LO) |
+				  (get_field32(instruction, INSN_FIELD_IMM12HI) << 5);
+		} else if (opcode == MATCH_LB ||
+				   opcode == (MATCH_FLH & ~INSN_FIELD_FUNCT3)) {
+			imm = get_field32(instruction, INSN_FIELD_IMM12);
+		} else {
+			return ERROR_FAIL;
```
en-sc: AFAIU, this block is unreachable. Consider replacing the return with an assertion.
```suggestion
			assert(false);
```

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,441 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	case MATCH_SB:
+	case MATCH_FSH & ~INSN_FIELD_FUNCT3:
+		if (opcode == MATCH_SB || opcode == (MATCH_FSH & ~INSN_FIELD_FUNCT3)) {
+			LOG_TARGET_DEBUG(target, "%x is store instruction", instruction);
+		} else if (opcode == MATCH_LB ||
+				   opcode == (MATCH_FLH & ~INSN_FIELD_FUNCT3)) {
+			LOG_TARGET_DEBUG(target, "%x is load instruction", instruction);
+		} else {
+			LOG_TARGET_DEBUG(target, "Unknown instruction");
+			return ERROR_FAIL;
```
en-sc: [same as above] AFAIU, this block is unreachable. Consider replacing the return with an assertion.
```suggestion
			assert(false);
```

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,441 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+		}
+		LOG_TARGET_DEBUG(target, "Memory address=0x%" PRIx64, mem_addr);
```
en-sc: [suggestion]
In many cases in this switch, there is no verification going on -- you are just reporting the info about the instruction via `LOG_TARGET_DEBUG(...)`.
IMHO, this is not necessary -- the value read from DPC will be present in the log and it can be decoded (e.g. using `spike-dasm`).
Consider dropping these prints. I think this will greatly simplify the function.

en-sc (approved):
LGTM. I have a few suggestions, but these are strictly optional.

en-sc (changes requested):
Please, [fix issues with code stile](
https://github.com/riscv-collab/riscv-openocd/actions/runs/8499805195/job/23383969672?pr=1029).
Info on Checkpatch: https://openocd.org/doc/doxygen/html/patchguide.html#checkpatch

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,429 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	case MATCH_C_FLW: // or MATCH_C_LD if xlen >= 64
+	case MATCH_C_FLD: // or MATCH_C_LQ if xlen == 128
+	case MATCH_C_SW:
+	case MATCH_C_FSW: // or MATCH_C_SD if xlen >= 64
+	case MATCH_C_FSD: // or MATCH_C_SQ if xlen == 128
+		rs = get_rs1_c(instruction);
+		break;
+
+	default:
+		LOG_TARGET_DEBUG(target, "%x is not a RV32I or \"C\" load or store",
```
en-sc: [nitpick] [here and around]
`riscv_insn_t` is `uint32_t`, so `%x` is not the correct format specifier.
```suggestion
		LOG_TARGET_DEBUG(target, "%" PRIx32 " is not a RV32I or \"C\" load or store",
```

en-sc (changes requested):
Please, take a look at the [Checkpatch failure](https://github.com/riscv-collab/riscv-openocd/actions/runs/8591059084/job/23571550511?pr=1029) and fix it.

Consider running `checkpatch` on your patches before submitting. This can be done automatically  via pre-push hook. However, this will only check the last commit.
To check all the commits you can use it on the output of the diff:
```
> git diff <base_version> | ./tools/scripts/checkpatch.pl
```
Or specify the commits explicitly:
```
>  ./tools/scripts/checkpatch.pl -g HEAD HEAD~ ...
```
There is also a minor suggestion, a nitpick really, about format specifiers.

src/target/riscv/riscv.c
MrAlexei: Replaced

MrAlexei:
> To check all the commits you can use it on the output of the diff

Thank you!

<details>
<summary>Check all the commits</summary>

```
total: 0 errors, 0 warnings, 0 checks, 517 lines checked

Your patch has no obvious style problems and is ready for submission.

NOTE: Ignored message types: AVOID_EXTERNS BLOCK_COMMENT_STYLE COMPLEX_MACRO CONST_STRUCT ENOSYS FILE_PATH_CHANGES GERRIT_CHANGE_ID LINE_SPACING LOGICAL_CONTINUATIONS MACRO_WITH_FLOW_CONTROL NEW_TYPEDEFS PARENTHESIS_ALIGNMENT PREFER_DEFINED_ATTRIBUTE_MACRO PREFER_FALLTHROUGH PREFER_KERNEL_TYPES SPLIT_STRING SSCANF_TO_KSTRTO SWITCH_CASE_INDENT_LEVEL TRACING_LOGGING VOLATILE
```

</details>

en-sc (dismissed):
LGTM

en-sc:
@TommyMurphyTM1234, could you please take a look if you have the time?
@JanMatCodasip, @MarekVCodasip, I would greatly appreciate your input.

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,430 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+{
+	uint16_t imm_4to2and7to6_bits =
+		get_field32(instruction, INSN_FIELD_C_UIMM8SPLO);
+	uint16_t imm_4to2_bits = imm_4to2and7to6_bits >> 2;
+	uint16_t imm_7to6_bits = imm_4to2and7to6_bits & 0x3;
+	uint16_t imm_b5_bit = get_field32(instruction, INSN_FIELD_C_UIMM8SPHI);
+	return (imm_4to2_bits << 2) + (imm_b5_bit << 5) + (imm_7to6_bits << 6);
+}
+
+//c.ldsp rd_n0 c_uimm9sphi c_uimm9splo - offset[5] offset[4:3|8:6]
```
TommyMurphyTM1234: Minor issue - should be a space between "//" and "c.ldsp" like other comments?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,430 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+{
+	uint16_t imm_4to3and8to6_bits =
+		get_field32(instruction, INSN_FIELD_C_UIMM9SPLO);
+	uint16_t imm_4to3_bits = imm_4to3and8to6_bits >> 3;
+	uint16_t imm_8to6_bits = imm_4to3and8to6_bits & 0x7;
+	uint16_t imm_b5_bit = get_field32(instruction, INSN_FIELD_C_UIMM9SPHI);
+	return (imm_4to3_bits << 3) + (imm_b5_bit << 5) + (imm_8to6_bits << 6);
+}
+
+//c.swsp c_rs2 c_uimm8sp_s - offset[5:2|7:6]
```
TommyMurphyTM1234: Minor issue - should be a space between "//" and "c.swsp" like other comments?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,430 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
 	return ERROR_OK;
 }
 
+/**
+ * These functions are needed to extract individual bits (for imm)
```
TommyMurphyTM1234: I have two observations here:

(1) The spec seems a bit confusing in that it refers to the relevant instruction bitfields as `imm`, `uimm` and `offset` in different places (and contexts?).

(2) Should the term `offset` take precedence here so that the functions should be named, for example, `get_offset_clwsp()` rather than `get_[u]imm_clwsp()`?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,430 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+ * from the instruction
+ */
+// c.lwsp rd_n0 c_uimm8sphi c_uimm8splo - offset[5] offset[4:2|7:6]
```
TommyMurphyTM1234, line 1623: I found this (and all other similar) comment(s) a bit confusing but maybe because of aforementioned use of `imm`, `uimm` and `offset` nomenclature in the spec.

I think that the relationship is:

`instruction[12]` = `uimm[5]`
`instruction[6:4]` = `uimm[4:2]`
`instruction[3:2]` = `uimm[7:6]`
`uimm[7:2]` = `{instruction[3:2], instruction[12], instruction[6:4]}`
`offset[7:0]` = `{uimm[7:2], 2'b00}`

I guess that there's only so much that can be explained in comments here but maybe the comments would be a bit clearer?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,430 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	uint16_t uimm_5to4and9to6_bits =
+		get_field32(instruction, INSN_FIELD_C_UIMM10SP_S);
+	uint16_t uimm_5to4_biits = uimm_5to4and9to6_bits >> 4;
+	uint16_t uimm_9to6_bits = uimm_5to4and9to6_bits & 0xf;
+	return (uimm_5to4_biits << 4) + (uimm_9to6_bits << 6);
+}
+
+static inline uint32_t get_rs1_c(riscv_insn_t instruction)
+{
+	return 8 + get_field32(instruction, INSN_FIELD_C_SREG1);
```
TommyMurphyTM1234: Maybe a comment/spec reference to explain this?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,430 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	default:
+		LOG_TARGET_DEBUG(target, "%" PRIx32 " is not a RV32I or \"C\" load or "
+						 "store", instruction);
+		return ERROR_FAIL;
+	}
+	*regid = rs;
+	return ERROR_OK;
+}
+
+static inline int get_storeload_memoffset(struct target *target,
```
TommyMurphyTM1234: `get_loadstore_memoffset()`?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,430 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+
+	case MATCH_C_LDSP: // if xlen >= 64 or MATCH_C_FLWSP:
+		if (riscv_xlen(target) > 32) { // MATCH_C_LDSP
+			imm = get_imm_cldsp(instruction);
+		} else { // MATCH_C_FLWSP
+			imm = get_imm_clwsp(instruction);
+		}
+		break;
+
+	case MATCH_C_FLDSP: // or MATCH_C_LQSP if RV128C
```
TommyMurphyTM1234: `// or MATCH_C_LQSP if xlen == 128`?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,430 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	/* opcode is first 7 bits of the instruction */
+	uint32_t opcode = instruction & INSN_FIELD_OPCODE;
+	if ((instruction & 0x03) < 0x03) { // opcode size RVC
+		/* RVC MASK_C = 0xe003 for load/store instructions */
+		opcode = instruction & MASK_C_LD;
+	}
+	return opcode;
+}
+
+static inline int get_storeload_membase_regno(struct target *target,
```
TommyMurphyTM1234: Isn't "load/store" a more common idiom than "store/load" or is there some specific reason to use the latter?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,430 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	case MATCH_LB:
+	case MATCH_FLH & ~INSN_FIELD_FUNCT3:
+	case MATCH_SB:
+	case MATCH_FSH & ~INSN_FIELD_FUNCT3:
+		rs = get_field32(instruction, INSN_FIELD_RS1);
+		break;
+
+	case MATCH_C_LWSP:
+	case MATCH_C_LDSP: // if xlen >= 64 or MATCH_C_FLWSP:
+	case MATCH_C_FLDSP: // or MATCH_C_LQSP if RV128C
```
TommyMurphyTM1234: `// or MATCH_C_LQSP if xlen == 128` for consistency with other comments?

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,430 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	case MATCH_C_LDSP: // if xlen >= 64 or MATCH_C_FLWSP:
+		if (riscv_xlen(target) > 32) { // MATCH_C_LDSP
+			if (get_field32(instruction, INSN_FIELD_RD) == 0) {
+				LOG_TARGET_DEBUG(target,
+					"The code points with rd = x0 are reserved for C.LDSP");
+				return ERROR_FAIL;
+			}
+		} else { // MATCH_C_FLWSP
+			if (!misa_f) {
+				LOG_TARGET_DEBUG(target, "Match C.FLWSP but target doesn\'t "
```
TommyMurphyTM1234: "Matched" rather than "Match"?
Ditto for other occurrences.

TommyMurphyTM1234 (commented):
Some observations on foot of the invitation by @en-sc to comment.
Hope this review helps.

JanMatCodasip:
@MrAlexei Thank you for the contribution and I am sorry for a not very timely review. I will be able to take a look at the changes tomorrow morning.

src/target/riscv/riscv.c
MrAlexei: I fixed the code a bit and did everything with the offset priority in the names.

src/target/riscv/riscv.c
MrAlexei: I was offered to use definitions for fields from encoding.h which is taken from the project https://github.com/riscv/riscv-opcodes

In the comments in the first part of the sentence describing the commands from riscv-opcodes,
the second part of the sentence contains how the offset for the instruction is encoded.

I fixed the code a bit and did everything with the offset priority in the names.

Thank you!

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,440 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+ * Volume I: Unprivileged ISA".
+ * */
+static inline uint32_t get_rs1_c(riscv_insn_t instruction)
+{
+	return GDB_REGNO_S0 + get_field32(instruction, INSN_FIELD_C_SREG1);
+}
+
+uint32_t get_opcode(const riscv_insn_t instruction)
+{
+	/* find out which memory address is accessed by the instruction at dpc */
```
JanMatCodasip: It looks like this comment line is unrelated to this function (perhaps just leftover after refactoring)?

```suggestion
```

src/target/riscv/riscv.c
```diff
@@ -1699,6 +2125,7 @@ static int riscv_hit_watchpoint(struct target *target, struct watchpoint **hit_w
 	 *
 	 * OpenOCD will behave as if this function had never been implemented i.e.
 	 * report the halt to GDB with no address information. */
+	LOG_TARGET_DEBUG(target, "Not hit address=%" TARGET_PRIxADDR, mem_addr);
```
JanMatCodasip: I'd recommend to rephrase the debug message for more clarity.

Prefix `0x` should be used for hexadecimal numbers.

Also it would be good to say in the message if this was load or store. In any case, the information about the type of the access (`is_load`) will be needed in this function anyway to fix [the above TODO](https://github.com/riscv-collab/riscv-openocd/pull/1029/files#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR2112).

```suggestion
	LOG_TARGET_DEBUG(target, "No watchpoint found that would cover %s 0x%" TARGET_PRIxADDR, is_load ? "load from" : "store to" , mem_addr);
```

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,440 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+						 "store", instruction);
+		return ERROR_FAIL;
+	}
+
+	if (rw == WPT_WRITE) {
+		LOG_TARGET_DEBUG(target, "%" PRIx32 " is store instruction",
+						 instruction);
+	} else {
+		LOG_TARGET_DEBUG(target, "%" PRIx32 " is load instruction",
+						 instruction);
```
JanMatCodasip: Please, don't forget to print `0x` in front of hex numbers to make them unambiguous for the log readers.

```suggestion
		LOG_TARGET_DEBUG(target, "0x%" PRIx32 " is store instruction",
						 instruction);
	} else {
		LOG_TARGET_DEBUG(target, "0x%" PRIx32 " is load instruction",
						 instruction);
```

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,440 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+								 "have the \"D\" extension");
+				return ERROR_FAIL;
+			}
+		}
+		rw = WPT_WRITE;
+		break;
+
+	default:
+		LOG_TARGET_DEBUG(target, "%" PRIx32 " is not a RV32I or \"C\" load or "
+						 "store", instruction);
```
```suggestion
		LOG_TARGET_DEBUG(target, "0x%" PRIx32 " is not a RV32I or \"C\" load or "
						 "store", instruction);
```

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,440 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+		return ERROR_FAIL;
+	}
+	*memoffset = offset;
+	return ERROR_OK;
+}
+
+static inline int verify_loadstore(struct target *target,
+		const riscv_insn_t instruction, const target_addr_t mem_addr)
```
JanMatCodasip: Please, remove the argument `mem_addr`, along with the debug print. 

The parameter as well as the debug print are both unrelated to the logic of the function verify_loadstore(), so they should not be there.

```suggestion
static inline int verify_loadstore(struct target *target,
		const riscv_insn_t instruction)
```

src/target/riscv/riscv.c
```diff
@@ -1636,59 +2070,51 @@ static int riscv_hit_watchpoint(struct target *target, struct watchpoint **hit_w
...
+
+	if (verify_loadstore(target, instruction, mem_addr) != ERROR_OK)
 		return ERROR_FAIL;
-	}
 
 	struct watchpoint *wp = target->watchpoints;
 	while (wp) {
-		/*TODO support length/mask */
-		if (wp->address == mem_addr) {
+		/* TODO support mask and check read/write/access */
```
JanMatCodasip, line 2115: Note: In RISC-V implementation in OpenOCD, watchpoints sensitive to the values (data + mask) are not supported at the moment. So no need to rush with that part.

On the other hand, the watchpoint type (read/write/access) should be checked. It can be done in a separate pull request. No need to make this one even longer :)

src/target/riscv/riscv.c
```diff
@@ -1636,59 +2070,51 @@ static int riscv_hit_watchpoint(struct target *target, struct watchpoint **hit_w
...
+		if (mem_addr >= wp->address &&
+			mem_addr < (wp->address + wp->length)) {
 			*hit_watchpoint = wp;
-			LOG_TARGET_DEBUG(target, "Hit address=%" TARGET_PRIxADDR, wp->address);
+			LOG_TARGET_DEBUG(target, "Hit address=%" TARGET_PRIxADDR,
+							 wp->address);
```
JanMatCodasip: I'd recommend to expand this message to make it clear what has actually been detected.

Proposal:

```suggestion
			LOG_TARGET_DEBUG(target, "WP hit found: %s 0x%" TARGET_PRIxADDR " covered by %s wp at address 0x%" TARGET_PRIxADDR,
				is_load ? "Load from" : "Store to",
				mem_addr,
				(wp->rw == WPT_READ ? "read" : (wp->rw == WPT_WRITE ? "write" : "access"),
				wp->address  );
```

JanMatCodasip:
@TommyMurphyTM1234 or @MrAlexei, please, don't forget to mark the above threads as resolved when done so. 

Thank you.

JanMatCodasip (commented):
@MrAlexei, thank you for the contribution.

I have done not-very-deep review and am sending few minor comments. I do not expect to have any further suggestions than that.

Also, @MrAlexei, how have you tested your changes?

Thank you.

src/target/riscv/riscv.c
```diff
@@ -1614,6 +1614,440 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
...
+	default:
+		LOG_TARGET_DEBUG(target, "%" PRIx32 " is not a RV32I or \"C\" load or "
+						 "store", instruction);
+		return ERROR_FAIL;
+	}
+	*regid = rs;
+	return ERROR_OK;
+}
+
+static inline int get_loadstore_memoffset(struct target *target,
```
aap-sc: minor. Mostly aesthetic.

1. I would remove `inline` from all these large functions. There is no benefit to inline these switches.
2. I would go as far and remove all `inline` directives. These are static functions. Compilers are quite good at figuring out what should and what should not be inlined.

MrAlexei:
@JanMatCodasip 
> Also, @MrAlexei, how have you tested your changes?

I've tested regression tests with Spike (w/o the hit bit in the mcontrol) using the riscv-test project and manual testing for RVC instructions.

src/target/riscv/riscv.c
```diff
@@ -1636,59 +2070,56 @@ static int riscv_hit_watchpoint(struct target *target, struct watchpoint **hit_w
 	}
 
 	riscv_reg_t dpc;
-	riscv_get_register(target, &dpc, GDB_REGNO_DPC);
+	if (riscv_get_register(target, &dpc, GDB_REGNO_DPC) != ERROR_OK)
```
aap-sc, line 2075: I've looked at this MR in a much greater detail. I have doubts regarding this. I think that this early exit is incorrect:

```
for (struct watchpoint *wp = target->watchpoints; wp; wp = wp->next) {
		if (wp->unique_id == r->trigger_hit) {
			*hit_watchpoint = wp;
			return ERROR_OK;
		}
	}
```

It looks like that this condition should be reversed. If the HW does not support hit bit. we should not decode anything, since we don't have guarantee that we hit a watch point.

I think we should perform decode only if hit bit is supported.

src/target/riscv/riscv.c
```diff
@@ -1636,59 +2070,56 @@ static int riscv_hit_watchpoint(struct target *target, struct watchpoint **hit_w
...
 		return ERROR_FAIL;
-	}
 
 	struct watchpoint *wp = target->watchpoints;
 	while (wp) {
-		/*TODO support length/mask */
-		if (wp->address == mem_addr) {
+		/* TODO support mask and check read/write/access */
+		if (mem_addr >= wp->address &&
+			mem_addr < (wp->address + wp->length)) {
```
aap-sc, line 2121: This condition is not quite correct. RISCV specification recommends to perform multi-value matches. That is we should take the size of underlying memory reference into account.

That is the recommended matching algorithm should be something like this:

```
if (intersects(mem_addr, mem_addr + ref_size, wp->address, wp->address + wp->length))
``` 

 Maybe this should be a target parameter.

src/target/riscv/riscv.c
```diff
@@ -1636,59 +2070,56 @@ static int riscv_hit_watchpoint(struct target *target, struct watchpoint **hit_w
...
-		mem_addr += imm;
-		LOG_TARGET_DEBUG(target, "Memory address=0x%" PRIx64, mem_addr);
-	} else {
-		LOG_TARGET_DEBUG(target, "%x is not a RV32I load or store", instruction);
+	if (verify_loadstore(target, instruction, &is_load) != ERROR_OK)
 		return ERROR_FAIL;
-	}
 
 	struct watchpoint *wp = target->watchpoints;
 	while (wp) {
```
aap-sc, line 2114: also, can we avoid reading of memory if no watch point is pending.

src/target/riscv/riscv.c
aap-sc: nevermind. I was wrong here. It seems that we just find the watchpoint, not the exact address. For the purposes of finding watch point - this is fine.

src/target/riscv/riscv.c
aap-sc: > also, can we avoid reading of memory if no watch point is pending.

I want to elaborate on this requirement. Currently, if target does not support hit bit , OpenOCD always reports that the target is halted due to watch point. And it seems that we always call watchpoint_hit function in this case. Thus the request to check if there is at least some watchpoint is pending before actually reading the memory

src/target/riscv/riscv.c
MrAlexei: Where can I read about this?

aap-sc: https://github.com/riscv/riscv-debug-spec/releases/download/1.0.0-rc2/riscv-debug-specification.pdf 

Search for the following lines:

```
This bit determines the contents of the XLEN-bit compare
values.
0 (address): There is at least one compare value and it
contains the lowest virtual address of the access. It is
recommended that there are additional compare values for
the other accessed virtual addresses. (E.g. on a 32-bit read
from 0x4000, the lowest address is 0x4000 and the other
addresses are 0x4001, 0x4002, and 0x4003.)
```

MrAlexei: Thanks for the clarification.
I think that to check this condition it will be necessary to add the setting of the "Select" bit when programming the trigger (as described in Debugger Implementation for Triggers). At the moment, I don't see any use of the "Select" bit in the OpenOCD code when programming the Triggers. 
Maybe another PR is needed to solve this issue?

aap-sc: > he "Select" bit when programming the trigger

Setting select bit is about data watchpoints. These are different. Current implementation does not set it because it always sets address watchpoints.

> Maybe another MR is needed to solve this issue?

Well, I'm fine with that - would you add a TODO then if it's not a bit dial?

MrAlexei: > Well, I'm fine with that - would you add a TODO then if it's not a bit dial?

Ok.
Thank you.

JanMatCodasip: I agree with @aap-sc that there should be a check for intersection of the access range and watchpoint range. 

And I also agree that it is perfectly fine to just add a `TODO` comment for later resolution. This PR is already complex enough :)

JanMatCodasip:
@aap-sc @MrAlexei Please, what is the status on the two last currently open threads ([1](https://github.com/riscv-collab/riscv-openocd/pull/1029#discussion_r1579574008), [2](https://github.com/riscv-collab/riscv-openocd/pull/1029#discussion_r1574955154)) in this PR? Thank you.

aap-sc:
@JanMatCodasip 

> @aap-sc @MrAlexei Please, what is the status on the two last currently open threads ([1](https://github.com/riscv-collab/riscv-openocd/pull/1029#discussion_r1579574008), [2](https://github.com/riscv-collab/riscv-openocd/pull/1029#discussion_r1574955154)) in this PR? Thank you.

My understanding is that these are not addressed (as of yet). However, these are not blockers. Given the size of the change I don't have strong objections against merging this. It's just we'll have to follow up with additional patches that should address these issues.

src/target/riscv/riscv.c
MrAlexei: Added.
Thank you!

src/target/riscv/riscv.c
MrAlexei: As I understand it, this issue is solved in https://github.com/riscv-collab/riscv-openocd/pull/1056

MrAlexei:
> @aap-sc @MrAlexei Please, what is the status on the two last currently open threads ([1](https://github.com/riscv-collab/riscv-openocd/pull/1029#discussion_r1579574008), [2](https://github.com/riscv-collab/riscv-openocd/pull/1029#discussion_r1574955154)) in this PR? Thank you.

I've finished working on these last threads. Thank you.

aap-sc (approved):

JanMatCodasip (approved):
@MrAlexei Thank you for the contribution.

MrAlexei:
Thank you all so much for the review!

PR #1031, opened 2024-03-21
fix confusing status messages during resume

aap-sc:
Recently, (after b503fdef02f) OpenOCD started to notify user about hart state updates. This causes confusion in some cases since some internal updates to the hart state should not be visible to the user as these are implementation details. For example situation like this:

```
> reset halt
JTAG tap: riscv.tap tap/device found: 0xdeadbeef ...
> resume
[riscv.cpu0] Found 4 triggers
riscv.cpu0 halted due to single-step.
[riscv.cpu1] Found 4 triggers
riscv.cpu1 halted due to single-step.
[riscv.cpu2] Found 4 triggers
riscv.cpu2 halted due to single-step.
[riscv.cpu3] Found 4 triggers
riscv.cpu3 halted due to single-step.
```
likely confuse people.

There is no issue with the resume functionality. It`s just that resume internally causes single-step that causes hart state to change.

This commit disable calling of user-specified (and default) callbacks during the "hidden" step operation disabling these confusing messages

Change-Id: I3412a089e2abdcd315d86cec7ee732fdd18c1601

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1709,15 +1709,26 @@ static int oldriscv_step(struct target *target, int current, uint32_t address,
 	return tt->step(target, current, address, handle_breakpoints);
 }
 
-static int old_or_new_riscv_step(struct target *target, int current,
-		target_addr_t address, int handle_breakpoints)
+static int riscv_openocd_step_impl(struct target *target, int current,
+		target_addr_t address, int handle_breakpoints, int handle_callbacks);
+
+static int old_or_new_riscv_step_impl(struct target *target, int current,
+		target_addr_t address, int handle_breakpoints, int handle_callbacks)
 {
 	RISCV_INFO(r);
 	LOG_TARGET_DEBUG(target, "handle_breakpoints=%d", handle_breakpoints);
 	if (!r->get_hart_state)
 		return oldriscv_step(target, current, address, handle_breakpoints);
 	else
-		return riscv_openocd_step(target, current, address, handle_breakpoints);
+		return riscv_openocd_step_impl(target, current, address, handle_breakpoints,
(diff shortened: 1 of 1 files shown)


aap-sc:
NOTE: this is more like a temporary solution to hide the problem. The situation with these "hidden" step and user-specified callbacks is not quite obvious. I'm going to clarify this with openocd maintainers in upstream.

Also, this commit only fixes 0.13 implementations. 0.11 is still affected. Will address this in future

en-sc (dismissed):
@JanMatCodasip, @MarekVCodasip, please, take a look at the patch.
I have already reviewed it internally.

src/target/riscv/riscv.c
```diff
@@ -3361,7 +3373,8 @@ int riscv_openocd_step(struct target *target, int current,
 		LOG_TARGET_ERROR(target, "Unable to restore the disabled breakpoint.");
 	}
 
-	if (success) {
+	// TODO: should we update debug_reason if no callback was called?
```
MarekVCodasip: I think that we should update the debug reason even if callbacks are not to be handled.

MarekVCodasip (commented):
Thank you for the patch.

I have commented my opinion on the TODO, please take a look.

Otherwise it looks good to me.

src/target/riscv/riscv.c

en-sc (approved):
LGTM

src/target/riscv/riscv.c
JanMatCodasip: @MarekVCodasip Marek, can you please quickly re-check this PR now that your suggestion has been addressed. Thanks.

MarekVCodasip (approved):
LGTM

Issue #981, opened 2023-12-14
riscv-013.c, several functions can't return error


timsifive:
Function list:
* read_abstract_arg
* write_abstract_arg
* sb_read_address
* riscv013_read_debug_buffer

Raised in #978

en-sc:
The issue with `read_abstract_arg` is addressed by #1033.

PR #1033, opened 2024-03-22
target/riscv: read abstract args using batch

en-sc:
This would eliminate the need for an extra nop in-between the two reads
in case of a 64-bit register.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -16,7 +16,7 @@
 /* Reserve extra room in the batch (needed for the last NOP operation) */
 #define BATCH_RESERVED_SCANS 1
 
-struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans, size_t idle)
+struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans)
 {
 	scans += BATCH_RESERVED_SCANS;
 	struct riscv_batch *out = calloc(1, sizeof(*out));
@@ -27,8 +27,9 @@ struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans, size_
 
 	out->target = target;
 	out->allocated_scans = scans;
-	out->idle_count = idle;
 	out->last_scan = RISCV_SCAN_TYPE_INVALID;
+	out->was_run = false;
+	out->used_idle_count = 0;
 
 	out->data_out = NULL;
 	out->data_in = NULL;
@@ -89,17 +90,44 @@ bool riscv_batch_full(struct riscv_batch *batch)
 	return riscv_batch_available_scans(batch) == 0;
 }
 
-int riscv_batch_run(struct riscv_batch *batch, bool resets_delays,
-		size_t reset_delays_after)
+static bool riscv_batch_was_scan_busy(const struct riscv_batch *batch,
+		size_t scan_idx)
 {
-	if (batch->used_scans == 0) {
-		LOG_TARGET_DEBUG(batch->target, "Ignoring empty batch.");
-		return ERROR_OK;
-	}
+	assert(batch->was_run);
+	assert(scan_idx < batch->used_scans);
+	const struct scan_field *field = batch->fields + scan_idx;
+	assert(field->in_value);
+	const uint64_t in = buf_get_u64(field->in_value, 0, field->num_bits);
+	return get_field(in, DTM_DMI_OP) == DTM_DMI_OP_BUSY;
+}
 
-	riscv_batch_add_nop(batch);
+static void add_idle_if_increased(struct riscv_batch *batch, size_t new_idle_count)
+{
+	if (!batch->was_run)
+		return;
+	if (batch->used_idle_count <= new_idle_count)
+		return;
+	const size_t idle_change = new_idle_count - batch->used_idle_count;
+	LOG_TARGET_DEBUG(batch->target,
+			"Idle count increased. Adding %zu idle cycles before the batch.",
+			idle_change);
+	jtag_add_runtest(idle_change, TAP_IDLE);
+}
 
-	for (size_t i = 0; i < batch->used_scans; ++i) {
+int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
+		size_t idle_count, bool resets_delays, size_t reset_delays_after)
+{
+	assert(batch->used_scans);
+	assert(batch->last_scan == RISCV_SCAN_TYPE_NOP);
+	assert(!batch->was_run || riscv_batch_was_scan_busy(batch, start_idx));
+	assert(start_idx == 0 || !riscv_batch_was_scan_busy(batch, start_idx - 1));
+
+	add_idle_if_increased(batch, idle_count);
+
+	LOG_TARGET_DEBUG(batch->target, "Running batch of scans [%zu, %zu)",
+			start_idx, batch->used_scans);
+
+	for (size_t i = start_idx; i < batch->used_scans; ++i) {
 		if (bscan_tunnel_ir_width != 0)
 			riscv_add_bscan_tunneled_scan(batch->target, batch->fields + i, batch->bscan_ctxt + i);
 		else
@@ -107,8 +135,8 @@ int riscv_batch_run(struct riscv_batch *batch, bool resets_delays,
 
 		const bool delays_were_reset = resets_delays
 			&& (i >= reset_delays_after);
-		if (batch->idle_count > 0 && !delays_were_reset)
-			jtag_add_runtest(batch->idle_count, TAP_IDLE);
+		if (idle_count > 0 && !delays_were_reset)
+			jtag_add_runtest(idle_count, TAP_IDLE);
 	}
 
 	keep_alive();
@@ -122,16 +150,18 @@ int riscv_batch_run(struct riscv_batch *batch, bool resets_delays,
 
 	if (bscan_tunnel_ir_width != 0) {
 		/* need to right-shift "in" by one bit, because of clock skew between BSCAN TAP and DM TAP */
-		for (size_t i = 0; i < batch->used_scans; ++i) {
+		for (size_t i = start_idx; i < batch->used_scans; ++i) {
 			if ((batch->fields + i)->in_value)
 				buffer_shr((batch->fields + i)->in_value, DMI_SCAN_BUF_SIZE, 1);
 		}
 	}
 
-	for (size_t i = 0; i < batch->used_scans; ++i)
-		riscv_decode_dmi_scan(batch->target, batch->idle_count, batch->fields + i,
+	for (size_t i = start_idx; i < batch->used_scans; ++i)
+		riscv_log_dmi_scan(batch->target, idle_count, batch->fields + i,
 				/*discard_in*/ false);
 
+	batch->was_run = true;
+	batch->used_idle_count = idle_count;
 	return ERROR_OK;
 }
 
@@ -208,14 +238,23 @@ size_t riscv_batch_available_scans(struct riscv_batch *batch)
 	return batch->allocated_scans - batch->used_scans - BATCH_RESERVED_SCANS;
 }
 
-bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch)
+bool riscv_batch_was_batch_busy(const struct riscv_batch *batch)
 {
-	if (batch->used_scans == 0)
-		/* Empty batch */
-		return false;
-
+	assert(batch->was_run);
+	assert(batch->used_scans);
 	assert(batch->last_scan == RISCV_SCAN_TYPE_NOP);
-	const struct scan_field *field = batch->fields + batch->used_scans - 1;
-	const uint64_t in = buf_get_u64(field->in_value, 0, field->num_bits);
-	return get_field(in, DTM_DMI_OP) == DTM_DMI_OP_BUSY;
+	return riscv_batch_was_scan_busy(batch, batch->used_scans - 1);
+}
+
+size_t riscv_batch_finished_scans(const struct riscv_batch *batch)
+{
+	if (!riscv_batch_was_batch_busy(batch)) {
+		/* Whole batch succeeded. */
+		return batch->used_scans;
+	}
+	assert(batch->used_scans);
+	size_t first_busy = 0;
+	while (!riscv_batch_was_scan_busy(batch, first_busy))
+		++first_busy;
+	return first_busy;
 }
--- a/src/target/riscv/batch.h
+++ b/src/target/riscv/batch.h
@@ -24,8 +24,6 @@ struct riscv_batch {
 	size_t allocated_scans;
 	size_t used_scans;
 
-	size_t idle_count;
-
 	uint8_t *data_out;
 	uint8_t *data_in;
 	struct scan_field *fields;
@@ -44,26 +42,42 @@ struct riscv_batch {
 	/* The read keys. */
 	size_t *read_keys;
 	size_t read_keys_used;
+
+	/* Flag indicating that the last run of the batch finished without an error
+	 * from the underlying JTAG layer of OpenOCD - all scans were performed.
+	 * However, RISC-V DMI "busy" condition could still have occurred.
+	 */
+	bool was_run;
+	/* Idle count used on the last run. Only valid after `was_run` is set. */
+	size_t used_idle_count;
 };
 
 /* Allocates (or frees) a new scan set.  "scans" is the maximum number of JTAG
- * scans that can be issued to this object, and idle is the number of JTAG idle
- * cycles between every real scan. */
-struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans, size_t idle);
+ * scans that can be issued to this object. */
+struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans);
 void riscv_batch_free(struct riscv_batch *batch);
 
 /* Checks to see if this batch is full. */
 bool riscv_batch_full(struct riscv_batch *batch);
 
-/* Executes this batch of JTAG DTM DMI scans.
+/* Executes this batch of JTAG DTM DMI scans, starting form "start" scan.
  *
- * If resets_delays is true, the algorithm will stop inserting idle cycles
- * (JTAG Run-Test Idle) after "reset_delays_after" number of scans is
+ * If batch is run for the first time, it is expected that "start" is zero.
+ * It is expected that the batch ends with a DMI NOP operation.
+ *
+ * "idle_count" is the number of JTAG Run-Test-Idle cycles to add in-between
+ * the scans.
+ *
+ * If "resets_delays" is true, the algorithm will stop inserting idle cycles
+ * (JTAG Run-Test-Idle) after "reset_delays_after" number of scans is
  * performed.  This is useful for stress-testing of RISC-V algorithms in
  * OpenOCD that are based on batches.
  */
-int riscv_batch_run(struct riscv_batch *batch, bool resets_delays,
-		size_t reset_delays_after);
+int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
+		size_t idle_count, bool resets_delays, size_t reset_delays_after);
+
+/* Get the number of scans successfully executed form this batch. */
+size_t riscv_batch_finished_scans(const struct riscv_batch *batch);
 
 /* Adds a DM register write to this batch. */
 void riscv_batch_add_dm_write(struct riscv_batch *batch, uint64_t address, uint32_t data,
@@ -83,12 +97,12 @@ void riscv_batch_add_nop(struct riscv_batch *batch);
 size_t riscv_batch_available_scans(struct riscv_batch *batch);
 
 /* Return true iff the last scan in the batch returned DMI_OP_BUSY. */
-bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch);
+bool riscv_batch_was_batch_busy(const struct riscv_batch *batch);
 
 /* TODO: The function is defined in `riscv-013.c`. This is done to reduce the
  * diff of the commit. The intention is to move the function definition to
  * a separate module (e.g. `riscv013-jtag-dtm.c/h`) in another commit. */
-void riscv_decode_dmi_scan(const struct target *target, int idle, const struct scan_field *field,
+void riscv_log_dmi_scan(const struct target *target, int idle, const struct scan_field *field,
 		bool discard_in);
 
 #endif
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -392,7 +392,7 @@ static unsigned int decode_dmi(const struct target *target, char *text, uint32_t
 	return 0;
 }
 
-void riscv_decode_dmi_scan(const struct target *target, int idle, const struct scan_field *field, bool discard_in)
+void riscv_log_dmi_scan(const struct target *target, int idle, const struct scan_field *field, bool discard_in)
 {
 	static const char * const op_string[] = {"-", "r", "w", "?"};
 	static const char * const status_string[] = {"+", "?", "F", "b"};
@@ -586,7 +586,7 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 
 	if (address_in)
 		*address_in = buf_get_u32(in, DTM_DMI_ADDRESS_OFFSET, info->abits);
-	riscv_decode_dmi_scan(target, idle_count, &field, /*discard_in*/ !data_in);
+	riscv_log_dmi_scan(target, idle_count, &field, /*discard_in*/ !data_in);
 	return buf_get_u32(in, DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH);
 }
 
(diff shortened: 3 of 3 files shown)


en-sc:
Note: this patch depends on #1028.
Addresses #981 (partially).

src/target/riscv/batch.h
```diff
@@ -55,16 +58,33 @@ void riscv_batch_free(struct riscv_batch *batch);
 /* Checks to see if this batch is full. */
 bool riscv_batch_full(struct riscv_batch *batch);
 
-/* Executes this batch of JTAG DTM DMI scans.
+/* Executes this batch of JTAG DTM DMI scans, starting form "start" scan.
+ * If batch is run for the first time, it is expected that "start" is zero.
+ * It is expected, that the bacth ends with a DMI NOP operation.
```
JanMatCodasip: Just a detail:

```suggestion
 * It is expected that the batch ends with a DMI NOP operation.
```

src/target/riscv/batch.c
```diff
@@ -89,17 +90,28 @@ bool riscv_batch_full(struct riscv_batch *batch)
 	return riscv_batch_available_scans(batch) == 0;
 }
 
-int riscv_batch_run(struct riscv_batch *batch, bool resets_delays,
-		size_t reset_delays_after)
+static bool riscv_batch_was_busy(const struct riscv_batch *batch, size_t scan_i)
```
JanMatCodasip: Detail: 

- You may consider renaming `scan_i` to `scan_idx`, `scan_index` or similar (also in the other functions).
- The function can be renamed to `riscv_batch_was_scan_busy()`, making it clearer that this refers to a particular scan, not the whole batch.

```suggestion
static bool riscv_batch_was_scan_busy(const struct riscv_batch *batch, size_t scan_index)
```

src/target/riscv/batch.c
```diff
@@ -210,12 +232,22 @@ size_t riscv_batch_available_scans(struct riscv_batch *batch)
 
 bool riscv_batch_dmi_busy_encountered(const struct riscv_batch *batch)
```
JanMatCodasip: This could be renamed to `riscv_batch_was_batch_busy()` to align the name with `riscv_batch_was_scan_busy()`.

src/target/riscv/batch.c
```diff
@@ -122,19 +134,29 @@ int riscv_batch_run(struct riscv_batch *batch, bool resets_delays,
...
-		for (size_t i = 0; i < batch->used_scans; ++i) {
+		for (size_t i = start_i; i < batch->used_scans; ++i) {
 			if ((batch->fields + i)->in_value)
 				buffer_shr((batch->fields + i)->in_value, DMI_SCAN_BUF_SIZE, 1);
 		}
 	}
 
-	for (size_t i = 0; i < batch->used_scans; ++i)
+	for (size_t i = start_i; i < batch->used_scans; ++i)
 		riscv_decode_dmi_scan(batch->target, batch->idle_count, batch->fields + i,
```
JanMatCodasip: If you like - this name would make it more clear what is happening:

```suggestion
		riscv_log_dmi_scan(batch->target, batch->idle_count, batch->fields + i,
```

src/target/riscv/batch.h
```diff
@@ -55,16 +58,33 @@ void riscv_batch_free(struct riscv_batch *batch);
...
  * OpenOCD that are based on batches.
  */
-int riscv_batch_run(struct riscv_batch *batch, bool resets_delays,
+int riscv_batch_run_from(struct riscv_batch *batch, size_t start, bool resets_delays,
 		size_t reset_delays_after);
 
+/* Change the nomber of JTAG Run-Test Idle cycles added in-between DMI operations,
+ * adding the difference to JTAG queue.
+ */
+void riscv_batch_update_idle(struct riscv_batch *batch, size_t new_idle);
```
JanMatCodasip: It looks like there are now multiple different ways how to configure the `idle` value:

- in riscv_batch_alloc()
- in riscv_batch_continue()
- in riscv_batch_update_idle()
- (but riscv_batch_run_from() does not have that argument)

I'd recommend to unify (and simplify) the API:

- Do not expose `riscv_batch_update_idle()`. Either remove it, or perhaps leave it as a helper function for the use inside the "batch" module.
- Remove `idle` argument from `riscv_batch_alloc()`.
- Have the `idle` argument only in functions that trigger execution of the batch:
  - riscv_batch_run_from()
  - riscv_batch_continue() - already present there

src/target/riscv/batch.h
```diff
@@ -55,16 +58,33 @@ void riscv_batch_free(struct riscv_batch *batch);
...
+
+/* Continue running the batch that was once run using "riscv_batch_run" and got
+ * a DMI busy response on "finished_scans" scan, updating the number of idle
+ * cycles used.
+ */
+int riscv_batch_continue(struct riscv_batch *batch, size_t finished_scans,
+		size_t new_idle, bool resets_delays, size_t reset_delays_after);
```
JanMatCodasip: If you like:

```suggestion
int riscv_batch_continue_after_busy(struct riscv_batch *batch, size_t finished_scans,
		size_t new_idle, bool resets_delays, size_t reset_delays_after);
```

src/target/riscv/batch.c
```diff
@@ -89,17 +90,28 @@ bool riscv_batch_full(struct riscv_batch *batch)
...
+}
 
-	riscv_batch_add_nop(batch);
+int riscv_batch_run_from(struct riscv_batch *batch, size_t start_i,
+		bool resets_delays, size_t reset_delays_after)
+{
+	assert(batch->used_scans);
+	assert(batch->last_scan == RISCV_SCAN_TYPE_NOP);
+	assert(!batch->was_run || riscv_batch_was_busy(batch, start_i));
+	assert(start_i == 0 || !riscv_batch_was_busy(batch, start_i - 1));
```
JanMatCodasip: I would recommend to remove these two assertions. The reason is that the user of the batch API may wish to re-run the batch for different reasons than DMI busy  -- e.g. due to encountering sbcs.sbbusyerror or abstractcs.cmderr==Busy.

```suggestion
```

src/target/riscv/batch.h
```diff
@@ -44,6 +44,9 @@ struct riscv_batch {
 	/* The read keys. */
 	size_t *read_keys;
 	size_t read_keys_used;
+
+	/* Flag, indicating that the last run of the batch was successful. */
+	bool was_run;
```
JanMatCodasip: Here I'd recommend to avoid the word "successful" because someone could interpret it as "busy not encountered", which is not the true meaning here.

I have tried to reword the comment:

```suggestion
	/* Flag indicating that in the last run of the batch, all requested scans were
	 * executed (regardless of the outcome - busy or non-busy) */
	bool was_run;
```

src/target/riscv/batch.c
```diff
@@ -210,12 +232,22 @@ size_t riscv_batch_available_scans(struct riscv_batch *batch)
...
+
+size_t riscv_batch_finished_scans(const struct riscv_batch *batch)
+{
+	if (!riscv_batch_dmi_busy_encountered(batch))
+		return batch->used_scans;
+	assert(batch->used_scans);
+	size_t first_busy = 0;
+	while (!riscv_batch_was_busy(batch, first_busy))
+		++first_busy;
+	return first_busy;
```
JanMatCodasip: Few cosmetic details - if you like:

```suggestion
	assert(batch->used_scans > 0);
	if (!riscv_batch_dmi_busy_encountered(batch))
		/* Whole batch succeeded */
		return batch->used_scans;

	size_t first_busy = 0;
	while (!riscv_batch_was_busy(batch, first_busy))
		++first_busy;
	return first_busy;
```

src/target/riscv/batch.c
```diff
@@ -89,17 +90,28 @@ bool riscv_batch_full(struct riscv_batch *batch)
...
+int riscv_batch_run_from(struct riscv_batch *batch, size_t start_i,
+		bool resets_delays, size_t reset_delays_after)
+{
+	assert(batch->used_scans);
+	assert(batch->last_scan == RISCV_SCAN_TYPE_NOP);
+	assert(!batch->was_run || riscv_batch_was_busy(batch, start_i));
+	assert(start_i == 0 || !riscv_batch_was_busy(batch, start_i - 1));
 
-	for (size_t i = 0; i < batch->used_scans; ++i) {
+	LOG_TARGET_DEBUG(batch->target, "Running scans [%zu, %zu)",
```
JanMatCodasip: If you like:

```suggestion
	LOG_TARGET_DEBUG(batch->target, "Running batch of scans [%zu, %zu)",
```

src/target/riscv/batch.h
JanMatCodasip: Actually, it looks like this function is declared but the definition is missing. Can you please double-check it.

src/target/riscv/riscv-013.c
```diff
@@ -2630,15 +2654,58 @@ static int sb_write_address(struct target *target, target_addr_t address,
...
+	/* TODO: To use `riscv_batch_finished_scans()` here, it is needed for
+	 * all scans to not discard input, meaning
+	 * "riscv_batch_add_dm_write(..., false)" should not be used. */
 	const size_t finished_scans = batch->used_scans;
 	decrement_reset_delays_counter(target, finished_scans);
 	return result;
 }
 
+/* It is expected that during creation of the batch
+ * "riscv_batch_add_dm_write(..., false)" was not used. */
```
JanMatCodasip: Please, can you explain why we have this constraint?

src/target/riscv/riscv-013.c
```diff
@@ -2630,15 +2654,58 @@ static int sb_write_address(struct target *target, target_addr_t address,
...
+static int batch_run_timeout(struct target *target, struct riscv_batch *batch)
+{
+	RISCV013_INFO(info);
+	size_t finished_scans = 0;
+	const time_t start = time(NULL);
+	riscv_batch_add_nop(batch);
+	int result = ERROR_FAIL;
```
JanMatCodasip: Cosmetics: 

- I'd recommend to move riscv_batch_add_nop() to the top, and only then define the variables needed in the loop.
- Variable `result` need not be initialized - the value is immediately overwritten, so let's not make the reader wonder why the assignment is there.

```suggestion
	RISCV013_INFO(info);

	riscv_batch_add_nop(batch);
	
	const time_t start = time(NULL);	
	size_t finished_scans = 0;	
	int result;
```

src/target/riscv/riscv-013.c
```diff
@@ -2630,15 +2654,58 @@ static int sb_write_address(struct target *target, target_addr_t address,
...
+			return ERROR_OK;
+		}
+		increase_dmi_busy_delay(target);
+		riscv_batch_update_idle(batch, info->dmi_busy_delay);
+	} while (time(NULL) - start < riscv_command_timeout_sec);
+	assert(result == ERROR_OK);
+	assert(riscv_batch_dmi_busy_encountered(batch));
+	/* Reset dmi_busy_delay, so the value doesn't get too big. */
+	LOG_TARGET_DEBUG(target, "dmi_busy_delay is reset.");
+	info->dmi_busy_delay = 0;
```
JanMatCodasip: Instead of resetting the dmi_busy_delay to zero, can we reset it to the original value - before the function `batch_run_timeout` was entered?

src/target/riscv/riscv-013.c
```diff
@@ -2630,15 +2654,58 @@ static int sb_write_address(struct target *target, target_addr_t address,
...
+	/* TODO: Introduce a separate timeout for this. */
+	LOG_TARGET_ERROR(target, "DMI operation didn't complete in %d seconds. "
+			"The target is either really slow or broken. You could increase "
+			"the timeout with riscv set_command_timeout_sec.",
+			riscv_command_timeout_sec);
+	return ERROR_TIMEOUT_REACHED;
+}
```
JanMatCodasip: This part of the code is written in a very dense style. Few blank lines will improve readability:

```suggestion
	} while (time(NULL) - start < riscv_command_timeout_sec);
	
	assert(result == ERROR_OK);
	assert(riscv_batch_dmi_busy_encountered(batch));
	
	/* Reset dmi_busy_delay, so the value doesn't get too big. */
	LOG_TARGET_DEBUG(target, "dmi_busy_delay is reset.");
	info->dmi_busy_delay = 0;

	/* TODO: Introduce a separate timeout for this. */
	LOG_TARGET_ERROR(target, "DMI operation didn't complete in %d seconds. "
			"The target is either really slow or broken. You could increase "
			"the timeout with riscv set_command_timeout_sec.",
			riscv_command_timeout_sec);
	return ERROR_TIMEOUT_REACHED;
}
```

src/target/riscv/riscv-013.c
```diff
@@ -2630,15 +2654,58 @@ static int sb_write_address(struct target *target, target_addr_t address,
...
+		}
+		increase_dmi_busy_delay(target);
+		riscv_batch_update_idle(batch, info->dmi_busy_delay);
+	} while (time(NULL) - start < riscv_command_timeout_sec);
+	assert(result == ERROR_OK);
+	assert(riscv_batch_dmi_busy_encountered(batch));
+	/* Reset dmi_busy_delay, so the value doesn't get too big. */
+	LOG_TARGET_DEBUG(target, "dmi_busy_delay is reset.");
+	info->dmi_busy_delay = 0;
+	/* TODO: Introduce a separate timeout for this. */
```
JanMatCodasip: `/* TODO: Introduce a separate timeout for this. */`

Do you mean to introduce a separate timeout just for batches? I wonder if that is worth it. I would recommend to not do it.

If the user is working on a very slow target - e.g. a slow simulation - they will most likely want to increase just one "master timeout" value (instead of configuring multiple timeouts).

src/target/riscv/riscv-013.c
```diff
@@ -977,27 +977,51 @@ static int execute_abstract_command(struct target *target, uint32_t command,
 	return ERROR_OK;
 }
 
-static riscv_reg_t read_abstract_arg(struct target *target, unsigned index,
-		unsigned size_bits)
+static void add_data_regs_read(struct riscv_batch *batch, unsigned int index,
+		unsigned int size_bits)
 {
+	const unsigned int size_in_words = size_bits / 32;
+	assert(size_bits % 32 == 0);
```
JanMatCodasip, line 984: I believe you can add one more assert here:

```suggestion
	assert(size_bits % 32 == 0);
	assert(size_bits >= 32);	
```

src/target/riscv/riscv-013.c
```diff
@@ -977,27 +977,51 @@ static int execute_abstract_command(struct target *target, uint32_t command,
 	return ERROR_OK;
 }
 
-static riscv_reg_t read_abstract_arg(struct target *target, unsigned index,
-		unsigned size_bits)
+static void add_data_regs_read(struct riscv_batch *batch, unsigned int index,
```
JanMatCodasip: I've tried to find a more descriptive name:

```suggestion
static void abstract_data_read_fill_batch(struct riscv_batch *batch, unsigned int index,
```

src/target/riscv/riscv-013.c
```diff
@@ -977,27 +977,51 @@ static int execute_abstract_command(struct target *target, uint32_t command,
...
+	const unsigned int size_in_words = size_bits / 32;
+	assert(size_bits % 32 == 0);
+	const unsigned int offset = index * size_in_words;
+	for (unsigned int i = 0; i < size_in_words; ++i) {
+		const unsigned int reg_address = DM_DATA0 + offset + i;
+		riscv_batch_add_dm_read(batch, reg_address);
+	}
+}
+
+static riscv_reg_t read_data_regs(struct riscv_batch *batch,
```
JanMatCodasip: (similar to previous)

```suggestion
static riscv_reg_t abstract_data_get_from_batch(struct riscv_batch *batch,
```

src/target/riscv/riscv-013.c
```diff
@@ -977,27 +977,51 @@ static int execute_abstract_command(struct target *target, uint32_t command,
...
+		unsigned int index, unsigned int size_bits)
+{
+	const unsigned int size_in_words = size_bits / 32;
+	assert(size_bits % 32 == 0);
+	assert(size_in_words * sizeof(uint32_t) <= sizeof(riscv_reg_t));
```
JanMatCodasip, line 999: I'd recommend to rewrite the assert to a more direct form, for example:

```suggestion
	assert(size_in_words == 1 || size_in_words == 2);
```
or

```suggestion
	assert(size_in_bits == 32 || size_in_bits == 64);
```

or, if you wish to check also the riscv_reg_t datatype (to prevent the unlikely of someone changing its size later):

```suggestion
	assert(size_in_words == 1 || size_in_words == 2);
	assert(sizeof(riscv_reg_t) == 8);	
```

src/target/riscv/riscv-013.c
```diff
@@ -977,27 +977,51 @@ static int execute_abstract_command(struct target *target, uint32_t command,
...
-		case 64:
-			if (dm_read(target, &v, DM_DATA0 + offset + 1) == ERROR_OK)
-				value |= ((uint64_t)v) << 32;
-			/* falls through */
-		case 32:
-			if (dm_read(target, &v, DM_DATA0 + offset) == ERROR_OK)
-				value |= v;
+	for (unsigned int i = 0; i < size_in_words; ++i) {
+		const uint32_t v = riscv_batch_get_dmi_read_data(batch, i);
+		value |= (riscv_reg_t)v << (i * 32);
```
JanMatCodasip: Here I'd not expect that readers remember the operator precedence ;)

```suggestion
		value |= ((riscv_reg_t)v) << (i * 32);
```

src/target/riscv/riscv-013.c
```diff
@@ -977,27 +977,51 @@ static int execute_abstract_command(struct target *target, uint32_t command,
...
+	assert(value);
+	RISCV013_INFO(info);
+	const unsigned char size_in_words = size_bits / 32;
+	assert(size_bits % 32 == 0);
+	struct riscv_batch * const batch = riscv_batch_alloc(target, size_in_words,
+			info->dmi_busy_delay);
+	add_data_regs_read(batch, index, size_bits);
+	int result = batch_run_timeout(target, batch);
+	if (result == ERROR_OK)
+		*value = read_data_regs(batch, index, size_bits);
```
JanMatCodasip: Cosmetics: The type of `value` is `uint64_t` whereas the return type of `read_data_regs()` is `riscv_reg_t`.

I'd recommend to use the same type everywhere (you can pick either of them, I do not have strong preference).

src/target/riscv/riscv-013.c
JanMatCodasip: Also, I believe the functions are bit more readable when the assertions on input parameters are the very first statements of the function body.

If you are in favor of that, you can make this change here and also in the other functions.

JanMatCodasip (changes requested):
Overall, this change looks all right to me. I am sending some suggestions.

src/target/riscv/riscv-013.c
en-sc: There is an issue with `riscv_batch_add_dm_write(..., /*read_back*/ false)` -- it discards the value read from the target.
Therefore the approach of "the next scan returns the result of the previous one" does not work.
Using `riscv_batch_add_dm_write(..., false)` requires another mechanism to determine which scan resulted in a busy response.

One approach would be to eliminate the `read_back` argument of `riscv_batch_add_dm_write()` and always read the response.
However, this is not desirable. Discarding the response significantly reduces USB traffic, which quite often is the bottleneck limiting the performance. And `riscv_batch_add_dm_write(..., /*read_back*/ false)` is exactly what is used in the most time-consuming operations -- memory writes.
E.g. when writing via Program Buffer there is no need to rely on DMI busy check for each scan. The DMI busy status can be determined by the response to the last NOP in the batch. And once it is known that the batch encountered DMI busy, OpenOCD reads target's register that contains the address of the element on which the busy status was encountered. An added benefit is that the algorithm is the same for `abstractcs.cmderr == busy`.

However, IMHO, adapting this functions to be usable in such cases can be beneficial. This can be done by the following changes:
`riscv_batch_was_scan_busy()` should use a callback in case `field->in_value` is zero. This callback will be optionally specified on batch allocation. The callback is specific to the content of the batch.
@JanMatCodasip, please let me know your thoughts on this.

src/target/riscv/batch.c
en-sc: I would like to leave the check here for now.
I think it's important to check none of the scans were skipped by the algorithm.
Also, the change to `riscv_batch_was_scan_busy()` suggested [here](https://github.com/riscv-collab/riscv-openocd/pull/1033#discussion_r1604781335) will fix the issue.

src/target/riscv/riscv-013.c
en-sc: Currently there are:
* `riscv_command_timeout_sec` changed by `riscv set_command_timeout_sec` TCL command.
* `riscv_reset_timeout_sec` changed by `riscv set_reset_timeout_sec` TCL command.

I think there should be a single `riscv set_timeout [command|reset]` command.
It would be quite easy to add `max`, `min`, `batch`, `dm_reset` after that.
It may be a bit excessive, but I think it would be better and would provide better configurability.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Is this really necessary?
I would like to avoid using arbitrary constants where possible (1 & 2, 32 & 64, 8).

Will address if you insist.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: > I believe you can add one more assert here:

Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/batch.c

src/target/riscv/batch.c

Though I left `assert(batch->used_scans > 0)` where it was. It's related to the loop condition, and I would like to leave the assertion as close to what it guards as possible.

src/target/riscv/batch.h
en-sc: Thanks for the suggestion!
The concern is valid.
However, I would word the comment differently. It's important, that the run didn't encounter an error.
Please, take a look.

src/target/riscv/batch.h
en-sc: Thanks for the catch! `riscv_bacth_continue()` was made redundant by adding `riscv_batch_run_from()`, but I've missed deleting the declaration.

src/target/riscv/batch.h

src/target/riscv/batch.c

src/target/riscv/batch.c

src/target/riscv/batch.c

src/target/riscv/batch.h

src/target/riscv/riscv-013.c
JanMatCodasip: As mentioned in the above comment, I am not in favor of introducing multiple timeout options. The reason is that I do not see any advantage for the user, only complications:

- The default timeout should be sufficient for vast majority of use cases.
- For special use cases (like slow hardware simulation) the user should ideally increase just a single timeout value (`riscv_command_timeout_sec <new_value>`) and all algorithms in OpenOCD should start using it. 
- More configurability in this case does not bring advantages but complicates the usage.

I would even recommend to deprecate the `riscv set_reset_timeout_sec` command and have only a single timeout value to configure: `riscv set_command_timeout_sec`.

src/target/riscv/batch.h
```diff
@@ -44,26 +42,36 @@ struct riscv_batch {
...
+/* Executes this batch of JTAG DTM DMI scans, starting form "start" scan.
+ * If batch is run for the first time, it is expected that "start" is zero.
+ * It is expected that the batch ends with a DMI NOP operation.
+ * "idle_count" is the number of JTAG Run-Test-Idle cycles to add in-between
+ * the scans.
+ * If "resets_delays" is true, the algorithm will stop inserting idle cycles
+ * (JTAG Run-Test-Idle) after "reset_delays_after" number of scans is
  * performed.  This is useful for stress-testing of RISC-V algorithms in
  * OpenOCD that are based on batches.
  */
```
JanMatCodasip: Detail: I have added few blank lines for readability (the comment is quite long).

```suggestion
/* Executes this batch of JTAG DTM DMI scans, starting form "start" scan.
 *
 * If batch is run for the first time, it is expected that "start" is zero.
 * It is expected that the batch ends with a DMI NOP operation.
 *
 * "idle_count" is the number of JTAG Run-Test-Idle cycles to add in-between
 * the scans.
 *
 * If "resets_delays" is true, the algorithm will stop inserting idle cycles
 * (JTAG Run-Test-Idle) after "reset_delays_after" number of scans is
 * performed.  This is useful for stress-testing of RISC-V algorithms in
 * OpenOCD that are based on batches.
 */
```

src/target/riscv/riscv-013.c
JanMatCodasip: The constants are not arbitrary - they are the list of allowed inputs to the function.

> Is this really necessary?

No, but I believe it helps code readability :)

> Will address if you insist.

Please, choose the style yourself, I won't insist on this one.

src/target/riscv/riscv-013.c
JanMatCodasip: Thank you for explanation.

> One approach would be to eliminate the read_back argument of riscv_batch_add_dm_write() and always read the response. However, this is not desirable.

I fully agree. The `read_back` argument was introduced for a good reason (reduction of traffic to the JTAG adapter) and I believe it should be kept.

> (...) This can be done by the following changes: riscv_batch_was_scan_busy() should use a callback in case field->in_value is zero. This callback will be optionally specified on batch allocation. The callback is specific to the content of the batch.

For simplicity of the batch API (and the code that uses it), I would recommend to not introduce any callbacks. Instead, I would strongly prefer to leave the responsibility on the caller to determine what part of the batch needs to be re-executed. That is, the caller inspects the output from the batch and decide if and what needs to re-execute. Then it simply calls `riscv_batch_run_from()` with the appropriate `start_idx` value.

src/target/riscv/batch.c
JanMatCodasip: This is related to https://github.com/riscv-collab/riscv-openocd/pull/1033#discussion_r1606923976, so I have continuted the discussion there. 

I am closing this thread.

src/target/riscv/batch.h
JanMatCodasip: I'm sorry to keep nit-picking on this comment. The current phrasing ("without an error") still may lead the reader to believe that "DMI busy was not encountered".

One more proposal:

```
/* Flag indicating that the last run of the batch finished without an error
 * from the underlying JTAG layer of OpenOCD - all scans were performed.
 */
```

src/target/riscv/batch.h

src/target/riscv/batch.h
en-sc: Addressed

src/target/riscv/riscv-013.c
en-sc: You are right. Removed the `TODO` comment.

JanMatCodasip (approved):
LGTM, thank you.

PR #1036, opened 2024-03-28
Merge up to a35e254c5383008cdacf7838a777f7f17af5eeb1 from upstream

en-sc:


en-sc:
`checkpatch` fails due to:
* `MACRO_ARG_REUSE` and `MACRO_ARG_PRECEDENCE` in `src/helper/list.h` (a213afad09e50402be8d02298f48757ff4e2e1cf).
* `MACRO_ARG_REUSE`  in `src/target/mips32.h` (561ea48d83ae4b83ff823888a80cbcc282b61333 and 01a797af143398c6f3aa1eb6f8949deff4ccf044)

Will adjust commit message to include `Checkpatch-ignore`

en-sc:
Adding appropriate `Checkpatch-ignore` to the commit message did not fix the issue. Seems like `checkpatch` does not properly support merge commits.
Please, ignore `checkpatch` failure.

src/flash/nor/nrf5.c
```diff
@@ -120,10 +95,10 @@ struct nrf52_ficr_info {
...
 
 enum nrf5_features {
-	NRF5_FEATURE_SERIES_51	= 1 << 0,
-	NRF5_FEATURE_SERIES_52	= 1 << 1,
-	NRF5_FEATURE_BPROT		= 1 << 2,
-	NRF5_FEATURE_ACL_PROT	= 1 << 3,
+	NRF5_FEATURE_SERIES_51	= BIT(0),
+	NRF5_FEATURE_SERIES_52	= BIT(1),
+	NRF5_FEATURE_BPROT		= BIT(2),
+	NRF5_FEATURE_ACL_PROT	= BIT(3),
```
MarekVCodasip, line 101: Maybe we could start using the BIT() macro in our code as well? What others think?

MarekVCodasip (approved):
LGTM

Please, see my comment about the use of the BIT() macro.

src/flash/nor/nrf5.c
JanMatCodasip: @MarekVCodasip Thank you for taking the time to review this upstream sync.

As for the BIT() macro, sure, we can use it whenever suitable. Though, I think we can handle most of our bit-masking cases directly by constants from files `debug_defines.h` and `encoding.h`.

PR #1045, opened 2024-04-19
target/riscv: fix segfault on RISC-V Debug v0.11

en-sc:
Change-Id: I5be9153f364a5ad5083e809f479aed5378e420c9

PR #1044, opened 2024-04-19
target/riscv: stop using register_get/set for 0.11 targets

en-sc:
Caching is somewhat handled in `riscv-011.c`. Handling it additionaly in `riscv.c` may cause problems. Sice there is no simulator that supports RISC-V Debug Specification v0.11, so it is not feaseable to automate testing.
This commit separates 0.11 register accesses and unlocks further development in this area.

Change-Id: I73ff17ef85106c4ababa38319f446f6c384a1750

PR #1022, opened 2024-02-29
[NFC] target/riscv: refactor `init_registers()`

en-sc:
The logic in `init_registers()` was quite convoluted. Initialization of each `struct reg` field is separated into function `gdb_regno_<field_name>()`.
IMHO, this makes it much easier to reason about the code.

Change-Id: Id7faa1464ce026cc5025585d0a6a95a01fb39cee

PR #1046, opened 2024-04-19
target/riscv/riscv-011.c: fix access to non-existent register

en-sc:
`reg` is a number in register cache, as evident by the following call to
`reg_cache_set()`. `CSR_DCSR` is `GDB_REGNO_DCSR - 65`. This results in
setting cache value for another register, which does not exist, and
causes a segfault if all non-existent registers are not allocated a
value (`reg->value == NULL`).

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1771,10 +1771,10 @@ static riscv_error_t handle_halt_routine(struct target *target)
 					reg = S0;
 					break;
 				case 31:
-					reg = CSR_DPC;
+					reg = GDB_REGNO_DPC;
 					break;
 				case 32:
-					reg = CSR_DCSR;
+					reg = GDB_REGNO_DCSR;
 					break;
 				default:
 					assert(0);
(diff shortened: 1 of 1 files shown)


en-sc:
@TommyMurphyTM1234, if it's not too much to ask, could you please also check this one?

It's an alternative to #1045.

I would have much preferred to merge this one instead.

The thing is, #1045 addresses the symptom, not the root cause. The segfault is caused by an access to a value of non-existent register:
```
#0  buf_set_u64 (value=1024, num=32, first=0, _buffer=0x0) at ./src/helper/binarybuffer.h:69
#1  reg_cache_set (number=number@entry=1969, value=1024, target=<optimized out>, target=<optimized out>) at src/target/riscv/riscv-011.c:1214
```
In #1022 I've changed the procedure allocating the buffer for cached register's value. I've stopped allocating the space for non-existent registers. I think this is reasonable. This is why the issue popped-up.

However, AFAIU, the issue was present, just silent. It seems like register cache is mismanaged in `riscv-011.c` -- value of `dcsr` is saved in cache, but using the wrong index (`CSR_DCSR` instead of `GDB_REGNO_DCSR`). So the value of `dcsr` was cached in two cache entries -- sometimes in one, sometimes in the other. This may be intentional, though I can't immediately see a reason for such strange behavior.

This patch should fix this.

TommyMurphyTM1234:
Yes I can check that but probably not until tomorrow if that's ok.

TommyMurphyTM1234:
I tried that and it works.
Just to be clear, I just ran OpenOCD so that it examined/connected to the HiFive1 board.
I didn't do any more extensive testing.
But if you need that (e.g. a run of the test suite?) I can do that.
Otherwise I can revert to looking at the original PR:

* https://github.com/riscv-collab/riscv-openocd/pull/1044

JanMatCodasip (approved):
Reviewed visually & LGTM, thank you.

PR #1054, opened 2024-04-23
target/riscv/riscv-011: pc and dpc should be cached at the same location

en-sc:
Prior to the commit, pc was cached at `info->dpc`, but dpc at register cache.

Change-Id: I369788441dbe21bcf8fc360d2e97e98096b25e3a


JanMatCodasip (approved):
Checked visually, LGTM, thank you.

PR #1055, opened 2024-04-23
target/riscv: use breakpoint_hw_set/watchpoint_set to properly initialize bp/wp descriptor

aap-sc:
This fixes the following issue:

```
> bp 0x80000008 4 hw
[riscv.cpu0] Found 4 triggers
breakpoint set at 0x80000008
> bp                
Hardware breakpoint(IVA): addr=0x80000008, len=0x4, num=1651340654
```

`num` is initialized for RISC-V targets.

Also see https://review.openocd.org/c/openocd/+/8181 for a discussion


MarekVCodasip (approved):
LGTM Thanks for the fix.

en-sc (approved):
LGTM

PR #1056, opened 2024-04-23
target/riscv: fix halt reason for targets that do not support hit bit on triggers

aap-sc:
Before this patch the following behavior is observed on targets that do not support hit bit:

```
bp 0x80000004 4 hw
resume 0x80000000
riscv.cpu halted due to watchpoint
```

This happens because the current implementation relies on the presence of hit bit way too much. While working on this patch few defects in hit bit-based trigger detection were discovered, added appropriate TODOs.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -34,6 +34,8 @@
 
 #define DBUS						0x11
 
+#define RISCV_TRIGGER_HIT_NOT_FOUND ((int64_t)-1)
+
 static uint8_t ir_dtmcontrol[4] = {DTMCONTROL};
 struct scan_field select_dtmcontrol = {
 	.in_value = NULL,
@@ -1550,17 +1552,24 @@ int riscv_remove_watchpoint(struct target *target,
 /**
  * Look at the trigger hit bits to find out which trigger is the reason we're
  * halted.  Sets *unique_id to the unique ID of that trigger. If *unique_id is
- * ~0, no match was found.
+ * RISCV_TRIGGER_HIT_NOT_FOUND, no match was found.
  */
-static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
+
+static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_id)
 {
+	/* FIXME: this function assumes that we have only one trigger that can
+	 * have hit bit set. Debug spec allows hit bit to bit set if a trigger has
+	 * matched but did not fire. Such targets will receive erroneous results.
+	 */
+
+	// FIXME: Add hit bits support detection and caching
 	RISCV_INFO(r);
 
 	riscv_reg_t tselect;
 	if (riscv_get_register(target, &tselect, GDB_REGNO_TSELECT) != ERROR_OK)
 		return ERROR_FAIL;
 
-	*unique_id = ~0;
+	*unique_id = RISCV_TRIGGER_HIT_NOT_FOUND;
 	for (unsigned int i = 0; i < r->trigger_count; i++) {
 		if (r->trigger_unique_id[i] == -1)
 			continue;
@@ -1594,15 +1603,15 @@ static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
 				hit_mask = CSR_ETRIGGER_HIT(riscv_xlen(target));
 				break;
 			default:
-				LOG_TARGET_DEBUG(target, "Trigger %d has unknown type %d", i, type);
+				LOG_TARGET_DEBUG(target, "Trigger %u has unknown type %d", i, type);
 				continue;
 		}
 
-		/* Note: If we ever use chained triggers, then this logic needs
-		 * to be changed to ignore triggers that are not the last one in
-		 * the chain. */
+		/* FIXME: this logic needs to be changed to ignore triggers that are not
+		 * the last one in the chain. */
 		if (tdata1 & hit_mask) {
-			LOG_TARGET_DEBUG(target, "Trigger %d (unique_id=%d) has hit bit set.", i, r->trigger_unique_id[i]);
+			LOG_TARGET_DEBUG(target, "Trigger %u (unique_id=%" PRIi64 ") has hit bit set.",
+				i, r->trigger_unique_id[i]);
 			if (riscv_set_register(target, GDB_REGNO_TDATA1, tdata1 & ~hit_mask) != ERROR_OK)
 				return ERROR_FAIL;
 
@@ -2268,6 +2277,33 @@ int riscv_flush_registers(struct target *target)
 	return ERROR_OK;
 }
 
+static enum target_debug_reason
+derive_debug_reason_without_hitbit(const struct target *target, riscv_reg_t dpc)
+{
+	/* TODO: if we detect that etrigger/itrigger/icount is set, we should
+	 * just report DBG_REASON_UNKNOWN, since we can't disctiguish these
+	 * triggers from BP/WP or from other triggers of such type. However,
+	 * currently this renders existing testsuite as failing. We need to
+	 * fix the testsuite first
+	 */
+	// TODO: the code below does not handle context-aware trigger types
+	for (const struct breakpoint *bp = target->breakpoints; bp; bp = bp->next) {
+		// TODO: investigate if we need to handle bp length
+		if (bp->type == BKPT_HARD && bp->is_set && bp->address == dpc) {
+			// FIXME: bp->linked_brp is uninitialized
+			if (bp->asid) {
+				LOG_TARGET_ERROR(target,
+					"can't derive debug reason for context-aware breakpoint: "
+					"unique_id = %" PRIu32 ", address = %" TARGET_PRIxADDR
+					", asid = %" PRIx32 ", linked = %d",
+					bp->unique_id, bp->address, bp->asid, bp->linked_brp);
+				return DBG_REASON_UNDEFINED;
+			}
+			return DBG_REASON_BREAKPOINT;
+		}
+	}
+	return DBG_REASON_WATCHPOINT;
+}
 /**
  * Set OpenOCD's generic debug reason from the RISC-V halt reason.
  */
@@ -2280,13 +2316,52 @@ static int set_debug_reason(struct target *target, enum riscv_halt_reason halt_r
 			target->debug_reason = DBG_REASON_BREAKPOINT;
 			break;
 		case RISCV_HALT_TRIGGER:
-			if (riscv_hit_trigger_hit_bit(target, &r->trigger_hit) != ERROR_OK)
+			target->debug_reason = DBG_REASON_UNDEFINED;
+			if (riscv_trigger_detect_hit_bits(target, &r->trigger_hit) != ERROR_OK)
 				return ERROR_FAIL;
-			target->debug_reason = DBG_REASON_WATCHPOINT;
-			/* Check if we hit a hardware breakpoint. */
-			for (struct breakpoint *bp = target->breakpoints; bp; bp = bp->next) {
-				if (bp->unique_id == r->trigger_hit)
+			// FIXME: handle multiple hit bits
+			if (r->trigger_hit != RISCV_TRIGGER_HIT_NOT_FOUND) {
+				/* We scan for breakpoints first. If no breakpoints are found we still
+				 * assume that debug reason is DBG_REASON_BREAKPOINT, unless
+				 * there is a watchpoint match - This is to take
+				 * ETrigger/ITrigger/ICount into account
+				 */
+				LOG_TARGET_DEBUG(target,
+					"Active hit bit is detected, trying to find trigger owner.");
+				for (struct breakpoint *bp = target->breakpoints; bp; bp = bp->next) {
+					if (bp->unique_id == r->trigger_hit) {
+						target->debug_reason = DBG_REASON_BREAKPOINT;
+						LOG_TARGET_DEBUG(target,
+							"Breakpoint with unique_id = %" PRIu32 " owns the trigger.",
+							bp->unique_id);
+					}
+				}
+				if (target->debug_reason == DBG_REASON_UNDEFINED) {
+					// by default we report all triggers as breakpoints
 					target->debug_reason = DBG_REASON_BREAKPOINT;
+					for (struct watchpoint *wp = target->watchpoints; wp; wp = wp->next) {
+						if (wp->unique_id == r->trigger_hit) {
+							target->debug_reason = DBG_REASON_WATCHPOINT;
+							LOG_TARGET_DEBUG(target,
+								"Watchpoint with unique_id = %" PRIu32 " owns the trigger.",
+								wp->unique_id);
+						}
+					}
+				}
+			} else {
+				LOG_TARGET_DEBUG(target,
+					"No trigger hit found, deriving debug reason without it.");
+				riscv_reg_t dpc;
+				if (riscv_get_register(target, &dpc, GDB_REGNO_DPC) != ERROR_OK)
+					return ERROR_FAIL;
+				/* Here we don't have the hit bit set (likely, HW does not support it).
+				 * We are trying to guess the state. But here comes the problem:
+				 * if we have etrigger/itrigger/icount raised - we can't really
+				 * distinguish it from the breakpoint or watchpoint. There is not
+				 * much we can do here, except for checking current PC against pending
+				 * breakpoints and hope for the best)
+				 */
+				target->debug_reason = derive_debug_reason_without_hitbit(target, dpc);
 			}
 			break;
 		case RISCV_HALT_INTERRUPT:
--- a/src/target/riscv/riscv.h
+++ b/src/target/riscv/riscv.h
@@ -159,11 +159,11 @@ struct riscv_info {
 	 * >= 0: unique_id of the breakpoint/watchpoint that is using it.
 	 * Note that in RTOS mode the triggers are the same across all harts the
 	 * target controls, while otherwise only a single hart is controlled. */
-	int trigger_unique_id[RISCV_MAX_HWBPS];
+	int64_t trigger_unique_id[RISCV_MAX_HWBPS];
 
 	/* The unique id of the trigger that caused the most recent halt. If the
 	 * most recent halt was not caused by a trigger, then this is -1. */
-	uint32_t trigger_hit;
+	int64_t trigger_hit;
 
 	/* The number of entries in the program buffer. */
 	int progbuf_size;


aap-sc:
Related to https://github.com/riscv-collab/riscv-openocd/pull/1029

aap-sc:
@JanMatCodasip , @MarekVCodasip could you take a look?

src/target/riscv/riscv.c
```diff
@@ -1815,6 +1823,33 @@ int riscv_flush_registers(struct target *target)
...
+					"unique_id = %" PRIu32 ", address = %" TARGET_PRIxADDR
+					", asid = %" PRIx32 ", linked = %d",
+					bp->unique_id, bp->address, bp->asid, bp->linked_brp);
+				return DBG_REASON_UNDEFINED;
+			}
+			return DBG_REASON_BREAKPOINT;
+		}
+	}
+	return DBG_REASON_WATCHPOINT;
+}
```
MarekVCodasip, line 2306: For function `derive_debug_reason_without_hitbit()`, I would recommend the following logic (pseudocode). It is similar your code, just extend it by the use of `riscv_hit_watchpoint()` for more clever watchpoint detection:

```
if (watchpoint hit using existing function riscv_hit_watchpoint()) {
    return DBG_REASON_WATCHPOINT;
}

if (breakpoint hit using existing logic) {
    if (bp->asid) {
         LOG_TARGET_ERROR(target, "BUG: unexpected context aware breakpoint in RISC-V target");
         return DBG_REASON_UNDEFINED;
    }
    return DBG_REASON_BREAKPOINT;
}

return DBG_REASON_UNDEFINED;
```

The output of `riscv_hit_watchpoint()` could be cached so that it does not have to be recomputed again later when the `gdb_server.c` - `gdb_signal_reply()` needs the information again.

src/target/riscv/riscv.c
```diff
@@ -1545,17 +1547,23 @@ int riscv_remove_watchpoint(struct target *target,
 /**
  * Look at the trigger hit bits to find out which trigger is the reason we're
  * halted.  Sets *unique_id to the unique ID of that trigger. If *unique_id is
- * ~0, no match was found.
+ * RISCV_TRIGGER_HIT_NOT_FOUND, no match was found.
  */
-static int riscv_hit_trigger_hit_bit(struct target *target, uint32_t *unique_id)
+
+static int riscv_hit_trigger_hit_bit(struct target *target, int64_t *unique_id)
```
MarekVCodasip: If  you like, I suggest a different name:

`riscv_trigger_detect_hit_bits()` or similar.

src/target/riscv/riscv.c
```diff
@@ -1545,17 +1547,23 @@ int riscv_remove_watchpoint(struct target *target,
...
 {
+	/* FIXME: this function assumes that we have only one trigger that can
+	 * have hit bit set. Debug spec allows hit bit to bit set if a trigger has
+	 * matched but did not fire. Such targets will receive erroneous results.
+	 */
```
MarekVCodasip, line 1563: In these cases. I think we should clear all hit bits. Since we cannot know which trigger triggered the target halt first and is target dependent. Can you please add it to the FIXME? Or if you really like, you can implement it right away. Thanks.

src/target/riscv/riscv.c
```diff
@@ -1827,13 +1862,40 @@ static int set_debug_reason(struct target *target, enum riscv_halt_reason halt_r
...
+					return ERROR_FAIL;
+				/* Here we don't have the hit bit set (likely, HW does not support it).
+				 * We are trying to guess the state. But here comes the problem:
+				 * if we have etrigger/itrigger/icount raised - we can't really
+				 * distinguish it from the breakpoint or watchpoint. There is not
+				 * much we can do here, except for checking current PC against pending
+				 * breakpoints and hope for the best)
+				 */
+				target->debug_reason = derive_debug_reason_without_hitbit(target, dpc);
 			}
```
MarekVCodasip: Long term, I think we should be caching if the target supports hit bits. This would be a new tri-state attribute (YES,NO,MAYBE). If we detect that the target supports hit bits, it would allow us to skip the deriving the debug reason without hitbits.

Can you please add a FIXME to remind us later? Something like `FIXME: Add hit bits support detection and caching`

src/target/riscv/riscv.c
```diff
@@ -1827,13 +1862,40 @@ static int set_debug_reason(struct target *target, enum riscv_halt_reason halt_r
...
+			target->debug_reason = DBG_REASON_UNDEFINED;
 			if (riscv_hit_trigger_hit_bit(target, &r->trigger_hit) != ERROR_OK)
 				return ERROR_FAIL;
-			target->debug_reason = DBG_REASON_WATCHPOINT;
-			/* Check if we hit a hardware breakpoint. */
-			for (struct breakpoint *bp = target->breakpoints; bp; bp = bp->next) {
-				if (bp->unique_id == r->trigger_hit)
+			// FIXME: handle multiple hit bits
+			if (r->trigger_hit != RISCV_TRIGGER_HIT_NOT_FOUND) {
+				/* We scan for breakpoints first. If no breakpoints are found we still
```
MarekVCodasip, line 2324: Could you add a LOG_TARGET_DEBUG here to inform us that the code is searching for hit bits?

Something like `LOG_TARGET_DEBUG(target, "Trigger hit found.");`

src/target/riscv/riscv.c
```diff
@@ -1827,13 +1862,40 @@ static int set_debug_reason(struct target *target, enum riscv_halt_reason halt_r
...
+				if (target->debug_reason == DBG_REASON_UNDEFINED) {
+					// by default we report all triggers as breakpoints
 					target->debug_reason = DBG_REASON_BREAKPOINT;
+					for (struct watchpoint *wp = target->watchpoints; wp; wp = wp->next) {
+						if (wp->unique_id == r->trigger_hit)
+							target->debug_reason = DBG_REASON_WATCHPOINT;
+					}
+				}
+			} else {
+				riscv_reg_t dpc;
```
MarekVCodasip, line 2354: And add LOG_TARGET_DEBUG here to inform us that no hit bits were found, but we are still searching for potential triggers.

Something like `LOG_TARGET_DEBUG(target, "No trigger hit found, deriving debug reason without a hitbit.");`

MarekVCodasip (changes requested):
Thanks for the patch.

It looks good. I just suggest some further changes/improvements.

src/target/riscv/riscv.c
aap-sc: For what it's worth. hit bits are cleared by OpenOCD already. I've checked this visually. by inspecting the code.

src/target/riscv/riscv.c
aap-sc: @MarekVCodasip I plan to re-write this in a somewhat similar manner (except for calling riscv_hit_watchpoint, more on that later). However, there are few issues.

1. One of them is this part of TODO above:

```
/* TODO: if we detect that etrigger/itrigger/icount is set, we should
	 * just report DBG_REASON_UNKNOWN, since we can't disctiguish these
	 * triggers from BP/WP or from other triggers of such type. However,
	 * currently this renders existing testsuite as failing. We need to
	 * fix the testsuite first
```

That is the testsuite needs to be fixed first. I plan to address this, but not a part of this PR.

2. I'm strongly against any usage of riscv_hit_watchpoint here (or anywhere else). In fact in a subsequent commit I'm going to add a new parameter to get rid of calling this function at all. This function is useless (IMHO) and dangerous .

The reason for such a strong opinion regarding `riscv_hit_watchpoint` is the following.

I. We need to read memory. This is NOT a simple operation. This requires us to figure out the context (like was it user Mode, was there hypervisor enabled, etc). Then (optionally) perform a SW page walk and only then to attempt decode instruction.

II. Decoding instruction by itself is NOT a trivial task. As https://github.com/riscv-collab/riscv-openocd/pull/1029 show it requires significant amount of code that is hard to test. Another problem comes that the list of memory operations depends on extensions supported by target. There is RVV extension, there is also support for stack manipulation extension or whatever abysmal extension will be ratified in future. Unless there is a separate library for this kind of task - I don't think that the respected disassembler belongs to  OpenOCD codebase.

III. Decoding instruction is one thing. We also need to figure out the virtual address. This is NOT a simple task. For some instructions (like indexed loads/stores from RVV) it requires us to know the semantic of instruction and (possibly) additional execution context. That is it requires us to have a mini simulator embedded in OpenOCD. I don't think this is reasonable.

Given points I, II and III I would drop the functionality `riscv_hit_watchpoint` entirely  for future versions and require all targets to support hit bits if they want this kind of functionality.

aap-sc: > I would drop the functionality `riscv_hit_watchpoint` entirely for future versions and require all targets to support hit bits if they want this kind of functionality.

TLDR: I really dislike `riscv_hit_watchpoint` :)

src/target/riscv/riscv.c
aap-sc: sure, I'll add a FIXME. Though I should mention that most likely hit bit support detection/derivation won't be trivial (if at all possible)

src/target/riscv/riscv.c
aap-sc: yeah... this would be useful. Will add.

aap-sc: addressed

src/target/riscv/riscv.c
aap-sc: addressed

src/target/riscv/riscv.c
aap-sc: added

src/target/riscv/riscv.c
aap-sc: renamed, thanks for the suggestion

aap-sc:
@MarekVCodasip could you take  a look once again?

src/target/riscv/riscv.c
```diff
@@ -1843,13 +1879,52 @@ static int set_debug_reason(struct target *target, enum riscv_halt_reason halt_r
...
+							target->debug_reason = DBG_REASON_WATCHPOINT;
+							LOG_TARGET_DEBUG(target,
+								"Watchpoint with unique_id = %" PRIu32 " owns the trigger.",
+								wp->unique_id);
+						}
+					}
+				}
+			} else {
+				LOG_TARGET_DEBUG(target,
+					"No trigger hit found, deriving debug reason without a hitbit.");
```
JanMatCodasip: Just a very minor detail:

```suggestion
					"No trigger hit found, deriving debug reason without it.");
```

src/target/riscv/riscv.c
JanMatCodasip: I can see that the trigger hit bit is cleared on line 1615: https://github.com/riscv-collab/riscv-openocd/pull/1056/files#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR1615

However, the loop is exited when the first hit is found, so subsequent hit bits will not be cleared.

We believe the algorithm should continue clearing the "hit" bits of all triggers.

src/target/riscv/riscv.c
JanMatCodasip: Hello @aap-sc, thank you for explaining your reasoning around `riscv_hit_watchpoint`.

> This function [riscv_hit_watchpoint] is useless (IMHO) and dangerous .

The function allows to inform GDB about watchpoint hit even on targets that do not support the trigger "hit" bits (see the `awatch`, `rwatch` and `watch` stop reasons in the GDB's `T` packet). 

The fact that OpenOCD users went great lengths to implement this mechanism shows that it was important to them. The mechanism need not be perfect to help users. If there are cases that we do not cover, like V extension, it would be good to warn the user.

> We need to read memory. This is NOT a simple operation. 

Reading memory -- one of the fundamental debugger operations -- should work reliably. If there are any concerns about certain cases not handled well, then that should be addressed as an independent issue (irrespective of riscv_hit_watchpoint()).

> Decoding instruction by itself is NOT a trivial task. As https://github.com/riscv-collab/riscv-openocd/pull/1029 show it requires significant amount of code that is hard to test. 

The code for decoding would be good candidate for unit-testing. I am considering whether to add unit-test capability to OpenOCD. 

> Another problem comes that the list of memory operations depends on extensions supported by target. (...)

> We also need to figure out the virtual address. This is NOT a simple task. For some instructions (like indexed loads/stores from RVV) it requires us to know the semantic of instruction and (possibly) additional execution context. (...)

The current version of the code (including https://github.com/riscv-collab/riscv-openocd/pull/1029) covers `I` and `C` extensions, which is a good baseline that will probably help significant amount of users. 

If an unsupported case is detected (like V extension), a warning can be issued. 

Alternatively, the user can warn the user anytime we detect that the user works with a core that does not support the "hit" bits.

---

TL,DR: I agree about the limitations of `riscv_hit_watchpoint()`, but still believe that the current version helps users. We should not strive to cover all possible cases. Instead, we can stick to the basics and warn the user about the limitations of the current algorithm.

aap-sc: > The fact that OpenOCD users went great lengths to implement this mechanism shows that it was important to them. 

Well, yeah. Probably there are cases when it is useful. Probably, my opinion is way too strong :).

> The mechanism need not be perfect to help users.

Yes, there are cases when it is terrible (see below). So I want to have an option to avoid these memory reads to reduce evasiveness. For example if we debug ONLY with OpenOCD without involvement of GDB we don't really need this functionality. 

> Reading memory -- one of the fundamental debugger operations -- should work reliably. 

The thing is that it **may not** not work reliably in all cases. For example: I constantly see designs that is capable to hang if there is something wrong with the system bus. For example, the underlying device does not send a response. We must not forget that people rarely use debugging with OpenOCD when everything is fine. One uses a debugger when something is (terribly) wrong. 

Reading memory is WAY too intrusive for SOME scenarios.

src/target/riscv/riscv.c
aap-sc: > I can see that the trigger hit bit is cleared on line 1615

Actually, I was talking about a different case. When we resume the target, we clear all hit bits anyway, since we temporally disable triggers when single-step is performed. If needed I can provide the exact lines. So technically we are fine for now.

> We believe the algorithm should continue clearing the "hit" bits of all triggers.

That being said, I agree that it would not hurt to continue the loop in this case. I'm just not sure if this should be part of a subsequent or this commit - will take a look once again.

aap-sc:
@JanMatCodasip for whatever reason my replies are not displayed properly, so I'll re-post them here. 

Regarding to your latest comments:

> I can see that the trigger hit bit is cleared on line 1615: 

Actually, I was talking about a different case. When we resume the target, we clear all hit bits anyway, since we temporally disable triggers when single-step is performed. If needed I can provide the exact lines. So technically we are fine for now.

> We believe the algorithm should continue clearing the "hit" bits of all triggers.

That being said, I agree that it would not hurt to continue the loop in this case. I'm just not sure if this should be part of a subsequent or this commit - will take a look once again.

> The fact that OpenOCD users went great lengths to implement this mechanism shows that it was important to them.

Well, yeah. Probably there are cases when it is useful. Probably, my opinion is way too strong :).

> The mechanism need not be perfect to help users.

Yes, there are cases when it is terrible (see below). So I want to have an option to avoid these memory reads to reduce invasiveness. For example if we debug ONLY with OpenOCD without involvement of GDB we don't really need this functionality.

> Reading memory -- one of the fundamental debugger operations -- should work reliably.

The thing is that it **may not** not work reliably in all cases. For example: I see design that is capable to hang if there is something wrong with the system bus. For example, the underlying device does not send a response after memory reference. We must not forget that people rarely use debugging with OpenOCD when everything is fine. One uses a debugger when something is (terribly) wrong. Reading memory may be considered way to intrusive for **some** scenarios.

src/target/riscv/riscv.c
JanMatCodasip: Thank you for further explaining your position. I basically agree with your concerns.

> So I want to have an option to avoid these memory reads to reduce evasiveness.

Considering the arguments you mentioned, I would even be OK with going a bit further:
- disable `riscv_hit_watchpoint()` by default
- create a command that allows the user to enable this alternative watchpoint hit detection algorithm
- if OpenOCD detects that "hit" bits are not implemented by the target:
  - produce a warning to the user
  - print a hint to the user how to enable the alternative watchpoint hit detection
  - print a warning to the user that the alternative detection has limitations (as we discussed above: it is not 100% reliable, not all ISA extensions are supported, extra memory reads are performed, etc.)

src/target/riscv/riscv.c
JanMatCodasip: Thank you for explanation. I missed the fact that the temporary disabling of watchpoints on resume or step clears the "hit" bits as a by-product.

I would still prefer the clearing of all hit bits to occur in `riscv_trigger_detect_hit_bits()`, because:
- It is very unclear to the code reader that the clearing of "hit" occurs at a different place.
- We could make debug prints when more watchpoints have been hit - for easier troubleshooting.
- The disabling of triggers before step/resume is somewhat questionable because the debugger (GDB, LLDB) is attempting the same operation, so we are often duplicate in OpenOCD what the debugger is already attempting. For that reason, the disabling of triggers on step/resume may one day change - in which case we would lose the "hit" bit clearing (perhaps without even noticing).

In any case, this can be done in a separate PR.

JanMatCodasip (dismissed):

src/target/riscv/riscv.c
aap-sc: finally addressed

JanMatCodasip (approved):

src/target/riscv/riscv.c
aap-sc: Just to wrap up this bit of  conversation (the part initiated by @JanMatCodasip)

I've looked at this in greater detail and here is my thoughts on the matter.

1. 

> I would still prefer the clearing of all hit bits to occur in riscv_trigger_detect_hit_bits(), because:

I don't think that `riscv_trigger_detect_hit_bits` should clear anything. It's job is to only detect hit bits. That's all. Clearing of hit bits should be handled elsewhere. Also I think we can make riscv_trigger_detect_hit_bits to detect all hit bits, not only the first one by allowing to call this function several times with a changed start index to make a lookup.

2.

> It is very unclear to the code reader that the clearing of "hit" occurs at a different place.

I agree that the way it is implemented now is confusing. I think that clearing of hit bits likely should happen in a dedicated function. 

3.

> The disabling of triggers before step/resume is somewhat questionable because the debugger 

It's not that questionable, actually. It is a matter of usability/convenience. In some debug scenarios it is not feasible to use high-level debugger like lldb or gdb because of side effects it has on a system (debugger loves to read memory here and there). Sometimes good-old telnet or tcl script is all we have.

Imagine a scenario when you hit a breakpoint or watchpoint. Now, you can read the processor state and want to move a little bit further using `step` command or even `resume`. If OpenOCD does not clear triggers - you'll end up stuck. In case of step you'll need to temporally disable the trigger. And you don't have a convenient interface to do that - you'll need to specify the virtual address than do the operation and then restore it! This is terrible for human and  inconvenient for scripting (since you'll need to parse text output of commands to get virtual address).  The  case of `resume` is similar (if not worse, since you won't have an opportunity to restore the trigger) - so the functionality to disable triggers has to stay.

That being said, RISC-V debug spec is incredibly quirky. For example it allows for various timings to be specified. Imagine that dreadful day when someone decides to implement **proper** support for timing "after"! In this case we may avoid disabling some triggers and we'll definitely require a dedicated functionality/function to clear the hit bit.

src/target/riscv/riscv.c
aap-sc: I've created https://github.com/riscv-collab/riscv-openocd/issues/1076 (I plan to start working on that as soon as this patch is landed).

aap-sc:
@MarekVCodasip there is a pending change request from your side...  @JanMatCodasip any chance you have capability to double-check if we should wait for Marek's reply or we can land the patch as is?

src/target/riscv/riscv.c
JanMatCodasip: > I think that clearing of hit bits likely should happen in a dedicated function.

I like that proposal. Agreed.

>> JanMatCodasip: The disabling of triggers before step/resume is somewhat questionable (...)

> aap-sc: It's not that questionable, actually. It is a matter of usability/convenience. (...) 

I agree that for convenience of TCL users, clearing of the breakpoints is strongly preferred. Side question: Would you know what other targets in OpenOCD do on TCL `resume` or `step` commands? Do they disable the breakpoints for the user or not? I wonder if the targets behave consistently in this regard. I haven't researched this yet.

What I was concerned (and considered questionable) is that riscv_openocd_step_impl() conditionally disables the current breakpoint (based on handle_breakpoints flag) but unconditionally disables all triggers - that doesn't seem consistent, and looks suspicious at the very least. For that reason, future changes in this part of the code are likely. And so we should not rely on disable_triggers() to clear the hit bits.

JanMatCodasip:
> MarekVCodasip there is a pending change request from your side... JanMatCodasip any chance you have capability to double-check if we should wait for Marek's reply or we can land the patch as is?

@aap-sc I have just checked with Marek and he does not have any further comments. Please feel free to go on with the merge. Thanks.

src/target/riscv/riscv.c
JanMatCodasip: Thank you :+1:

src/target/riscv/riscv.c
JanMatCodasip: (No need to block the merge request due to this open thread with broader discussion.)

MarekVCodasip (approved):

src/target/riscv/riscv.c
aap-sc:
```
Would you know what other targets in OpenOCD do on TCL resume or step commands? Do they disable the breakpoints for the user or not?
```

AFAIK most of the targets do that. Last time I checked (at the time I've created https://github.com/riscv-collab/riscv-openocd/pull/741).  These were: arm, aarch64, mips, stm - all the major ones.

I don't see that other targets do any special processing with respect to trigger-like objects (maybe they don't need it? more on that below). Some targets have a code that handles masking of ISR, though. 

You can search it like this:

```
git grep handle_breakpoints | grep if | awk -F: '{ print $1 }' | sort | uniq
```

> What I was concerned (and considered questionable) is that riscv_openocd_step_impl() conditionally disables the current breakpoint (based on handle_breakpoints flag) but unconditionally disables all triggers - that doesn't seem consistent, and looks suspicious at the very least.

The reason is quite simple, I guess. For triggers you can't figure out which instruction is affected . I guess this is the reason.  Whether or not other target **require** such behavior for trigger-like entities is a separate question, since it heavily depends on how single-step is implemented by hw/architecture (like target architecture can explicitly state that watch points are not fired during hw-assisted single step if a certain bit is set).  It should be noted that the code that disabled triggers for RISC-V openocd was there for a while and I always took it for granted.

**To summarize**:

1. I think we can conclude that the removal of SW breakpoint at the current address during step/resume is reasonable. Other targets do that too.
2. As for other types of triggers... Now that you've mentioned this I kind of share the concern. I think we should at least investigate the behavior for other targets. And I think it would  not hurt to ask in OpenOCD mailing list regarding the expected semantics of 'step/resume' for the case of watchpoints. I'll create a separate OS issue.

JanMatCodasip: Thank you for the reply and I agree with your conclusions/summary.

aap-sc: Ticket for further investigations: https://github.com/riscv-collab/riscv-openocd/issues/1079

PR #1061, opened 2024-04-26
target/riscv: only `dmactive` can be written if `dmactive` is low

en-sc:
There was an error introduced by
8319eee9e1ffc601b94b4223958180b49f8b8188.

According to RISC-V Debug Spec 1.0.0-rc1 [3.14.2. Debug Module Contro]:
> 0 (inactive): The module’s state, including authentication mechanism,
takes its reset values (the dmactive bit is the only bit which can be written to something other than its reset value).

`dmactive` was written together with `hartsel` and `hasel` in 8319eee9e1ffc601b94b4223958180b49f8b8188.

Change-Id: I11fba35cb87f8261c0a4a45e28b2813a5a086078

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1952,6 +1952,81 @@ static int wait_for_idle_if_needed(struct target *target)
 	return ERROR_OK;
 }
 
+static int reset_dm(struct target *target)
+{
+	/* TODO: This function returns an error when a DMI operation fails.
+	 * However, [3.14.2. Debug Module Control] states:
+	 * > 0 (inactive): ... Any accesses to the module may fail.
+	 *
+	 * Ignoring failures may introduce incompatibility with 0.13.
+	 * See https://github.com/riscv/riscv-debug-spec/issues/1021
+	 */
+	dm013_info_t *dm = get_dm(target);
+	assert(dm && "DM is expected to be already allocated.");
+	assert(!dm->was_reset && "Attempt to reset an already-reset debug module.");
+	/* `dmcontrol.hartsel` should be read first, in order not to
+	 * change it when requesting the reset, since changing it
+	 * without checking that `abstractcs.busy` is low is
+	 * prohibited.
+	 */
+	uint32_t dmcontrol;
+	int result = dm_read(target, &dmcontrol, DM_DMCONTROL);
+	if (result != ERROR_OK)
+		return result;
+
+	if (get_field32(dmcontrol, DM_DMCONTROL_DMACTIVE)) {
+		/* `dmcontrol.hartsel` is not changed. */
+		dmcontrol = (dmcontrol & DM_DMCONTROL_HARTSELLO) |
+			(dmcontrol & DM_DMCONTROL_HARTSELHI);
+		LOG_TARGET_DEBUG(target, "Initiating DM reset.");
+		result = dm_write(target, DM_DMCONTROL, dmcontrol);
+		if (result != ERROR_OK)
+			return result;
+
+		const time_t start = time(NULL);
+		LOG_TARGET_DEBUG(target, "Waiting for the DM to acknowledge reset.");
+		do {
+			result = dm_read(target, &dmcontrol, DM_DMCONTROL);
+			if (result != ERROR_OK)
+				return result;
+
+			if (time(NULL) - start > riscv_reset_timeout_sec) {
+				/* TODO: Introduce a separate timeout for this. */
+				LOG_TARGET_ERROR(target, "DM didn't acknowledge reset in %d s. "
+						"Increase the timeout with 'riscv set_reset_timeout_sec'.",
+						riscv_reset_timeout_sec);
+				return ERROR_TIMEOUT_REACHED;
+			}
+		} while (get_field32(dmcontrol, DM_DMCONTROL_DMACTIVE));
+		LOG_TARGET_DEBUG(target, "DM reset initiated.");
+	}
+
+	LOG_TARGET_DEBUG(target, "Activating the DM.");
+	result = dm_write(target, DM_DMCONTROL, DM_DMCONTROL_DMACTIVE);
+	if (result != ERROR_OK)
+		return result;
+
+	const time_t start = time(NULL);
+	LOG_TARGET_DEBUG(target, "Waiting for the DM to come out of reset.");
+	do {
+		result = dm_read(target, &dmcontrol, DM_DMCONTROL);
+		if (result != ERROR_OK)
+			return result;
+
+		if (time(NULL) - start > riscv_reset_timeout_sec) {
+			/* TODO: Introduce a separate timeout for this. */
+			LOG_TARGET_ERROR(target, "Debug Module did not become active in %d s. "
+					"Increase the timeout with 'riscv set_reset_timeout_sec'.",
+					riscv_reset_timeout_sec);
+			return ERROR_TIMEOUT_REACHED;
+		}
+	} while (!get_field32(dmcontrol, DM_DMCONTROL_DMACTIVE));
+
+	LOG_TARGET_DEBUG(target, "DM successfully reset.");
+	dm->was_reset = true;
+	return ERROR_OK;
+}
+
 static int examine_dm(struct target *target)
 {
 	dm013_info_t *dm = get_dm(target);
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -1952,6 +1952,81 @@ static int wait_for_idle_if_needed(struct target *target)
...
+		const time_t start = time(NULL);
+		LOG_TARGET_DEBUG(target, "Waiting for the DM to acknowledge reset.");
+		do {
+			result = dm_read(target, &dmcontrol, DM_DMCONTROL);
+			if (result != ERROR_OK)
+				return result;
+
+			if (time(NULL) - start > riscv_reset_timeout_sec) {
+				/* TODO: Introduce a separate timeout for this. */
+				LOG_TARGET_ERROR(target, "DM didn't acknowledge reset in %ds. "
```
JanMatCodasip: Pedantic detail:

```suggestion
				LOG_TARGET_ERROR(target, "DM didn't acknowledge reset in %d s. "
```

src/target/riscv/riscv-013.c
```diff
@@ -1952,6 +1952,81 @@ static int wait_for_idle_if_needed(struct target *target)
...
+	const time_t start = time(NULL);
+	LOG_TARGET_DEBUG(target, "Waiting for the DM to come out of reset.");
+	do {
+		result = dm_read(target, &dmcontrol, DM_DMCONTROL);
+		if (result != ERROR_OK)
+			return result;
+
+		if (time(NULL) - start > riscv_reset_timeout_sec) {
+			/* TODO: Introduce a separate timeout for this. */
+			LOG_TARGET_ERROR(target, "Debug Module did not become active in %ds. "
```
```suggestion
			LOG_TARGET_ERROR(target, "Debug Module did not become active in %d s. "
```

src/target/riscv/riscv-013.c
```diff
@@ -1952,6 +1952,81 @@ static int wait_for_idle_if_needed(struct target *target)
...
+
+	if (get_field32(dmcontrol, DM_DMCONTROL_DMACTIVE)) {
+		/* `dmcontrol.hartsel` is not changed. */
+		dmcontrol = (dmcontrol & DM_DMCONTROL_HARTSELLO) |
+			(dmcontrol & DM_DMCONTROL_HARTSELHI);
+		LOG_TARGET_DEBUG(target, "Initiating DM reset.");
+		result = dm_write(target, DM_DMCONTROL, dmcontrol);
+		if (result != ERROR_OK)
+			return result;
+
```
JanMatCodasip: As already noted in the related debug spec thread (https://github.com/riscv/riscv-debug-spec/issues/1021#issuecomment-2093183331), I tend to believe that the requirement to not change the hartsel bits was meant for standard operation of the DM, not for master-reset of the whole DM.

I'd recommend to try to not change the hartsel bits, but if that is not possible, just go ahead and reset the debug module.

The dmactive bit is a master-reset of the DM, that means it is a mechanism how to bring the DM to a known initial state. So it would be very strange to add constraints, saying that this DM reset cannot be used under certain circumstances, even though the strict interpretation of the spec is saying exactly that.

~~In pseudo-code, my proposal would be:~~ Edit May-13-2024: I have posted a corrected version below.

```
read dmcontrol --> dmcontrol_val;
if (not successful) {
    LOG_TARGET_WARNING(target, "Could not read dmcontrol register. Attempting debug module reset anyway.");
    dmcontrol_val = 0x0;
}

if (dmcontrol.dmactive) {
    dmcontrol_val &= ~dmactive;
    write dmcontrol <-- dmcontrol_val;
    wait for DM reset acknowledge;
}

write dmcontrol <-- 0x1;
wait for DM activation after reset;

LOG_TARGET_DEBUG(target, "DM successfully reset.");
```

JanMatCodasip: Another point that supports doing the reset always is that (to my knowledge) no user has so far objected against that behavior of riscv-openocd.

en-sc: Please, see my [comment in the related riscv-spec discussion](https://github.com/riscv/riscv-debug-spec/issues/1021#issuecomment-2093259615)

en-sc: > In pseudo-code, my proposal would be:
...

Your proposal is great. However, I would like to implement it as a separate commit.

This commit intends to address a different issue that I have introduced in 8319eee9e1ffc601b94b4223958180b49f8b8188:

I've merged these two writes:

https://github.com/riscv-collab/riscv-openocd/commit/8319eee9e1ffc601b94b4223958180b49f8b8188#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63L1978
https://github.com/riscv-collab/riscv-openocd/commit/8319eee9e1ffc601b94b4223958180b49f8b8188#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63L1985-L1987

Into one:

https://github.com/riscv-collab/riscv-openocd/commit/8319eee9e1ffc601b94b4223958180b49f8b8188#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63R1977-R1979

This is not allowed by the spec:
> 0 (inactive): The module’s state, including authentication mechanism,
takes its reset values (the dmactive bit is the only bit which can be written to something other than its reset value).

And addressing this issue is the intention behind this commit.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -1952,6 +1952,81 @@ static int wait_for_idle_if_needed(struct target *target)
...
+	/* TODO: This function returns an error when a DMI operation fails.
+	 * However, [3.14.2. Debug Module Control] states:
+	 * > 0 (inactive): ... Any accesses to the module may fail.
+	 *
+	 * Ignoring failures may introduce incompatibility with 0.13.
+	 * See https://github.com/riscv/riscv-debug-spec/issues/1021
+	 */
+	dm013_info_t *dm = get_dm(target);
+	assert(dm && "DM is expected to be already allocated.");
+	assert(!dm->was_reset && "DM is reset only once.");
```
JanMatCodasip: Detail - if you'd like:

```suggestion
	assert(!dm->was_reset && "Attempt to reset an already-reset debug module.");
```

src/target/riscv/riscv-013.c
JanMatCodasip: Hello @en-sc,

I am sorry, I have made a mistake in the pseudo-code above. Here is the corrected version:

```
read dmcontrol --> dmcontrol_val;
if (not successful) {
    LOG_TARGET_WARNING(target, "Could not read dmcontrol register. Attempting debug module reset anyway.");
    /* We could not read dmcontrol. For robustness, assume that the debug module is active
     * and perform the full DM reset. */
    dmcontrol_val = 0x1;
}

if (dmcontrol.dmactive) {
    dmcontrol_val &= ~dmactive;
    write dmcontrol <-- dmcontrol_val;
    wait for DM reset acknowledge;
}

write dmcontrol <-- 0x1;
wait for DM activation after reset;

LOG_TARGET_DEBUG(target, "DM successfully reset.");
```

JanMatCodasip: > en-sc: This is not allowed by the spec:
>> 0 (inactive): The module’s state, including authentication mechanism,
>> takes its reset values (the dmactive bit is the only bit which can be written to something other than its reset value).

The sentence "the dmactive bit is the only bit which can be written to something other than its reset value" appears ambiguous to me. In my opinion, it can mean two things:

- 1) The external debugger must not write any bit of dmcontrol (or other registers), apart from dmactive, to any other value than the reset value.
- 2) The DM must not allow any bit of dmcontrol (or other registers), apart from dmactive, to be written to a non-reset value by the external debugger.

I believe that 2) is what the spec intended, but it looks the wording could have been more clear. I have filed a question/ticket for it: https://github.com/riscv/riscv-debug-spec/issues/1029

JanMatCodasip: In other words, I do not believe that your current implementation is not necessarily incorrect, but I would also prefer to activate the debug module by writing 0x1 to dmactive (that is, no other dmcontrol bit is set at the time when the debug module is being activated).

JanMatCodasip: > I would like to implement it as a separate commit.

Sure, feel free to merge this one as is, and further refinement can take place in a separate PR. 

Thank you.

JanMatCodasip (dismissed):

src/target/riscv/riscv-013.c

JanMatCodasip (approved):

PR #1072, opened 2024-05-28
target/riscv: do not emit warnings when a non-existent CSR is hidden

aap-sc:
hide_csrs should not emit warnings on an attempt to hide non-exitents CSR. hide_csrs funcitonality is intended to be used for scenarios when we don`t want certain groups of registers to be available in GDB. Typically this is needed to simplify integration with various IDE. In such scenarious it may be impractical/unfeseable to figure out which register is present on a target. So reporting a situation when a user wants to hide a non-existent register creates way too much noise. This commit reduces severity of relevant debug message to LOG_TARGET_DEBUG


en-sc (approved):
LGTM

JanMatCodasip (approved):
Thank you.

PR #1073, opened 2024-05-29
target/riscv: write registers using batch

en-sc:
This allows to eliminate up to two DMI NOPs.

Change-Id: I09a18bd896fce2392d1b65d4efb38b53e334a358

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -29,11 +29,12 @@ struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans)
 	out->allocated_scans = scans;
 	out->last_scan = RISCV_SCAN_TYPE_INVALID;
 	out->was_run = false;
-	out->used_idle_count = 0;
+	out->used_delay = 0;
 
 	out->data_out = NULL;
 	out->data_in = NULL;
 	out->fields = NULL;
+	out->delay_classes = NULL;
 	out->bscan_ctxt = NULL;
 	out->read_keys = NULL;
 
@@ -53,6 +54,11 @@ struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans)
 	out->fields = malloc(sizeof(*out->fields) * scans);
 	if (!out->fields) {
 		LOG_ERROR("Failed to allocate fields in RISC-V batch.");
+		goto alloc_error;
+	}
+	out->delay_classes = malloc(sizeof(*out->delay_classes) * scans);
+	if (!out->delay_classes) {
+		LOG_ERROR("Failed to allocate delay_classes in RISC-V batch.");
 		goto alloc_error;
 	}
 	if (bscan_tunnel_ir_width != 0) {
@@ -80,6 +86,7 @@ void riscv_batch_free(struct riscv_batch *batch)
 	free(batch->data_in);
 	free(batch->data_out);
 	free(batch->fields);
+	free(batch->delay_classes);
 	free(batch->bscan_ctxt);
 	free(batch->read_keys);
 	free(batch);
@@ -101,28 +108,51 @@ static bool riscv_batch_was_scan_busy(const struct riscv_batch *batch,
 	return get_field(in, DTM_DMI_OP) == DTM_DMI_OP_BUSY;
 }
 
-static void add_idle_if_increased(struct riscv_batch *batch, size_t new_idle_count)
+static void add_idle_before_batch(const struct riscv_batch *batch, size_t start_idx,
+		struct riscv_scan_delays delays)
 {
 	if (!batch->was_run)
 		return;
-	if (batch->used_idle_count <= new_idle_count)
+	/* Get the delay type of the scan that resulted in the busy response.
+	 * Since DMI interactions always end with a NOP, if "start_idx" is zero
+	 * the base delay value is used.
+	 */
+	const enum riscv_scan_delay_class delay_class = start_idx > 0
+		? batch->delay_classes[start_idx - 1]
+		: RISCV_DELAY_BASE;
+	const unsigned int new_delay = riscv_scan_get_delay(delays, delay_class);
+	if (new_delay <= batch->used_delay)
 		return;
-	const size_t idle_change = new_idle_count - batch->used_idle_count;
-	LOG_TARGET_DEBUG(batch->target,
-			"Idle count increased. Adding %zu idle cycles before the batch.",
+	const unsigned int idle_change = new_delay - batch->used_delay;
+	LOG_TARGET_DEBUG(batch->target, "Adding %u idle cycles before the batch.",
 			idle_change);
+	assert(idle_change <= INT_MAX);
 	jtag_add_runtest(idle_change, TAP_IDLE);
 }
 
+static int get_delay(const struct riscv_batch *batch, size_t scan_idx,
+		struct riscv_scan_delays delays)
+{
+	assert(batch);
+	assert(scan_idx < batch->used_scans);
+	const enum riscv_scan_delay_class delay_class =
+		batch->delay_classes[scan_idx];
+	const unsigned int delay =  riscv_scan_get_delay(delays, delay_class);
+	assert(delay <= INT_MAX);
+	return delay;
+}
+
 int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
-		size_t idle_count, bool resets_delays, size_t reset_delays_after)
+		struct riscv_scan_delays delays, bool resets_delays,
+		size_t reset_delays_after)
 {
 	assert(batch->used_scans);
 	assert(batch->last_scan == RISCV_SCAN_TYPE_NOP);
 	assert(!batch->was_run || riscv_batch_was_scan_busy(batch, start_idx));
 	assert(start_idx == 0 || !riscv_batch_was_scan_busy(batch, start_idx - 1));
 
-	add_idle_if_increased(batch, idle_count);
+	if (batch->was_run)
+		add_idle_before_batch(batch, start_idx, delays);
 
 	LOG_TARGET_DEBUG(batch->target, "Running batch of scans [%zu, %zu)",
 			start_idx, batch->used_scans);
@@ -135,8 +165,10 @@ int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
 
 		const bool delays_were_reset = resets_delays
 			&& (i >= reset_delays_after);
-		if (idle_count > 0 && !delays_were_reset)
-			jtag_add_runtest(idle_count, TAP_IDLE);
+		const int delay = get_delay(batch, i, delays);
+
+		if (!delays_were_reset)
+			jtag_add_runtest(delay, TAP_IDLE);
 	}
 
 	keep_alive();
@@ -156,17 +188,19 @@ int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
 		}
 	}
 
-	for (size_t i = start_idx; i < batch->used_scans; ++i)
-		riscv_log_dmi_scan(batch->target, idle_count, batch->fields + i,
+	for (size_t i = start_idx; i < batch->used_scans; ++i) {
+		const int delay = get_delay(batch, i, delays);
+		riscv_log_dmi_scan(batch->target, delay, batch->fields + i,
 				/*discard_in*/ false);
+	}
 
 	batch->was_run = true;
-	batch->used_idle_count = idle_count;
+	batch->used_delay = get_delay(batch, batch->used_scans - 1, delays);
 	return ERROR_OK;
 }
 
 void riscv_batch_add_dm_write(struct riscv_batch *batch, uint64_t address, uint32_t data,
-	bool read_back)
+		bool read_back, enum riscv_scan_delay_class delay_class)
 {
 	assert(batch->used_scans < batch->allocated_scans);
 	struct scan_field *field = batch->fields + batch->used_scans;
@@ -179,11 +213,13 @@ void riscv_batch_add_dm_write(struct riscv_batch *batch, uint64_t address, uint3
 	} else {
 		field->in_value = NULL;
 	}
+	batch->delay_classes[batch->used_scans] = delay_class;
 	batch->last_scan = RISCV_SCAN_TYPE_WRITE;
 	batch->used_scans++;
 }
 
-size_t riscv_batch_add_dm_read(struct riscv_batch *batch, uint64_t address)
+size_t riscv_batch_add_dm_read(struct riscv_batch *batch, uint64_t address,
+		enum riscv_scan_delay_class delay_class)
 {
 	assert(batch->used_scans < batch->allocated_scans);
 	struct scan_field *field = batch->fields + batch->used_scans;
@@ -192,6 +228,7 @@ size_t riscv_batch_add_dm_read(struct riscv_batch *batch, uint64_t address)
 	field->in_value  = (void *)(batch->data_in  + batch->used_scans * DMI_SCAN_BUF_SIZE);
 	riscv_fill_dm_read(batch->target, (char *)field->out_value, address);
 	riscv_fill_dm_nop(batch->target, (char *)field->in_value);
+	batch->delay_classes[batch->used_scans] = delay_class;
 	batch->last_scan = RISCV_SCAN_TYPE_READ;
 	batch->used_scans++;
 
@@ -228,6 +265,9 @@ void riscv_batch_add_nop(struct riscv_batch *batch)
 	field->in_value  = (void *)(batch->data_in  + batch->used_scans * DMI_SCAN_BUF_SIZE);
 	riscv_fill_dm_nop(batch->target, (char *)field->out_value);
 	riscv_fill_dm_nop(batch->target, (char *)field->in_value);
+	/* DMI NOP never triggers any debug module operation,
+	 * so the shortest (base) delay can be used. */
+	batch->delay_classes[batch->used_scans] = RISCV_DELAY_BASE;
 	batch->last_scan = RISCV_SCAN_TYPE_NOP;
 	batch->used_scans++;
 }
--- a/src/target/riscv/batch.h
+++ b/src/target/riscv/batch.h
@@ -14,6 +14,67 @@ enum riscv_scan_type {
 	RISCV_SCAN_TYPE_WRITE,
 };
 
+/* These types are used to specify how many JTAG RTI cycles to add after a
+ * scan.
+ */
+enum riscv_scan_delay_class {
+	/* Delay needed for accessing debug module registers: */
+	RISCV_DELAY_BASE,
+	/* Delay for execution of an abstract command: */
+	RISCV_DELAY_ABSTRACT_COMMAND,
+	/* Delay for System Bus read operation: */
+	RISCV_DELAY_SYSBUS_READ,
+	/* Delay for System Bus write operation: */
+	RISCV_DELAY_SYSBUS_WRITE,
+};
+
+struct riscv_scan_delays {
+	/* The purpose of these delays is to be passed to "jtag_add_runtest()",
+	 * which accepts an "int".
+	 * Therefore, they should be no greater then "INT_MAX".
+	 */
+	unsigned int base_delay;
+	unsigned int ac_delay;
+	unsigned int sb_read_delay;
+	unsigned int sb_write_delay;
+};
+
+static inline unsigned int riscv_scan_get_delay(struct riscv_scan_delays delays,
+		enum riscv_scan_delay_class delay_class)
+{
+	switch (delay_class) {
+	case RISCV_DELAY_BASE:
+		return delays.base_delay;
+	case RISCV_DELAY_ABSTRACT_COMMAND:
+		return delays.ac_delay;
+	case RISCV_DELAY_SYSBUS_READ:
+		return delays.sb_read_delay;
+	case RISCV_DELAY_SYSBUS_WRITE:
+		return delays.sb_write_delay;
+	}
+	return 0;
+}
+
+static inline void riscv_scan_set_delay(struct riscv_scan_delays *delays,
+		enum riscv_scan_delay_class delay_class, unsigned int delay)
+{
+	assert(delay <= INT_MAX);
+	switch (delay_class) {
+	case RISCV_DELAY_BASE:
+		delays->base_delay = delay;
+		return;
+	case RISCV_DELAY_ABSTRACT_COMMAND:
+		delays->ac_delay = delay;
+		return;
+	case RISCV_DELAY_SYSBUS_READ:
+		delays->sb_read_delay = delay;
+		return;
+	case RISCV_DELAY_SYSBUS_WRITE:
+		delays->sb_write_delay = delay;
+		return;
+	}
+}
+
 /* A batch of multiple JTAG scans, which are grouped together to avoid the
  * overhead of some JTAG adapters when sending single commands.  This is
  * designed to support block copies, as that's what we actually need to go
@@ -27,6 +88,7 @@ struct riscv_batch {
 	uint8_t *data_out;
 	uint8_t *data_in;
 	struct scan_field *fields;
+	enum riscv_scan_delay_class *delay_classes;
 
 	/* If in BSCAN mode, this field will be allocated (one per scan),
 	   and utilized to tunnel all the scans in the batch.  If not in
@@ -48,8 +110,10 @@ struct riscv_batch {
 	 * However, RISC-V DMI "busy" condition could still have occurred.
 	 */
 	bool was_run;
-	/* Idle count used on the last run. Only valid after `was_run` is set. */
-	size_t used_idle_count;
+	/* Number of RTI cycles used by the last scan on the last run.
+	 * Only valid when `was_run` is set.
+	 */
+	unsigned int used_delay;
 };
 
 /* Allocates (or frees) a new scan set.  "scans" is the maximum number of JTAG
@@ -65,8 +129,8 @@ bool riscv_batch_full(struct riscv_batch *batch);
  * If batch is run for the first time, it is expected that "start" is zero.
  * It is expected that the batch ends with a DMI NOP operation.
  *
- * "idle_count" is the number of JTAG Run-Test-Idle cycles to add in-between
- * the scans.
+ * "idle_counts" specifies the number of JTAG Run-Test-Idle cycles to add
+ * after each scan depending on the delay class of the scan.
  *
  * If "resets_delays" is true, the algorithm will stop inserting idle cycles
  * (JTAG Run-Test-Idle) after "reset_delays_after" number of scans is
@@ -74,19 +138,21 @@ bool riscv_batch_full(struct riscv_batch *batch);
  * OpenOCD that are based on batches.
  */
 int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
-		size_t idle_count, bool resets_delays, size_t reset_delays_after);
+		struct riscv_scan_delays delays, bool resets_delays,
+		size_t reset_delays_after);
 
 /* Get the number of scans successfully executed form this batch. */
 size_t riscv_batch_finished_scans(const struct riscv_batch *batch);
 
 /* Adds a DM register write to this batch. */
 void riscv_batch_add_dm_write(struct riscv_batch *batch, uint64_t address, uint32_t data,
-	bool read_back);
+	bool read_back, enum riscv_scan_delay_class delay_class);
 
 /* DM register reads must be handled in two parts: the first one schedules a read and
  * provides a key, the second one actually obtains the result of the read -
  * status (op) and the actual data. */
-size_t riscv_batch_add_dm_read(struct riscv_batch *batch, uint64_t address);
+size_t riscv_batch_add_dm_read(struct riscv_batch *batch, uint64_t address,
+		enum riscv_scan_delay_class delay_class);
 unsigned int riscv_batch_get_dmi_read_op(const struct riscv_batch *batch, size_t key);
 uint32_t riscv_batch_get_dmi_read_data(const struct riscv_batch *batch, size_t key);
 
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -719,12 +719,6 @@ static int dmi_write(struct target *target, uint32_t address, uint32_t value)
 	return dmi_op(target, NULL, NULL, DMI_OP_WRITE, address, value, false, true);
 }
 
-static int dmi_write_exec(struct target *target, uint32_t address,
-		uint32_t value, bool ensure_success)
-{
-	return dmi_op(target, NULL, NULL, DMI_OP_WRITE, address, value, true, ensure_success);
-}
-
 static uint32_t riscv013_get_dmi_address(const struct target *target, uint32_t address)
 {
 	assert(target);
@@ -780,16 +774,6 @@ static int dm_write(struct target *target, uint32_t address, uint32_t value)
 	if (!dm)
 		return ERROR_FAIL;
 	return dmi_write(target, address + dm->base, value);
-}
-
-static int dm_write_exec(struct target *target, uint32_t address,
-		uint32_t value, bool ensure_success)
-{
-	dm013_info_t *dm = get_dm(target);
-	if (!dm)
-		return ERROR_FAIL;
-	dm->abstract_cmd_maybe_busy = true;
-	return dmi_write_exec(target, address + dm->base, value, ensure_success);
 }
 
 static bool check_dbgbase_exists(struct target *target)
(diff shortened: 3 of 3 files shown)


src/target/riscv/batch.c
```diff
@@ -34,6 +34,7 @@ struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans)
 	out->data_out = NULL;
 	out->data_in = NULL;
 	out->fields = NULL;
+	out->delay_types = NULL;
```
JanMatCodasip: Detail - optional:

I believe `delay class` is somewhat more descriptive than `delay type`.

```suggestion
	out->delay_classes = NULL;
```

src/target/riscv/batch.h
```diff
@@ -14,6 +14,17 @@ enum riscv_scan_type {
 	RISCV_SCAN_TYPE_WRITE,
 };
 
+/* These types are used to specify how many JTAG RTI cycles to add after a
+ * scan.
+ */
+enum riscv_scan_delay_type {
+	RISCV_SCAN_DELAY_DM,
```
JanMatCodasip: Optional - if you like:

```suggestion
	RISCV_SCAN_DELAY_BASE,
```

src/target/riscv/batch.h
```diff
@@ -14,6 +14,17 @@ enum riscv_scan_type {
 	RISCV_SCAN_TYPE_WRITE,
 };
 
+/* These types are used to specify how many JTAG RTI cycles to add after a
+ * scan.
+ */
+enum riscv_scan_delay_type {
```
JanMatCodasip: Optional - if you like:

```suggestion
enum riscv_scan_delay_class {
```

src/target/riscv/batch.h
```diff
@@ -14,6 +14,17 @@ enum riscv_scan_type {
...
+	RISCV_SCAN_DELAY_DM,
+	RISCV_SCAN_DELAY_ABSTRACT_COMMAND,
+	RISCV_SCAN_DELAY_SYSBUS_READ,
+	RISCV_SCAN_DELAY_SYSBUS_WRITE,
+	RISCV_SCAN_N_DELAY_TYPES,
```
JanMatCodasip: If you like:

```suggestion
	/* Learned delay needed for accessing debug module registers: */
	RISCV_SCAN_DELAY_BASE,
	/* Learned delay for execution of an abstract command: */
	RISCV_SCAN_DELAY_ABSTRACT_COMMAND,
	/* Learned delay for System Bus read operation: */
	RISCV_SCAN_DELAY_SYSBUS_READ,
	/* Learned delay for System Bus write operation: */
	RISCV_SCAN_DELAY_SYSBUS_WRITE,
	/* Number of enum items: */
	RISCV_SCAN_N_DELAY_TYPES,
```

Furthermore, the constants could be shortened this way: `RISCV_SCAN_DELAY_*` --> `RISCV_DELAY_*`, if you like.

```suggestion
	/* Learned delay needed for accessing debug module registers: */
	RISCV_DELAY_BASE,
	/* Learned delay for execution of an abstract command: */
	RISCV_DELAY_ABSTRACT_COMMAND,
	/* Learned delay for System Bus read operation: */
	RISCV_DELAY_SYSBUS_READ,
	/* Learned delay for System Bus write operation: */
	RISCV_DELAY_SYSBUS_WRITE,
	/* Number of enum items: */
	RISCV_N_DELAY_TYPES,
```

src/target/riscv/riscv-013.c
```diff
@@ -2692,12 +2743,27 @@ static int sb_write_address(struct target *target, target_addr_t address,
...
+{
+	RISCV013_INFO(info);
+	idle_counts[RISCV_SCAN_DELAY_DM] = info->dmi_busy_delay;
+	idle_counts[RISCV_SCAN_DELAY_ABSTRACT_COMMAND] = info->dmi_busy_delay +
+		info->ac_busy_delay;
+	idle_counts[RISCV_SCAN_DELAY_SYSBUS_READ] = info->dmi_busy_delay +
+		info->bus_master_read_delay;
+	idle_counts[RISCV_SCAN_DELAY_SYSBUS_WRITE] = info->dmi_busy_delay +
+		info->bus_master_write_delay;
+}
```
JanMatCodasip: My recommendation would be to define a structure to hold all the delay valued ans then both store and pass it as the structure type (instead of an array).

This way:
- The representation of the delays will be the same everywhere.
- Work with the structure will be somewhat safer (less chance of overflow).
- The overall code would be easier to read - all values encapsulated into the struct and passed as a whole.
- It will be easier to inspect the delay values in a debugger (to see which item means what delay).

Please let me know if you like this proposal.

```c
typedef struct {
    size_t dmi_busy_delay;
    size_t ac_busy_delay;
    /* ... */
} riscv_learned_delays_t;  // or similar name

typedef struct {
    /* ... */
    
    riscv_learned_delays_t delays;  // <--- Use the struct for storage of the delays.

} riscv013_info_t;

/* ... */

riscv_batch_run_from(batch, finished_scans,
    /* riscv_learned_delays_t */ delays, reset_delays, reset_delays_wait);

size_t riscv013_get_delay(const riscv_learned_delays_t *delays, riscv_delay_class_t delay_class)
{
    switch (delay_class) {
        case RISCV_SCAN_DELAY_DM:
            return delays->dmi_busy_delay;
        case RISCV_SCAN_DELAY_ABSTRACT_COMMAND:
            return delays->dmi_busy_delay + info->ac_busy_delay;
        /* ... */
    }
}

```

src/target/riscv/batch.h
```diff
@@ -48,8 +60,10 @@ struct riscv_batch {
 	 * However, RISC-V DMI "busy" condition could still have occurred.
 	 */
 	bool was_run;
-	/* Idle count used on the last run. Only valid after `was_run` is set. */
-	size_t used_idle_count;
+	/* Idle count used by the last scan on the last run. Only valid after
+	 * `was_run` is set.
```
JanMatCodasip: Detail:

```suggestion
	/* Idle count used by the last scan on the last run. Only valid when
	 * `was_run` is set.
```

src/target/riscv/batch.h
```diff
@@ -65,28 +79,30 @@ bool riscv_batch_full(struct riscv_batch *batch);
  * If batch is run for the first time, it is expected that "start" is zero.
  * It is expected that the batch ends with a DMI NOP operation.
  *
- * "idle_count" is the number of JTAG Run-Test-Idle cycles to add in-between
- * the scans.
+ * "idle_counts" specifies the number of JTAG Run-Test-Idle cycles to add
+ * in-between the scans for each "riscv_batch_delay_type".
```
JanMatCodasip: If you like:

```suggestion
 * "idle_counts" specifies the number of JTAG Run-Test-Idle cycles to add
 * after each scan depending on the delay class of the scan.
```

JanMatCodasip (commented):
@en-sc Thank you for the changes. I have managed to review approx. half of the code and am sending my comments so far. I will return to this tomorrow.

src/target/riscv/batch.h

src/target/riscv/batch.h

src/target/riscv/batch.h
en-sc: Addressed. Though I have dropped "Learned" from the comments, since IMHO it's too specific. It is the delays to use. They are not necessarily learned -- they could be obtained through other means in the future.

src/target/riscv/batch.h

src/target/riscv/batch.h

src/target/riscv/batch.c

en-sc:
@JanMatCodasip, thank you for the review!

Please note, apart from addressing your comments I've changed the approach in `register_write_abstract()`.
Now both `register_write_abstract()` and `execute_abstract_command()` re-use a new function: `execute_abstract_command_in_batch()`.

src/target/riscv/batch.h
```diff
@@ -48,8 +65,10 @@ struct riscv_batch {
 	 * However, RISC-V DMI "busy" condition could still have occurred.
 	 */
 	bool was_run;
-	/* Idle count used on the last run. Only valid after `was_run` is set. */
-	size_t used_idle_count;
+	/* Idle count used by the last scan on the last run. Only valid when
+	 * `was_run` is set.
+	 */
+	int used_idle_count;
```
JanMatCodasip: Please, why is a signed type used here - `int`? Is that because jtag_add_runtest() accepts `int` as its argument? 

I recommend to stick with an unsigned type -- `unsigned int`, `size_t` or similar -- for all variables holding delay counts. 

(Note: There are multiple occurrences throughout this pull request. I haven't commented on each of them.)

src/target/riscv/batch.c
```diff
@@ -101,28 +108,46 @@ static bool riscv_batch_was_scan_busy(const struct riscv_batch *batch,
...
 	if (!batch->was_run)
 		return;
-	if (batch->used_idle_count <= new_idle_count)
-		return;
-	const size_t idle_change = new_idle_count - batch->used_idle_count;
-	LOG_TARGET_DEBUG(batch->target,
-			"Idle count increased. Adding %zu idle cycles before the batch.",
+	/* Get the delay type of the scan that resulted in the busy response.
+	 * Since DMI interactions always end with a NOP, if "start_idx" is zero
+	 * "RISCV_SCAN_DELAY_DM" is used
```
JanMatCodasip:
```suggestion
	 * "RISCV_SCAN_DELAY_BASE" is used
```

or 

```suggestion
	 * the base delay value is used.
```

src/target/riscv/batch.c
```diff
@@ -135,7 +160,10 @@ int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
 
 		const bool delays_were_reset = resets_delays
 			&& (i >= reset_delays_after);
-		if (idle_count > 0 && !delays_were_reset)
+		const int idle_count = get_idle_count(batch, i, idle_counts);
+
+		assert(idle_count >= 0);
+		if (!delays_were_reset)
 			jtag_add_runtest(idle_count, TAP_IDLE);
```
JanMatCodasip: As noted in a previous comment, please prefer to use an unsigned type for the idle counts. 

The fact that jtag_add_runtest() uses a singed parameter looks like a mistake/oversight to me (it is 15 years old code).

```suggestion
		const unsigned int idle_count = get_idle_count(batch, i, idle_counts);

		if (!delays_were_reset)
			jtag_add_runtest((int)idle_count, TAP_IDLE);
```

src/target/riscv/riscv-013.c
```diff
@@ -4668,22 +4736,25 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
+						RISCV_DELAY_SYSBUS_WRITE);
 			}
 			if (size > 4) {
 				sbvalue[1] = ((uint32_t)p[4]) |
 						(((uint32_t)p[5]) << 8) |
 						(((uint32_t)p[6]) << 16) |
 						(((uint32_t)p[7]) << 24);
-				riscv_batch_add_dm_write(batch, DM_SBDATA1, sbvalue[1], false);
+				riscv_batch_add_dm_write(batch, DM_SBDATA1, sbvalue[1], false,
+						RISCV_DELAY_SYSBUS_WRITE);
```
JanMatCodasip: The writes to `SBDATA3`, `2` and `1` don't start any system bus operation, and so the delay should be changed to `RISCV_DELAY_BASE`. 

Only the last write to `SBDATA0` register triggers the bus write operation (and so it should keep `RISCV_DELAY_SYSBUS_WRITE)`.

src/target/riscv/riscv-013.c
```diff
@@ -2865,10 +2928,10 @@ static int sample_memory_bus_v1(struct target *target,
 			break;
 		}
 
-		size_t sbcs_read_index = riscv_batch_add_dm_read(batch, DM_SBCS);
+		size_t sbcs_read_index = riscv_batch_add_dm_read(batch, DM_SBCS,
+				RISCV_DELAY_SYSBUS_READ);
```
JanMatCodasip: The read of SBCS register does not start any bus operation, so I believe the delay should be changed to RISCV_DELAY_BASE.

```suggestion
		size_t sbcs_read_index = riscv_batch_add_dm_read(batch, DM_SBCS,
				RISCV_DELAY_BASE);
```

src/target/riscv/riscv-013.c
```diff
@@ -2834,7 +2890,8 @@ static int sample_memory_bus_v1(struct target *target,
 						sbcs_write |= DM_SBCS_SBREADONDATA;
 					sbcs_write |= sb_sbaccess(config->bucket[i].size_bytes);
 					if (!sbcs_valid || sbcs_write != sbcs) {
-						riscv_batch_add_dm_write(batch, DM_SBCS, sbcs_write, true);
+						riscv_batch_add_dm_write(batch, DM_SBCS, sbcs_write,
+								true, RISCV_DELAY_SYSBUS_READ);
```
JanMatCodasip: Writing to SBCS does not start any bus operation, so RISCV_DELAY_BASE can be used:

```suggestion
						riscv_batch_add_dm_write(batch, DM_SBCS, sbcs_write,
								true, RISCV_DELAY_BASE);
```

src/target/riscv/riscv-013.c
```diff
@@ -2843,18 +2900,24 @@ static int sample_memory_bus_v1(struct target *target,
 							(!sbaddress1_valid ||
 							sbaddress1 != config->bucket[i].address >> 32)) {
 						sbaddress1 = config->bucket[i].address >> 32;
-						riscv_batch_add_dm_write(batch, DM_SBADDRESS1, sbaddress1, true);
+						riscv_batch_add_dm_write(batch, DM_SBADDRESS1,
+								sbaddress1, true,
+								RISCV_DELAY_SYSBUS_READ);
```
JanMatCodasip: Similarly to previous comment - writing to SBADDRESS1 does not start the bus operation yet.

```suggestion
						riscv_batch_add_dm_write(batch, DM_SBADDRESS1,
								sbaddress1, true,
								RISCV_DELAY_BASE);
```

src/target/riscv/riscv-013.c
```diff
@@ -2692,12 +2731,27 @@ static int sb_write_address(struct target *target, target_addr_t address,
 		(uint32_t)address, false, ensure_success);
 }
 
-static int batch_run(struct target *target, struct riscv_batch *batch,
-		size_t idle_count)
+/* TODO: replace "info->*_delay" with such array and remove this function. */
```
JanMatCodasip: @en-sc  

:one: 
Would you like to handle this TODO in a separate merge request?


:two: 
 `... "info->*_delay" with such array ...`

I would prefer to have a structure with descriptively named scalars. But even an array within the structure would still be better than passing the array directly.

src/target/riscv/riscv-013.c
```diff
@@ -1145,15 +1171,28 @@ static int register_write_abstract(struct target *target, enum gdb_regno number,
...
 			AC_ACCESS_REGISTER_WRITE);
+	LOG_DEBUG_REG(target, AC_ACCESS_REGISTER, command);
+	assert(size_bits % 32 == 0);
+	const unsigned int size_in_words = size_bits / 32;
+	/** The batch consists of:
+	 * * A write of "data*" ("size_in_words" scans).
+	 * * A write of "command" (1 scan).
+	 * * A read of "abstractcs" (1 scan).
+	 * * An implicit "nop".
+	 */
```
JanMatCodasip: Nitpick: 
- The comment does not need to start with the doxygen mark `/**`.
- Using `-` as bullet points makes it visually more clear than the stars that are already used at the beginning of each comment line.

```suggestion
	/* The batch consists of:
	 * - A write of "data*" ("size_in_words" scans).
	 * - A write of "command" (1 scan).
	 * - A read of "abstractcs" (1 scan).
	 * - An implicit "nop".
	 */
```

src/target/riscv/riscv-013.c
```diff
@@ -1023,6 +1033,23 @@ static int read_abstract_arg(struct target *target, riscv_reg_t *value,
...
+	const unsigned int offset = index * size_in_words;
+
+	for (unsigned int i = 0; i < size_in_words; ++i) {
+		const unsigned int reg_address = DM_DATA0 + offset + i;
+
+		riscv_batch_add_dm_write(batch, reg_address, value,
+				/* read_back */ true, RISCV_DELAY_BASE);
+		value >>= 32;
+	}
+}
```
JanMatCodasip: This function assumes that the abstract command is not yet executed - that abstractauto.autoexecdata[n] bits are zero.

Please consider mentioning it in a comment, for example like this:

```
/**
 * Queue scans into a batch that write the value to abstract data registers:
 * data[index] (and data[index+1] in case of 64-bit value).
 *
 * No extra DTM delay is added after the write to data[N]. It is assumed that
 * this is a one-shot abstract command, that means no auto-execution is set up
 * (abstractauto.autoexecdata bits are zero).
 */
```

src/target/riscv/riscv-013.c
```diff
@@ -973,10 +962,33 @@ static int execute_abstract_command(struct target *target, uint32_t command,
...
+	struct riscv_batch *batch = riscv_batch_alloc(target,
+			2 /*write command, read abstractcs*/);
+	const int res = execute_abstract_command_in_batch(target, batch, command,
+			cmderr);
+	riscv_batch_free(batch);
+	return res;
+}
+
 static void abstract_data_read_fill_batch(struct riscv_batch *batch, unsigned int index,
 		unsigned int size_bits)
```
JanMatCodasip, line 1005: This function makes the same assumption about abstractauto as mentioned in abstract_data_write_fill_batch().

src/target/riscv/batch.c
```diff
@@ -228,6 +261,7 @@ void riscv_batch_add_nop(struct riscv_batch *batch)
 	field->in_value  = (void *)(batch->data_in  + batch->used_scans * DMI_SCAN_BUF_SIZE);
 	riscv_fill_dm_nop(batch->target, (char *)field->out_value);
 	riscv_fill_dm_nop(batch->target, (char *)field->in_value);
+	batch->delay_classes[batch->used_scans] = RISCV_DELAY_BASE;
```
JanMatCodasip, line 270: If you like:

```suggestion
	/* DMI NOP never triggers any debug module operation, so the shortest (base) delay can be used. */
	batch->delay_classes[batch->used_scans] = RISCV_DELAY_BASE;
```

src/target/riscv/riscv-013.c
```diff
@@ -934,34 +918,39 @@ static int dm013_select_target(struct target *target)
...
-				LOG_DEBUG_REG(target, AC_ACCESS_REGISTER, command);
-				break;
-			default:
-				LOG_TARGET_DEBUG(target, "command=0x%x", command);
-				break;
-		}
-	}
+	riscv_batch_add_dm_write(batch, DM_COMMAND, command, /* read_back */ true,
+			RISCV_DELAY_BASE);
+	riscv_batch_add_dm_read(batch, DM_ABSTRACTCS, RISCV_DELAY_ABSTRACT_COMMAND);
```
JanMatCodasip: The write to DM_COMMAND triggers the abstract command execution, so it looks to me that the delays should be the other way around:

```suggestion
	riscv_batch_add_dm_write(batch, DM_COMMAND, command, /* read_back */ true,
			RISCV_DELAY_ABSTRACT_COMMAND);
	riscv_batch_add_dm_read(batch, DM_ABSTRACTCS, RISCV_DELAY_BASE);
```

src/target/riscv/batch.h
```diff
@@ -14,6 +14,22 @@ enum riscv_scan_type {
...
 };
 
+/* These types are used to specify how many JTAG RTI cycles to add after a
+ * scan.
+ */
+enum riscv_scan_delay_class {
+	/* Delay needed for accessing debug module registers: */
+	RISCV_DELAY_BASE,
+	/* Delay for execution of an abstract command: */
+	RISCV_DELAY_ABSTRACT_COMMAND,
```
JanMatCodasip, line 24: It is likely that different abstract commands will have different delays in real world-hardware, and so eventually multiple classes should be defined instead of the single `RISCV_DELAY_ABSTRACT_COMMAND`:

- Abstract register read/write
- Register access via program buffer
- Memory access via program buffer
- Abstract memory access

This is just a note and can be handled later - when the time comes.

src/target/riscv/riscv-013.c
```diff
@@ -1023,6 +1033,23 @@ static int read_abstract_arg(struct target *target, riscv_reg_t *value,
...
+		riscv_reg_t value, unsigned int index, unsigned int size_bits)
+{
+	assert(size_bits % 32 == 0);
+	const unsigned int size_in_words = size_bits / 32;
+	const unsigned int offset = index * size_in_words;
+
+	for (unsigned int i = 0; i < size_in_words; ++i) {
+		const unsigned int reg_address = DM_DATA0 + offset + i;
+
+		riscv_batch_add_dm_write(batch, reg_address, value,
```
JanMatCodasip: Here I'd recommend to explicitly truncate the `value` - for code readers' benefit:

```suggestion
		riscv_batch_add_dm_write(batch, reg_address, (uint32_t)value,
```

src/target/riscv/riscv-013.c
```diff
@@ -1145,15 +1171,28 @@ static int register_write_abstract(struct target *target, enum gdb_regno number,
...
 
-	if (write_abstract_arg(target, 0, value, size) != ERROR_OK)
-		return ERROR_FAIL;
-
+	abstract_data_write_fill_batch(batch, value, /*index*/ 0, size_bits);
 	uint32_t cmderr;
-	int result = execute_abstract_command(target, command, &cmderr);
+	const int result = execute_abstract_command_in_batch(target, batch, command,
+			&cmderr);
+	riscv_batch_free(batch);
```
JanMatCodasip: I am afraid the division of the responsibilities between the functions, as it is proposed now, makes the code harder to understand:

- The caller must allocate the batch - it must know the proper size, but it is not the caller who fills the scan items. 
- Abstract_data_write_fill_batch() adds the scans to write dataN registers. This is ok - one function does one thing.
- Execute_abstract_command_in_batch() does too many things - it adds further scans, then executes the batch, and finally it checks the abstract command result.


For these reasons, please consider different approach: Create multiple functions that will handle the execution of the whole abstract command batches. It would be:

- `execute_abstract_cmd_as_batch_w0(...)` - allocate the batch of needed size, fill the batch with write to `arg0`, write to command, then execute the batch, then check the result and finally free the batch
- `execute_abstract_cmd_as_batch_r0(...)` - same as above, but instead of writing to `arg0`, read `arg0` after the command is executed
- `execute_abstract_cmd_as_batch_w1r0(...)` - same, but write `arg1` before the command and read `arg0` after (needed for abstract memory read)
- `execute_abstract_cmd_as_batch_w1w0(...)` - same, but write `arg1` and `arg0` before the command (needed for abstract memory write)
- `execute_abstract_cmd_as_batch(...)` - only execute the abstract command and don't care about reading or writing the abstract arguments (dataN registers)

(The naming convention can of course be different - I just tried to present the idea.)

This way the functions would handle all the processing, including the management of the batches. And the code of the caller would be cleaner, shorter and easier to read - more focused on the actual algorithm, not at details of abstract commands or batches.

src/target/riscv/riscv-013.c
```diff
@@ -934,34 +918,39 @@ static int dm013_select_target(struct target *target)
...
-		else
-			LOG_TARGET_DEBUG(target, "command 0x%" PRIx32 " failed (unknown fatal error %d)", command, wait_result);
-		return wait_result;
+	int result = batch_run_timeout(target, batch);
+	if (result != ERROR_OK) {
+		riscv_batch_free(batch);
+		return result;
+	}
+	/* The only read in the batch is of "abstractcs". */
+	uint32_t abstractcs = riscv_batch_get_dmi_read_data(batch, 0);
```
JanMatCodasip: This makes strong assumptions about what scans the caller has put into the batch before calling this function.

IMO, this is yet another reason why the split of the responsibility between the functions should be different (please see my other comment).

JanMatCodasip (changes requested):
I have finished my first round of review.

 @en-sc Thank you for taking the time to improve batches and the abstract command processing.

Since this is touching some of the fundamental parts of the RISC-V processing and is not a simple change, I have multiple comments and suggestions.

src/target/riscv/riscv-013.c
```diff
@@ -1023,6 +1033,23 @@ static int read_abstract_arg(struct target *target, riscv_reg_t *value,
 	return result;
 }
 
+static void abstract_data_write_fill_batch(struct riscv_batch *batch,
+		riscv_reg_t value, unsigned int index, unsigned int size_bits)
+{
+	assert(size_bits % 32 == 0);
+	const unsigned int size_in_words = size_bits / 32;
+	const unsigned int offset = index * size_in_words;
```
JanMatCodasip, line 1062: If you like:

```suggestion
	const unsigned int offset = index * size_in_words;
	assert(value <= 0xffffffffu || size_in_words > 1);	
```

There are probably few other functions (not necessarily touched in this merge request) where the same assertion could be added.

JanMatCodasip (changes requested):

src/target/riscv/riscv-013.c
en-sc: Thanks for the suggestion. Addressed.

src/target/riscv/batch.h

src/target/riscv/batch.c

src/target/riscv/batch.c
en-sc: Addressed. Although here it's still an `int` to protect against integer overflow.

src/target/riscv/riscv-013.c
en-sc: 1. Yes. It's ready, I'll post it in a couple of days.
2. Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Addressed.

src/target/riscv/batch.c
en-sc: Addressed.

src/target/riscv/riscv-013.c
en-sc: Thanks for the catch! Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Addressed (only in this function).

I don't think similar assertion is necessary in `register_write_abstract()`, since it just passes `value` to `abstract_data_write_fill_batch()`.

As for other cases, I will keep in mind such assertions going forward.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: You are right, the function was poorly designed.
Please, take a look at current approach.

I didn't go forward with your suggestion since there is an issue.

Consider an example: `execute_abstract_cmd_as_batch_w1[r/w]0()` used to access memory via abstract command in case address auto increment is not supported.
Here it is actually neaded to merge as much consequent abstract command executions into a single batch, since it's a memory read/write -- a long operation.
With the approach you suggest it is not possible.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -4881,8 +4985,10 @@ static target_addr_t write_memory_progbuf_fill_batch(struct riscv_batch *batch,
 		log_memory_access64(address, value, size, /*is_read*/ false);
 		if (writes_per_element == 2)
 			riscv_batch_add_dm_write(batch, DM_DATA1,
-					(uint32_t)(value >> 32), false);
-		riscv_batch_add_dm_write(batch, DM_DATA0, (uint32_t)value, false);
+					(uint32_t)(value >> 32), false,
+					RISCV_DELAY_ABSTRACT_COMMAND);
```
JanMatCodasip: The write to DATA1 does not trigger the abstract command, so I believe `RISCV_DELAY_BASE` should be used here:

```suggestion
					(uint32_t)(value >> 32), false,
					RISCV_DELAY_BASE);
```

src/target/riscv/riscv-013.c
```diff
@@ -928,6 +912,49 @@ static int dm013_select_target(struct target *target)
 	return dm013_select_hart(target, info->index);
 }
 
+#define EXECUTE_ABSTRACT_COMMAND_BATCH_SIZE 2
+
+static size_t execute_abstract_command_fill_batch(struct riscv_batch *batch,
```
JanMatCodasip: I would recommend to change the prefix of the functions: `execute_abstract_command_*` ---> `abstract_command_batch_*` or `abstract_cmd_batch_*`.

The reason is that starting the functions names with `execute_` may lead the reader to think that the functions actually start the operation - which is not the case.

Proposal:

- `execute_abstract_command_fill_batch` --> `abstract_cmd_batch_fill` or `abstract_cmd_fill_batch`
- `execute_abstract_command_check_and_clear_cmderr` -> `abstract_cmd_batch_check_and_clear_cmderr` or `abstract_cmd_batch_check_result` or `abstract_cmd_check_batch_result`
- `execute_abstract_command` --> `execute_abstract_cmd` (if you like)

src/target/riscv/riscv-013.c
```diff
@@ -1017,6 +1046,32 @@ static int read_abstract_arg(struct target *target, riscv_reg_t *value,
...
 
+/**
+ * Queue scans into a batch that write the value to abstract data registers:
+ * data[index] (and data[index+1] in case of 64-bit value).
+ *
+ * No extra DTM delay is added after the write to data[N]. It is assumed that
+ * this is a one-shot abstract command, that means no auto-execution is set up
+ * (abstractauto.autoexecdata bits are zero).
+ */
+static void abstract_data_write_fill_batch(struct riscv_batch *batch,
```
JanMatCodasip, line 1056: To align it with the proposed names above, this one could be named:

```suggestion
static void abstract_cmd_batch_fill_data_write(struct riscv_batch *batch,
```

src/target/riscv/batch.h
```diff
@@ -14,6 +14,67 @@ enum riscv_scan_type {
...
+	case RISCV_DELAY_SYSBUS_WRITE:
+		return delays.sb_write_delay;
+	}
+	return 0;
+}
+
+static inline void riscv_scan_set_delay(struct riscv_scan_delays *delays,
+		enum riscv_scan_delay_class delay_class, unsigned int delay)
+{
+	assert(delay <= INT_MAX);
```
JanMatCodasip, line 61: The exponential back-off for the delays `next_delay = curr_delay + 1 + round_down(curr_delay * 0.1)` takes currently approx 200 rounds to exceed INT_MAX on 32 bits. 

So perhaps it is not impossible to reach.

For that reason, I would recommend to replace the assertion with saturation. The max value should be chosen so that the int don't overflow on the next increment. For example:

```suggestion
	/* Safety: Limit the delay to a value that is guaranteed to not overflow the int datatype on the next increment. */
	if (delay >= RISCV_MAX_DELAY) {  
		delay = RISCV_MAX_DELAY;
		// RISCV_MAX_DELAY can be e.g. 1000000000,
		// which is approx half of INT_MAX on 32 bits
	}	
```

src/target/riscv/riscv-013.c
JanMatCodasip: Good point about the abstract memory access and the issue with my proposal - I agree.

The current approach looks good to me. Thank you.

I have only proposed different names for some of the new functions. See one of my other comments.

JanMatCodasip (commented):
@en-sc Thank you for addressing the suggestions so far.

I have posted few last comments and don't expect to have more than that.

src/target/riscv/riscv-013.c
en-sc: I'd like to reuse it in `write_abstract_arg`. I think it will be quite confusing if it were named `abstract_cmd_batch_fill_data_write`

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Addressed. Left `execute_abstract_command` as is, since there are quite a few uses.

src/target/riscv/batch.h
en-sc: This is a valid concern. I've tried to handle it in #1087. Please, take a look

src/target/riscv/riscv-013.c
JanMatCodasip: Thank you.

src/target/riscv/riscv-013.c
JanMatCodasip: Ok, let's leave the name as is, then.

src/target/riscv/batch.h
JanMatCodasip: Thank you. I have posted my comments under #1087. (And this thread can be closed.)

JanMatCodasip (approved):
Thank you, LGTM.

PR #1081, opened 2024-06-04
 target/riscv: use batch interface in `read_memory_bus_v1()`

### LINKED ISSUE #1080: `read_memory_bus_v1()` is broken
Reproduction:
* Run Spike:
```
> spike --rbb-port=7777 --dm-progsize=4 --dm-sba=32 -p1 ...
```
* OpenOCD config:
```
> cat spike.cfg
adapter driver remote_bitbang
remote_bitbang port 7777
remote_bitbang host 127.0.0.1

set _CHIPNAME riscv
jtag newtap $_CHIPNAME cpu -irlen 5

target create $_TARGETNAME riscv -chain-position $_TARGETNAME
```
* Read via Program Buffer (reference):
```
> ACC=progbuf; openocd -f spike.cfg -c init -c 'reset halt' -c "riscv set_mem_access $ACC" -c "mdw [lindex [reg pc] 2] 2" -c shutdown
...
0x00001000: 00000297 02028593
```
* Read via System Bus (**the first word is zeroes**):
```
> ACC=sysbus; openocd -f spike.cfg -c init -c 'reset halt' -c "riscv set_mem_access $ACC" -c "mdw [lindex [reg pc] 2] 2" -c shutdown
...
0x00001000: 00000000 02028593
```

en-sc:
Fixes #1080

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3570,6 +3570,9 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
 	target_addr_t next_address = address;
 	target_addr_t end_address = address + (increment ? count : 1) * size;
 
+	/* TODO: Reading all the elements in a single batch will boost the
+	 * performance.
+	 */
 	while (next_address < end_address) {
 		uint32_t sbcs_write = set_field(0, DM_SBCS_SBREADONADDR, 1);
 		sbcs_write |= sb_sbaccess(size);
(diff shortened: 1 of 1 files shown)


en-sc:
This PR is based on #1073

src/target/riscv/riscv-013.c
```diff
@@ -3607,84 +3610,46 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
...
-					}
-				}
-				next_read_j = j;
-				next_read = address + i * increment + next_read_j * 4;
-				buffer_offset = i * size + next_read_j * 4;
+			const uint32_t size_in_words = DIV_ROUND_UP(size, 4);
+			struct riscv_batch *batch = riscv_batch_alloc(target, size_in_words);
+			riscv_batch_add_dm_read(batch, sbdata[0], RISCV_DELAY_SYSBUS_READ);
+			for (uint32_t j = 1; j < size_in_words; ++j)
+				riscv_batch_add_dm_read(batch, sbdata[j], RISCV_DELAY_BASE);
```
JanMatCodasip: Here I believe read of `sbdata0` should be done as the last:

```suggestion
			/* Read of sbdata0 must be performed as last because it starts the new bus data transfer
			 * (in case sbcs.sbreadondata was set above). We don't want to start the next bus read before 
			 * we fetch all the data from the last bus read. */
			for (uint32_t j = size_in_words - 1; j >= 1; j--)
				riscv_batch_add_dm_read(batch, sbdata[j], RISCV_DELAY_BASE);
			riscv_batch_add_dm_read(batch, sbdata[0], RISCV_DELAY_SYSBUS_READ);				
```

src/target/riscv/riscv-013.c
```diff
@@ -3607,84 +3610,46 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
...
-				dmi_status_t status = dmi_scan(target, NULL, &sbvalue[0], DMI_OP_NOP, 0, 0, false);
-				if (status == DMI_STATUS_BUSY)
-					increase_dmi_busy_delay(target);
-				else if (status == DMI_STATUS_SUCCESS)
-					break;
-				else
-					return ERROR_FAIL;
+			for (size_t k = 0; k < batch->read_keys_used; ++k) {
+				sbvalue[k] = riscv_batch_get_dmi_read_data(batch, k);
+				buf_set_u32(buffer + i * size + k * 4, 0, 8 * size, sbvalue[k]);
```
JanMatCodasip: (When the order of sbdataN accesses is changed in the above loop, this one will of course need to change, too.)

src/target/riscv/riscv-013.c
```diff
@@ -3607,84 +3610,46 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
...
 			}
-			buf_set_u32(buffer + buffer_offset, 0, 8 * MIN(size, 4), sbvalue[0]);
+			riscv_batch_free(batch);
+			next_read = address + i * increment;
 			log_memory_access(next_read, sbvalue, size, true);
+			memset(sbvalue, 0, size);
```
JanMatCodasip: Detail: It looks like this memset is redundant. 

If that is correct understanding, I'd recommend to remove it so that future code readers don't again wonder whether it has any significance.

```suggestion
```

JanMatCodasip (commented):
Overall this change looks all right - thanks. I like how the use of the batch API simplifies the code in read_memory_bus_v1() and makes it easier to reason about.

Two things:

- I have not really grasped where exactly in the original code was the problem causing #1080, and what had to happen for it to occur. Could you please explain it.
- I believe the order of sbdata0 reads should be different - please see my comments.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
en-sc: Thanks a lot for the catch!
Addressed.

en-sc:
> I have not really grasped where exactly in the original code was the problem causing [`read_memory_bus_v1()` is broken #1080](https://github.com/riscv-collab/riscv-openocd/issues/1080), and what had to happen for it to occur. Could you please explain it.

## How the issue occurs?

* The delay is not enough, so the first read results in a busy response:
https://github.com/riscv-collab/riscv-openocd/blob/c6bb902629924eb66aae2a08c0ab8654261c9d71/src/target/riscv/riscv-013.c#L3700
* `next_address = 0x1004` (read from `sbaddress*`).
https://github.com/riscv-collab/riscv-openocd/blob/riscv/src/target/riscv/riscv-013.c#L3708
* The first element is read from `sbdata` correctly.
https://github.com/riscv-collab/riscv-openocd/blob/c6bb902629924eb66aae2a08c0ab8654261c9d71/src/target/riscv/riscv-013.c#L3714-L3715
```
Debug: 647 10 riscv-013.c:3379 log_memory_access64(): M[0x1000] reads 0x00000297
```
* `continue` transfers control to the start of the outermost loop:
https://github.com/riscv-collab/riscv-openocd/blob/c6bb902629924eb66aae2a08c0ab8654261c9d71/src/target/riscv/riscv-013.c#L3722
https://github.com/riscv-collab/riscv-openocd/blob/c6bb902629924eb66aae2a08c0ab8654261c9d71/src/target/riscv/riscv-013.c#L3582
* On this iteration, the loop that reads an element is not entered, since `next_address - address == 4`, `i = (next_address - address) / size == 1`, `i < (count - 1 == 1)` is false.:
https://github.com/riscv-collab/riscv-openocd/blob/c6bb902629924eb66aae2a08c0ab8654261c9d71/src/target/riscv/riscv-013.c#L3615
* However, DMI NOP is executed outside the loop, since the count is greater than 1:
https://github.com/riscv-collab/riscv-openocd/blob/c6bb902629924eb66aae2a08c0ab8654261c9d71/src/target/riscv/riscv-013.c#L3660-L3675
* This NOP returns some garbage (the previous DMI OP was a NOP) and this garbage is assumed to be the result of the read (and overrides the previous correct result):
```
Debug: 661 11 riscv-013.c:3379 log_memory_access64(): M[0xfff] reads 0x00000000
```
* The reported address is equal to `0xfff` due to initialization of `next_read`.
https://github.com/riscv-collab/riscv-openocd/blob/c6bb902629924eb66aae2a08c0ab8654261c9d71/src/target/riscv/riscv-013.c#L3612

### TL;DR:
The condition https://github.com/riscv-collab/riscv-openocd/blob/c6bb902629924eb66aae2a08c0ab8654261c9d71/src/target/riscv/riscv-013.c#L3660
Shoudl guarantee that the body of the loop
https://github.com/riscv-collab/riscv-openocd/blob/c6bb902629924eb66aae2a08c0ab8654261c9d71/src/target/riscv/riscv-013.c#L3615
is executed at least once for the read value to be correct.

## How the commit fixes the issue?

The execution of the NOP is moved into the loop (the last operation in the batch).

JanMatCodasip (dismissed):
@en-sc Thank you for explaining it in detail :+1:. I have understood the original problem now.

LGTM!

en-sc:
@JanMatCodasip, please, take a look once again. There was a conflict with aa9a3fa34898ebda6835897ee1b816c605b111a2.

JanMatCodasip (approved):

src/target/riscv/riscv-013.c
```diff
@@ -3600,77 +3603,44 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
...
-					}
-				}
-				next_read_j = j;
-				next_read = address + i * increment + next_read_j * 4;
-				buffer_offset = i * size + next_read_j * 4;
+			const size_t last_key = batch->read_keys_used - 1;
+			for (size_t k = 0; k <= last_key; ++k) {
+				sbvalue[k] = riscv_batch_get_dmi_read_data(batch,
+						last_key - k);
+				buf_set_u32(buffer + i * size + k * 4, 0, 8 * size, sbvalue[k]);
```
aap-sc, line 3635: @en-sc you have an error here. size can be 8 here. 

```
#7  0x000055555609d04c in buf_set_u32 (_buffer=0x5555569ca720 "\277", first=0, num=64, value=191)                           
    at /home/nikita/projects/openocd/src/helper/binarybuffer.h:43 
#8  0x00005555560b1d13 in read_memory_bus_v1 (target=0x5555569e5390, address=2147491969, size=8, count=2, 
    buffer=0x5555569ca720 "\277", increment=8) at ../openocd/src/target/riscv/riscv-013.c:3410
#9  0x00005555560b58c0 in read_memory (target=0x5555569e5390, address=2147491969, size=8, count=2, 
    buffer=0x5555569ca720 "\277", increment=8) at ../openocd/src/target/riscv/riscv-013.c:4397
#10 0x0000555555c9b5c8 in riscv_read_memory (target=0x5555569e5390, address=2147491969, size=8, count=2, 
    buffer=0x5555569ca720 "\277") at ../openocd/src/target/riscv/riscv.c:3088
#11 0x0000555555c2daee in target_read_memory (target=0x5555569e5390, address=2147491969, size=8, count=2, 
    buffer=0x5555569ca720 "\277") at ../openocd/src/target/target.c:1245
```

JanMatCodasip: (For reference, in case anyone wonders.)

This is being fixed in https://github.com/riscv-collab/riscv-openocd/pull/1109

PR #1082, opened 2024-06-04
target/riscv: simplify `sbcs` read in `write_memory_bus_v1()`

en-sc:


src/target/riscv/riscv-013.c
```diff
@@ -4702,27 +4702,22 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
 		if (result != ERROR_OK)
 			return result;
 
-		/* Read sbcs value.
-		 * At the same time, detect if DMI busy has occurred during the batch write. */
-		bool dmi_busy_encountered;
-		if (dm_op(target, &sbcs, &dmi_busy_encountered, DMI_OP_READ,
-				DM_SBCS, 0, false, true) != ERROR_OK)
-			return ERROR_FAIL;
+		bool dmi_busy_encountered = riscv_batch_was_batch_busy(batch);
```
JanMatCodasip, line 4810: Let's suppose the batch_run() on line 4699 above ends with "busy" result. Can you please help me understand where the subsequent dtmcs.dmireset takes place? My understanding is that:

- No dmireset is performed as part of the batch.
- The subsequent dm_read() on line 4712 starts with DMI busy being still set. 
- The dm_op_timeout(), called from the dm_read(), is bound to fail once due to the busy. Then the dmi reset happens and the operation is retried within dm_op_timeout(), finally having a chance to succeed.

So fortunately, it looks like everything will work in the end, but we have broken the rule that whoever causes DMI busy should also clean up after itself (by dtmcs.dmireset).

To fix this, should we call `increase_dmi_busy_delay()` from `batch_run()`? I believe so.

Note that the other function `batch_run_timeout()` already contains such a call to `increase_dmi_busy_delay()`.

src/target/riscv/riscv-013.c
```diff
@@ -4702,27 +4702,22 @@ static int write_memory_bus_v1(struct target *target, target_addr_t address,
...
 			if (time(NULL) - start > riscv_command_timeout_sec) {
 				LOG_TARGET_ERROR(target, "Timed out after %ds waiting for sbbusy to go low (sbcs=0x%x). "
 						  "Increase the timeout with riscv set_command_timeout_sec.",
 						  riscv_command_timeout_sec, sbcs);
 				return ERROR_FAIL;
 			}
-			if (dm_read(target, &sbcs, DM_SBCS) != ERROR_OK)
-				return ERROR_FAIL;
-		}
+		} while (get_field(sbcs, DM_SBCS_SBBUSY));
```
JanMatCodasip: This whole loop can be replaced by `read_sbcs_nonbusy()` which contains the exact same logic (including the error print).

JanMatCodasip (commented):
Overall I like this idea. Please, if you could take a look at my two comments.

src/target/riscv/riscv-013.c
en-sc: You are right.
Though, I believe this will broaden the scope of this commit (now it is confined to `write_memory_bus_v1()`).
Moreover, this behavior was already there and is not introduced by the change and was already present before.
So I'd like to address the issue by a separate commit: #1096

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
JanMatCodasip: Sure - by all means.

JanMatCodasip (dismissed):
Thank you.

aap-sc:
Omg. https://github.com/riscv-collab/riscv-openocd/compare/e0066b80c76493a7e94052fe204f767afec8c225..4455f7f3c8cf6efbca9c56a9b6e500f73ca51c41 . I feel the pain.

en-sc:
@JanMatCodasip, please, take a look once more. I've missed a use-after-free. I'm terribly sorry...

JanMatCodasip (approved):
No problem. Thanks for catching it.

PR #1083, opened 2024-06-05
target/riscv: deprecate `riscv set_reset_timeout_sec`

en-sc:


src/target/riscv/riscv-011.c
```diff
@@ -739,7 +739,7 @@ static int wait_for_debugint_clear(struct target *target, bool ignore_first)
 
 		if (!bits.interrupt)
 			return ERROR_OK;
-		if (time(NULL) - start > riscv_command_timeout_sec) {
+		if (time(NULL) - start > riscv_command_timeout_sec()) {
```
JanMatCodasip: Detail:

Since this is now a function (getter), not a plain variable, you may consider adding the "_get_" part to its name.

```suggestion
		if (time(NULL) - start > riscv_get_command_timeout_sec()) {
```

src/target/riscv/riscv.c
```diff
@@ -5135,7 +5141,7 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.handler = riscv_set_reset_timeout_sec,
 		.mode = COMMAND_ANY,
 		.usage = "[sec]",
-		.help = "Set the wall-clock timeout (in seconds) after reset is deasserted"
+		.help = "DEPRECATED. Use 'riscv set_command_timeout_sec' instead."
```
JanMatCodasip, line 5075: [Just a note, no change requested]

In vanilla upstream OpenOCD, it seems to be a convention to put the deprecated commands into `startup.tcl` and implement them as TCL proc wrappers on top of the non-deprecated commands. However, that approach does not seem to work well for your case where you'd like to store both the values (and take the higher of them).

So I am fine with the current state, and hope upstream reviewers will be, too.

src/target/riscv/riscv.h
```diff
@@ -21,8 +21,7 @@ struct riscv_program;
 #define RISCV_MAX_HWBPS 16
 #define RISCV_MAX_DMS 100
 
-#define DEFAULT_COMMAND_TIMEOUT_SEC		2
-#define DEFAULT_RESET_TIMEOUT_SEC		30
+#define DEFAULT_COMMAND_TIMEOUT_SEC 30
```
JanMatCodasip: I understand that you have chosen 30 to be defensive and to use the higher of the two original values.

Still, that seems an excessive to me. 

Please, consider lowering it to 5-10 sec (or a similar value), so that the user gets a timely feedback if an operation takes unexpectedly long to complete - without the need to wait the whole 30 seconds. Generally, if any operation takes more than a second (or few seconds), that means the target is really slow or broken, and in either case it would be good if the user learns about it relatively quickly.

Once the user learns about the timeouts, they can take an action - increase the timeout or investigate.

src/target/riscv/riscv.c
JanMatCodasip, line 5075: Please, also remove the deprecated command from the documentation (openocd.texi).

src/target/riscv/riscv.c
```diff
@@ -3922,13 +3927,14 @@ COMMAND_HANDLER(riscv_set_command_timeout_sec)
...
 
-	riscv_command_timeout_sec = timeout;
+	riscv_command_timeout_sec_value = timeout;
 
 	return ERROR_OK;
 }
 
 COMMAND_HANDLER(riscv_set_reset_timeout_sec)
 {
+	LOG_WARNING("This function is deprecated! Please, use 'riscv set_command_timeout_sec'.");
```
```suggestion
	LOG_WARNING("The command 'riscv set_reset_timeout_sec' is deprecated! Please, use 'riscv set_command_timeout_sec'.");
```

src/target/riscv/riscv.h

src/target/riscv/riscv.c
en-sc: Addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv-011.c
en-sc: Addressed.

JanMatCodasip (dismissed):
LGTM, thank you.

JanMatCodasip (approved):

PR #1084, opened 2024-06-05
target/riscv: separate register cache stuff into files

en-sc:
This commit creates file structure for register cache related functions.
Specifically:

* `riscv_reg.h` -- general interface to registers. Safe to use after register cache initialization is successful.
* `riscv_reg_impl.h` -- helper functions to use while implementing register cache initialization.
* `riscv_reg.c` -- definitions of functions from `riscv_reg.h` and `riscv_reg_impl.h`.
* `riscv-011_reg.h` -- register cache interface specific to 0.11 targets.
* `riscv-013_reg.h` -- register cache interface specific to 0.13+ targets.
* `riscv-011/0.13.h` -- version-specific methods used to access registers. Will be extended as needed once other functionality (not related to register access) is separated (e.g. DM/DTM specific stuff).

### DIFF
--- a/src/target/riscv/riscv-011.h
+++ b/src/target/riscv/riscv-011.h
@@ -1,0 +1,15 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+
+#ifndef OPENOCD_TARGET_RISCV_RISCV_011_H
+#define OPENOCD_TARGET_RISCV_RISCV_011_H
+
+#include "riscv.h"
+#include "gdb_regs.h"
+#include "target/target.h"
+
+int riscv011_get_register(struct target *target, riscv_reg_t *value,
+		enum gdb_regno regid);
+int riscv011_set_register(struct target *target, enum gdb_regno regid,
+		riscv_reg_t value);
+
+#endif /*OPENOCD_TARGET_RISCV_RISCV_011_H*/
--- a/src/target/riscv/riscv-011_reg.c
+++ b/src/target/riscv/riscv-011_reg.c
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "riscv-011_reg.h"
+
+#include "riscv_reg_impl.h"
+#include "riscv-011.h"
+
+static int riscv011_reg_get(struct reg *reg)
+{
+	struct target * const target = riscv_reg_impl_get_target(reg);
+	riscv_reg_t value;
+	const int result = riscv011_get_register(target, &value, reg->number);
+	if (result != ERROR_OK)
+		return result;
+	buf_set_u64(reg->value, 0, reg->size, value);
+	return ERROR_OK;
+}
+
+static int riscv011_reg_set(struct reg *reg, uint8_t *buf)
+{
+	const riscv_reg_t value = buf_get_u64(buf, 0, reg->size);
+	struct target * const target = riscv_reg_impl_get_target(reg);
+	return riscv011_set_register(target, reg->number, value);
+}
+
+static const struct reg_arch_type *riscv011_gdb_regno_reg_type(uint32_t regno)
+{
+	static const struct reg_arch_type riscv011_reg_type = {
+		.get = riscv011_reg_get,
+		.set = riscv011_reg_set
+	};
+	return &riscv011_reg_type;
+}
+
+static int riscv011_init_reg(struct target *target, uint32_t regno)
+{
+	return riscv_reg_impl_init_one(target, regno, riscv011_gdb_regno_reg_type(regno));
+}
+
+int riscv011_reg_init_all(struct target *target)
+{
+	if (riscv_reg_impl_init_cache(target) != ERROR_OK)
+		return ERROR_FAIL;
+
+	init_shared_reg_info(target);
+
+	for (uint32_t regno = 0; regno < target->reg_cache->num_regs; ++regno)
+		if (riscv011_init_reg(target, regno) != ERROR_OK)
+			return ERROR_FAIL;
+
+	if (riscv_reg_impl_expose_csrs(target) != ERROR_OK)
+		return ERROR_FAIL;
+
+	riscv_reg_impl_hide_csrs(target);
+
+	return ERROR_OK;
+}
--- a/src/target/riscv/riscv-011_reg.h
+++ b/src/target/riscv/riscv-011_reg.h
@@ -1,0 +1,19 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+
+#ifndef OPENOCD_TARGET_RISCV_RISCV_REG_011_H
+#define OPENOCD_TARGET_RISCV_RISCV_REG_011_H
+
+#include "target/target.h"
+
+/**
+ * This file describes additional register cache interface available to the
+ * RISC-V Debug Specification v0.11 targets.
+ */
+
+/**
+ * Initialize register cache. After this function all registers can be
+ * safely accessed via functions described here and in `riscv_reg.h`.
+ */
+int riscv011_reg_init_all(struct target *target);
+
+#endif /*OPENOCD_TARGET_RISCV_RISCV_REG_011_H*/
--- a/src/target/riscv/riscv-013.h
+++ b/src/target/riscv/riscv-013.h
@@ -1,0 +1,23 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+
+#ifndef OPENOCD_TARGET_RISCV_RISCV_013_H
+#define OPENOCD_TARGET_RISCV_RISCV_013_H
+
+#include "riscv.h"
+
+/* TODO: These functions should be replaced here by access methods that can be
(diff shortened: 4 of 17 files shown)


en-sc:
Please, disregard the failure of Checkpatch. It is due to missing or SPDX-License-Identifier tags.

en-sc:
> Please, disregard the failure of Checkpatch. It is due to missing or SPDX-License-Identifier tags.

Can be fixed by #1085.

src/target/riscv/riscv-011.h
```diff
@@ -0,0 +1,13 @@
+#ifndef OPENOCD_TARGET_RISCV_RISCV_011_H
```
JanMatCodasip, line 3: Please, add the SPDX license marker to the top of the file.

Also check the other newly created files - I have not commented on all of them.

```suggestion
/* SPDX-License-Identifier: GPL-2.0-or-later */

#ifndef OPENOCD_TARGET_RISCV_RISCV_011_H
```

src/target/riscv/riscv-011_reg.c
```diff
@@ -0,0 +1,60 @@
+#ifdef HAVE_CONFIG_H
```
JanMatCodasip, line 3: [Just a note, no change requested]

It is possible that it would make sense to put more register-related functions into the `riscv-011_reg` and `riscv-013_reg` modules. However, I do not yet have a clear opinion on this topic.

So let's leave it as is in this PR. The first step for the code division has been made and it can be further refined as we go.

src/target/riscv/riscv-011_reg.c
```diff
@@ -0,0 +1,60 @@
...
+}
+
+static int riscv011_init_reg(struct target *target, uint32_t regno)
+{
+	return init_reg(target, regno, riscv011_gdb_regno_reg_type(regno));
+}
+
+int riscv011_init_registers(struct target *target)
+{
+	//TODO move to init
```
JanMatCodasip: If you'd like to keep the comment here (and resolve it in a later PR), could you please make it more clear what should be moved where.

src/target/riscv/riscv-011_reg.h
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * This file describes additional register cache interface available to the
+ * RISC-V Debug Specification v0.11 targets.
+ */
+
+/**
+ * Initialize register cache. After this function all registers can be
+ * safely accessed via functions described here and in `riscv_reg.h`.
+ */
+int riscv011_init_registers(struct target *target);
```
JanMatCodasip: I would like to propose a unified naming convention for the files and "public" functions provided by them:

For `riscv-011_<name>.h/c`, all the functions would be called `riscv011_<name>_*`.

In this case, it could be:

```suggestion
int riscv011_reg_init_registers(struct target *target);
```
or
```suggestion
int riscv011_reg_init(struct target *target);
```
or similar.

What is your opinion on this?

src/target/riscv/riscv-013_reg.c
```diff
@@ -0,0 +1,161 @@
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "riscv-013_reg.h"
+
+#include "riscv_reg.h" // TODO: should be removed
```
JanMatCodasip: Could you please help me understand this TODO comment: what prevents you from removing this include right now?

src/target/riscv/riscv-013_reg.c
```diff
@@ -0,0 +1,161 @@
...
+}
+
+static int riscv013_init_reg(struct target *target, uint32_t regno)
+{
+	return init_reg(target, regno, riscv013_gdb_regno_reg_type(regno));
+}
+
+int riscv013_init_registers(struct target *target)
+{
+	//TODO move to init
```
JanMatCodasip: (same as above)

This TODO comment is unclear, it would be better to reword it.

src/target/riscv/riscv.h
```diff
@@ -9,8 +9,9 @@ struct riscv_program;
 #include "opcodes.h"
```
JanMatCodasip, line 9: Could you please change the include guards on lines 3-4 si that they match the convention used in the other files:

```
// current:
#ifndef RISCV_H
#define RISCV_H

// proposed:
#ifndef OPENOCD_TARGET_RISCV_RISCV_H
#define OPENOCD_TARGET_RISCV_RISCV_H
```

Please, could you check also the other existing header files - I have not commented on all of them.

src/target/riscv/riscv_reg_impl.h
```diff
@@ -0,0 +1,211 @@
+#ifndef OPENOCD_TARGET_RISCV_RISCV_REG_IMPL_H
```
JanMatCodasip, line 3: 1) How about calling this `riscv_reg_helpers.h`, `riscv_reg_utils.h` or similar?

2) I would prefer to avoid the `static inline` function definitions in the header file. Instead, I'd recommend to keep the .h/.c structure for this module, too, so that all the modules look the same way and no irregularities are introduced.

3) If you like the naming convention for functions (mentioned in a previous comment), I'd recommend to follow it also here: `riscv_reg_utils_*()`.

JanMatCodasip (commented):
I have done a superficial review and am sending suggestions that I have so far. 

I will have to return to this PR in the next week.

src/target/riscv/riscv-011_reg.c
en-sc: The idea was to move the calls to `riscv_init_reg_cache()` and `init_shared_reg_info()` to `riscv_init_target()`, since they do not depend on target type and do not require communication with target.
However, I don't think it improves anything and it will result in one more function in register cache interface, so, I don't want to do this anymore :).
Dropped the `TODO`.

src/target/riscv/riscv-011.h
en-sc: Addressed.

src/target/riscv/riscv-011_reg.h
en-sc: IMHO, it's a great suggestion and will significantly improve readability, thanks!
Addressed.

src/target/riscv/riscv-013_reg.c
en-sc: I've dropped the `TODO`.
I was thinking target-type-specific implementations (`riscv-01(1|3)_reg.c`) should only use private interface (`riscv_reg_impl.h`) and avoid using public one.
However, when I think of it now, it seems OK.

src/target/riscv/riscv-013_reg.c

src/target/riscv/riscv.h
en-sc: I believe it's better to address separately.
Filed an issue: https://github.com/riscv-collab/riscv-openocd/issues/1097

src/target/riscv/riscv_reg_impl.h
en-sc: 1. I'd like to leave the `_impl` suffix.
IMHO, `_utils` is generally used for useful functionality used throughout the codebase.
Here `_impl` denotes that these headers should only be included in implementations of register cache functionality, not for general use.

2. There are currently 3 `static inline` functions here. I'd like to leave it as-is.
      1. `riscv_reg_impl_is_initialized()` -- it's quite short and I'd like for it to be immediately clear what it means for the register to be "initialized".
      2. `riscv_reg_impl_init_vector_reg_type()` -- as mentioned in a `TODO` there, I'd like to move that function. So, leaving it in the header makes it more of an eyesore and makes me want to address the `TODO` quicker :). If you insist, I'll move it.
      3. `riscv_reg_impl_gdb_regno_cacheable()` -- this one will get removed by future changes. If you insist, I'll move it, but I'd like to leave it there for a reason similar to (2.ii).
3. Addressed, considering (1).

src/target/riscv/riscv-011_reg.c
JanMatCodasip: Ok, thank you for explanation.

src/target/riscv/riscv-013.h
```diff
@@ -0,0 +1,22 @@
...
+
+#ifndef OPENOCD_TARGET_RISCV_RISCV_013_H
+#define OPENOCD_TARGET_RISCV_RISCV_013_H
+
+#include "riscv.h"
+
+/**
+ * TODO: This functions will be and replaced here by specific access methods
+ * (e.g. abstract commands, program buffer writes and so on.), while specifics
+ * on how to use them to get a register's value will be in `riscv-013_reg.c`.
```
JanMatCodasip: Detail: It seems there are some words missing in the first sentence, and as a result of that, I am not sure I have understood the comment.

src/target/riscv/riscv_reg_impl.h
JanMatCodasip: 1) I've now understood why the `_impl` suffix makes more sense in this case, thanks.

2) Understood, let's leave it as is.

3) Thanks.

src/target/riscv/riscv-013.h
en-sc: Addressed.

JanMatCodasip: Thanks.

JanMatCodasip (approved):

PR #1085, opened 2024-06-05
.github/workflows: check git revisions instead of a diff

en-sc:
When running on diff, Checkpatch ignores `Checkpatch-ignore` directives in the commit message.

Change-Id: Ib296d5e972408973fb381fafc51f59569a01d1f0


JanMatCodasip (approved):
This looks all right to me.

PR #1087, opened 2024-06-06
 target/riscv: replace `info->*_delay` with `riscv_scan_delays`

en-sc:
* Improves error handling.
* Handles possible overflow.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -29,7 +29,7 @@ struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans)
 	out->allocated_scans = scans;
 	out->last_scan = RISCV_SCAN_TYPE_INVALID;
 	out->was_run = false;
-	out->used_delay = 0;
+	out->last_scan_delay = 0;
 
 	out->data_out = NULL;
 	out->data_in = NULL;
@@ -109,7 +109,7 @@ static bool riscv_batch_was_scan_busy(const struct riscv_batch *batch,
 }
 
 static void add_idle_before_batch(const struct riscv_batch *batch, size_t start_idx,
-		struct riscv_scan_delays delays)
+		const struct riscv_scan_delays *delays)
 {
 	if (!batch->was_run)
 		return;
@@ -121,9 +121,9 @@ static void add_idle_before_batch(const struct riscv_batch *batch, size_t start_
 		? batch->delay_classes[start_idx - 1]
 		: RISCV_DELAY_BASE;
 	const unsigned int new_delay = riscv_scan_get_delay(delays, delay_class);
-	if (new_delay <= batch->used_delay)
+	if (new_delay <= batch->last_scan_delay)
 		return;
-	const unsigned int idle_change = new_delay - batch->used_delay;
+	const unsigned int idle_change = new_delay - batch->last_scan_delay;
 	LOG_TARGET_DEBUG(batch->target, "Adding %u idle cycles before the batch.",
 			idle_change);
 	assert(idle_change <= INT_MAX);
@@ -131,19 +131,19 @@ static void add_idle_before_batch(const struct riscv_batch *batch, size_t start_
 }
 
 static int get_delay(const struct riscv_batch *batch, size_t scan_idx,
-		struct riscv_scan_delays delays)
+		const struct riscv_scan_delays *delays)
 {
 	assert(batch);
 	assert(scan_idx < batch->used_scans);
 	const enum riscv_scan_delay_class delay_class =
 		batch->delay_classes[scan_idx];
-	const unsigned int delay =  riscv_scan_get_delay(delays, delay_class);
+	const unsigned int delay = riscv_scan_get_delay(delays, delay_class);
 	assert(delay <= INT_MAX);
 	return delay;
 }
 
 int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
-		struct riscv_scan_delays delays, bool resets_delays,
+		const struct riscv_scan_delays *delays, bool resets_delays,
 		size_t reset_delays_after)
 {
 	assert(batch->used_scans);
@@ -195,7 +195,7 @@ int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
 	}
 
 	batch->was_run = true;
-	batch->used_delay = get_delay(batch, batch->used_scans - 1, delays);
+	batch->last_scan_delay = get_delay(batch, batch->used_scans - 1, delays);
 	return ERROR_OK;
 }
 
--- a/src/target/riscv/batch.h
+++ b/src/target/riscv/batch.h
@@ -25,40 +25,66 @@ enum riscv_scan_delay_class {
 	/* Delay for System Bus read operation: */
 	RISCV_DELAY_SYSBUS_READ,
 	/* Delay for System Bus write operation: */
-	RISCV_DELAY_SYSBUS_WRITE,
+	RISCV_DELAY_SYSBUS_WRITE
 };
 
+static inline const char *
+riscv_scan_delay_class_name(enum riscv_scan_delay_class delay_class)
+{
+	switch (delay_class) {
+	case RISCV_DELAY_BASE:
+		return "DM access";
+	case RISCV_DELAY_ABSTRACT_COMMAND:
+		return "Abstract Command";
+	case RISCV_DELAY_SYSBUS_READ:
+		return "System Bus read";
+	case RISCV_DELAY_SYSBUS_WRITE:
+		return "System Bus write";
+	}
+	assert(0);
+	return NULL;
+}
+
+/* The scan delay values are passed to "jtag_add_runtest()", which accepts an
+ * "int".  Therefore, the passed value should be no greater than "INT_MAX".
+ *
+ * Since the resulting delay value can be a sum of two individual delays,
+ * individual delays are limited to "INT_MAX / 2" to prevent overflow of the
+ * final sum.
+ */
+#define RISCV_SCAN_DELAY_MAX (INT_MAX / 2)
+
 struct riscv_scan_delays {
-	/* The purpose of these delays is to be passed to "jtag_add_runtest()",
-	 * which accepts an "int".
-	 * Therefore, they should be no greater then "INT_MAX".
-	 */
 	unsigned int base_delay;
 	unsigned int ac_delay;
 	unsigned int sb_read_delay;
 	unsigned int sb_write_delay;
 };
 
-static inline unsigned int riscv_scan_get_delay(struct riscv_scan_delays delays,
+static inline unsigned int
+riscv_scan_get_delay(const struct riscv_scan_delays *delays,
 		enum riscv_scan_delay_class delay_class)
 {
 	switch (delay_class) {
 	case RISCV_DELAY_BASE:
-		return delays.base_delay;
+		return delays->base_delay;
 	case RISCV_DELAY_ABSTRACT_COMMAND:
-		return delays.ac_delay;
+		return delays->base_delay + delays->ac_delay;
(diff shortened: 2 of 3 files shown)


en-sc:
Based on #1073.

src/target/riscv/batch.h
```diff
@@ -48,8 +159,10 @@ struct riscv_batch {
 	 * However, RISC-V DMI "busy" condition could still have occurred.
 	 */
 	bool was_run;
-	/* Idle count used on the last run. Only valid after `was_run` is set. */
-	size_t used_idle_count;
+	/* Number of RTI cycles used by the last scan on the last run.
+	 * Only valid when `was_run` is set.
+	 */
+	unsigned int used_delay;
```
JanMatCodasip: I'd recommend to name this "last_scan_delay" (or similar) to make its meaning more clear to the reader.

```suggestion
	unsigned int last_scan_delay;
```

src/target/riscv/batch.h
```diff
@@ -14,6 +14,116 @@ enum riscv_scan_type {
...
+	case RISCV_DELAY_ABSTRACT_COMMAND:
+		return "Abstract Command";
+	case RISCV_DELAY_SYSBUS_READ:
+		return "System Bus read";
+	case RISCV_DELAY_SYSBUS_WRITE:
+		return "System Bus write";
+	case RISCV_N_DELAY_CLASSES:
+		assert(0);
+	}
+	return NULL;
```
JanMatCodasip, line 45: For robustness, I'd recommend to use `assert(0);` also in the case none of the switch branches match.

```suggestion
	assert(0);
	return NULL;
```

src/target/riscv/riscv-013.c
```diff
@@ -479,15 +470,38 @@ static int dtmcontrol_scan(struct target *target, uint32_t out, uint32_t *in_ptr
...
+	for (enum riscv_scan_delay_class delay_class_i = 0;
+			delay_class_i < RISCV_N_DELAY_CLASSES;
+			++delay_class_i) {
+		int new_res = riscv_scan_increase_delay(&info->learned_delays,
+				delay_class_i, RISCV_DELAY_BASE);
+		if (res == ERROR_OK)
+			res = new_res;
+	}
+	return res;
+}
```
JanMatCodasip: Please, consider storing the "raw" values in `info->learned_delays`. That means, do not add automatically add RISCV_DELAY_BASE to the other RISCV_DELAY_* classes. 

As a general principle, I believe all data should be stored in its original (raw) form, not in pre-computed form. (Apart from cases of caching.)

This way:

- The API of `riscv_scan_increase_delay` will get simpler, the last parameter can be dropped. The resulting API will be clearer to the reader.
- The functions reading the delay will read two items: base delay + the other delay. That will result in clearer code - the reader will explicitly see what is happening and that combination of two delays is used.

JanMatCodasip: For example, there could be two functions:

- `riscv_scan_get_raw_delay()` 
- `riscv_scan_get_delay()`
 
The second one would provide the ready-to-use delay value, including the base.

JanMatCodasip: This will make the `int` overflow check slightly more difficult, but lowering the limit to `INT_MAX / 2` will do the trick.

Reducing the limit from `INT_MAX` to `INT_MAX / 2` corresponds to only around 7 rounds of the exponential back-off, so there should be no issue with that.

src/target/riscv/batch.c
```diff
@@ -101,28 +108,51 @@ static bool riscv_batch_was_scan_busy(const struct riscv_batch *batch,
...
 }
 
+static int get_delay(const struct riscv_batch *batch, size_t scan_idx,
+		struct riscv_scan_delays delays)
+{
+	assert(batch);
+	assert(scan_idx < batch->used_scans);
+	const enum riscv_scan_delay_class delay_class =
+		batch->delay_classes[scan_idx];
+	const unsigned int delay =  riscv_scan_get_delay(delays, delay_class);
```
JanMatCodasip: Nitpick:

```suggestion
	const unsigned int delay = riscv_scan_get_delay(delays, delay_class);
```

src/target/riscv/riscv-013.c
```diff
@@ -555,9 +567,9 @@ static dmi_status_t dmi_scan(struct target *target, uint32_t *address_in,
 		jtag_add_dr_scan(target->tap, 1, &field, TAP_IDLE);
 	}
 
-	int idle_count = info->dmi_busy_delay;
-	if (exec)
-		idle_count += info->ac_busy_delay;
+	int idle_count = exec
+		? learned_delay(target, RISCV_DELAY_ABSTRACT_COMMAND)
+		: learned_delay(target, RISCV_DELAY_BASE);
```
JanMatCodasip: Please, consider dropping the function `learned_delay()`. 

In all functions where it is currently used, it can be simply replaced by calling `riscv_scan_get_delay(info, ...)` and the pointer to `riscv013_info_t` is available in all places where learned_delay() is currently used.

For the sake of code readability, it is IMO better to avoid having multiple functions that do (almost) the same thing. 

```suggestion
	int idle_count = exec
		? riscv_scan_get_delay(info, RISCV_DELAY_ABSTRACT_COMMAND)
		: riscv_scan_get_delay(info, RISCV_DELAY_BASE);
```

src/target/riscv/riscv-013.c
```diff
@@ -928,6 +927,51 @@ static int dm013_select_target(struct target *target)
 	return dm013_select_hart(target, info->index);
 }
 
+#define EXECUTE_ABSTRACT_COMMAND_BATCH_SIZE 2
```
JanMatCodasip: This could be called just:

```suggestion
#define ABSTRACT_COMMAND_BATCH_SIZE 2
```

src/target/riscv/batch.h
```diff
@@ -14,6 +14,116 @@ enum riscv_scan_type {
...
+		return delays.ac_delay;
+	case RISCV_DELAY_SYSBUS_READ:
+		return delays.sb_read_delay;
+	case RISCV_DELAY_SYSBUS_WRITE:
+		return delays.sb_write_delay;
+	case RISCV_N_DELAY_CLASSES:
+		assert(0);
+	}
+	return 0;
+}
```
JanMatCodasip: Here I'd recommend to pass the structure via const pointer.

I'd also appears reasonable to add `assert(false)` prior to the `return`.

```suggestion
static inline unsigned int riscv_scan_get_delay(const struct riscv_scan_delays *delays,
		enum riscv_scan_delay_class delay_class)
{
	assert(delays);
	switch (delay_class) {
	case RISCV_DELAY_BASE:
		return delays->base_delay;
	case RISCV_DELAY_ABSTRACT_COMMAND:
		return delays->ac_delay;
	case RISCV_DELAY_SYSBUS_READ:
		return delays->sb_read_delay;
	case RISCV_DELAY_SYSBUS_WRITE:
		return delays->sb_write_delay;
	case RISCV_N_DELAY_CLASSES:
		assert(false);
	}
	assert(false);
	return 0;
}
```

JanMatCodasip (changes requested):
Thank you for the changes. I have completed my first review round and posting my suggestions.

src/target/riscv/batch.h

src/target/riscv/batch.h
en-sc: Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/batch.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/batch.h

src/target/riscv/riscv-013.c
```diff
@@ -479,15 +470,25 @@ static int dtmcontrol_scan(struct target *target, uint32_t out, uint32_t *in_ptr
...
+	res = riscv_scan_increase_delay(&info->learned_delays,
+			RISCV_DELAY_BASE);
+	return res;
+}
+
+static void reset_learned_delays(struct target *target)
+{
+	RISCV013_INFO(info);
+	assert(info);
+	memset(&info->learned_delays, 0, sizeof(info->learned_delays));
```
JanMatCodasip, line 491: [Just a note; no change requested]

I typically prefer to reset all elements of a structure by assigning zeros to each member independently.

It is more typing, but the advantages are:

- static analysis tools are more likely to understand that than the memset
- when searching for all uses of variables (structure members), it is easy to miss the memset on the whole structure

src/target/riscv/batch.h
```diff
@@ -25,40 +25,64 @@ enum riscv_scan_delay_class {
...
+	}
+	assert(0);
+	return NULL;
+}
+
+/* The purpose of these delays is to be passed to "jtag_add_runtest()",
+ * which accepts an "int".
+ * Therefore, they should be no greater then "INT_MAX" after adding the base
+ * delay.
+ */
```
JanMatCodasip: I have tried to reword the comment for to make it even clearer. Please let me know if you like this wording.

```suggestion
/* The scan delay values are passed to "jtag_add_runtest()", which accepts an "int".
 * Therefore, the passed value should be no greater than "INT_MAX".
 *
 * Since the resulting delay value can be a sum of two individual delays, we limit the 
 * individual delays to INT_MAX / 2 to prevent overflow of the final sum.
 */
```

src/target/riscv/batch.h
```diff
@@ -73,6 +97,25 @@ static inline void riscv_scan_set_delay(struct riscv_scan_delays *delays,
...
+		enum riscv_scan_delay_class delay_class)
+{
+	const unsigned int delay = riscv_scan_get_delay(delays, delay_class);
+	const unsigned int delay_step = delay / 10 + 1;
+	if (delay > RISCV_SCAN_DELAY_MAX - delay_step) {
+		/* It's not clear if this issue actually occurs in real
+		 * use-cases, so stick with the simple solution until the
+		 * first bug report.
+		 */
+		LOG_ERROR("BUG: Delay %s (%d) is not increased anymore.",
```
JanMatCodasip: Detail: I have also reworded the message a little. I would also prefer to not call this "BUG" since this can happen due to reasons outside of OpenOCD's developer control.

I would keep this as an error to make sure the message is very visible, should this ever happen.

```suggestion
		LOG_ERROR("Delay for %s (%d) is not increased anymore (maximum was reached).",
```

src/target/riscv/batch.h
```diff
@@ -73,6 +97,25 @@ static inline void riscv_scan_set_delay(struct riscv_scan_delays *delays,
...
+}
+
+static inline int riscv_scan_increase_delay(struct riscv_scan_delays *delays,
+		enum riscv_scan_delay_class delay_class)
+{
+	const unsigned int delay = riscv_scan_get_delay(delays, delay_class);
+	const unsigned int delay_step = delay / 10 + 1;
+	if (delay > RISCV_SCAN_DELAY_MAX - delay_step) {
+		/* It's not clear if this issue actually occurs in real
+		 * use-cases, so stick with the simple solution until the
```
JanMatCodasip: Pedantic detail:

```suggestion
		 * use-cases, so stick with a simple solution until the
```

JanMatCodasip (dismissed):
LGTM, thank you.

I have posted few last optional suggestions.

src/target/riscv/batch.h
en-sc: I believe it's better, thanks! Addressed.

src/target/riscv/batch.h
en-sc: Adressed.

src/target/riscv/batch.h

JanMatCodasip (approved):
LGTM, thank you.

PR #1089, opened 2024-06-13
target/riscv: select DMI IR on batch access

en-sc:
Without the selection the TAP can be left in bypass.

Change-Id: I79c6bf74802dc9c9475947d1787a3d0b797f3952

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2777,6 +2777,7 @@ struct riscv_scan_delays get_scan_delays(struct target *target)
 static int batch_run(struct target *target, struct riscv_batch *batch)
 {
 	RISCV_INFO(r);
+	select_dmi(target);
 	riscv_batch_add_nop(batch);
 	const int result = riscv_batch_run_from(batch, 0,
 			get_scan_delays(target),
@@ -2797,6 +2798,7 @@ static int batch_run_timeout(struct target *target, struct riscv_batch *batch)
 {
 	RISCV013_INFO(info);
 
+	select_dmi(target);
 	riscv_batch_add_nop(batch);
 
 	size_t finished_scans = 0;


en-sc:
@JanMatCodasip, can you please prioritize looking at this one. It's an uncaught bug in #1073.

JanMatCodasip (dismissed):
@en-sc I have checked this fix visually and it looks good. Thank you.

Few additional notes (perhaps for future pull requests):

- The calls to `select_dmi()` in `assert_reset()`, `deassert_reset()` and `read_memory_progbuf()` appear to be redundant. The underlying batch API or dm_read/dm_write() API should always take care of that.
- I would be in favor of having assertions that the DMI has been selected, however the BSCAN tunnel support complicates that - the assertions would not be simple.
- Furthermore, I am afraid we don't have any way how to test the BSCAN tunnel support. (Or do we?)
- I would eventually like to move all the JTAG DTM related items to their own module (riscv013-jtag-dtm.c/h). Then the select_dmi() could be called directly from batch.c, which would put the select_dmi() call closer to the actual execution of the JTAG scans.

en-sc:
@JanMatCodasip, please, take a look again.
I would like to separate the check in `select_dmi()` into #1091 and test it some more.

en-sc:
> * The calls to `select_dmi()` in `assert_reset()`, `deassert_reset()` and `read_memory_progbuf()` appear to be redundant. The underlying batch API or dm_read/dm_write() API should always take care of that.

I think you are right. Filed #1092.

> * I would be in favor of having assertions that the DMI has been selected, however the BSCAN tunnel support complicates that - the assertions would not be simple.

I'm not sure it's worthwhile. We do have assertions already JTAG common interface that cover this.

> * Furthermore, I am afraid we don't have any way how to test the BSCAN tunnel support. (Or do we?)

AFAIK, we don't.

> * I would eventually like to move all the JTAG DTM related items to their own module (riscv013-jtag-dtm.c/h). Then the select_dmi() could be called directly from batch.c, which would put the select_dmi() call closer to the actual execution of the JTAG scans.

I'd like that to. This is essentially the goal of all these commits -- to tidy up the work with DM and move it to a separate file after this.

JanMatCodasip (approved):
LGTM

JanMatCodasip:
> en-sc: I would like to separate the check in select_dmi() into https://github.com/riscv-collab/riscv-openocd/pull/1091 and test it some more.

Sure, no problem!

> JanMatCodsip: I would be in favor of having assertions that the DMI has been selected (...)

> en-sc: I'm not sure it's worthwhile. We do have assertions already JTAG common interface that cover this.

Ok, let's leave it as is, without additional asserts.

> JanMatCodasip: I would eventually like to move all the JTAG DTM related items to their own module (...)

> en-sc: I'd like that to. This is essentially the goal of all these commits (...)

Great :)

PR #1090, opened 2024-06-14
target/riscv: Ensure other TAPs are in BYPASS for batch DR scan

rbradford:
In configurations with multiple TAPs and targets it is necessary to ensure that the TAPs other than the one chosen for the batch are in BYPASS. This can be done by doing an IR scan which will ensure that all other than the active TAP are in BYPASS.

Make the select_dmi function non-static and use it from the batch code to avoid duplication.

This resolves an assertion when running with a configuration with multiple TAPs:

src/jtag/drivers/driver.c:156: interface_jtag_add_dr_scan: Assertion `active == tap' failed.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -157,6 +157,8 @@ int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
 	LOG_TARGET_DEBUG(batch->target, "Running batch of scans [%zu, %zu)",
 			start_idx, batch->used_scans);
 
+	select_dmi(batch->target);
+
 	for (size_t i = start_idx; i < batch->used_scans; ++i) {
 		if (bscan_tunnel_ir_width != 0)
 			riscv_add_bscan_tunneled_scan(batch->target, batch->fields + i, batch->bscan_ctxt + i);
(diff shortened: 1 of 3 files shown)


JanMatCodasip:
@rbradford Hi Rob, thank you for the patch. It looks like your pull request is a slightly different fix for the issue already being addressed by https://github.com/riscv-collab/riscv-openocd/pull/1089. Is that correct?

If so, can you please check if https://github.com/riscv-collab/riscv-openocd/pull/1089 addresses your issue?

en-sc:
@rbradford, thank you for the patch! I'm sorry for introducing the bug in #1073.
If you don't mind, I would like to use #1089.
The reason is, `riscv_batch_run_from()` is called repeatedly by `batch_run_timeout()`. That will result in unnecessary extra IR scans (unless the code of `dtmcontrol_scan()` is adjusted).
https://github.com/riscv-collab/riscv-openocd/blob/fc5cd940ed25f370bd0bafc1067c3ef9069d4744/src/target/riscv/riscv-013.c#L466

rbradford:
Thank you both! I was able to test the patch from #1089 and it resolved the issue for me! :+1:

PR #1091, opened 2024-06-14
target/riscv: avoid unnecessary IR scans

en-sc:


JanMatCodasip:
I have checked the change in select_dmi() visually and it looks all right to me. It is equivalent to what majority (or all?) other targets do.

I can see this PR is still marked as a draft. @en-sc, do you have any doubts about this one or any further testing that you'd like to carry out?

(Also the PR needs to be rebased.)

en-sc:
@JanMatCodasip, I've tested the commit and it seems fine. Please, take a look.

JanMatCodasip (approved):

en-sc:
This commit introduces an issue for targets with `IRLEN >= 8` (the issue is not in the commit itself, but in `buf_cmp()`).
The fix is being merged to mainline OpenOCD (huge thanks to @jrtc27): https://review.openocd.org/c/openocd/+/8489

PR #1094, opened 2024-06-25
Merge up to ad87fbd1cf28760795c4e18f3318a2d720e5a8a6 from upstream

en-sc:
Conflicts:

* `doc/openocd.texi`: due to https://github.com/riscv-collab/riscv-openocd/commit/d382c95d57c0ad9ed2dcc83c95404babb7647708, resolved by selecting the upstream version.

* `src/server/gdb_server.c`: between https://github.com/riscv-collab/riscv-openocd/commit/944fe66f104e356c5fcd2b5c25200cebef9b389c and https://github.com/riscv-collab/riscv-openocd/commit/92e8823ebdb6d01b41bb5d79af49501d525acd1d. Resolved by adopting the use of `LOG_TARGET_*`.
* `src/target/target.c`: between https://github.com/riscv-collab/riscv-openocd/commit/639e68a621b7ae8c4a296ca7e45b47075268fded and https://github.com/riscv-collab/riscv-openocd/commit/c5358c84ad0d3e7497498e0457cec7785f72910a, selected the version from `riscv-openocd`.

PR #1095, opened 2024-06-30
ch347 usb2jtag driver integrated + Quad SPI Driver for Riscduino SOC …

dineshannayya:
I have Integrated following functionality
1) src/jtag/drivers/ch347.c : USB To JTAG CH347 Driver : https://www.wch-ic.com/products/CH347.html
2)  src/flash/nor/rqspi.c :  Quad SPI Driver for Riscduino SOC 
These driver needed for following RISCV SOC platform 
Single 32 Bit RISCV Core: https://github.com/riscduino/riscduino
Dual 32 Bit RISCV Core: https://github.com/riscduino/riscduino_dcore
Quad 32 RISCV Core: https://github.com/riscduino/riscduino_qcore

Let me know any additional information needed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -130,7 +130,8 @@ m4_define([USB1_ADAPTERS],
 	[[armjtagew], [Olimex ARM-JTAG-EW Programmer], [ARMJTAGEW]],
 	[[rlink], [Raisonance RLink JTAG Programmer], [RLINK]],
 	[[usbprog], [USBProg JTAG Programmer], [USBPROG]],
-	[[esp_usb_jtag], [Espressif JTAG Programmer], [ESP_USB_JTAG]]])
+	[[esp_usb_jtag], [Espressif JTAG Programmer], [ESP_USB_JTAG]],
+        [[ch347], [Mode 3 of the CH347 devices], [CH347]]])
 
 m4_define([HIDAPI_ADAPTERS],
 	[[[cmsis_dap], [CMSIS-DAP Compliant Debugger], [CMSIS_DAP_HID]],
--- a/src/flash/nor/Makefile.am
+++ b/src/flash/nor/Makefile.am
@@ -80,6 +80,7 @@ NOR_DRIVERS = \
 	%D%/w600.c \
 	%D%/xcf.c \
 	%D%/xmc1xxx.c \
+	%D%/rqspi.c \
 	%D%/xmc4xxx.c
 
 NORHEADERS = \
--- a/src/flash/nor/driver.h
+++ b/src/flash/nor/driver.h
@@ -309,5 +309,6 @@ extern const struct flash_driver w600_flash;
 extern const struct flash_driver xcf_flash;
 extern const struct flash_driver xmc1xxx_flash;
 extern const struct flash_driver xmc4xxx_flash;
+extern const struct flash_driver rqspi_flash;
 
 #endif /* OPENOCD_FLASH_NOR_DRIVER_H */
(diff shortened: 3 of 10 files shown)


TommyMurphyTM1234:
Should changes like this, not involving any files in the `src/target/riscv` directory, maybe be done in the upstream OpenOCD project and then back ported from there?

TommyMurphyTM1234:
> Should changes like this, not involving any files in the `src/target/riscv` directory, maybe be done in the upstream OpenOCD project and then back ported from there?

Actually, was this already added upstream:

- https://sourceforge.net/p/openocd/mailman/search/?q=ch347

and, as such, should probably be integrated from there via a PR like this:

- https://github.com/riscv-collab/riscv-openocd/pull/1094

Or am I missing something here?

dineshannayya:
@TommyMurphyTM1234  Not clear with your response, Do you want me push these changes into https://github.com/en-sc/riscv-openocd.git Repo first ?

TommyMurphyTM1234:
> @TommyMurphyTM1234 Not clear with your response, Do you want me push these changes into https://github.com/en-sc/riscv-openocd.git Repo first ?

No - I never mentioned that repo.

What I'm wondering is...

1. If these changes were already made upstream (https://openocd.org/pages/repos.html) in which case maybe they should be integrated here in the `riscv-openocd` via a PR such as this: https://github.com/riscv-collab/riscv-openocd/pull/1094
2. Assuming that these changes have NOT been made upstream already then maybe that's where they should be done since they are not RISC-V specific (i.e. don't touch any files in `src/target/riscv`)?

Others such as @en-sc might be able to comment?

JanMatCodasip (changes requested):
Hi @dineshannayya, thank you for your contribution. However, there are multiple issues with your pull request.

Please, don't integrate multiple (semi-)independent features into a single merge request.

As for the CH347 driver:

- That change is RISC-V independent.
- It is already being reviewed in vanilla OpenOCD upstream: https://review.openocd.org/c/openocd/+/7937. 
- If you are interested in this change, please help the review process in upstream. 
- Once this becomes part of vanilla OpenOCD, we will get this change into riscv-openocd via a periodic merge ([here is an example](https://github.com/riscv-collab/riscv-openocd/pull/1094) of how we synchronize with vanilla OpenOCD).

As for the rqspi flash driver:

- Will this change work with vanilla OpenOCD? If so, please create a merge request there. Then we will get the change automatically via a periodic sync.
- I have quickly looked at the code and it does not follow the OpenOCD code style (mixture of spaces and tabs for indentation, inconsistent use of spaces `if(cond)` vs `if (cond)`).

As for the config file riscduino-dcore.cfg:

- There seems to be commented code for SMP mode, and it looks like the user should uncomment things based on if they want SMP or not.
- This is not how the configuration files work. Imagine OpenOCD is installed into a read-only location where the user modifications are not possible. Instead, there should be two configuration files (for SMP and non-SMP), which can possibly both include a third file with a common code (if that is suitable). Or at least one configuration file with a boolean flag that change the mode. Or some wise combination of both these approaches.

TL;DR:

- This should be split into three parts.
- The first part (CH347) is already being dealt with in upstream, and should not be re-submitted here.
- The other two parts need more work and cannot be merged as is. Furthermore, it is unclear whether they actually depend on riscv-openocd or could instead go into vanilla OpenOCD upstream.

JanMatCodasip:
Closing this stale pull request as there has not been any development for the past 2 months.

PR #1096, opened 2024-07-01
target/riscv: reset `dmi.busy` after batches

en-sc:
Additionally, avoid calling `riscv_batch_finished_scans()` / decrementing reset counter if the batch run failed.

Change-Id: I3eb7b23e4dc029090e92e3e543719824add623e1

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2769,12 +2769,16 @@ static int batch_run(struct target *target, struct riscv_batch *batch)
 	const int result = riscv_batch_run_from(batch, 0, &info->learned_delays,
 			/*resets_delays*/  r->reset_delays_wait >= 0,
 			r->reset_delays_wait);
+	if (result != ERROR_OK)
+		return result;
 	/* TODO: To use `riscv_batch_finished_scans()` here, it is needed for
 	 * all scans to not discard input, meaning
 	 * "riscv_batch_add_dm_write(..., false)" should not be used. */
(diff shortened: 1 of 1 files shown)


en-sc:
Note that this PR conflicts with #1087, since #1087 makes it so `increase_dmi_busy_delay()` can return an error.
I'd prefer for #1087 to be merged first.

src/target/riscv/riscv-013.c
```diff
@@ -2783,12 +2783,16 @@ static int batch_run(struct target *target, struct riscv_batch *batch)
...
+		return result;
 	/* TODO: To use `riscv_batch_finished_scans()` here, it is needed for
 	 * all scans to not discard input, meaning
 	 * "riscv_batch_add_dm_write(..., false)" should not be used. */
 	const size_t finished_scans = batch->used_scans;
 	decrement_reset_delays_counter(target, finished_scans);
-	return result;
+	if (riscv_batch_was_batch_busy(batch))
+		increase_dmi_busy_delay(target);
+	return ERROR_OK;
```
```suggestion
	if (riscv_batch_was_batch_busy(batch))
		return increase_dmi_busy_delay(target);
	return ERROR_OK;
```

en-sc (commented):
This are the changes that will be needed after #1087 is merged.

JanMatCodasip (dismissed):
LGTM

src/target/riscv/riscv-013.c

JanMatCodasip (approved):

Issue #1026, opened 2024-03-04
Rewrite read_memory_bus_v1() to not use low-level JTAG functions


JanMatCodasip:
Note to self:

Function for System Bus read operation `read_memory_bus_v1()` uses low-level JTAG operations (functions `jtag_*()`) to carry out the DMI operations.

Rewrite this function to use the proper DMI access functions (`dm_read()`, `dm_write()`, `riscv_batch_*()` etc.).

See:
https://github.com/riscv-collab/riscv-openocd/blob/riscv/src/target/riscv/riscv-013.c#L3219

PR #1099, opened 2024-07-03
 target/riscv: write SB address using batch

en-sc:
Reduces the number of JTAG queue flushes.

Requires #1081, #1082

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -731,17 +731,6 @@ static uint32_t riscv013_get_dmi_address(const struct target *target, uint32_t a
 	return address + base;
 }
 
-static int dm_op(struct target *target, uint32_t *data_in,
-		bool *dmi_busy_encountered, int op, uint32_t address,
-		uint32_t data_out, bool exec, bool ensure_success)
-{
-	dm013_info_t *dm = get_dm(target);
-	if (!dm)
-		return ERROR_FAIL;
-	return dmi_op(target, data_in, dmi_busy_encountered, op, address + dm->base,
-			data_out, exec, ensure_success);
-}
-
 static int dm_read(struct target *target, uint32_t *value, uint32_t address)
 {
 	dm013_info_t *dm = get_dm(target);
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -2743,21 +2732,42 @@ static uint32_t sb_sbaccess(unsigned int size_bytes)
 	return 0;
 }
 
-static int sb_write_address(struct target *target, target_addr_t address,
-							bool ensure_success)
+static unsigned int sbaddress_n_writes(const struct target *target)
```
MarekVCodasip: I would prefer if this function had a more descriptive name.

Something like `get_sbaddress_word_count()` or `get_sbaadress_reg_count()`

src/target/riscv/riscv-013.c
```diff
@@ -2743,21 +2732,42 @@ static uint32_t sb_sbaccess(unsigned int size_bytes)
...
-		dm_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS2, 0, false, false);
-	if (sbasize > 32)
-		dm_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS1,
-			(uint32_t)(address >> 32), false, false);
-	return dm_op(target, NULL, NULL, DMI_OP_WRITE, DM_SBADDRESS0,
-		(uint32_t)address, false, ensure_success);
+	assert(sizeof(target_addr_t) == sizeof(uint64_t));
+	const uint32_t addresses[] = {DM_SBADDRESS0, DM_SBADDRESS1, DM_SBADDRESS2, DM_SBADDRESS3};
+	const uint32_t values[] = {(uint32_t)address, (uint32_t)(address >> 32), 0, 0};
+	const unsigned int n_writes = sbaddress_n_writes(target);
```
MarekVCodasip: If you decide to change the function name, I feel like this variable name should be changed as well.

src/target/riscv/riscv-013.c
```diff
@@ -2743,21 +2732,42 @@ static uint32_t sb_sbaccess(unsigned int size_bytes)
...
+	assert(n_writes > 0);
+	assert(n_writes <= ARRAY_SIZE(addresses));
+	assert(ARRAY_SIZE(addresses) == ARRAY_SIZE(values));
+	for (unsigned int i = n_writes - 1; i > 0; --i)
```
```suggestion
	assert(ARRAY_SIZE(addresses) == ARRAY_SIZE(values));
	
	for (unsigned int i = n_writes - 1; i > 0; --i)
```

If you prefer, I would add an extra empty line, for clarity.

src/target/riscv/riscv-013.c
```diff
@@ -3581,96 +3594,51 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
...
-				buffer_offset = i * size + next_read_j * 4;
+			const size_t last_key = batch->read_keys_used - 1;
+			for (size_t k = 0; k <= last_key; ++k) {
+				sbvalue[k] = riscv_batch_get_dmi_read_data(batch,
+						last_key - k);
+				buf_set_u32(buffer + i * size + k * 4, 0, 8 * size, sbvalue[k]);
 			}
+			riscv_batch_free(batch);
+			const target_addr_t read_addr = address + i * increment;
+			log_memory_access(read_addr, sbvalue, size, true);
```
MarekVCodasip: Up to your consideration: I think this functionality should be separated into it's own data read function. Something like`sb_read_data()`, in the same vein as `sb_write_address()` It could be also reused later in `sample_memory_bus_v1()`

MarekVCodasip (commented):
Thanks for the patch. This is a nice improvement. I have a few suggestions, please take a look.

Also a small question, did you consider also using the improvement in `sample_memory_bus_v1()`? Or do you plan to implement it in a separate patch?

src/target/riscv/riscv-013.c
en-sc: As mentioned in the `TODO`, IMHO the whole function can be refactored, improving the performance and readability using bactches and `riscv_batch_run_from()` https://github.com/riscv-collab/riscv-openocd/blob/bc3d9606962c0e7056f10d59a880e786de9d351e/src/target/riscv/riscv-013.c#L3573-L3575

I'd like for this to be addressed in a separate commit.

The main point of #1081, #1082, #1083 and this one is to address #1026 and unlock #1100 (by eliminating unwrapped use of `dmi_op`)

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
MarekVCodasip: Alright, sounds reasonable.

MarekVCodasip (commented):
Are you leaving the `sample_memory_bus_v1()` for later patch then?

en-sc:
> Are you leaving the `sample_memory_bus_v1()` for later patch then?

AFAIU, `sample_memory_bus_v1` is a bit different -- there is no need to override the whole address, it's enough to touch `sbaddr0` to trigger the next read, so `sb_write_address` shouldn't be used there.

MarekVCodasip:
> > Are you leaving the `sample_memory_bus_v1()` for later patch then?
> 
> AFAIU, `sample_memory_bus_v1` is a bit different -- there is no need to override the whole address, it's enough to touch `sbaddr0` to trigger the next read, so `sb_write_address` shouldn't be used there.

Yes, you are correct. Approving.

MarekVCodasip (dismissed):

src/target/riscv/riscv-013.c
```diff
@@ -2716,21 +2705,43 @@ static uint32_t sb_sbaccess(unsigned int size_bytes)
...
 
-static int sb_write_address(struct target *target, target_addr_t address,
-							bool ensure_success)
+static unsigned int get_sbaadress_reg_count(const struct target *target)
 {
 	RISCV013_INFO(info);
-	unsigned int sbasize = get_field(info->sbcs, DM_SBCS_SBASIZE);
+	const unsigned int sbasize = get_field(info->sbcs, DM_SBCS_SBASIZE);
+	assert(sbasize % 32 == 0);
+	return sbasize / 32;
```
JanMatCodasip: Detail: I would recommend to not assume that `sbasize` is divisible by 32 as there is no such requirement in the spec. I understand that it is not very likely to happen in practice, but still, it is IMO worth to count with such cases:

```suggestion
	assert(DIV_ROUND_UP(sbasize, 32) <= 4);  // if you like - for extra safety
	return DIV_ROUND_UP(sbasize, 32);
```

JanMatCodasip (commented):
Just one detail that I came across, otherwise this PR looks fine.

src/target/riscv/riscv-013.c
en-sc: Thanks for the catch!
Addressed. Didn't add the assertion, since IMHO it's better placed in `batch_fill_sb_write_address()` and it's already there:
https://github.com/riscv-collab/riscv-openocd/blob/6d4ad0033ed4e57c1b64998992dcc7a858d7ccd5/src/target/riscv/riscv-013.c#L2725

JanMatCodasip (approved):

MarekVCodasip (approved):

PR #1100, opened 2024-07-03
 target/riscv: single DMI accesses via batch

en-sc:
* Eliminates the use of VLA, which is prohibited by `doc/manual/style.txt`:
Link: https://github.com/riscv-collab/riscv-openocd/blob/c6bb902629924eb66aae2a08c0ab8654261c9d71/doc/manual/style.txt#L164-L166

* Unifies DMI access interface.

* Reduces code duplication.

Requires: #1099, #1083

### DIFF
--- a/src/target/riscv/batch.h
+++ b/src/target/riscv/batch.h
@@ -190,14 +190,32 @@ int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
 size_t riscv_batch_finished_scans(const struct riscv_batch *batch);
 
 /* Adds a DM register write to this batch. */
-void riscv_batch_add_dm_write(struct riscv_batch *batch, uint64_t address, uint32_t data,
+void riscv_batch_add_dmi_write(struct riscv_batch *batch, uint64_t address, uint32_t data,
 	bool read_back, enum riscv_scan_delay_class delay_class);
+
+static inline void
+riscv_batch_add_dm_write(struct riscv_batch *batch, uint64_t address, uint32_t data,
+	bool read_back, enum riscv_scan_delay_class delay_type)
+{
+	return riscv_batch_add_dmi_write(batch,
+			riscv_get_dmi_address(batch->target, address), data,
+			read_back, delay_type);
+}
 
 /* DM register reads must be handled in two parts: the first one schedules a read and
  * provides a key, the second one actually obtains the result of the read -
  * status (op) and the actual data. */
-size_t riscv_batch_add_dm_read(struct riscv_batch *batch, uint64_t address,
+size_t riscv_batch_add_dmi_read(struct riscv_batch *batch, uint64_t address,
 		enum riscv_scan_delay_class delay_class);
+
+static inline size_t
+riscv_batch_add_dm_read(struct riscv_batch *batch, uint64_t address,
+		enum riscv_scan_delay_class delay_type)
+{
+	return riscv_batch_add_dmi_read(batch,
+			riscv_get_dmi_address(batch->target, address), delay_type);
+}
+
 unsigned int riscv_batch_get_dmi_read_op(const struct riscv_batch *batch, size_t key);
 uint32_t riscv_batch_get_dmi_read_data(const struct riscv_batch *batch, size_t key);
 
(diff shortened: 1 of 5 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -731,29 +516,52 @@ static uint32_t riscv013_get_dmi_address(const struct target *target, uint32_t a
...
+	struct riscv_batch *batch = riscv_batch_alloc(target, 1);
+	riscv_batch_add_dm_read(batch, address, RISCV_DELAY_ABSTRACT_COMMAND);
 	dm->abstract_cmd_maybe_busy = true;
-	return dmi_read_exec(target, value, address + dm->base);
+	int res = batch_run_timeout(target, batch);
+	if (res == ERROR_OK && value)
+		*value = riscv_batch_get_dmi_read_data(batch, 0);
+	riscv_batch_free(batch);
+	return res;
+}
```
MarekVCodasip, line 550: I think this functionality (`dmi_read()` and `dm_read_exec()`) could further be unified by a function like `dmi_read_batch_delay(target, value, address, delay)`

What do you think?

MarekVCodasip (commented):
Thanks for the patch. This is a big improvement.

I have a suggestion for you to consider, please take a look.

src/target/riscv/riscv-013.c
en-sc: The issue is with `dm->abstract_cmd_maybe_busy = true;`. 
1. It can be either left in `dm_read_exec()`, which would create a possibility of `dm->abstract_cmd_maybe_busy` being left high after an error before `batch_run_timeout()` is called, so that there is no abstract command running on the target right now. This is not a big deal -- assuming an abstract command is running shouldn't break anything, but this is a possible misrepresentation of target's state that can be avoided. Moreover, the whole `get_dm()` related part would be duplicated so the diff will increase.
2. Alternatively, `dm->abstract_cmd_maybe_busy = true;` can be moved into the proposed `dmi_read_batch_delay()`. Hovewer, it would be placed there under an `if` on delay type. I believe this would not make the code simpler.

src/target/riscv/batch.h
```diff
@@ -190,14 +190,32 @@ int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
...
 
+static inline void
+riscv_batch_add_dm_write(struct riscv_batch *batch, uint64_t address, uint32_t data,
+	bool read_back, enum riscv_scan_delay_class delay_type)
+{
+	return riscv_batch_add_dmi_write(batch,
+			riscv_get_dmi_address(batch->target, address), data,
+			read_back, delay_type);
+}
+
```
JanMatCodasip, line 204: [Just a note - no change requested]

I tend to not use static inline functions in header files. For the sake of unified code style, I prefer the usual layout where possible: Prototype in header file and implementation in the corresponding .c file.

I'd also leave the optimizations for the compiler (avoid micro-optimization) unless there is a significant gain.

src/target/riscv/riscv-013.c
JanMatCodasip: I don't mind the separate functions as it is now. It looks clear enough to me.

Alternatively, `dmi_read_exec()` could be created so that the call graph has the same structure in all three cases and is very easy to comprehend for code readers:

- `dm_read_exec()` -> `dmi_read_exec()` -> batch functions
- `dm_read()` -> `dmi_read()` -> batch functions
- `dm_write()` -> `dmi_write()` -> batch functions

But that's really a detail (optional suggestion).

JanMatCodasip (dismissed):

src/target/riscv/riscv-013.c
en-sc: I'd like to leave it as-is.

Regarding `dmi_read_exec()` -- I can't imagine a scenario where it would be useful, outside `dm_read_exec()`, so would prefer not to add it.

src/target/riscv/batch.h
en-sc: My intention here is not to micro-optimize. The main point of declaring `_dm_` functions together their definitions is to highlight that the only difference from `_dmi_` counterparts is the address conversion.

src/target/riscv/riscv-013.c
JanMatCodasip: Ok, let's leave it as is.

src/target/riscv/batch.h
JanMatCodasip: Hmm, ok, so the reason for `static inline` here is to make it very clear that the `_dm_` function is almost the same thing (a simple wrapper) over the `_dmi_` function. 

I'd still prefer to not use static inline, but that's a detail and matter of personal taste.

So let's leave it as is.

JanMatCodasip (dismissed):

MarekVCodasip (dismissed):

JanMatCodasip (approved):

PR #1101, opened 2024-07-04
target/riscv: reg cache entry is initialized before access

en-sc:
* Register file examination is separated.
* Allow to access registers through cache as early as possible to re-use general register access interface and propely track state of the register.
* Reduces the number of operations: S0 and S1 are saved/restored only when needed (targets without abstract CSR access).

Change-Id: I2e205ae4e88733a5c792f8a35cf30325c68d96b2

### DIFF
--- a/src/target/riscv/riscv-013_reg.c
+++ b/src/target/riscv/riscv-013_reg.c
@@ -9,6 +9,7 @@
 #include "riscv_reg.h"
 #include "riscv_reg_impl.h"
 #include "riscv-013.h"
+#include "debug_defines.h"
 #include <helper/time_support.h>
 
 static int riscv013_reg_get(struct reg *reg)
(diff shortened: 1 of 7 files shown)


src/target/riscv/riscv-013_reg.c
```diff
@@ -92,26 +93,165 @@ static const struct reg_arch_type *riscv013_gdb_regno_reg_type(uint32_t regno)
...
+}
+
+/* As defined by RISC-V V extension specification:
+ * https://github.com/riscv/riscv-v-spec/blob/2f68ef7256d6ec53e4d2bd7cb12862f406d64e34/v-spec.adoc?plain=1#L67-L72 */
+#define VLEN_MAX 65536
+
+static int examine_vlenb(struct target *target)
+{
+	RISCV_INFO(r);
+	/* Reading "vlenb" requires "mstatus.vs" to be set, so "mstatus" should
```
MarekVCodasip:
```suggestion
	RISCV_INFO(r);
	
	/* Reading "vlenb" requires "mstatus.vs" to be set, so "mstatus" should
```
Detail: I believe it would help readability.

src/target/riscv/riscv-013_reg.c
```diff
@@ -92,26 +93,165 @@ static const struct reg_arch_type *riscv013_gdb_regno_reg_type(uint32_t regno)
...
+	if (vlenb_val > vlenb_max) {
+		LOG_TARGET_WARNING(target, "'vlenb == %" PRIu64
+				"' is greater than maximum allowed by specification (%u); vector register access won't work.",
+				vlenb_val, vlenb_max);
+		r->vlenb = 0;
+		return ERROR_OK;
+	}
+	assert(vlenb_max <= UINT_MAX);
+	r->vlenb = (unsigned int)vlenb_val;
+	LOG_TARGET_INFO(target, "Vector support with vlenb=%u", r->vlenb);
```
MarekVCodasip:
```suggestion
	r->vlenb = (unsigned int)vlenb_val;
	
	LOG_TARGET_INFO(target, "Vector support with vlenb=%u", r->vlenb);
```
Detail: I believe it would help readability.

src/target/riscv/riscv-013_reg.c
```diff
@@ -92,26 +93,165 @@ static const struct reg_arch_type *riscv013_gdb_regno_reg_type(uint32_t regno)
...
+	} else if (riscv_reg_get(target, &value, GDB_REGNO_MTOPEI) != ERROR_OK) {
+		LOG_TARGET_INFO(target, "S?aia detected without IMSIC");
+		r->mtopei_readable = false;
+	} else {
+		LOG_TARGET_INFO(target, "S?aia detected with IMSIC");
+	}
+	res = riscv_reg_impl_set_exist(target, GDB_REGNO_MTOPI, r->mtopi_readable);
+	if (res != ERROR_OK)
+		return res;
+	return riscv_reg_impl_set_exist(target, GDB_REGNO_MTOPEI, r->mtopei_readable);
```
MarekVCodasip:
```suggestion
		return res;
		
	return riscv_reg_impl_set_exist(target, GDB_REGNO_MTOPEI, r->mtopei_readable);
```
Detail: I believe it would help readability.

src/target/riscv/riscv-013_reg.c
```diff
@@ -92,26 +93,165 @@ static const struct reg_arch_type *riscv013_gdb_regno_reg_type(uint32_t regno)
...
+{
+	RISCV_INFO(r);
+	int res = init_cache_entry(target, GDB_REGNO_MISA);
+	if (res != ERROR_OK)
+		return res;
+
+	res = riscv_reg_get(target, &r->misa, GDB_REGNO_MISA);
+	if (res != ERROR_OK)
+		return res;
+	return ERROR_OK;
```
MarekVCodasip:
```suggestion
	RISCV_INFO(r);
	
	int res = init_cache_entry(target, GDB_REGNO_MISA);
	if (res != ERROR_OK)
		return res;

	res = riscv_reg_get(target, &r->misa, GDB_REGNO_MISA);
	if (res != ERROR_OK)
		return res;
		
	return ERROR_OK;
```
Detail: I believe it would help readability.

src/target/riscv/riscv-013_reg.c
```diff
@@ -92,26 +93,165 @@ static const struct reg_arch_type *riscv013_gdb_regno_reg_type(uint32_t regno)
...
+			GDB_REGNO_S0, /* size */ 64, AC_ACCESS_REGISTER_TRANSFER);
+	int res = riscv013_execute_abstract_command(target, command, &cmderr);
+	if (res == ERROR_OK) {
+		r->xlen = 64;
+		return ERROR_OK;
+	}
+	if (res == ERROR_TIMEOUT_REACHED)
 		return ERROR_FAIL;
+	r->xlen = 32;
+	return ERROR_OK;
```
MarekVCodasip:
```suggestion
	RISCV_INFO(r);
	unsigned int cmderr;
	
	const uint32_t command = riscv013_access_register_command(target,
			GDB_REGNO_S0, /* size */ 64, AC_ACCESS_REGISTER_TRANSFER);
	int res = riscv013_execute_abstract_command(target, command, &cmderr);
	if (res == ERROR_OK) {
		r->xlen = 64;
		return ERROR_OK;
	}
	if (res == ERROR_TIMEOUT_REACHED)
		return ERROR_FAIL;
	r->xlen = 32;
	
	return ERROR_OK;
```
Detail: I believe it would help readability.

src/target/riscv/riscv-013_reg.c
```diff
@@ -92,26 +93,165 @@ static const struct reg_arch_type *riscv013_gdb_regno_reg_type(uint32_t regno)
...
+	 * be accessible.*/
+	int res = init_cache_entry(target, GDB_REGNO_MSTATUS);
+	if (res != ERROR_OK)
+		return res;
+
+	res = init_cache_entry(target, GDB_REGNO_VLENB);
+	if (res != ERROR_OK)
+		return res;
+	uint64_t vlenb_val;
+	if (riscv_reg_get(target, &vlenb_val, GDB_REGNO_VLENB) != ERROR_OK) {
```
JanMatCodasip: Would it make sense to add an assertion inside `riscv_reg_get()` that the corresponding cache entry is initialized?

src/target/riscv/riscv-013_reg.c
```diff
@@ -92,26 +93,165 @@ static const struct reg_arch_type *riscv013_gdb_regno_reg_type(uint32_t regno)
...
 
-static int riscv013_init_reg(struct target *target, uint32_t regno)
+static int init_cache_entry(struct target *target, uint32_t regno)
 {
-	return riscv_reg_impl_init_one(target, regno, riscv013_gdb_regno_reg_type(regno));
+	return riscv_reg_impl_init_cache_entry(target, regno, riscv013_gdb_regno_reg_type(regno));
 }
 
-int riscv013_reg_init_all(struct target *target)
+static int examine_xlen(struct target *target)
```
JanMatCodasip, line 118: I wonder if `riscv013_reg_examine_all()` and its companion `examine_*` functions should rather go to `riscv-013.c`.

The reason is that `riscv-013_reg` module provides the register access (lower-level functions) but the examine logic is the user of it (higher-level algorithms). So to me they belong to a higher level module -- `riscv-013.c` or in future maybe `riscv-013_examine.c`.

@en-sc Please let me know what you think about it. Thanks.

JanMatCodasip (commented):
Just two remarks, otherwise it looks good.

MarekVCodasip (commented):
Thanks for the patch, I like the improvements.

I have found a few details concerning extra lines for clarity, please, take a look.

src/target/riscv/riscv-013_reg.c
en-sc: IMHO, `riscv013_reg_examine_all()` deals with initialization of the register cache (essentially a constructor of RISC-V register cache). It needs to know much more about the inner workings of register access on RISC-V then any other interface function. So I'd like to leave `riscv013_reg_examine_all()` in `riscv-013_reg.c`.

src/target/riscv/riscv-013_reg.c

src/target/riscv/riscv-013_reg.c

src/target/riscv/riscv-013_reg.c

src/target/riscv/riscv-013_reg.c

src/target/riscv/riscv-013_reg.c

src/target/riscv/riscv-013_reg.c

src/target/riscv/riscv_reg_impl.h
```diff
@@ -37,8 +37,18 @@ static inline bool riscv_reg_impl_is_initialized(const struct reg *reg)
...
 
 /** Initialize register. */
-int riscv_reg_impl_init_one(struct target *target, uint32_t regno,
-		const struct reg_arch_type *reg_type);
+int riscv_reg_impl_init_cache_entry(struct target *target, uint32_t regno,
+		bool exist, const struct reg_arch_type *reg_type);
+
+/**
+ * For most registers, returns whether they exist or not.
+ * For some registers the "exist" bit should be set explisitly.
```
aap-sc: typo. `**explisitly**

src/target/riscv/riscv-013_reg.c
JanMatCodasip: Ok, let's leave it as is.

JanMatCodasip (dismissed):

JanMatCodasip:
@MarekVCodasip Please, don't forget to close your review threads if you are fine with the resolution - thanks. ;-)

src/target/riscv/riscv_reg_impl.h
en-sc: Addressed.

JanMatCodasip (approved):

MarekVCodasip (approved):
LGTM

PR #1105, opened 2024-07-17
add support for 64-bit system profiling

yuriishutkin:
Hello riscv-collab!

Current profile implementation looks outdated and doesn't support 64-bit systems, so I've implemented this support for my purposes and want to share it.

To store 64-bit addresses I used the same uint32_t samples array do keep lower 32-bits of each sample address, and higher 32-bit of address is passed though a separate sample_address_hi32 pointer. The latter does not store 32 bit word for every sample but only stores a single uint32_t for all samples assuming that all samples fit into a single 32-bit slice and have equal higher 32 bits. Bigger slice will not fit into reasonable buckets array anyway.

I've also added start_address and end_address arguments to collection function arguments, so it can filter out samples that are not interesting and make sure that sample_address_hi32 calculated properly.

Please review the proposed change and accept it if it looks useful. I'm not sure if it should go to riscv repo or to the original openocd repo, I've not found the guidelines on how to choose between them.

### DIFF
--- a/src/target/cortex_m.c
+++ b/src/target/cortex_m.c
@@ -2202,8 +2202,9 @@ void cortex_m_deinit_target(struct target *target)
 	free(cortex_m);
 }
 
-int cortex_m_profiling(struct target *target, uint32_t *samples,
-			      uint32_t max_num_samples, uint32_t *num_samples, uint32_t seconds)
+int cortex_m_profiling(struct target *target, uint32_t *samples, uint32_t* sample_address_hi32,
+			bool with_range, uint64_t start_address, uint64_t end_address,
+			uint32_t max_num_samples, uint32_t *num_samples, uint32_t seconds)
 {
 	struct timeval timeout, now;
 	struct armv7m_common *armv7m = target_to_armv7m(target);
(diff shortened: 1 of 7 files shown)


TommyMurphyTM1234:
Why are there changes to `cortex_m.{c,h}` and `or1k.c` if this PR is for RISC-V support for the feature?

yuriishutkin:
> Why are there changes to `cortex_m.{c,h}` and `or1k.c` if this PR is for RISC-V support for the feature?

Hi, this is actually not only for riscv, I've changed default sample collection function that applies to all targets. I've changed the signature of this function, and cortex_m and or1k have its own collection functions that overwrite the default one, so their signatures are also changed.

TommyMurphyTM1234:
I suspect that such a change should probably be done upstream in the master OpenOCD project rather than here. 

- https://github.com/openocd-org/openocd/blob/master/HACKING

It will ultimately require acceptance there anyway so better to get it there first and then integrate it here via one of the semi-regular merges from upstream? E.g.:

- https://github.com/riscv-collab/riscv-openocd/pull/1094

yuriishutkin:
Thanks, I'll try that way.

JanMatCodasip:
@yuriishutkin Per the previous comment, can this pull request be closed?

Thanks.

yuriishutkin:
Yes, I have a pending CL on openocd repo for this.

PR #1109, opened 2024-08-05
target/riscv: sys bus v1 fix for sizes greater than 4

### LINKED ISSUE #1107: read_memory_bus_v1 does not work properly for sizes greater than 4
The issue is illustrated here:

https://github.com/riscv-collab/riscv-openocd/pull/1081/files#r1702661010

```
#7  0x000055555609d04c in buf_set_u32 (_buffer=0x5555569ca720 "\277", first=0, num=64, value=191)                           
    at /home/nikita/projects/openocd/src/helper/binarybuffer.h:43 
#8  0x00005555560b1d13 in read_memory_bus_v1 (target=0x5555569e5390, address=2147491969, size=8, count=2, 
    buffer=0x5555569ca720 "\277", increment=8) at ../openocd/src/target/riscv/riscv-013.c:3410
#9  0x00005555560b58c0 in read_memory (target=0x5555569e5390, address=2147491969, size=8, count=2, 
    buffer=0x5555569ca720 "\277", increment=8) at ../openocd/src/target/riscv/riscv-013.c:4397
#10 0x0000555555c9b5c8 in riscv_read_memory (target=0x5555569e5390, address=2147491969, size=8, count=2, 
    buffer=0x5555569ca720 "\277") at ../openocd/src/target/riscv/riscv.c:3088
#11 0x0000555555c2daee in target_read_memory (target=0x5555569e5390, address=2147491969, size=8, count=2, 
    buffer=0x5555569ca720 "\277") at ../openocd/src/target/target.c:1245
```

aap-sc:


en-sc (dismissed):
LGTM

src/target/riscv/riscv-013.c
```diff
@@ -3405,9 +3408,12 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
 
 			const size_t last_key = batch->read_keys_used - 1;
 			for (size_t k = 0; k <= last_key; ++k) {
-				sbvalue[k] = riscv_batch_get_dmi_read_data(batch,
-						last_key - k);
-				buf_set_u32(buffer + i * size + k * 4, 0, 8 * size, sbvalue[k]);
+				sbvalue[k] = riscv_batch_get_dmi_read_data(batch, last_key - k);
+				/* This can be written as: n_bits = MIN(32, 8 * size), but I think
+				 * we'd better to be explicit on how n_bits is calculated */
+				const size_t n_bits = (k != last_key) ? 32 : (8 * size - k * 4 * 8);
```
MarekVCodasip: Myself, I think the `MIN(32, 8 * size)` is fine to use.

MarekVCodasip (commented):
I would prefer if @JanMatCodasip takes a look on it as well. He should be able to next week.

src/target/riscv/riscv-013.c
aap-sc: I feel like MIN(32, 8 * size) is quite a hacky way to do this. This is the reason for the comment above. 

I mean:

- if we use MIN () call the code become somewhat difficult to read, since in order to understand it you need to know the surrounding context quite well
- if we use `8 * size - k * 4 * 8` the code becomes a little more verbose, however it becomes quite clear and most readers can understand what is going on.

This looks like a matter of personal preference and not a blocker, am I right? Alternatively I can maybe  re-organize the code and avoid the ternary operator  (by moving the last `buf_set_u32` out of the loop)

JanMatCodasip: Thank you for your opinions.

After reading the code, I'd be bit more inclined to use `MIN(32, 8 * size)` but I don't feel strongly. 

Please, feel free to use whichever option. 

(Note: Please don't wait for any further comments from me as I will be away until the start of September.)

JanMatCodasip (dismissed):

src/target/riscv/riscv-013.c
aap-sc: @MarekVCodasip , @en-sc - ~~could you take a looks once again? I've refactored the code a liittle bit. I use MIN(...) now, but moved the last iteration out of the loop. I think it's marginally better now~~. Nevermind. I just changed to MIN().

en-sc (approved):

src/target/riscv/riscv-013.c
aap-sc: @MarekVCodasip jentle ping ...

aap-sc: @MarekVCodasip ping once again)

JanMatCodasip (approved):

src/target/riscv/riscv-013.c
MarekVCodasip: Sorry for that.

MarekVCodasip (approved):

Issue #1108, opened 2024-08-05
There are cases when WPs disabled by `disable_triggers()` can not be re-enabled by `enable_triggers()`. 


en-sc:
Currently, `disable_triggers()` just goes through all the WPs and disables them one-by-one (the case when `r->manual_hwbp_set` is `false`):
https://github.com/riscv-collab/riscv-openocd/blob/5afed58fcdb67ff87842dad9c3ab7f73ca0c8cd8/src/target/riscv/riscv.c#L2487-L2499
`enable_triggers()` works accordingly:
https://github.com/riscv-collab/riscv-openocd/blob/5afed58fcdb67ff87842dad9c3ab7f73ca0c8cd8/src/target/riscv/riscv.c#L2526-L2536

Now consider the following situation:
* A target has 2 triggers. The 1-st trigger supports configurations required by WP 1, WP 2, WP 3. The 2-nd trigger only supports configuration required by WP-2.
* The users sets WP 1. WP 1 occupies the 1-st trigger.
* The user sets WP 2. WP 2 occupies the 2-nd trigger,
* The user removes WP 1, the 1-st trigger is free.
* The user sets WP 3. WP 3 occupies the 1-st trigger.
* The situation is as follows:
    * `taget->watchpoints` points to WP 2.
    * `taget->watchpoints->next` points to WP 3.
* `disable_triggers()` is called (e.g. during `step`).
* `enable_triggers()` will try to re-enable the disabled WP, but will fail, since WP 2 will occupy 1-st trigger and 2-nd triggerdoes not support WP 3.

en-sc:
Moreover, seems like the `state` array is redundant here -- all watchpoints are enabled/disabled at once.

JanMatCodasip:
It sounds like a reasonable fix would be to:

- <del>Keep track of watchpoint --> trigger slot mapping in `target->watchpoints` or similar data structure.</del>
  - This is already in `trigger_unique_id` in `struct riscv_info`.
- Only (temporarily) disable the watchpoints and do not remove them from the `target->watchpoints` structure in the process.
- When re-enabling watchpoints, ensure they are placed back into the exact "trigger slots" as originally.

PR #1111, opened 2024-08-07
target/riscv: manage triggers available to OpenOCD for internal use

en-sc:
Before the change, if the user wrote to any `tdata*` register, OpenOCD would sometimes start to disable all the triggers (by writing zeroes to `tdata1`) and re-enable them again (by witing all trigger registers to the values read before for each `tselect` value), e.g. on `step` (see `disable/enable_triggers()`).

There are a couple of issues with such approach:
1. RISC-V Debug Specification does not require custom register types to support re-enabling by such sequence of writes (e.g. some custom trigger type may require writing a custom CSR to enable it).
2. OpenOCD may still overwrite these triggers when a user asks to set a new WP.

This commit introduces `riscv reserve_trigger ...` command to explicitly mark the triggers OpenOCD should not touch.

Such approach allows to separate management of custom triggers and offload it onto the user (e.g. disable/enable such triggers by setting up an event handler on `step`-related events).

Change-Id: I3339000445185ab221368442a070f412bf44bfab

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -11524,6 +11524,21 @@ as in the mie CSR (defined in the RISC-V Privileged Spec).
 For details on this trigger type, see the RISC-V Debug Specification.
 @end deffn
 
+@deffn {Command} {riscv reserve_trigger} [index @option{on|off}]
+Manages the set of reserved triggers. Reserving a trigger results in OpenOCD
+not using it internally (e.g. skipping it when setting a watchpoint or a
+hardware breakpoint), so that the user or the application has unfettered
+control over the trigger. By default there are no reserved triggers.
+
+@enumerate
+@item @var{index} specifies the index of a trigger to reserve or free up.
+@item The second argument specifies whether the trigger should be reserved
+(@var{on}) or a prior reservation cancelled (@var{off}).
+@item If called without parameters, returns indices of reserved triggers.
+@end enumerate
+
+@end deffn
+
 @deffn {Command} {riscv itrigger clear}
 Clear the type 4 trigger that was set using @command{riscv itrigger set}.
 @end deffn
--- a/src/target/riscv/riscv-013_reg.c
+++ b/src/target/riscv/riscv-013_reg.c
@@ -44,7 +44,6 @@ static int riscv013_reg_get(struct reg *reg)
 static int riscv013_reg_set(struct reg *reg, uint8_t *buf)
 {
 	struct target *target = riscv_reg_impl_get_target(reg);
-	RISCV_INFO(r);
 
 	char *str = buf_to_hex_str(buf, reg->size);
 	LOG_TARGET_DEBUG(target, "Write 0x%s to %s (valid=%d).", str, reg->name,
@@ -56,17 +55,6 @@ static int riscv013_reg_set(struct reg *reg, uint8_t *buf)
 			riscv_supports_extension(target, 'E') &&
 			buf_get_u64(buf, 0, reg->size) == 0)
 		return ERROR_OK;
-
-	if (reg->number == GDB_REGNO_TDATA1 ||
-			reg->number == GDB_REGNO_TDATA2) {
-		r->manual_hwbp_set = true;
-		/* When enumerating triggers, we clear any triggers with DMODE set,
-		 * assuming they were left over from a previous debug session. So make
-		 * sure that is done before a user might be setting their own triggers.
-		 */
-		if (riscv_enumerate_triggers(target) != ERROR_OK)
-			return ERROR_FAIL;
-	}
 
 	if (reg->number >= GDB_REGNO_V0 && reg->number <= GDB_REGNO_V31) {
 		if (riscv013_set_register_buf(target, reg->number, buf) != ERROR_OK)
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -527,6 +527,8 @@ static void riscv_deinit_target(struct target *target)
 	if (!info)
 		return;
 
+	free(info->reserved_triggers);
+
 	range_list_t *entry, *tmp;
 	list_for_each_entry_safe(entry, tmp, &info->hide_csr, list) {
 		free(entry->name);
@@ -620,6 +622,15 @@ static int find_first_trigger_by_id(struct target *target, int unique_id)
 static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdata1, riscv_reg_t tdata2,
 	riscv_reg_t tdata1_ignore_mask)
 {
+	RISCV_INFO(r);
+	assert(r->reserved_triggers);
+	assert(idx < r->trigger_count);
+	if (r->reserved_triggers[idx]) {
+		LOG_TARGET_DEBUG(target,
+				"Trigger %u is reserved by 'reserve_trigger' command.", idx);
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	}
+
 	riscv_reg_t tdata1_rb, tdata2_rb;
 	// Select which trigger to use
 	if (riscv_reg_set(target, GDB_REGNO_TSELECT, idx) != ERROR_OK)
@@ -2453,87 +2464,48 @@ static int riscv_deassert_reset(struct target *target)
 	return tt->deassert_reset(target);
 }
 
-/* state must be riscv_reg_t state[RISCV_MAX_HWBPS] = {0}; */
-static int disable_triggers(struct target *target, riscv_reg_t *state)
+/* "wp_is_set" array must have at least "r->trigger_count" items. */
+static int disable_watchpoints(struct target *target, bool *wp_is_set)
 {
 	RISCV_INFO(r);
-
 	LOG_TARGET_DEBUG(target, "Disabling triggers.");
 
-	if (riscv_enumerate_triggers(target) != ERROR_OK)
-		return ERROR_FAIL;
-
-	if (r->manual_hwbp_set) {
-		/* Look at every trigger that may have been set. */
-		riscv_reg_t tselect;
-		if (riscv_reg_get(target, &tselect, GDB_REGNO_TSELECT) != ERROR_OK)
-			return ERROR_FAIL;
-		for (unsigned int t = 0; t < r->trigger_count; t++) {
-			if (riscv_reg_set(target, GDB_REGNO_TSELECT, t) != ERROR_OK)
+	/* TODO: The algorithm is flawed and may result in a situation described in
+	 * https://github.com/riscv-collab/riscv-openocd/issues/1108
+	 */
+	memset(wp_is_set, false, r->trigger_count);
+	struct watchpoint *watchpoint = target->watchpoints;
+	int i = 0;
+	while (watchpoint) {
+		LOG_TARGET_DEBUG(target, "Watchpoint %" PRIu32 ": set=%s",
+				watchpoint->unique_id,
+				wp_is_set[i] ? "true" : "false");
+		wp_is_set[i] = watchpoint->is_set;
+		if (watchpoint->is_set) {
+			if (riscv_remove_watchpoint(target, watchpoint) != ERROR_OK)
 				return ERROR_FAIL;
-			riscv_reg_t tdata1;
-			if (riscv_reg_get(target, &tdata1, GDB_REGNO_TDATA1) != ERROR_OK)
-				return ERROR_FAIL;
-			if (tdata1 & CSR_TDATA1_DMODE(riscv_xlen(target))) {
-				state[t] = tdata1;
-				if (riscv_reg_set(target, GDB_REGNO_TDATA1, 0) != ERROR_OK)
-					return ERROR_FAIL;
-			}
 		}
-		if (riscv_reg_set(target, GDB_REGNO_TSELECT, tselect) != ERROR_OK)
-			return ERROR_FAIL;
-
-	} else {
-		/* Just go through the triggers we manage. */
-		struct watchpoint *watchpoint = target->watchpoints;
-		int i = 0;
-		while (watchpoint) {
-			LOG_TARGET_DEBUG(target, "Watchpoint %d: set=%d", i, watchpoint->is_set);
-			state[i] = watchpoint->is_set;
-			if (watchpoint->is_set) {
-				if (riscv_remove_watchpoint(target, watchpoint) != ERROR_OK)
-					return ERROR_FAIL;
-			}
-			watchpoint = watchpoint->next;
-			i++;
-		}
+		watchpoint = watchpoint->next;
+		i++;
 	}
 
 	return ERROR_OK;
 }
 
-static int enable_triggers(struct target *target, riscv_reg_t *state)
+static int enable_watchpoints(struct target *target, bool *wp_is_set)
 {
-	RISCV_INFO(r);
-
-	if (r->manual_hwbp_set) {
-		/* Look at every trigger that may have been set. */
-		riscv_reg_t tselect;
-		if (riscv_reg_get(target, &tselect, GDB_REGNO_TSELECT) != ERROR_OK)
-			return ERROR_FAIL;
-		for (unsigned int t = 0; t < r->trigger_count; t++) {
-			if (state[t] != 0) {
-				if (riscv_reg_set(target, GDB_REGNO_TSELECT, t) != ERROR_OK)
-					return ERROR_FAIL;
-				if (riscv_reg_set(target, GDB_REGNO_TDATA1, state[t]) != ERROR_OK)
-					return ERROR_FAIL;
-			}
+	struct watchpoint *watchpoint = target->watchpoints;
+	int i = 0;
+	while (watchpoint) {
+		LOG_TARGET_DEBUG(target, "Watchpoint %" PRIu32
+				": %s to be re-enabled.", watchpoint->unique_id,
+				wp_is_set[i] ? "needs " : "does not need");
+		if (wp_is_set[i]) {
+			if (riscv_add_watchpoint(target, watchpoint) != ERROR_OK)
+				return ERROR_FAIL;
 		}
-		if (riscv_reg_set(target, GDB_REGNO_TSELECT, tselect) != ERROR_OK)
-			return ERROR_FAIL;
-
-	} else {
-		struct watchpoint *watchpoint = target->watchpoints;
-		int i = 0;
-		while (watchpoint) {
-			LOG_TARGET_DEBUG(target, "Watchpoint %d: cleared=%" PRId64, i, state[i]);
-			if (state[i]) {
-				if (riscv_add_watchpoint(target, watchpoint) != ERROR_OK)
-					return ERROR_FAIL;
-			}
-			watchpoint = watchpoint->next;
-			i++;
-		}
+		watchpoint = watchpoint->next;
+		i++;
 	}
 
 	return ERROR_OK;
@@ -3781,9 +3753,16 @@ static int riscv_openocd_step_impl(struct target *target, int current,
 			return ERROR_FAIL;
 	}
 
-	riscv_reg_t trigger_state[RISCV_MAX_HWBPS] = {0};
-	if (disable_triggers(target, trigger_state) != ERROR_OK)
+	if (riscv_enumerate_triggers(target) != ERROR_OK)
 		return ERROR_FAIL;
+
+	RISCV_INFO(r);
+	bool *wps_to_enable = calloc(sizeof(*wps_to_enable), r->trigger_count);
+	if (disable_watchpoints(target, wps_to_enable) != ERROR_OK) {
+		LOG_TARGET_ERROR(target, "Failed to temporarily disable "
+				"watchpoints before single-step.");
+		return ERROR_FAIL;
+	}
 
 	bool success = true;
 	uint64_t current_mstatus;
@@ -3814,9 +3793,10 @@ static int riscv_openocd_step_impl(struct target *target, int current,
 		}
 
 _exit:
-	if (enable_triggers(target, trigger_state) != ERROR_OK) {
+	if (enable_watchpoints(target, wps_to_enable) != ERROR_OK) {
 		success = false;
-		LOG_TARGET_ERROR(target, "Unable to enable triggers.");
+		LOG_TARGET_ERROR(target, "Failed to re-enable watchpoints "
+				"after single-step.");
 	}
 
 	if (breakpoint && (riscv_add_breakpoint(target, breakpoint) != ERROR_OK)) {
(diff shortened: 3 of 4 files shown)


en-sc:
This one needs a patch to `riscv-tests`: https://github.com/riscv-software-src/riscv-tests/pull/575

src/target/riscv/riscv-013_reg.c
```diff
@@ -57,17 +56,6 @@ static int riscv013_reg_set(struct reg *reg, uint8_t *buf)
...
-			reg->number == GDB_REGNO_TDATA2) {
-		r->manual_hwbp_set = true;
-		/* When enumerating triggers, we clear any triggers with DMODE set,
-		 * assuming they were left over from a previous debug session. So make
-		 * sure that is done before a user might be setting their own triggers.
-		 */
-		if (riscv_enumerate_triggers(target) != ERROR_OK)
-			return ERROR_FAIL;
-	}
-
```
MarekVCodasip, line 70: I think we should reserve the trigger (together with an INFO or a WARNING print if it isn't already reserved) if the user touches it via `reg_set()`

Trough, this raises a question if a watchpoint is set on said trigger, if we should remove it from memory.

What do others think?

src/target/riscv/riscv.c
```diff
@@ -5287,6 +5304,14 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.usage = "[('eq'|'napot'|'ge_lt'|'all') ('wp'|'none')]",
 		.help = "Control whether OpenOCD is allowed to use certain RISC-V trigger features for watchpoints."
 	},
+	{
+		.name = "reserve_trigger",
+		.handler = handle_reserve_trigger,
+		/* TODO: Move this to COMMAND_ANY */
```
MarekVCodasip, line 5324: Is there a reason this has not been done yet?

MarekVCodasip (commented):
Thanks for the patch, but I do think at least one of the changes needs further discussion.

src/target/riscv/riscv.c
en-sc: To reserve a trigger when a trigger module is not available or when the number of available triggers is not enough for the given index to be valid seems like an error.
Therefore, `handle_reserve_trigger` calls `riscv_enumerate_triggers()` which is only available after `examine()` (i.e. after TCL's `init` is called, i.e. in `COMMAND_EXEC` mode).

This can be mitigated by adding RISC-V specific target configure options that will allow the user to specify the number of available triggers in `COMMAND_CONFIG` mode, making `examine` just check that the configured parameter is correct (or even do nothing at all).

src/target/riscv/riscv-013_reg.c
en-sc: I'd like to avoid it.
1. This will require reading `tselect` to know which trigger to reserve. IMHO `reg` command should strive to minimize the number of operations it does.
2. TBH, I think Trigger Module registers should probably be marked as hidden and should not be accessible to user  by default.

aap-sc: > What do others think?

1. My vote is to always print big ugly error message when user tries to access trigger module registers explicitly. 
2. I would go as far as to render trigger module registers as unavailable to the user by default.

Otherwise it may create way too much hassle because of some obscure corner case when a user wants to mess up with the trigger module while having bp/wp set. I don't think any sane person want to do that. If user is crazy enough to mess with this stuff - he can unlock them an do whatever she/he wants (without any baby-sitting from OpenOCD side).

That being said I would recommend to file an issue for this functionality and merge this patch as is, since in my opinion this patch makes things a little bit better compared to what we have now.

doc/openocd.texi
```diff
@@ -11561,6 +11561,21 @@ riscv exec_progbuf 0x0330000f 0x0000100f
...
+
+@enumerate
+@item @var{index} specifies the index of a trigger to reserve or free up.
+@item The optional second argument specifies whether the trigger should be
+reserved. Defaults to @option{on}.
+@item The command always returns indices of reserved triggers.
+@end enumerate
+
+@end deffn
+
```
JanMatCodasip: Please, move the newly added content into the proper subsection -- `RISC-V Trigger Commands` (that is, just few lines up within this same file).

src/target/riscv/riscv.c
```diff
@@ -5031,6 +5005,49 @@ COMMAND_HANDLER(riscv_set_enable_trigger_feature)
...
+	RISCV_INFO(r);
+	if (t >= r->trigger_count) {
+		command_print(CMD, "Error: trigger %" PRIu64
+				" does not exist, there are only %u triggers on the target.",
+				t, r->trigger_count);
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	}
+	bool reserve = true;
+	if (CMD_ARGC == 2)
+		COMMAND_PARSE_ON_OFF(CMD_ARGV[1], reserve);
```

I would recommend to make the second argument (`on|off`) mandatory. The reason is to reduce the variability, to be more explicit and avoid having two different command variants that mean the same thing.

src/target/riscv/riscv.c
JanMatCodasip: For now, the current solution looks fine to me: only allow `reserve_trigger` in runtime, that is when the number of available triggers is known.

Should this ever become a limitation to someone, the implementation can be extended:
- remember the indexes of reserved triggers during configuration, and
- validate them later after target examination.

But, IMO, it does not seem worth doing at the moment.

src/target/riscv/riscv.c
```diff
@@ -5031,6 +5005,49 @@ COMMAND_HANDLER(riscv_set_enable_trigger_feature)
...
+	r->reserved_triggers[t] = reserve;
```
JanMatCodasip: What if someone tries to reserve a trigger that is already used by an OpenOCD's breakpoint or watchpoint?

I would prefer to detect that situation and print an error.

doc/openocd.texi
```diff
@@ -11561,6 +11561,21 @@ riscv exec_progbuf 0x0330000f 0x0000100f
 riscv exec_progbuf 0x94a20405
 @end example
 
+@deffn {Command} {riscv reserve_trigger} [index [@option{on|off}]]
+Manages the set of reserved triggers. Reserving a trigger results in OpenOCD
+not using it internally (e.g. skipping it when setting a watchpoint or a
+hardware breakpoint), so that the user or the application being run has
+unfettered control over the trigger. By default there are no reserved triggers.
```
JanMatCodasip: If you like, this can be added: `If called without parameters, the current list of reserved triggers is reported.`

src/target/riscv/riscv-013_reg.c
JanMatCodasip: I am happy to see the `manual_hwbp_set` heuristics removed, since it silently changed the behavior of enabling/disabling triggers when the user touched any of the trigger registers.

> I think we should reserve the trigger if (...) What do others think?

My preference would be to keep the trigger CSRs normally visible to the user, and don't perform any extra action if the user modifies any of the trigger CSRs. OpenOCD is a low-level debug tool, and as such it gives the user generally a lot of freedom about what it can do, and at the same time it requires cautious use, otherwise things may go awry.

I'd apply the same logic to the trigger manipulation, and not try to make OpenOCD any more "clever" - just do what the user says. :)

JanMatCodasip (commented):
Overall, I like the `reserve_trigger` feature. Please, if you can take a look at my comments. Thank you.

doc/openocd.texi
en-sc: Addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

doc/openocd.texi
en-sc: IMHO the mention in the list is a bit more informative:
https://github.com/riscv-collab/riscv-openocd/blob/9a935fcd5e1cd906c96281fbff762b4b4938f22f/doc/openocd.texi#L11574

src/target/riscv/riscv.c
```diff
@@ -5031,6 +5005,51 @@ COMMAND_HANDLER(riscv_set_enable_trigger_feature)
...
+	RISCV_INFO(r);
+	if (t >= r->trigger_count) {
+		command_print(CMD, "Error: trigger %" PRIu64
+				" does not exist, there are only %u triggers on the target.",
+				t, r->trigger_count);
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	}
+	if (r->trigger_unique_id[t] != -1) {
+		command_print(CMD, "Error: trigger %" PRIu64
+				" is already in use.", t);
```
JanMatCodasip: Optional detail:

I tried to make the error message even more clear.

```suggestion
		command_print(CMD, "Error: trigger with index %" PRIu64
				" is already in use, cannot reserve it.", t);
```

src/target/riscv/riscv.c
```diff
@@ -5031,6 +5005,51 @@ COMMAND_HANDLER(riscv_set_enable_trigger_feature)
...
+
+	if (riscv_enumerate_triggers(target) != ERROR_OK)
+		return ERROR_FAIL;
+	RISCV_INFO(r);
+	if (t >= r->trigger_count) {
+		command_print(CMD, "Error: trigger %" PRIu64
+				" does not exist, there are only %u triggers on the target.",
+				t, r->trigger_count);
+		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
+	}
```
JanMatCodasip: Optional detail:

I tried to make the trigger index more clear to the user - the 0..N-1 range. 

```suggestion
	if (r->trigger_count == 0) {
		command_print(CMD, "Error: There are no triggers in the target.");
		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
	}
	if (t >= r->trigger_count) {
		command_print(CMD, "Error: trigger with index %" PRIu64
				" does not exist. There are only %u triggers on the target (with indexes 0 .. %u).",
				t, r->trigger_count, r->trigger_count - 1);
		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
	}
```

doc/openocd.texi
JanMatCodasip: Thank you for pointing it out - I missed that part.

I'd recommend to only print the list of reserved triggers when the command is called without parameters.

If the parameters are provided, I'd prefer to just perform the operation. If successful, do not print anything, as it is common for *nix commands. Most (but sadly not all) OpenOCD TCL commands already follow this convention, e.g. `bp`, `wp` or `set_reg`.

JanMatCodasip (commented):
Ah, I had few review comments that I thought I submitted but I forgot. I'm sorry and sending them now.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

doc/openocd.texi
en-sc: Addressed.

src/target/riscv/riscv.h
```diff
@@ -272,9 +272,7 @@ struct riscv_info {
 	struct reg_data_type_union vector_union;
 	struct reg_data_type type_vector;
 
-	/* Set when trigger registers are changed by the user. This indicates we need
-	 * to beware that we may hit a trigger that we didn't realize had been set. */
-	bool manual_hwbp_set;
+	bool *reserved_triggers;
```
JanMatCodasip, line 275: Just a detail that I noticed - for your consideration:

If you like, the `reserved_triggers` array can be allocated statically:

```suggestion
	bool reserved_triggers[RISCV_MAX_TRIGGERS];
```

The advantages would be:

- Consistency -- the same approach is already used for `trigger_info` in `struct riscv_info`.
- No need for malloc and free - one less thing to worry about and reduced area for coding mistakes.

src/target/riscv/riscv.c
```diff
@@ -620,6 +622,15 @@ static int find_first_trigger_by_id(struct target *target, int unique_id)
 static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdata1, riscv_reg_t tdata2,
 	riscv_reg_t tdata1_ignore_mask)
 {
+	RISCV_INFO(r);
+	assert(r->reserved_triggers);
```
JanMatCodasip, line 626: This assertion will become unnecessary if `reserved_triggers` are allocated statically.

doc/openocd.texi
```diff
@@ -11524,6 +11524,20 @@ as in the mie CSR (defined in the RISC-V Privileged Spec).
...
+
+@enumerate
+@item @var{index} specifies the index of a trigger to reserve or free up.
+@item The second argument specifies whether the trigger should be reserved.
```
JanMatCodasip: I have tried to make the command documentation even clearer:

```suggestion
@item The second argument specifies whether the trigger should be reserved (@var{on}) or a prior reservation cancelled (@var{off}).
```

doc/openocd.texi
```diff
@@ -11524,6 +11524,20 @@ as in the mie CSR (defined in the RISC-V Privileged Spec).
 For details on this trigger type, see the RISC-V Debug Specification.
 @end deffn
 
+@deffn {Command} {riscv reserve_trigger} [index [@option{on|off}]]
+Manages the set of reserved triggers. Reserving a trigger results in OpenOCD
+not using it internally (e.g. skipping it when setting a watchpoint or a
+hardware breakpoint), so that the user or the application being run has
```
JanMatCodasip: Detail: "being run" can be dropped which makes the sentence easier to read and no meaning is lost.

```suggestion
hardware breakpoint), so that the user or the application has
```

src/target/riscv/riscv.c
```diff
@@ -2453,87 +2464,46 @@ static int riscv_deassert_reset(struct target *target)
 	return tt->deassert_reset(target);
 }
 
-/* state must be riscv_reg_t state[RISCV_MAX_HWBPS] = {0}; */
-static int disable_triggers(struct target *target, riscv_reg_t *state)
+/* state must be riscv_reg_t state[r->trigger_count] = {false}; */
```
JanMatCodasip: Have you meant this?

```suggestion
/* wp_is_set array must have at least "r->trigger_count" items. */
```
Also I'd prefer to zero-initialize the array inside disable_watchpoints(), as opposed to relying on the caller to do it.  This way the likelihood of coding mistakes would be reduced a little (defensive programming).

```
static int disable_watchpoints(struct target *target, bool *wp_is_set)
{
	LOG_TARGET_DEBUG(target, "Disabling triggers.");

	/* Initialize all wp_is_set items to false */
	for (unsigned int k; k < r->trigger_count; k++)
		wp_is_set[k] = false;

	// ...
```

src/target/riscv/riscv.c
```diff
@@ -2453,87 +2464,46 @@ static int riscv_deassert_reset(struct target *target)
...
-			if (state[i]) {
-				if (riscv_add_watchpoint(target, watchpoint) != ERROR_OK)
-					return ERROR_FAIL;
-			}
-			watchpoint = watchpoint->next;
-			i++;
+	struct watchpoint *watchpoint = target->watchpoints;
+	int i = 0;
+	while (watchpoint) {
+		LOG_TARGET_DEBUG(target, "Watchpoint %" PRIu32 ": cleared=%s",
```
JanMatCodasip: If you'd like: This should hopefully make it clearer to the log reader what is actually happening:

```suggestion
		LOG_TARGET_DEBUG(target, "Watchpoint %" PRIu32 ": needs_re-enable=%s",
```

src/target/riscv/riscv.c
```diff
@@ -3814,7 +3788,7 @@ static int riscv_openocd_step_impl(struct target *target, int current,
 		}
 
 _exit:
-	if (enable_triggers(target, trigger_state) != ERROR_OK) {
+	if (enable_watchpoints(target, wps_to_enable) != ERROR_OK) {
 		success = false;
 		LOG_TARGET_ERROR(target, "Unable to enable triggers.");
```
JanMatCodasip: If you like: I have tried to make the error message even more clear:

```suggestion
		LOG_TARGET_ERROR(target, "Failed to re-enable watchpoint triggers after single-step.");
```

src/target/riscv/riscv.c
```diff
@@ -3781,8 +3751,12 @@ static int riscv_openocd_step_impl(struct target *target, int current,
...
 
-	riscv_reg_t trigger_state[RISCV_MAX_HWBPS] = {0};
-	if (disable_triggers(target, trigger_state) != ERROR_OK)
+	if (riscv_enumerate_triggers(target) != ERROR_OK)
+		return ERROR_FAIL;
+
+	RISCV_INFO(r);
+	bool *wps_to_enable = calloc(sizeof(*wps_to_enable), r->trigger_count);
+	if (disable_watchpoints(target, wps_to_enable) != ERROR_OK)
 		return ERROR_FAIL;
```
JanMatCodasip: I believe it would be good to also display an error message in case disable_watchpoints() fails:

```suggestion
	if (disable_watchpoints(target, wps_to_enable) != ERROR_OK) {
		LOG_TARGET_ERROR(target, "Failed to temporarily disable watchpoint triggers before single-step.");
		return ERROR_FAIL;
	}
```

src/target/riscv/riscv.c
```diff
@@ -3781,8 +3751,12 @@ static int riscv_openocd_step_impl(struct target *target, int current,
 			return ERROR_FAIL;
```
JanMatCodasip, line 3753: I'm wondering if this algorithm for temporary disabling of breakpoints suffers from the same issue as the watchpoint disabling algorithm, as described here: https://github.com/riscv-collab/riscv-openocd/issues/1108

My understanding is that yes -- the same issue is there, too.

doc/openocd.texi
```diff
@@ -11524,6 +11524,20 @@ as in the mie CSR (defined in the RISC-V Privileged Spec).
 For details on this trigger type, see the RISC-V Debug Specification.
 @end deffn
 
+@deffn {Command} {riscv reserve_trigger} [index [@option{on|off}]]
```
JanMatCodasip: Looking at the code of `COMMAND_HANDLER(handle_reserve_trigger)`, the second argument (`on|off`) is actually mandatory:

```suggestion
@deffn {Command} {riscv reserve_trigger} [index @option{on|off}]
```

src/target/riscv/riscv.c
```diff
@@ -5031,6 +5005,56 @@ COMMAND_HANDLER(riscv_set_enable_trigger_feature)
...
+{
+	RISCV_INFO(r);
+	if (riscv_enumerate_triggers(target) != ERROR_OK)
+		return ERROR_FAIL;
+	const char *separator = "";
+	for (riscv_reg_t t = 0; t < r->trigger_count; ++t)
+		if (r->reserved_triggers[t]) {
+			command_print_sameline(CMD, "%s%" PRIu64, separator, t);
+			separator = " ";
+		}
```
JanMatCodasip: Detail: If checkpatch allows it, I'd wrap the for-loop body into another pair of curly braces.

```suggestion
	for (riscv_reg_t t = 0; t < r->trigger_count; ++t) {
		if (r->reserved_triggers[t]) {
			command_print_sameline(CMD, "%s%" PRIu64, separator, t);
			separator = " ";
		}
	}
```

src/target/riscv/riscv.c
```diff
@@ -5287,6 +5311,14 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.usage = "[('eq'|'napot'|'ge_lt'|'all') ('wp'|'none')]",
 		.help = "Control whether OpenOCD is allowed to use certain RISC-V trigger features for watchpoints."
 	},
+	{
+		.name = "reserve_trigger",
+		.handler = handle_reserve_trigger,
+		/* TODO: Move this to COMMAND_ANY */
+		.mode = COMMAND_EXEC,
+		.usage = "[index [('on'|'off')]]",
```
```suggestion
		.usage = "[index ('on'|'off')]",
```

src/target/riscv/riscv.c
```diff
@@ -5287,6 +5311,14 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
+		.help = "Controlls which triggers can be used by OpenOCD.",
```
JanMatCodasip: I've fixed a typo ("controls") and tried to reword the sentence to make it even more clear.

```suggestion
		.help = "Controls which RISC-V triggers shall not be used (not touched anyhow) by OpenOCD.",
```

JanMatCodasip (commented):
Overall, I like the `reserve_triggers` feature very much. This merge request looks very close to completion. 

When I read the code again, I made some minor comments and suggestions. If you can please look at them. I don't expect to have any further review findings than what I posted so far.

Thank you.

src/target/riscv/riscv.h
en-sc: `RISCV_MAX_TRIGGERS == 32`  and `RISCV_MAX_HWBPS == 16`  seem like arbitrary limitations to me.
AFAIK, the specification supports up to `2^XLEN` triggers (limited by `tselect`).
Therefore, I'd like to remove this constants in the future.

doc/openocd.texi
en-sc: Addressed.

doc/openocd.texi
en-sc: Addressed.

doc/openocd.texi
en-sc: Addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: I don't think so.
The thing is -- only one breakpoint is removed (the one that was hit) (this will probably result in issues once context-aware BPs are supported). So there is no issue of re-arrangement -- the Bp will be placed on the same trigger if the issue with WPs is fixed and the order of operations is as follows:
1. remove BP
2. remove WPs
3. restore WPs (if the issue is fixed, WPs occupy the same triggers they used to before removal)
4. restore BP

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: Addressed. Changed to make a bit shorter. Please, take a look.

src/target/riscv/riscv.h
JanMatCodasip: That's a fair point, so let's leave it as is. (Although I believe that hardware having so many triggers would be rare.)

src/target/riscv/riscv.c
JanMatCodasip: You're right. I mis-read the code and thought all breakpoints would be removed, which is not true.

Thanks for explaining the situation to me.

src/target/riscv/riscv.c
JanMatCodasip: Looks good and very clear - thanks.

JanMatCodasip (approved):

PR #1112, opened 2024-08-09
 target/riscv: check `misa` value before reporting

en-sc:
Currently, during register file examination:
1. A read of an XPR is attempted via 64-bit abstract access.
2. If such a read fails (e.g. connection unstable) XLEN is assumed to be
   32.
3. Then `misa` is read. Since `misa` is a CSR and it may be only
   readable via program buffer, `s0` should be readable beforehand (at
   least some assumption about `xlen` should be made).
4. Before the commit, the `misa.mxl` field was not checked against
   `xlen`, therefore erroneous info may have been reported to the user.
   Moreover, the `examine()` would pass indicating no error at all.
5. After the commit, `misa.mxl` is checked against `xlen` value.

### DIFF
--- a/src/target/riscv/riscv-013_reg.c
+++ b/src/target/riscv/riscv-013_reg.c
@@ -5,6 +5,7 @@
 #endif
 
 #include "riscv-013_reg.h"
+#include "field_helpers.h"
 
 #include "riscv_reg.h"
 #include "riscv_reg_impl.h"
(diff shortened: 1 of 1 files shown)


en-sc:
Depends on #1101

src/target/riscv/riscv-013_reg.c
```diff
@@ -173,6 +174,64 @@ static int examine_vlenb(struct target *target)
...
+
+	if (r->misa == 0) {
+		LOG_TARGET_WARNING(target, "'misa' register is read as zero."
+				"OpenOCD will not be able to determine some hart's capabilities.");
+		return ERROR_OK;
+	}
+	const unsigned int xlen = riscv_xlen(target);
+	assert(xlen <= sizeof(riscv_reg_t) * CHAR_BIT);
+	assert(xlen >= 2);
+	riscv_reg_t misa_mxl_field = (riscv_reg_t)0x3 << (xlen - 2);
```
JanMatCodasip: Details really:

- This can be const
- If you like, it can be called "mask" instead of "field" for more clarity

`const riscv_reg_t misa_mxl_mask = (riscv_reg_t)0x3 << (xlen - 2);`

src/target/riscv/riscv-013_reg.c
```diff
@@ -85,33 +87,247 @@ static int riscv013_reg_set(struct reg *reg, uint8_t *buf)
...
+	const unsigned int mxl = get_field(r->misa, misa_mxl_field);
+	/* RISC-V Debug Spec. requires "DXLEN" to be the widest. */
+	if (mxl == MISA_MXL_INVALID || xlen_to_mxl(xlen) < mxl) {
+		LOG_TARGET_ERROR(target,
+				"'misa.mxl' field's value of %u does not "
+				" coincide with 'xlen' of %u.",
+				mxl, xlen);
+		return ERROR_FAIL;
+	}
```
JanMatCodasip: 1) If you like, the situation could be made more clear to the reader by producing different error messages (shown below).

2) Also it seems more natural to convert MXL to MXLEN and then make the comparison of the *XLEN values. That's because the privileged spec says `XLEN is never greater than MXLEN`.

3) While making changes here, it may be good to check for MXLEN = 128. If detected, report an error to the user, saying that OpenOCD currently does not support it. 

```suggestion
	/* RISC-V Debug Spec. requires "DXLEN" to be the widest. */
	if (mxl == MISA_MXL_INVALID) {
		LOG_TARGET_ERROR(target, "Invalid value of MXL: misa.mxl == 0");
		return ERROR_FAIL;
	}	
	if (xlen > mxl_to_xlen(mxl)) {
		LOG_TARGET_ERROR(target,
				"The detected XLEN (%u) exceeds the XLEN reported in misa.mxl field (%u)",
				...);
		return ERROR_FAIL;
	}
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -85,33 +87,247 @@ static int riscv013_reg_set(struct reg *reg, uint8_t *buf)
...
+	/** NOTE:
+	 * The value of "misa.mxl" may stil not coincide with "xlen".
+	 * "misa[26:XLEN-3]" bits are marked as WIRI in at least version 1.10
+	 * of RISC-V Priveleged Spec. Therefore, if "xlen" is erroneously
+	 * assumed to be 32 when it actually is 64, "actual_mxl" will be read
```
JanMatCodasip: > if "xlen" is erroneously assumed to be 32 when it actually is 64 (...)

How could that happen? Do you mean some sort of a sporadic issue that could interfere with the 64-bit read of s0 (in examine()), making OpenOCD believe that the target is a 32-bit one?

JanMatCodasip (commented):
Overall, this looks fine to me, thank you. I have only posted few minor comments.

src/target/riscv/riscv-013_reg.c
aap-sc: @JanMatCodasip yes. We sometimes observe such situations. For example if the system was under non-OpenOCD induced reset on some designs really weird things may happen.

JanMatCodasip: Thanks for the explanation.

src/target/riscv/riscv-013_reg.c

src/target/riscv/riscv-013_reg.c
en-sc: 1,2 - Addressed.

> it may be good to check for MXLEN = 128

AFAIU, this can not be done here: if `misa` is read assuming DXLEN is 64, when it is actually 128, `misa.mxl` will be read from `misa[61:63]`, which will be zero (or possibly non-zero in case of Priv. Spec 1.10).

JanMatCodasip: 1,2 - Thank you.

3 - You're right. The detection of MXLEN=128 would have to be done differently than what I suggested (for instance by attempting a 128-bit abstract register read).

JanMatCodasip (approved):

aap-sc (approved):

PR #1115, opened 2024-08-14
target/riscv: restrict BSCAN-related commands to before-`init`

en-sc:
Logically, BSCAN tunneling is used to establish a connection, therefore it should be set up before the communication starts (i.e. before `init`).

Moreover, current implementation does not support changing `bscan_tunnel_ir_width` after `init`. This is evident by RISC-V handler of the `init` itself.
Link: https://github.com/riscv-collab/riscv-openocd/blob/9a23c9e67978f77d9166102cefc7b537b714b561/src/target/riscv/riscv.c#L467-L481

Change-Id: I817c6a996f7f7171b2286e181daf1092bd358f69

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -11354,10 +11354,15 @@ Display/set the current core displayed in GDB. This is needed only if
 @code{riscv smp} was used.
 @end deffn
 
-@deffn {Command} {riscv use_bscan_tunnel} value
+@deffn {Command} {riscv use_bscan_tunnel} width [type]
 Enable or disable use of a BSCAN tunnel to reach the Debug Module. Supply the
-width of the DM transport TAP's instruction register to enable. Supply a
-value of 0 to disable.
+@var{width} of the DM transport TAP's instruction register to enable. The
+@var{width} should fit into 7 bits. Supply a value of 0 to disable.
+Pass a second argument (optional) to indicate Bscan Tunnel Type:
+@enumerate
+@item 0:(default) NESTED_TAP
+@item 1: DATA_REGISTER
+@end enumerate
 
 This BSCAN tunnel interface is specific to SiFive IP. Anybody may implement
 it, but currently there is no good documentation on it. In a nutshell, this
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -54,7 +54,8 @@ struct scan_field select_idcode = {
 };
 
 static bscan_tunnel_type_t bscan_tunnel_type;
-int bscan_tunnel_ir_width; /* if zero, then tunneling is not present/active */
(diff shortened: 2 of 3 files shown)


en-sc:
@Dolu1990, @GregSavin, @command-paul could you please take a look if time permits?

GregSavin:
The change from COMMAND_ANY to COMMAND_CONFIG looks good and correct to me.

JanMatCodasip (approved):

JanMatCodasip:
@en-sc The code change looks all right to me, thank you.

Broader question to all participants in this thread (@Dolu1990, @GregSavin, @command-paul) - 

I am concerned about maintenance of this the proprietary BSCAN tunnel functionality in OpenOCD as we do not have a way to test it. If you have any suggestions in this area, that would be most welcome.

GregSavin:
Regarding testing, that is a fair point, and actually there are now 2 slightly different proprietary variants covered by that tunnel feature (nested TAP and tunneled DR).  Presumably testable on their own respective FPGA-based environments but not on Spike.  Modifications to Spike might be possible, as build-time and run-time options, but I don't know if these tunneling schemes are currently codified in RISC-V standards or drafts, so that might be a barrier for getting that approach accepted into the main Spike branch.  (I used to work for one of the companies that implemented one of the BSCAN tunnel schemes on some of their FPGA images for RISC-V based systems, but now I work for a different company).

Dolu1990:
Hi,

So i would say, we don't need any hardware to test it. Simulation can be good enough.
In my case, i use some Scala + SpinalHDL+Verilator based simulation testbench which let me connect real openocd to the simulated system through a jtag over TCP connection.
It isn't "fast", but fast enough to connect in 3 seconds and load a binary at a rate of about 1KB/s.
The installation requirements are : Java + Verilator + SBT. Could even remove the SBT requirements by releasing a JAR file.

Let me know if interrested, i can setup something.
(based on https://github.com/SpinalHDL/VexiiRiscv/blob/dev/src/main/tcl/openocd/vexiiriscv_sim_tunneled.tcl)

JanMatCodasip:
Hello @GregSavin, @Dolu1990,

Thank you for replying (and sorry for my late reaction due to being OoO).

If the preference would be to go with Spike (as @GregSavin suggests), maybe a thin wrapper over Spike could be written to simulate both the BSCAN tunnel modes. Perhaps something along [these lines](https://github.com/riscv-software-src/riscv-tests/blob/master/debug/rbb_daisychain.py). This way Spike could remain unmodified.

The solution from @Dolu1990 to have a simulated HW with the BSCAN tunnel also sounds fine to me.

Dolu1990:
I think the best is probably to try the spike solution first, as it would be more compliant / much faster to simulate.

PR #1118, opened 2024-08-21
target/riscv: re-apply patch do stop avoid warnings when a non-existent CSR is hidden

aap-sc:
the original fix was introduced in https://github.com/riscv-collab/riscv-openocd/commit/b201a5db23c0db34c0e10fd1c7c08fc73a5ec3fc but was lost in https://github.com/riscv-collab/riscv-openocd/commit/3883b03aaafef5e1eff423664f7ea2e1d8f14ba4

### DIFF
--- a/src/target/riscv/riscv_reg.c
+++ b/src/target/riscv/riscv_reg.c
@@ -730,7 +730,7 @@ int riscv_reg_impl_expose_csrs(const struct target *target)
 			struct reg * const reg = riscv_reg_impl_cache_entry(target, regno);
 			const unsigned int csr_number = regno - GDB_REGNO_CSR0;
 			if (reg->exist) {
-				LOG_TARGET_WARNING(target,
+				LOG_TARGET_DEBUG(target,
 						"Not exposing CSR %d: register already exists.",
 						csr_number);
 				continue;


aap-sc:
@MarekVCodasip , @en-sc I'm going to wait till Friday and if no objections - I'll proceed with merging this. We already had this patch reviewed - it was occasionally lost in https://github.com/riscv-collab/riscv-openocd/commit/3883b03aaafef5e1eff423664f7ea2e1d8f14ba4

MarekVCodasip (approved):
LGTM

en-sc:
@aap-sc, this patch is not the same as b201a5db23c0db34c0e10fd1c7c08fc73a5ec3fc -- it changes the warning in `riscv_reg_impl_expose_csrs()` instead of the one in `riscv_reg_impl_hide_csrs()`. I'd like to revert it, since the commit message is extremely misleading.

The original issue (the one that b201a5db23c0db34c0e10fd1c7c08fc73a5ec3fc addressed) is still present in the current version:
https://github.com/riscv-collab/riscv-openocd/blob/c85d4e1858840fe90a2fa0c521528338952d49da/src/target/riscv/riscv_reg.c#L758-L763

PR #1121, opened 2024-08-30
enable additional platforms in CI

aap-sc:


.github/workflows/spike-openocd-tests.yml
```diff
@@ -115,7 +115,8 @@ jobs:
               --gcc /opt/riscv/toolchain/bin/riscv-none-elf-gcc \
               --gdb /opt/riscv/toolchain/bin/riscv-none-elf-gdb \
               --sim_cmd /opt/riscv/spike/bin/spike \
-              --server_cmd $GITHUB_WORKSPACE/src/openocd
+              --remotelogfile-enable \
+              --server_cmd "$GITHUB_WORKSPACE/src/openocd -d3"
```
en-sc, line 119: I would prefer not to add `-d3` here. The reason is: `d3` could change OpenOCD behavior (there are many cases of `if (debug_level < 3) return;` ).

aap-sc: 1. I'm strongly against not testing -d3 . This is the mode when we use when run this testsuite internally. And this is the mode we use when we ask users when there are problems.
2. The fact that there is these kinds of if statement in the codebase is a major problem, but I think it is a separate one. If it comes to that - such if statements should be eradicated. At least from riscv codebase for now.

en-sc: > This is the mode when we use when run this testsuite internally.

Exactly. After this change there will be no testing of OpenOCD without `-d3`.

Anyway, I don't have a strong opinion on this so I'll approve the change regardless.

aap-sc: @JanMatCodasip , @MarekVCodasip do you have an opinion on the matter? The concern raised by @en-sc is valid. 

I'm inclined to extend testing even more by enabling both variants: "-d3" is enabled and "-d3" is omitted. Tests may be run a little bit slower now, however I think that we are better safe than sorry

en-sc: > I'm inclined to extend testing even more by enabling both variants

Please, don't. This is pre-commit testing. I believe it's better to run such tests at least in post-commit.

JanMatCodasip: There will always remain a little risk that adding `-d3` will change behavior -- IMO very small but still present.

My recommendation would be to:

- 1) Remove suspicious/risky statements `if (debug_level < XYZ ) return;` (in a separate merge request).
- 2) Pick one verbosity level for the tests that run in the CI on per-commit basis (I prefer `-d3` slightly more but don't feel strongly).
- 3) Optionally, if you like: Run the tests with the other verbosity level, too, but less frequently. For example:
  - When comitting/merging to the main `riscv` branch.
  - Or every night on the main `riscv` branch ([schedule options in Github Actions](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule)).
  - Or similar
- 4) Optionally, alternate approach: Use different logging level on different targets. This would at least introduce more variability into the testing.

.github/workflows/spike-openocd-tests.yml
```diff
@@ -125,7 +126,30 @@ jobs:
...
+      - name: Run Spike32-2-hwthread Tests
+        if: success() || steps.spike32-tests.conclusion == 'failure'
+        run: |
+          cd riscv-tests/debug
+          ./gdbserver.py targets/RISC-V/spike32-2-hwthread.py --print-failures \
+              --gcc /opt/riscv/toolchain/bin/riscv-none-elf-gcc \
+              --gdb /opt/riscv/toolchain/bin/riscv-none-elf-gdb \
+              --sim_cmd /opt/riscv/spike/bin/spike \
+              --remotelogfile-enable \
+              --server_cmd "$GITHUB_WORKSPACE/src/openocd -d3"
```
JanMatCodasip, line 152: In past, I have observed that some of the riscv-tests were unstable.

For that reason, I'd recommend to run these newly added tests many times (even as much as 50 times on your local machine) to detect any potential instability, or to gain high level of confidence that the tests are very stable.

(Maybe you have already done that.)

.github/workflows/spike-openocd-tests.yml
```diff
@@ -125,7 +126,30 @@ jobs:
               --gcc /opt/riscv/toolchain/bin/riscv-none-elf-gcc \
               --gdb /opt/riscv/toolchain/bin/riscv-none-elf-gdb \
               --sim_cmd /opt/riscv/spike/bin/spike \
-              --server_cmd $GITHUB_WORKSPACE/src/openocd
+              --remotelogfile-enable \
+              --server_cmd "$GITHUB_WORKSPACE/src/openocd -d3"
+
+      - name: Run Spike64-2-hwthread Tests
```
JanMatCodasip, line 132: Now that the number of of tests has increased, it may be worth to split them to multiple jobs that would then run in parallel, leading to overall shorter test time.

JanMatCodasip (approved):
I have posted several comments but otherwise the changes LGTM.

Thank you.

.github/workflows/spike-openocd-tests.yml
aap-sc: @en-sc 

> Please, don't. This is pre-commit testing. I believe it's better to run such tests at least in post-commit.

I don't believe we should reduce amount when creating a commit in open source. We already had few regressions that could be detected by a more rigorous testing. This commit is a direct response on one of such regression.  I don't mind waiting 40 minutes instead of 20 to get testing results. I don't think this is a bit problem. 

@JanMatCodasip 

> Remove suspicious/risky statements if (debug_level < XYZ ) return; (in a separate merge request)

The thing is it is not that simple and may require a significant effort. Things are getting more complicated since the decision to use such approach was promoted in upstream several months ago. So unless the policy regarding such condition is changed in upstream - the issue shall persist.  But we definitely should revise existing codebase regarding the matter. 

> Pick one verbosity level for the tests that run in the CI on per-commit basis (I prefer -d3 slightly more but don't feel strongly).

My pick would be -d3 too since usually this logging level allows to debug test.

> Optionally, if you like: Run the tests with the other verbosity level, too, but less frequently.

I'll investigate this . Thanks for the idea.

> Optionally, alternate approach: Use different logging level on different targets. This would at least introduce more variability into the testing.

As for messing up with different logging levels, I believe this may be an over complication. However, your suggestion regarding splitting test into multiple job may help with this one.

.github/workflows/spike-openocd-tests.yml
aap-sc: > In past, I have observed that some of the riscv-tests were unstable.
> (Maybe you have already done that.)

Well actually we've already done that. The tests are quite stable now (at least on a decent machines). Internally in Syntacore I've analysed the stability of the teststuite and run the whole testsuite more than a 1000 times. Not a single failure was detected. That being said there is still this https://github.com/riscv-collab/riscv-openocd/issues/104 that still may haunt us under certain conditions - I know the root cause, but still could not find time to fix this.

.github/workflows/spike-openocd-tests.yml
aap-sc: yes, this is definitely a way to go. Will investigate how to do this.

.github/workflows/spike-openocd-tests.yml
JanMatCodasip: Thank you.

.github/workflows/spike-openocd-tests.yml
JanMatCodasip: It looks to me that the current job can be split to multiple parallel jobs quite easily using [strategy-matrix feature](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow) of Github Actions.

Here is a concrete example: https://github.com/HonzaMat/PyOpenocdClient/blob/main/.github/workflows/integration_tests.yml#L7

JanMatCodasip:
Hello @aap-sc,
this PR looks like a good improvement to me. Are you interested in continuing with it?

In the [first open thread](https://github.com/riscv-collab/riscv-openocd/pull/1121#discussion_r1802626643) it looks like we reached the consensus.
The [second open thread](https://github.com/riscv-collab/riscv-openocd/pull/1121#discussion_r1803071877) is just about an optional improvement.

Thanks.

aap-sc:
> Hello @aap-sc, this PR looks like a good improvement to me. Are you interested in continuing with it?
> 
> In the [first open thread](https://github.com/riscv-collab/riscv-openocd/pull/1121#discussion_r1802626643) it looks like we reached the consensus. The [second open thread](https://github.com/riscv-collab/riscv-openocd/pull/1121#discussion_r1803071877) is just about an optional improvement.
> 
> Thanks.

Yep . I'm looking at this PR every week and try to force myself to do the final push (failing miserably every time). I understand that this is an optional improvement, however, since I agree with it I'm not feeling comfortable not addressing it.

PR #1124, opened 2024-09-02
target/riscv: Mismatch napot when mcontrol.maskmax is zero

sunnyzhu-learning:
Reference release specification (https://github.com/riscv/riscv-debug-spec/blob/release/riscv-debug-release.pdf), in section 5.2.9 introduce maskmx：
A value of 0 indicates that only exact value matches are supported (one byte range).

so. when tdata1.type=2 and napot is unsupported, `try_setup_single_match_trigger` should skip `match=1` .

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -835,6 +835,10 @@ static int try_setup_single_match_trigger(struct target *target,
 	for (unsigned int idx = 0;
 			find_next_free_trigger(target, trigger_type, false, &idx) == ERROR_OK;
 			++idx) {
+		if (trigger_type == CSR_TDATA1_TYPE_MCONTROL
+				&& (get_field(trig_info.tdata1, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT)
+				&& r->trigger_maskmax[idx] == 0)
+				continue;
 		ret = try_use_trigger_and_cache_result(target, idx, trig_info.tdata1, trig_info.tdata2,
 			trig_info.tdata1_ignore_mask);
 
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv.c
```diff
@@ -5729,6 +5733,11 @@ int riscv_enumerate_triggers(struct target *target)
 		riscv_reg_t tdata1;
 		if (riscv_reg_get(target, &tdata1, GDB_REGNO_TDATA1) != ERROR_OK)
 			return ERROR_FAIL;
+		int type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
+		if (type == CSR_TDATA1_TYPE_MCONTROL) {
```
en-sc, line 5736: What if the trigger supports multiple types and is disabled at the moment of enumeration?
AFAIU, you it will not be possible to use such trigger for a NAPOT match.

en-sc (changes requested):
AFAIU, the approach suggested in the commit will not be suitable for some HW configurations (please, see the comment on `riscv_enumerate_triggers()`).

I'd suggest to look at `set_trigger()` instead.
In particular, `tdata1_config_denied` can reflect that the value in `mcontrol.maskmax` fits the requested range.
e.g.:
1. Rename `trigger_request_info::tdata1_ignore_mask` -> `trigger_request_info::maskmax`.
2.  Add the something like the following to `set_trigger()`:
```
const uint32_t type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
if (type == CSR_TDATA1_TYPE_MCONTROL &&
		(get_field(tdata1, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT))
		tdata1_config_denied |= get_field(tdata1_rb, CSR_MCONTROL_MASKMAX) < (get_least_significant_zero_index(tdata2) + 1).
}
```
(`get_least_significant_zero_index()` is not yet implemented).

sunnyzhu-learning:
thanks a lot, I will create new PR addressed this question

PR #1125, opened 2024-09-03
target/riscv: remove duplicate of progbufsize field

fkhaidari:
Remove duplicate of progbufsize field
* removed `progbuf_size`  field from `riscv_info`; added getter
* moved `impebreak` field from `riscv_info` to 'riscv013_info' as implementation dependent field; added getter

### DIFF
--- a/src/target/riscv/program.c
+++ b/src/target/riscv/program.c
@@ -183,8 +183,8 @@ int riscv_program_ebreak(struct riscv_program *p)
 {
 	struct target *target = p->target;
 	RISCV_INFO(r);
-	if (p->instruction_count == riscv_progbuf_size(p->target) &&
-			r->impebreak) {
+	if (p->instruction_count == riscv_progbuf_size(target) &&
+			r->get_impebreak(target)) {
 		return ERROR_OK;
 	}
 	return riscv_program_insert(p, ebreak());
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -67,6 +67,8 @@ static int read_memory(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, uint8_t *buffer, uint32_t increment);
 static int write_memory(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, const uint8_t *buffer);
+static bool riscv013_get_impebreak(const struct target *target);
+static unsigned int riscv013_get_progbufsize(const struct target *target);
 
 typedef enum {
 	HALT_GROUP,
@@ -153,7 +155,9 @@ typedef struct {
 	/* Number of abstract command data registers. */
 	unsigned datacount;
 	/* Number of words in the Program Buffer. */
-	unsigned progbufsize;
+	unsigned int progbufsize;
+	/* Hart contains an implicit ebreak at the end of the program buffer. */
+	bool impebreak;
 
 	/* We cache the read-only bits of sbcs here. */
 	uint32_t sbcs;
@@ -1310,9 +1314,7 @@ static unsigned int register_size(struct target *target, enum gdb_regno number)
 static bool has_sufficient_progbuf(struct target *target, unsigned size)
 {
 	RISCV013_INFO(info);
-	RISCV_INFO(r);
-
-	return info->progbufsize + r->impebreak >= size;
+	return info->progbufsize + info->impebreak >= size;
 }
 
 /**
@@ -2037,14 +2039,13 @@ static int examine(struct target *target)
 	LOG_TARGET_INFO(target, "datacount=%d progbufsize=%d",
 			info->datacount, info->progbufsize);
 
-	RISCV_INFO(r);
-	r->impebreak = get_field(dmstatus, DM_DMSTATUS_IMPEBREAK);
+	info->impebreak = get_field(dmstatus, DM_DMSTATUS_IMPEBREAK);
 
 	if (!has_sufficient_progbuf(target, 2)) {
 		LOG_TARGET_WARNING(target, "We won't be able to execute fence instructions on this "
 				"target. Memory may not always appear consistent. "
 				"(progbufsize=%d, impebreak=%d)", info->progbufsize,
-				r->impebreak);
+				info->impebreak);
 	}
 
 	if (info->progbufsize < 4 && riscv_enable_virtual) {
@@ -2060,6 +2061,8 @@ static int examine(struct target *target)
 	enum riscv_hart_state state_at_examine_start;
 	if (riscv_get_hart_state(target, &state_at_examine_start) != ERROR_OK)
 		return ERROR_FAIL;
+
+	RISCV_INFO(r);
 	const bool hart_halted_at_examine_start = state_at_examine_start == RISCV_STATE_HALTED;
 	if (!hart_halted_at_examine_start) {
 		r->prepped = true;
@@ -2072,10 +2075,6 @@ static int examine(struct target *target)
 
 	target->state = TARGET_HALTED;
 	target->debug_reason = hart_halted_at_examine_start ? DBG_REASON_UNDEFINED : DBG_REASON_DBGRQ;
-
-	/* Without knowing anything else we can at least mess with the
-	 * program buffer. */
-	r->progbuf_size = info->progbufsize;
 
 	result = riscv013_reg_examine_all(target);
 	if (result != ERROR_OK)
@@ -2817,6 +2816,8 @@ static int init_target(struct command_context *cmd_ctx,
 	generic_info->read_memory = read_memory;
 	generic_info->data_bits = &riscv013_data_bits;
 	generic_info->print_info = &riscv013_print_info;
+	generic_info->get_impebreak = &riscv013_get_impebreak;
+	generic_info->get_progbufsize = &riscv013_get_progbufsize;
 
 	generic_info->handle_became_unavailable = &handle_became_unavailable;
 	generic_info->tick = &tick;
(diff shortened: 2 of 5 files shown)


en-sc (dismissed):
LGTM (reviewed internally). @JanMatCodasip, @MarekVCodasip, please take a look.

src/target/riscv/riscv.h
```diff
@@ -236,6 +230,9 @@ struct riscv_info {
 	int (*dmi_read)(struct target *target, uint32_t *value, uint32_t address);
 	int (*dmi_write)(struct target *target, uint32_t address, uint32_t value);
 
+	bool (*get_impebreak)(const struct target *target);
```
JanMatCodasip, line 233: Detail, if you like:

This callback + the actual functions could be called `has_impebreak` to make it even clearer that they provide a bool value.

```suggestion
	bool (*has_impebreak)(const struct target *target);
```

src/target/riscv/riscv-013.c
```diff
@@ -2060,6 +2061,8 @@ static int examine(struct target *target)
 	enum riscv_hart_state state_at_examine_start;
 	if (riscv_get_hart_state(target, &state_at_examine_start) != ERROR_OK)
 		return ERROR_FAIL;
+
+	struct riscv_info *r = riscv_info(target);
```
JanMatCodasip: Detail, if you like:

You may use RISCV_INFO() here. It is the prevailing approach (it is used more often than riscv_info) so it'd be preferable for code consistency.

```suggestion
	RISCV_INFO(r);
```

src/target/riscv/program.c
```diff
@@ -182,9 +182,8 @@ int riscv_program_fence_rw_rw(struct riscv_program *p)
 int riscv_program_ebreak(struct riscv_program *p)
 {
 	struct target *target = p->target;
-	RISCV_INFO(r);
-	if (p->instruction_count == riscv_progbuf_size(p->target) &&
-			r->impebreak) {
+	if (p->instruction_count == riscv_progbuf_size(target) &&
+			riscv_info(target)->get_impebreak(target)) {
```
JanMatCodasip: (same as in previous comment)

I'd use RISCV_INFO() here - I'd prefer consistency over a saved line of code.

```suggestion
	RISCV_INFO(r);
	if (p->instruction_count == riscv_progbuf_size(target) &&
			r->get_impebreak(target)) {
```

src/target/riscv/riscv.c
```diff
@@ -5568,8 +5568,7 @@ static enum riscv_halt_reason riscv_halt_reason(struct target *target)
 
 size_t riscv_progbuf_size(struct target *target)
 {
-	RISCV_INFO(r);
-	return r->progbuf_size;
+	return riscv_info(target)->get_progbufsize(target);
```
JanMatCodasip: (same as in the previous comment)

I'd prefer RISCV_INFO() for consistency with the other functions around.

```suggestion
	RISCV_INFO(r);
	return r->get_progbufsize(target);
```

src/target/riscv/riscv-013.c
```diff
@@ -2817,6 +2816,8 @@ static int init_target(struct command_context *cmd_ctx,
 	generic_info->read_memory = read_memory;
 	generic_info->data_bits = &riscv013_data_bits;
 	generic_info->print_info = &riscv013_print_info;
+	generic_info->get_impebreak = &riscv013_get_impebreak;
+	generic_info->get_progbufsize = &riscv013_get_progbufsize;
```
JanMatCodasip, line 2820: (Just a side note, no change requested)

I would like to get rid of all these callbacks; hopefully I'll find the time to do that soon.

These callbacks are needed now due to improper division of the code. Most notably, some pieces of code that are specific to RISC-V 0.13 still reside in riscv.c. For example, riscv-011 code should never call `riscv_progbuf_size()` and all progbuf operations should be confined into riscv-013 portion of the code.

Once the code is properly relocated and cleaned up, these function pointers / callbacks should become redundant, and direct function calls will be possible.

src/target/riscv/riscv-013.c
```diff
@@ -4872,6 +4873,16 @@ static int write_memory(struct target *target, target_addr_t address,
...
 
+static bool riscv013_get_impebreak(const struct target *target)
+{
+	return get_info(target)->impebreak;
+}
+
+static unsigned int riscv013_get_progbufsize(const struct target *target)
+{
+	return get_info(target)->progbufsize;
+}
```

RISCV013_INFO() is more frequenly used than get_info(). So for code consistency, you may consider this:

```suggestion
static bool riscv013_get_impebreak(const struct target *target)
{
	RISCV013_INFO(info);
	return info->impebreak;
}

static unsigned int riscv013_get_progbufsize(const struct target *target)
{
	RISCV013_INFO(info);
	return info->progbufsize;
}
```

JanMatCodasip (dismissed):
Thank you for the cleanup, it looks good.

I have posted few optional comments for your consideration, mostly related to code consistency.

MarekVCodasip (dismissed):
Nothing to add to @JanMatCodasip

TommyMurphyTM1234:
Small (possibly dumb) question...
What does "[NFC]" in the PR title mean ?
The only thing that I can think of is "non-functional change" - meaning a code change that cleans something up but doesn't deliberately change the logic/functionality (although surely any code change impacts logic/functionality?).
Is that it or is it something else? 
Thanks.

fkhaidari:
@TommyMurphyTM1234, you are right. NFC means non-functional change. I can rename PR if such naming is confusing

TommyMurphyTM1234:
> @TommyMurphyTM1234, you are right. NFC means non-functional change. I can rename PR if such naming is confusing

Thanks @fk-sc - I just hadn't come across such nomenclature before and couldn't find anything explanatory by searching. But maybe it's a commonly used shorthand.

src/target/riscv/riscv.h
fkhaidari: Thanks for the offer, but I will leave previous name to keep it consistent with `get_progbufsize`

src/target/riscv/riscv-013.c
fkhaidari: Addressed. To be honest, I think that explicit way of variable declaration is better than macro. But I understand that keeping consistency is important. Maybe I'll refactor whole codebase to get rid of such macros in future

src/target/riscv/program.c
fkhaidari: Addressed

src/target/riscv/riscv.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
fkhaidari: Wouldn't it be confusing? And how to guarantee that function called for the appropriate type? Just assume that this is always correct or use assert?

src/target/riscv/riscv.h
JanMatCodasip: Ok, let's leave the name as is.

src/target/riscv/riscv-013.c
JanMatCodasip: > I think that explicit way of variable declaration is better than macro.

My position to this is neutral - I don't mind either way as long as the code is consistent.

> Maybe I'll refactor whole codebase to get rid of such macros in future

Yes, that is possible. If that happens, I'd prefer to make a bulk replace across the whole codebase.

src/target/riscv/riscv-013.c
JanMatCodasip: Most of the RISC-V code is specific to one of the versions: riscv-0.11 vs. riscv-0.13. For that reason the code for 0.11 should directly call the 0.11 functions, and the same is true for 0.13.

To use `get_impebreak` and `get_progbufsize` as examples:

These callbacks are specific for riscv-0.13. For that reason:
- The code for 0.11 should never use any of them.
- The code for 0.13 should call directly `riscv013_get_impebreak()` and `riscv013_get_progbufsize()` (once the code is properly re-structured into functions and modules).

I'm not sure if I managed to explain my point clearly. Anyway, I hope to make a PR with the proposed changes sometime soon.

src/target/riscv/riscv-013.c
JanMatCodasip: It looks like this change has not been pushed - can you please double-check it? Thanks.

src/target/riscv/riscv-013.c
```diff
@@ -153,7 +155,9 @@ typedef struct {
 	/* Number of abstract command data registers. */
 	unsigned datacount;
 	/* Number of words in the Program Buffer. */
-	unsigned progbufsize;
+	unsigned int progbufsize;
```
MarekVCodasip, line 158:
```suggestion
	unsigned progbufsize;
```
Please, stick to just using `unsigned`

src/target/riscv/riscv.h
```diff
@@ -236,6 +230,9 @@ struct riscv_info {
...
+	unsigned int (*get_progbufsize)(const struct target *target);
```
MarekVCodasip, line 234:
```suggestion
	unsigned (*get_progbufsize)(const struct target *target);
```
And here as well.

MarekVCodasip (commented):
I have two small nitpicks I additionally noticed. Please take a look.

src/target/riscv/riscv-013.c
en-sc: `unsigned int` is preferable, according to style guide:
https://github.com/riscv-collab/riscv-openocd/blob/22bbdd2a5e9388376c8097ac9ea0cbfede3680d6/doc/manual/style.txt#L177

Moreover, Checkpatch will issue a warning.

src/target/riscv/riscv.h
en-sc: Please, see https://github.com/riscv-collab/riscv-openocd/pull/1125#discussion_r1743901085

src/target/riscv/riscv-013.c
fkhaidari: Agree with @en-sc

src/target/riscv/riscv-013.c
fkhaidari: You are right. Updated commit

src/target/riscv/riscv-013.c
fkhaidari: OK, thank you for the explanation!

src/target/riscv/riscv-013.c
JanMatCodasip: That's right, OpenOCD code style asks for `unsigned int`.

JanMatCodasip (approved):

src/target/riscv/riscv-013.c
MarekVCodasip: oh, in that case, we should change it elsewhere

MarekVCodasip (approved):

PR #1127, opened 2024-09-05
target/riscv: Mismatch napot when mcontrol.maskmax is not expected

sunnyzhu-learning:
From https://github.com/riscv-collab/riscv-openocd/pull/1124,
1.Rename `trigger_request_info::tdata1_ignore_mask` -> `trigger_request_info::maskmax`.
2.Add the something like the following to `set_trigger()`:

>  const uint32_t type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
if (type == CSR_TDATA1_TYPE_MCONTROL &&
		(get_field(tdata1, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT))
		tdata1_config_denied |= get_field(tdata1_rb, CSR_MCONTROL_MASKMAX) < (get_least_significant_zero_index(tdata2) + 1).
}

addressed
@en-sc Thanks for your suggestion, internal discussion a lot,  and first push code to github,delay....please review~

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -621,8 +621,17 @@ static int find_first_trigger_by_id(struct target *target, int unique_id)
 	return -1;
 }
 
-static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdata1, riscv_reg_t tdata2,
-	riscv_reg_t tdata1_ignore_mask)
+static unsigned int count_trailing_ones(riscv_reg_t reg)
+{
+	assert(sizeof(riscv_reg_t) * 8 == 64);
+	for (unsigned int i = 0; i < 64; i++) {
+		if ((1 & (reg >> i)) == 0)
+			return i;
+	}
+	return 64;
+}
+
+static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdata1, riscv_reg_t tdata2)
 {
 	RISCV_INFO(r);
 	assert(r->reserved_triggers);
@@ -655,20 +664,51 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
 		return ERROR_FAIL;
 	if (riscv_reg_get(target, &tdata2_rb, GDB_REGNO_TDATA2) != ERROR_OK)
 		return ERROR_FAIL;
-	bool tdata1_config_denied = (tdata1 & ~tdata1_ignore_mask) != (tdata1_rb & ~tdata1_ignore_mask);
-	bool tdata2_config_denied = tdata2 != tdata2_rb;
-	if (tdata1_config_denied || tdata2_config_denied) {
+
+	const uint32_t type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
+	const bool is_mcontrol = type == CSR_TDATA1_TYPE_MCONTROL;
+
+	/* Determine if tdata1 supports what we need.
+	 * For mcontrol triggers, we don't care about
+	 * the value in the read-only "maskmax" field.
+	 */
+	const riscv_reg_t tdata1_ignore_mask = is_mcontrol ? CSR_MCONTROL_MASKMAX(riscv_xlen(target)) : 0;
+	const bool tdata1_config_denied = (tdata1 & ~tdata1_ignore_mask) != (tdata1_rb & ~tdata1_ignore_mask);
+
+	/* Determine if tdata1.maxmask is sufficient
+	 * (only relevant for mcontrol triggers and NAPOT match type)
+	 */
+	bool unsupported_napot_range = false;
+	riscv_reg_t maskmax_value = 0;
+	if (!tdata1_config_denied) {
+		const bool is_napot_match = get_field(tdata1_rb, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT;
+		if (is_mcontrol && is_napot_match) {
+			maskmax_value = get_field(tdata1_rb, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
+			const unsigned int napot_size = count_trailing_ones(tdata2) + 1;
+			if (maskmax_value < napot_size)
+				unsupported_napot_range = true;
+		}
+	}
+
+	const bool tdata2_config_denied = tdata2 != tdata2_rb;
+	if (tdata1_config_denied || tdata2_config_denied || unsupported_napot_range) {
 		LOG_TARGET_DEBUG(target, "Trigger %u doesn't support what we need.", idx);
 
 		if (tdata1_config_denied)
 			LOG_TARGET_DEBUG(target,
-				"After writing 0x%" PRIx64 " to tdata1 it contains 0x%" PRIx64 "; tdata1_ignore_mask=0x%" PRIx64,
-				tdata1, tdata1_rb, tdata1_ignore_mask);
+				"After writing 0x%" PRIx64 " to tdata1 it contains 0x%" PRIx64,
+				tdata1, tdata1_rb);
 
 		if (tdata2_config_denied)
 			LOG_TARGET_DEBUG(target,
-				"wrote 0x%" PRIx64 " to tdata2 but read back 0x%" PRIx64,
+				"After writing 0x%" PRIx64 " to tdata2 it contains 0x%" PRIx64,
 				tdata2, tdata2_rb);
+
+		if (unsupported_napot_range)
+			LOG_TARGET_DEBUG(target,
+				"The requested NAPOT match range (tdata2=0x%" PRIx64 ") exceeds maskmax_value=0x%" PRIx64,
+				tdata2, maskmax_value);
+
 		if (riscv_reg_set(target, GDB_REGNO_TDATA1, 0) != ERROR_OK)
 			return ERROR_FAIL;
 		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
@@ -717,7 +757,7 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
 	tdata1 = set_field(tdata1, bpcontrol_bpaction, 0); /* cause bp exception */
 	tdata1 = set_field(tdata1, bpcontrol_bpmatch, 0); /* exact match */
 	tdata2 = trigger->address;
-	ret = set_trigger(target, idx, tdata1, tdata2, 0);
+	ret = set_trigger(target, idx, tdata1, tdata2);
 	if (ret != ERROR_OK)
 		return ret;
 	r->trigger_unique_id[idx] = trigger->unique_id;
@@ -727,13 +767,11 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
 struct trigger_request_info {
 	riscv_reg_t tdata1;
 	riscv_reg_t tdata2;
-	riscv_reg_t tdata1_ignore_mask;
 };
 
 static void log_trigger_request_info(struct trigger_request_info trig_info)
 {
-	LOG_DEBUG("tdata1=%" PRIx64 ", tdata2=%" PRIx64 ", tdata1_ignore_mask=%" PRIx64,
-			trig_info.tdata1, trig_info.tdata2, trig_info.tdata1_ignore_mask);
+	LOG_DEBUG("tdata1=%" PRIx64 ", tdata2=%" PRIx64, trig_info.tdata1, trig_info.tdata2);
 };
 
 static struct tdata1_cache *tdata1_cache_alloc(struct list_head *tdata1_cache_head, riscv_reg_t tdata1)
@@ -816,12 +854,12 @@ static bool wp_triggers_cache_search(struct target *target, unsigned int idx,
 }
 
 static int try_use_trigger_and_cache_result(struct target *target, unsigned int idx, riscv_reg_t tdata1,
-	riscv_reg_t tdata2, riscv_reg_t tdata1_ignore_mask)
+	riscv_reg_t tdata2)
 {
 	if (wp_triggers_cache_search(target, idx, tdata1, tdata2))
 		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
 
-	int ret = set_trigger(target, idx, tdata1, tdata2, tdata1_ignore_mask);
+	int ret = set_trigger(target, idx, tdata1, tdata2);
 
 	/* Add these values to the cache to remember that they are not supported. */
 	if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
@@ -844,8 +882,7 @@ static int try_setup_single_match_trigger(struct target *target,
 	for (unsigned int idx = 0;
 			find_next_free_trigger(target, trigger_type, false, &idx) == ERROR_OK;
 			++idx) {
-		ret = try_use_trigger_and_cache_result(target, idx, trig_info.tdata1, trig_info.tdata2,
-			trig_info.tdata1_ignore_mask);
+		ret = try_use_trigger_and_cache_result(target, idx, trig_info.tdata1, trig_info.tdata2);
 
 		if (ret == ERROR_OK) {
 			r->trigger_unique_id[idx] = trigger->unique_id;
@@ -873,16 +910,14 @@ static int try_setup_chained_match_triggers(struct target *target,
 	for (unsigned int idx = 0;
 			find_next_free_trigger(target, trigger_type, true, &idx) == ERROR_OK;
 			++idx) {
-		ret = try_use_trigger_and_cache_result(target, idx, t1.tdata1, t1.tdata2,
-			t1.tdata1_ignore_mask);
+		ret = try_use_trigger_and_cache_result(target, idx, t1.tdata1, t1.tdata2);
 
 		if (ret == ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
 			continue;
 		else if (ret != ERROR_OK)
 			return ret;
 
-		ret = try_use_trigger_and_cache_result(target, idx + 1, t2.tdata1, t2.tdata2,
-			t2.tdata1_ignore_mask);
+		ret = try_use_trigger_and_cache_result(target, idx + 1, t2.tdata1, t2.tdata2);
 
 		if (ret == ERROR_OK) {
 			r->trigger_unique_id[idx] = trigger->unique_id;
@@ -890,7 +925,7 @@ static int try_setup_chained_match_triggers(struct target *target,
 			return ERROR_OK;
 		}
 		/* Undo the setting of the previous trigger */
-		int ret_undo = set_trigger(target, idx, 0, 0, 0);
+		int ret_undo = set_trigger(target, idx, 0, 0);
 		if (ret_undo != ERROR_OK)
 			return ret_undo;
 
@@ -918,7 +953,6 @@ struct match_triggers_tdata1_fields {
 		riscv_reg_t ge;
 		riscv_reg_t eq;
 	} match;
-	riscv_reg_t tdata1_ignore_mask;
 };
 
 static struct match_triggers_tdata1_fields fill_match_triggers_tdata1_fields_t2(struct target *target,
@@ -951,8 +985,7 @@ static struct match_triggers_tdata1_fields fill_match_triggers_tdata1_fields_t2(
 			.lt = field_value(CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_LT),
 			.ge = field_value(CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_GE),
 			.eq = field_value(CSR_MCONTROL_MATCH, CSR_MCONTROL_MATCH_EQUAL)
-		},
-		.tdata1_ignore_mask = CSR_MCONTROL_MASKMAX(riscv_xlen(target))
+		}
 	};
 	return result;
 }
@@ -989,8 +1022,7 @@ static struct match_triggers_tdata1_fields fill_match_triggers_tdata1_fields_t6(
 			.lt = field_value(CSR_MCONTROL6_MATCH, CSR_MCONTROL6_MATCH_LT),
 			.ge = field_value(CSR_MCONTROL6_MATCH, CSR_MCONTROL6_MATCH_GE),
 			.eq = field_value(CSR_MCONTROL6_MATCH, CSR_MCONTROL6_MATCH_EQUAL)
-		},
-		.tdata1_ignore_mask = 0
+		}
 	};
 	return result;
 }
@@ -1009,8 +1041,7 @@ static int maybe_add_trigger_t2_t6_for_wp(struct target *target,
 				struct trigger_request_info napot = {
 					.tdata1 = fields.common | fields.size.any |
 						fields.chain.disable | fields.match.napot,
-					.tdata2 = trigger->address | ((trigger->length - 1) >> 1),
-					.tdata1_ignore_mask = fields.tdata1_ignore_mask
+					.tdata2 = trigger->address | ((trigger->length - 1) >> 1)
 				};
 				ret = try_setup_single_match_trigger(target, trigger, napot);
 				if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
@@ -1026,14 +1057,12 @@ static int maybe_add_trigger_t2_t6_for_wp(struct target *target,
 			struct trigger_request_info ge_1 = {
 				.tdata1 = fields.common | fields.size.any | fields.chain.enable |
 					fields.match.ge,
-				.tdata2 = trigger->address,
-				.tdata1_ignore_mask = fields.tdata1_ignore_mask
+				.tdata2 = trigger->address
 			};
 			struct trigger_request_info lt_2 = {
 				.tdata1 = fields.common | fields.size.any | fields.chain.disable |
 					fields.match.lt,
-				.tdata2 = trigger->address + trigger->length,
-				.tdata1_ignore_mask = fields.tdata1_ignore_mask
+				.tdata2 = trigger->address + trigger->length
 			};
 			ret = try_setup_chained_match_triggers(target, trigger, ge_1, lt_2);
 			if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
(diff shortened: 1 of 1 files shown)


en-sc:
Checkpatch fails:
```
> Run ./tools/scripts/checkpatch.pl --no-signoff --git FETCH_HEAD..HEAD
WARNING:COMMIT_MESSAGE: Missing commit description - Add an appropriate one
...
```
Please, add an empty line between commit title and description.
See https://git-scm.com/docs/git-commit#_discussion

src/target/riscv/riscv.c
```diff
@@ -619,8 +619,17 @@ static int find_first_trigger_by_id(struct target *target, int unique_id)
 	return -1;
 }
 
+static inline uint64_t get_least_significant_zero_index(riscv_reg_t tdata2)
```
en-sc: 1. Since the definition is not in a header I believe `inline` to be redundant.
2. `unsigned int` seems more appropriate here as a return type:
https://github.com/riscv-collab/riscv-openocd/blob/c85d4e1858840fe90a2fa0c521528338952d49da/doc/manual/style.txt#L168-L169
3. IMHO, `tdata2` is better renamed to`reg` or some other generic name -- there is nothing specific to `tdata2` about this function.
4. If you like: `get_least_significant_zero_index` -> `count_trailing_ones` (seems to me more specific and closer to naming scheme of compiler builtins).
```suggestion
static unsigned int count_trailing_ones(riscv_reg_t reg)
```

en-sc (commented):
Only minor comments, nothing significant.

On a side note:
@sunnyzhu-learning, have you tested the change? What was the simulator/HW used?

sunnyzhu-learning:
Q:have you tested the change? What was the simulator/HW used?
A:tested in HW env,
following is gdb + OpenOCD test：
```
$ riscv64-unknown-elf-gdb hello_world.elf
Reading symbols from hello_world.elf...
(gdb) target remote:3333
Remote debugging using :3333
_start () at ../****/SoC/****/Common/Source/GCC/startup_***.S:111
111         csrc MSTATUS, MSTATUS_MIE
(gdb) load
Loading section .init, size 0x18e lma 0x80000000
Loading section .text, size 0xe38c lma 0x80000400
Loading section .rodata, size 0xd00 lma 0x8000e790
Loading section .init_array, size 0x4 lma 0x8000f490
Loading section .data, size 0x728 lma 0x800104a0
Loading section .amo, size 0x90 lma 0x80011f58
Start address 0x80000000, load size 63958
Transfer rate: 34 KB/sec, 7106 
(gdb) monitor wp 0x800104a0 0x2 r
[riscv.cpu] Found 4 triggers
[riscv.cpu] Could not set a trigger that will match a whole address range. As a fallback, this trigger (and maybe others) will only match against the first address of the range.
(gdb) monitor reg tselect 0
tselect (/32): 0x00000000
(gdb) monitor reg tdata1
tdata1 (/32): 0x28001049
(gdb) monitor reg tdata2
tdata2 (/32): 0x800104a0

File: src/target/riscv/riscv.c (hunk 1 above)
Comment (sunnyzhu-learning) [reply src/target/riscv/riscv.c]: addressed

**Review by en-sc — DISMISSED** [review]
LGTM

File: src/target/riscv/riscv.c [hunk 2]
```diff
@@ -619,8 +619,17 @@ static int find_first_trigger_by_id(struct target *target, int unique_id)
 	return -1;
 }
 
+static unsigned int count_trailing_ones(riscv_reg_t reg)
+{
+	for (int i = 0; i < 64; i++) {
```
JanMatCodasip: Very small detail:

This helper function assumes that `riscv_reg_t` is an alias to `uint64_t`. 

Please, consider putting an assertion there:

```suggestion
	assert(sizeof(riscv_reg_t) * 8 == 64);

	for (int i = 0; i < 64; i++) {
```

src/target/riscv/riscv.c
```diff
@@ -716,13 +731,13 @@ static int maybe_add_trigger_t1(struct target *target, struct trigger *trigger)
 struct trigger_request_info {
 	riscv_reg_t tdata1;
 	riscv_reg_t tdata2;
-	riscv_reg_t tdata1_ignore_mask;
+	riscv_reg_t maskmax;
```
JanMatCodasip: I am not in favor of this rename. It can mis-lead the code reader to think that this field contains a value to be written to `maskmax` field, which is however not true.

Please, consider one of the following:

- a) Keep the original name `tdata1_ignore_mask` which very well describes the meaning of this variable.
- b) Replace this field by a simple `bool` flag that says whether the special handling of maxmask should be activated or not. E.g.: `bool check_tdata1_maxmask;`

Thank you.

src/target/riscv/riscv.c
```diff
@@ -644,15 +653,21 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
...
 
 		if (tdata1_config_denied)
 			LOG_TARGET_DEBUG(target,
-				"After writing 0x%" PRIx64 " to tdata1 it contains 0x%" PRIx64 "; tdata1_ignore_mask=0x%" PRIx64,
-				tdata1, tdata1_rb, tdata1_ignore_mask);
+				"After writing 0x%" PRIx64 " to tdata1 it contains 0x%" PRIx64 "; maskmax=0x%" PRIx64,
+				tdata1, tdata1_rb, maskmax);
 
 		if (tdata2_config_denied)
 			LOG_TARGET_DEBUG(target,
```
JanMatCodasip, line 703: Detail but helpful:

While making changes there, it would be good to unify the debug messages for tdata1 and tdata2:

```
	if (tdata2_config_denied)
			LOG_TARGET_DEBUG(target,
				"After writing 0x%" PRIx64 " to tdata2 it contains 0x%" PRIx64,
				tdata2, tdata2_rb);
```

src/target/riscv/riscv.c
```diff
@@ -644,15 +653,21 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
...
 	if (riscv_reg_get(target, &tdata2_rb, GDB_REGNO_TDATA2) != ERROR_OK)
 		return ERROR_FAIL;
-	bool tdata1_config_denied = (tdata1 & ~tdata1_ignore_mask) != (tdata1_rb & ~tdata1_ignore_mask);
+	bool tdata1_config_denied = (tdata1 & ~maskmax) != (tdata1_rb & ~maskmax);
+	const uint32_t type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
+	if (type == CSR_TDATA1_TYPE_MCONTROL &&
+	    (get_field(tdata1, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT)) {
+	    tdata1_config_denied |= (get_field(tdata1_rb, CSR_MCONTROL_MASKMAX(riscv_xlen(target)))
+				    < (count_trailing_ones(tdata2) + 1));
+	}
```
JanMatCodasip: 1) A third boolean flag for should be introduced here, e.g. `unsupported_napot_range`. The existing flag `tdata1_config_denied should` should not be re-used because these are two different error situations. There also should be different log messages for each (please see below).

2) Also the type of the trigger (line 657) should be taken from `tdata1_rb`. That's because the target can decide to disable the trigger (by changing the trigger type to `15`), and in that case we cannot get any of the `mcontrol` fields (`mcontrol.match`, `mcontrol.maskmax`) - all those checks become irrelevant.

3) Since the conditions are quite complex, you can improve the code readability a lot by creating helper variables like `is_mcontrol`, `is_napot_match`, ...

```suggestion
	bool tdata1_config_denied = (tdata1 & ~maskmax) != (tdata1_rb & ~maskmax);
	
	bool unsupported_napot_range = false;
	const uint32_t type = get_field(tdata1_rb, CSR_TDATA1_TYPE(riscv_xlen(target)));
	const bool is_mcontrol = type == CSR_TDATA1_TYPE_MCONTROL;
	const bool is_napot_match = get_field(tdata1, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT;
	if (is_mcontrol && is_napot_match) {
		const riscv_reg_t maskmax = get_field(tdata1_rb, CSR_MCONTROL_MASKMAX(riscv_xlen(target));
		const unsigned napot_size = (count_trailing_ones(tdata2) + 1);
		if (maskmax < napot_size) {
			unsupported_napot_range = true;			
		}
	}
```

Error message for `unsupported_napot_range`:

```c
if (tdata1_config_denied || tdata2_config_denied || unsupported_napot_range) {
	// ...

	if (unsupported_napot_range)
		LOG_TARGET_DEBUG(target,
			"The requested NAPOT match range (tdata2=0x%" PRIx64 ") exceeds maxmask=0x%" PRIx64,
			... );
	// ...
}
```

JanMatCodasip (changes requested):
Hi @sunnyzhu-learning,

thanks for your contribution! Overall, I like the improvement.

I do have several review comments, though. If you can please look at them.

Thank you.

src/target/riscv/riscv.c
JanMatCodasip: Small detail: It would be good to make the data type match the return type of the function.

```suggestion
	for (unsigned int i = 0; i < 64; i++) {
```

src/target/riscv/riscv.c
en-sc: Maybe dropping this field altogether is better?
`maskmax`/`tdata1_ignore_mask` is used when `tdata1.type` is `mcontrol` (2).
This commit already derives `type` from `tdata1` in `set_trigger()`.
https://github.com/riscv-collab/riscv-openocd/blob/f0bbdb37724f2327bce6f4603273266a02a1d9b6/src/target/riscv/riscv.c#L657

Wouldn't ignoring `mcontrol.maskmax` based on `tdat1.type` be better:
```
	const uint32_t type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
	const riscv_reg_t tdata1_ignore_mask = (type == CSR_TDATA1_TYPE_MCONTROL) ?
			CSR_MCONTROL_MASKMAX(riscv_xlen(target)) : 0;
	bool tdata1_config_denied = (tdata1 & ~tdata1_ignore_mask) != (tdata1_rb & ~tdata1_ignore_mask);
	if (type == CSR_TDATA1_TYPE_MCONTROL &&
	    (get_field(tdata1, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT)) {
	    tdata1_config_denied |= (get_field(tdata1_rb, CSR_MCONTROL_MASKMAX(riscv_xlen(target)))
				    < (count_trailing_ones(tdata2) + 1));
	}
```
@JanMatCodasip, what are your thoughts on this?

JanMatCodasip: I like that solution :+1:. It is probably better to go that way, compared to what I originally suggested.

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: I have dropped maskmax,this value derives in `set_trigger()`

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

en-sc:
@sunnyzhu-learning, what is the reason for closing the PR?
Please, consider reopening. I believe the patch to be very valuable.

sunnyzhu-learning:
Q:what is the reason for closing the PR?
Please, consider reopening. I believe the patch to be very valuable.

A:my branch updated,this PR deleted,and not reopen again ,so I recreated a PR(https://github.com/riscv-collab/riscv-openocd/pull/1133) for this
image

sunnyzhu-learning:
> @sunnyzhu-learning, what is the reason for closing the PR? Please, consider reopening. I believe the patch to be very valuable.

reopen this PR,solved code merge conflict,I'm very sorry,poor use of PR,@en-sc @JanMatCodasip please code review,thanks....

src/target/riscv/riscv.c
```diff
@@ -633,7 +642,7 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
 		return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
 	}
 
-	riscv_reg_t tdata1_rb, tdata2_rb;
+	riscv_reg_t tdata1_rb, tdata2_rb, maskmax;
```

The variable `maskmax` can be declared very close to the place of the use, which improves code readability.

```suggestion
	riscv_reg_t tdata1_rb, tdata2_rb;
```

src/target/riscv/riscv.c
```diff
@@ -655,20 +664,41 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
 		return ERROR_FAIL;
 	if (riscv_reg_get(target, &tdata2_rb, GDB_REGNO_TDATA2) != ERROR_OK)
 		return ERROR_FAIL;
-	bool tdata1_config_denied = (tdata1 & ~tdata1_ignore_mask) != (tdata1_rb & ~tdata1_ignore_mask);
+	bool unsupported_napot_range = false;
+	bool tdata1_config_denied = true;
+	const uint32_t type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
+	const bool is_mcontrol = type == CSR_TDATA1_TYPE_MCONTROL;
+	const bool is_napot_match = get_field(tdata1, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT;
```
JanMatCodasip: This check should use the real value (read back from the register), not the one that we attempted to write there.

```suggestion
	const bool is_napot_match = get_field(tdata1_rb, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT;
```

JanMatCodasip: (same as previous)

```suggestion
	const bool is_napot_match = get_field(tdata1_rb, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT;
```

JanMatCodasip: (same)

```suggestion
	const bool is_napot_match = get_field(tdata1_rb, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT;
```

src/target/riscv/riscv.c
```diff
@@ -655,20 +664,41 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
...
+		if (maskmax < napot_size)
+			unsupported_napot_range = true;
+		tdata1_config_denied = (tdata1 & ~maskmax) != (tdata1_rb & ~maskmax);
+	} else if (is_mcontrol && !is_napot_match) {
+		maskmax = get_field(tdata1_rb, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
+		tdata1_config_denied = (tdata1 & ~maskmax) != (tdata1_rb & ~maskmax);
+	} else {
+		tdata1_config_denied = tdata1 != tdata1_rb;
+	}
 	bool tdata2_config_denied = tdata2 != tdata2_rb;
```
JanMatCodasip: Detail: The use of `const` is helpful for code readers.

```suggestion
	const bool tdata2_config_denied = tdata2 != tdata2_rb;
```

src/target/riscv/riscv.c
```diff
@@ -655,20 +664,41 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
...
+	bool tdata1_config_denied = true;
+	const uint32_t type = get_field(tdata1, CSR_TDATA1_TYPE(riscv_xlen(target)));
+	const bool is_mcontrol = type == CSR_TDATA1_TYPE_MCONTROL;
+	const bool is_napot_match = get_field(tdata1, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT;
+	if (is_mcontrol && is_napot_match) {
+		maskmax = get_field(tdata1_rb, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
+		const unsigned int napot_size = count_trailing_ones(tdata2) + 1;
+		if (maskmax < napot_size)
+			unsupported_napot_range = true;
+		tdata1_config_denied = (tdata1 & ~maskmax) != (tdata1_rb & ~maskmax);
```
JanMatCodasip: While refactoring the code, a bug was introduced: There would need to be separate variables for `maskmax_value` and `maskmax_mask`, as shown below.

Instead of just fixing it, please, improve the overall structure of the code as shown in the below comment.

```c
		maskmax_mask = CSR_MCONTROL_MASKMAX(riscv_xlen(target));
		maskmax_value = get_field(tdata1_rb, maskmax_mask);
		const unsigned int napot_size = count_trailing_ones(tdata2) + 1;
		if (maskmax_value < napot_size)
			unsupported_napot_range = true;
		tdata1_config_denied = (tdata1 & ~maskmax_mask) != (tdata1_rb & ~maskmax_mask);
```

src/target/riscv/riscv.c
```diff
@@ -655,20 +664,41 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
...
+	} else if (is_mcontrol && !is_napot_match) {
+		maskmax = get_field(tdata1_rb, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
+		tdata1_config_denied = (tdata1 & ~maskmax) != (tdata1_rb & ~maskmax);
+	} else {
+		tdata1_config_denied = tdata1 != tdata1_rb;
+	}
```
JanMatCodasip: This part of the code combines two almost-independent checks:

- check for the TDATA1 value
- check for unsupported mcontrol NAPOT range

For that reason, please, split it into two separate checks. Don't try to merge both the checks into a single if-else cascade. That way it the code will be easier to comprehend, and bugs like the above will hopefully become less likely.

Also please use blank lines to visually separate related blocks of code. 

Thanks.

```c
/* Determine if tdata1 supports what we need.
 * For mcontrol triggers, we don't care about the value in the read-only "maskmax" field. */
const riscv_reg_t tdata1_ignore_mask = is_mcontrol ? CSR_MCONTROL_MASKMAX(riscv_xlen(target)) ? 0;
const riscv_reg_t tdata1_config_denied = (tdata1 & ~tdata1_ignore_mask) != (tdata1_rb & ~tdata1_ignore_mask);

/* Determine if tdata1.maxmask is sufficient
 * (only relevant for mcontrol triggers and NAPOT match type) */
riscv_reg_t unsupported_napot_range = false;
if (is_mcontrol && is_napot_match) {
	// ...
}
```

src/target/riscv/riscv.c
JanMatCodasip: Thanks.

src/target/riscv/riscv.c
```diff
@@ -655,20 +664,41 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
...
 			LOG_TARGET_DEBUG(target,
-				"wrote 0x%" PRIx64 " to tdata2 but read back 0x%" PRIx64,
+				"After writing 0x%" PRIx64 " to tdata2 it contains 0x%" PRIx64,
 				tdata2, tdata2_rb);
+
+		if (unsupported_napot_range)
+			LOG_TARGET_DEBUG(target,
+				"The requested NAPOT match range (tdata2=0x%" PRIx64 ") exceeds maskmax=0x%" PRIx64,
+				tdata2, maskmax);
 		if (riscv_reg_set(target, GDB_REGNO_TDATA1, 0) != ERROR_OK)
```
JanMatCodasip: Detail: Blank line can be used here to maintain consistent style. 

```suggestion
				tdata2, maskmax);

		if (riscv_reg_set(target, GDB_REGNO_TDATA1, 0) != ERROR_OK)
```

JanMatCodasip (changes requested):
@sunnyzhu-learning Thank you for addressing our feedback so far. 

I have several more comments. If you can please look at them.

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: thanks a lot, before:line：677 value `maskmax` is a bug

src/target/riscv/riscv.c
```diff
@@ -655,20 +664,50 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
...
+
+	/* Determine if tdata1.maxmask is sufficient
+	 * (only relevant for mcontrol triggers and NAPOT match type)
+	 */
+	bool unsupported_napot_range = false;
+	const bool is_napot_match = get_field(tdata1_rb, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT;
+	riscv_reg_t maskmax_mask, maskmax_value;
+	if (is_mcontrol && is_napot_match) {
+		maskmax_mask = CSR_MCONTROL_MASKMAX(riscv_xlen(target));
+		maskmax_value = get_field(tdata1_rb, maskmax_mask);
```
JanMatCodasip: For safety, please initialize `maskmax_value`.

Variable `maskmax_mask` can be const. Alternatively, since it is now used in a single place only, it can be removed completely. (Please pick which solution you prefer.)

```suggestion
	riscv_reg_t maskmax_value = 0;
	if (is_mcontrol && is_napot_match) {
		const riscv_reg_t maskmax_mask = CSR_MCONTROL_MASKMAX(riscv_xlen(target));
		maskmax_value = get_field(tdata1_rb, maskmax_mask);
```
or

```suggestion
	riscv_reg_t maskmax_value = 0;
	if (is_mcontrol && is_napot_match) {
		maskmax_value = get_field(tdata1_rb, CSR_MCONTROL_MASKMAX(riscv_xlen(target)));
```

src/target/riscv/riscv.c
```diff
@@ -655,20 +664,50 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
...
+		const unsigned int napot_size = count_trailing_ones(tdata2) + 1;
+		if (maskmax_value < napot_size)
+			unsupported_napot_range = true;
+	}
```
JanMatCodasip: One more suggestion: Please wrap the whole maxmask check into `if (!tdata1_config_denied) { ... }`.

Reasons:

- If it is already determined that tdata1 does not support the trigger options that we need, it would be logically incorrect to continue with other checks. 
- Furthermore, `tdata1.type` may even change to something else (e.g. disabled trigger), in which case mcontrol.maskmax field won't even be available for us to read.

```c
	bool unsupported_napot_range = false;
	riscv_reg_t maskmax_value = 0;
	if (!tdata1_config_denied) {
		const bool is_napot_match = get_field(tdata1_rb, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT;
		if (is_mcontrol && is_napot_match) {
			// ...
		}
	}
```

or

```c
	bool unsupported_napot_range = false;
	riscv_reg_t maskmax_value = 0;
	const bool is_napot_match = get_field(tdata1_rb, CSR_MCONTROL_MATCH) == CSR_MCONTROL_MATCH_NAPOT;
	if (!tdata1_config_denied && is_mcontrol && is_napot_match) {
		// ...
	}
```

(Please pick one of the variants.)

JanMatCodasip (commented):
@sunnyzhu-learning Thank you for addressing the review feedback so far.

I have posted two last comments (and that should be all from me).

src/target/riscv/riscv.c
sunnyzhu-learning: Thanks for the explanation, this is a very rigorous implementation, I will wrap whole maskmax in next patch

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

JanMatCodasip (approved):
LGTM.

@sunnyzhu-learning, thank you for your contribution.

src/target/riscv/riscv.c
```diff
@@ -621,8 +621,17 @@ static int find_first_trigger_by_id(struct target *target, int unique_id)
 	return -1;
 }
 
-static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdata1, riscv_reg_t tdata2,
-	riscv_reg_t tdata1_ignore_mask)
+static unsigned int count_trailing_ones(riscv_reg_t reg)
+{
+	assert(sizeof(riscv_reg_t) * 8 == 64);
+	for (unsigned int i = 0; i < 64; i++) {
```
en-sc, line 627:
```suggestion
	const unsigned int riscv_reg_bits = sizeof(riscv_reg_t) * CHAR_BIT;
	for (unsigned int i = 0; i < riscv_reg_bits; i++) {
```

src/target/riscv/riscv.c
```diff
@@ -621,8 +621,17 @@ static int find_first_trigger_by_id(struct target *target, int unique_id)
...
+		if ((1 & (reg >> i)) == 0)
+			return i;
+	}
+	return 64;
```
en-sc, line 631:
```suggestion
	return riscv_reg_bits;
```

en-sc (approved):

sunnyzhu-learning:
@en-sc  Sorry，No timely feedback，I took a seven-day vacation，next PR will modify above changes~

PR #1128, opened 2024-09-05
try all triggers when setting itrigger/etrigger/icount breakpoints

aap-sc:
in addition this PR includes some minor cleanups related to trigger disabling

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2036,13 +2036,6 @@ static int examine(struct target *target)
 				info->impebreak);
 	}
 
-	if (info->progbufsize < 4 && riscv_virt2phys_mode_is_hw(target)) {
-		LOG_TARGET_ERROR(target, "software address translation "
-				"is not available on this target. It requires a "
-				"program buffer size of at least 4. (progbufsize=%d) "
-				"Use `riscv set_enable_virtual off` to continue.", info->progbufsize);
-	}
-
 	/* Don't call any riscv_* functions until after we've counted the number of
 	 * cores and initialized registers. */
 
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv.c
```diff
@@ -1133,27 +1161,30 @@ static int maybe_add_trigger_t3(struct target *target, bool vs, bool vu,
...
+					/* chained */ false, &idx) == ERROR_OK;
+			++idx) {
+		ret = sdtrig_set_t1(target, idx, tdata1, /* tdata1 ignore mask */ 0);
+		if (ret == ERROR_OK) {
+			r->trigger_unique_id[idx] = unique_id;
+			return ERROR_OK;
+		}
+		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
+			break;
+	}
```
JanMatCodasip: The loop can be re-written to avoid the complex expressions inside the `for`. I also tried to add few more comments to make the code easier to grasp for newcomers.

The resulting code is bit longer but hopefully easier to read.

Please let me know if you like it.

```suggestion
	unsigned int idx = 0;
	while (true) {
		int ret = find_next_free_trigger(target, CSR_TDATA1_TYPE_ICOUNT,
					/* chained */ false, &idx);
		if (ret != ERROR_OK) {
			/* No more free triggers, or an error occurred. */
			return ret;
		}
		
		/* A free trigger has been found whose tselect index is now in idx. */
		ret = sdtrig_set_t1(target, idx, tdata1, /* tdata1 ignore mask */ 0);
		if (ret == ERROR_OK) {
			/* Success: the trigger has been set */
			r->trigger_unique_id[idx] = unique_id;
			return ERROR_OK;
		}
		
		if (ret != ERROR_TARGET_RESOURCE_NOT_AVAILABLE)
			/* An error occurred during setting the trigger */
			return ret;
		
		/* The trigger did not support the options that we required.
		 * Try the next free trigger. */
		 idx++;
	}
```

src/target/riscv/riscv.c
```diff
@@ -666,6 +669,33 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
...
+			tdata1_ignore_mask, /* update tdata2 */ false);
+}
+
+static int sdtrig_set_t1t2(struct target *target, unsigned int idx,
+		riscv_reg_t tdata1, riscv_reg_t tdata2, riscv_reg_t tdata1_ignore_mask)
+{
+	return stdtrig_set_tx_impl(target, idx, tdata1, tdata2, tdata1_ignore_mask,
+			/* update tdata2 */ true);
+}
+
```
JanMatCodasip: I'd recommend to rename the functions to say `tdata1`, `tdata2` instead of `t1`, `t2`. The reason is that the `tN` convention is already used to denote trigger types in `maybe_add_trigger_tN`.

Possible names could be:

- `sdtrig_set_t1()` --> `sdtrig_set_tdata1()`
- `sdtrig_set_t1t2()` --> `sdtrig_set_tdata1_2()`
- `stdtrig_set_tx_impl()` --> `stdtrig_set_tdata_impl()`

Or you can come up with another naming convention; my point is just to have different names for `maybe_add_trigger_*()` and `sdtrig_set_*()`.

src/target/riscv/riscv.c
```diff
@@ -1420,30 +1458,46 @@ static int remove_trigger(struct target *target, int unique_id)
...
+		LOG_TARGET_ERROR(target,
+				"BUG: could not find the hardware resources used by bp/wp %d",
+				unique_id);
+		ret = ERROR_FAIL;
 	}
 
-	riscv_reg_set(target, GDB_REGNO_TSELECT, tselect);
+	if (riscv_reg_set(target, GDB_REGNO_TSELECT, tselect) != ERROR_OK) {
+		LOG_TARGET_ERROR(target, "Coudld not restore tselect after bp/wp %d "
+				"removal attempt", unique_id);
```
JanMatCodasip: Typo fix: Could

I also tried to simplify the message a little while retaining the meaning.

```suggestion
		LOG_TARGET_ERROR(target, "Could not restore tselect after removal of bp/wp %d ", unique_id);
```

JanMatCodasip (commented):
I have only done superficial review so far and am sending several early comments. Overall the changes look fine - thank you.

I will need to return to this PR next week to re-read all the code properly, though.

src/target/riscv/riscv.c
aap-sc: I like the naming scheme you've proposed. Will try to use it.

As for `maybe_add_trigger_`  - I don't like this inconsistency too.. But I think it would be worth to move all trigger-related code to a separate file and do rename during this activity.

src/target/riscv/riscv.c
aap-sc: oh, thanks

src/target/riscv/riscv.c
aap-sc: I think it is definitely easier to read, though my personal aesthetic feelings are hurt whenever I see something like `while(true)/while(1)`. I'll try to piggy-back on your suggestion and either adopt it or rework a little bit

JanMatCodasip:
Hello @aap-sc, I just wanted to note that this merge request looks like a nice improvement, so I believe it is very worth to continue with it (if you are interested and have the capacity, of course).

aap-sc:
> Hello @aap-sc, I just wanted to note that this merge request looks like a nice improvement, so I believe it is very worth to continue with it (if you are interested and have the capacity, of course).

I plan to continue this, but in this specific case I was quite busy with other stuff, I do hope that I'll wrap this up soon.

aap-sc:
Ouch. Closed by mistake. Will try to re-animate this MR

PR #1129, opened 2024-09-06
target/riscv: Fix calloc calls.

rtwfroody:
This was pointed out by gcc. Presumably it's a newer warning. I doubt it has any effect on anything.


aap-sc (approved):
@rtwfroody welcome back! :)

@en-sc FYI - the warning was introduced in 5a8697b3cf5c4e4abde77ac1726a76877592dfa9

JanMatCodasip (approved):

PR #1130, opened 2024-09-10
Revert "target/riscv: re-apply patch do stop avoid warnings when a no… …n-existent CSR is hidden"

en-sc:
This PR fixes the issue with #1118 by reverting commit https://github.com/riscv-collab/riscv-openocd/commit/e56dc61697e91cf7273476ec3126078692a5e387 and re-applying https://github.com/riscv-collab/riscv-openocd/commit/b201a5db23c0db34c0e10fd1c7c08fc73a5ec3fc

The reverted commit claims to be the same as https://github.com/riscv-collab/riscv-openocd/commit/b201a5db23c0db34c0e10fd1c7c08fc73a5ec3fc, but it's not -- it changes the warning in `riscv_reg_impl_expose_csrs()` instead of the one in `riscv_reg_impl_hide_csrs()`.


aap-sc (approved):
thanks

JanMatCodasip (approved):

PR #1134, opened 2024-09-20
target/riscv: early exit support for memory access operations

fkhaidari:
(1) Error code and `skip_reason` string were replaced with memory access
    status. It allows to specify whether OpenOCD should exit the access
    early.
(2) Slightly refactored `read_memory` and 'write_memory' functions.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2168,8 +2168,8 @@ static unsigned riscv013_data_bits(struct target *target)
 	RISCV013_INFO(info);
 	RISCV_INFO(r);
 
-	for (unsigned int i = 0; i < RISCV_NUM_MEM_ACCESS_METHODS; i++) {
-		int method = r->mem_access_methods[i];
+	for (unsigned int i = 0; i < r->num_enabled_mem_access_methods; i++) {
+		riscv_mem_access_method_t method = r->mem_access_methods[i];
 
 		if (method == RISCV_MEM_ACCESS_PROGBUF) {
 			if (has_sufficient_progbuf(target, 3))
@@ -2190,9 +2190,9 @@ static unsigned riscv013_data_bits(struct target *target)
 			 * take those into account as well.  For now we assume abstract commands
 			 * support XLEN-wide accesses. */
 			return riscv_xlen(target);
-		} else if (method == RISCV_MEM_ACCESS_UNSPECIFIED)
-			/* No further mem access method to try. */
-			break;
+		} else {
+			assert(false);
+		}
 	}
 	LOG_TARGET_ERROR(target, "Unable to determine supported data bits on this target. Assuming 32 bits.");
 	return 32;
@@ -3389,7 +3389,7 @@ static int read_memory_bus_v1(struct target *target, target_addr_t address,
 	return ERROR_OK;
 }
 
-static void log_mem_access_result(struct target *target, bool success, int method, bool is_read)
+static void log_mem_access_result(struct target *target, bool success, riscv_mem_access_method_t method, bool is_read)
 {
 	RISCV_INFO(r);
 	bool warn = false;
@@ -3404,18 +3404,8 @@ static void log_mem_access_result(struct target *target, bool success, int metho
 
 	/* Determine the log message severity. Show warnings only once. */
 	if (!success) {
-		if (method == RISCV_MEM_ACCESS_PROGBUF) {
-			warn = r->mem_access_progbuf_warn;
-			r->mem_access_progbuf_warn = false;
-		}
-		if (method == RISCV_MEM_ACCESS_SYSBUS) {
-			warn = r->mem_access_sysbus_warn;
-			r->mem_access_sysbus_warn = false;
-		}
-		if (method == RISCV_MEM_ACCESS_ABSTRACT) {
-			warn = r->mem_access_abstract_warn;
-			r->mem_access_abstract_warn = false;
-		}
+		warn = r->mem_access_warn[method];
+		r->mem_access_warn[method] = false;
 	}
 
 	if (warn)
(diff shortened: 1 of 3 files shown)


fkhaidari:
@MarekVCodasip, @JanMatCodasip, please take a look

en-sc (dismissed):
LGTM (reviewed internally).

src/target/riscv/riscv.h
```diff
@@ -272,13 +270,13 @@ struct riscv_info {
 	bool *reserved_triggers;
 
 	/* Memory access methods to use, ordered by priority, highest to lowest. */
-	int mem_access_methods[RISCV_NUM_MEM_ACCESS_METHODS];
+	riscv_mem_access_method_t mem_access_methods[RISCV_MEM_ACCESS_MAX_METHODS_NUM];
+
+	unsigned int mem_access_method_num;
```
JanMatCodasip: If you like:

This can be renamed to `num_enabled_mem_access_methods` which is longer but hopefully clearer.

The original name could be interpreted by a reader as "number (ID) of a method".

```suggestion
	unsigned int num_enabled_mem_access_methods;
```

src/target/riscv/riscv.c
```diff
@@ -5451,9 +5449,8 @@ static void riscv_info_init(struct target *target, struct riscv_info *r)
 	r->mem_access_methods[1] = RISCV_MEM_ACCESS_SYSBUS;
 	r->mem_access_methods[2] = RISCV_MEM_ACCESS_ABSTRACT;
 
-	r->mem_access_progbuf_warn = true;
-	r->mem_access_sysbus_warn = true;
-	r->mem_access_abstract_warn = true;
+	r->mem_access_method_num = RISCV_MEM_ACCESS_MAX_METHODS_NUM;
+	memset(r->mem_access_warn, true, RISCV_MEM_ACCESS_MAX_METHODS_NUM);
```
JanMatCodasip: Detail but important:

This assumes that `bool` is 1 byte wide which need not be true (https://stackoverflow.com/a/8014188).

For that reason, please replace this by:

```suggestion
	for (unsigned int i = 0; i < r->mem_access_method_num; i++)
		r->mem_access_warn[i] = true;
```

Generally, it'd be my preference to not use `memset` for initialization of structures (or arrays) for these reasons:

- It is error-prone w.r.t. the object size
- Static analysis tools may not properly recognize this form of initialization
- When searching for assignments to structure or array items, this is easy to miss

src/target/riscv/riscv.c
```diff
@@ -3907,9 +3907,7 @@ COMMAND_HANDLER(riscv_set_mem_access)
 	}
 
 	/* Reset warning flags */
-	r->mem_access_progbuf_warn = true;
-	r->mem_access_sysbus_warn = true;
-	r->mem_access_abstract_warn = true;
+	memset(r->mem_access_warn, true, RISCV_MEM_ACCESS_MAX_METHODS_NUM);
```
JanMatCodasip: (same as in previous comment)

src/target/riscv/riscv-013.c
```diff
@@ -3424,114 +3412,189 @@ static void log_mem_access_result(struct target *target, bool success, int metho
...
 }
 
-static bool mem_should_skip_progbuf(struct target *target, target_addr_t address,
-		uint32_t size, bool is_read, char **skip_reason)
+typedef enum {
+	MEM_ACCESS_STATUS_KIND_OK,
+	MEM_ACCESS_STATUS_KIND_DISABLED,
+	MEM_ACCESS_STATUS_KIND_SKIPPED,
+	MEM_ACCESS_STATUS_KIND_FAILED
+} mem_access_status_kind_t;
```
JanMatCodasip: If you like: `result` is more descriptive than `status`.

```suggestion
typedef enum {
	MEM_ACCESS_RESULT_KIND_OK,
	MEM_ACCESS_RESULT_KIND_DISABLED,
	MEM_ACCESS_RESULT_KIND_SKIPPED,
	MEM_ACCESS_RESULT_KIND_FAILED
} mem_access_result_kind_t;
```

JanMatCodasip: Also you may consider `_RESULT_TYPE_` instead of `_RESULT_KIND_` (but I don't feel strongly).

src/target/riscv/riscv-013.c
```diff
@@ -3424,114 +3412,189 @@ static void log_mem_access_result(struct target *target, bool success, int metho
...
+
+#define LIST_OF_MEM_ACCESS_STATUSES                                                                \
+	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_STATUS_OK,                                            \
+			MEM_ACCESS_STATUS_KIND_OK, "ok")                                           \
+	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_STATUS_DISABLED,                                      \
+			MEM_ACCESS_STATUS_KIND_DISABLED, "disabled")                               \
+	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_STATUS_SKIPPED_ABSTRACT_ACCESS_CMDERR,                \
+			MEM_ACCESS_STATUS_KIND_SKIPPED, "skipped (abstract access cmderr)")        \
+	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_STATUS_SKIPPED_PROGBUF_INSUFFICIENT,                  \
+			MEM_ACCESS_STATUS_KIND_SKIPPED, "skipped (insufficient progbuf)")          \
```
JanMatCodasip: Here it'd be good to introduce also "no progbuf" status:

```
	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_STATUS_SKIPPED_PROGBUF_NOT_PRESENT,                  \
			MEM_ACCESS_STATUS_KIND_SKIPPED, "skipped (progbuf not present)")          \
```

This is however an improvement for a separate PR.

src/target/riscv/riscv-013.c
```diff
@@ -3424,114 +3412,189 @@ static void log_mem_access_result(struct target *target, bool success, int metho
...
+	MEM_ACCESS_STATUS_KIND_FAILED
+} mem_access_status_kind_t;
+
+#define LIST_OF_MEM_ACCESS_STATUSES                                                                \
+	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_STATUS_OK,                                            \
+			MEM_ACCESS_STATUS_KIND_OK, "ok")                                           \
+	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_STATUS_DISABLED,                                      \
+			MEM_ACCESS_STATUS_KIND_DISABLED, "disabled")                               \
+	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_STATUS_SKIPPED_ABSTRACT_ACCESS_CMDERR,                \
+			MEM_ACCESS_STATUS_KIND_SKIPPED, "skipped (abstract access cmderr)")        \
```
JanMatCodasip: If you'd like to shorten the identifiers, you can make the prefix shorter `MEM_ACCESS_STATUS_` --> `MEM_ACCESS_` and they still remain very clear and readable.

```suggestion
	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_OK,                                            \
			MEM_ACCESS_STATUS_KIND_OK, "ok")                                           \
	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_DISABLED,                                      \
			MEM_ACCESS_STATUS_KIND_DISABLED, "disabled")                               \
	MEM_ACCESS_STATUS_HANDLER(MEM_ACCESS_SKIPPED_ABSTRACT_ACCESS_CMDERR,                \
			MEM_ACCESS_STATUS_KIND_SKIPPED, "skipped (abstract access cmderr)")        \
```

src/target/riscv/riscv-013.c
```diff
@@ -3424,114 +3412,189 @@ static void log_mem_access_result(struct target *target, bool success, int metho
...
+
+bool is_mem_access_failed(mem_access_status_t status)
+{
+	#define MEM_ACCESS_STATUS_HANDLER(name, kind, msg) \
+		case name: return kind == MEM_ACCESS_STATUS_KIND_FAILED;
+	switch (status) {
+		LIST_OF_MEM_ACCESS_STATUSES
+	}
+	#undef MEM_ACCESS_STATUS_HANDLER
+	LOG_ERROR("Unknown memory access status: %d", status);
```
JanMatCodasip, line 3471: If you like:

```suggestion
	assert(false);
```

Or both the assertion and error message can be kept:

```suggestion
	LOG_ERROR("Unknown memory access status: %d", status);
	assert(false);	
```

src/target/riscv/riscv-013.c
```diff
@@ -3424,114 +3412,189 @@ static void log_mem_access_result(struct target *target, bool success, int metho
...
+
+bool is_mem_access_skipped(mem_access_status_t status)
+{
+	#define MEM_ACCESS_STATUS_HANDLER(name, kind, msg) \
+		case name: return kind == MEM_ACCESS_STATUS_KIND_SKIPPED;
+	switch (status) {
+		LIST_OF_MEM_ACCESS_STATUSES
+	}
+	#undef MEM_ACCESS_STATUS_HANDLER
+	LOG_ERROR("Unknown memory access status: %d", status);
```
JanMatCodasip, line 3484: Same as in previous comment - `assert(false)` would be good to have here.

src/target/riscv/riscv-013.c
```diff
@@ -3424,114 +3412,189 @@ static void log_mem_access_result(struct target *target, bool success, int metho
...
 {
-	assert(skip_reason);
+	#define MEM_ACCESS_STATUS_HANDLER(name, kind, msg) \
+		[name] = msg,
+	static const char * const table[] = {
+		LIST_OF_MEM_ACCESS_STATUSES
+	};
+	#undef MEM_ACCESS_STATUS_HANDLER
+
+	return table[status];
```
JanMatCodasip, line 3499: If you like:

```suggestion
	assert(status < ARRAY_SIZE(table));
	return table[status];
```

src/target/riscv/riscv-013.c
```diff
@@ -3556,7 +3620,7 @@ static int read_memory_abstract(struct target *target, target_addr_t address,
 			result = write_abstract_arg(target, 1, address + c * size, riscv_xlen(target));
 			if (result != ERROR_OK) {
 				LOG_TARGET_ERROR(target, "Failed to write arg1.");
-				return result;
+				return MEM_ACCESS_STATUS_FAILED_ABSTRACT_ARG_WRITE_FAILED;
```
JanMatCodasip: There could be just one generic return status `MEM_ACCESS_FAILED_DM_ACCESS_ERROR` that would cover all cases when access to the DM registers failed.

src/target/riscv/riscv-013.c
```diff
@@ -3592,32 +3656,42 @@ static int read_memory_abstract(struct target *target, target_addr_t address,
...
+			return MEM_ACCESS_STATUS_FAILED_ABSTRACT_ARG_READ_FAILED;
 		buf_set_u64(p, 0, 8 * size, value);
 
 		if (info->has_aampostincrement == YNM_YES)
 			updateaddr = false;
 		p += size;
 	}
 
-	return result;
+	return (result == ERROR_OK) ? MEM_ACCESS_STATUS_OK : MEM_ACCESS_STATUS_FAILED;
```
JanMatCodasip: It looks like the check for `result == ERROR_OK` is redundant.

If I read the code correctly, the result cannot be different than ERROR_OK when we reach the line 3677.

src/target/riscv/riscv-013.c
```diff
@@ -3685,15 +3759,18 @@ static int write_memory_abstract(struct target *target, target_addr_t address,
...
-			return result;
+			return MEM_ACCESS_STATUS_SKIPPED_ABSTRACT_ACCESS_CMDERR;
 
 		if (info->has_aampostincrement == YNM_YES)
 			updateaddr = false;
 		p += size;
 	}
 
-	return result;
+	return (result == ERROR_OK) ? MEM_ACCESS_STATUS_OK : MEM_ACCESS_STATUS_FAILED;
```

It looks that the check for `result == ERROR_OK` is redundant here.

src/target/riscv/riscv-013.c
```diff
@@ -4274,62 +4373,57 @@ static int read_memory(struct target *target, target_addr_t address,
...
+					read_memory_sysbus(target, address,
+						size, count, buffer, increment);
+				break;
+			case RISCV_MEM_ACCESS_ABSTRACT:
+				skip_reason[method] =
+					read_memory_abstract(target, address,
+						size, count, buffer, increment);
+				break;
+			default:
+				LOG_TARGET_ERROR(target, "Unknown memory access method: %d", method);
```
JanMatCodasip, line 4418: An assert can be used here:

```suggestion
				LOG_TARGET_ERROR(target, "Unknown memory access method: %d", method);
				assert(false);
```

src/target/riscv/riscv-013.c
```diff
@@ -4810,73 +4904,95 @@ static int write_memory_progbuf(struct target *target, target_addr_t address,
...
+					write_memory_sysbus(target, address,
+						size, count, buffer);
+				break;
+			case RISCV_MEM_ACCESS_ABSTRACT:
+				skip_reason[method] =
+					write_memory_abstract(target, address,
+						size, count, buffer);
+				break;
+			default:
+				LOG_TARGET_ERROR(target, "Unknown memory access method: %d", method);
```
JanMatCodasip, line 4989: (same as above)

assert(false) can be used here.

```suggestion
				LOG_TARGET_ERROR(target, "Unknown memory access method: %d", method);
				assert(false);
```

JanMatCodasip (commented):
Thank you for the refactor. Overall it looks good.

I have so far managed approx. 50 % of the review and will continue at the end of this week. I am sending the comments I have so far.

src/target/riscv/riscv.h
Tako-San: Renamed

src/target/riscv/riscv.c
Tako-San: Addressed

src/target/riscv/riscv.h
fkhaidari: Renamed

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
fkhaidari: Addressed

src/target/riscv/riscv-013.c
fkhaidari: Addressed

src/target/riscv/riscv-013.c
fkhaidari: Addressed

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
fkhaidari: Addressed

src/target/riscv/riscv-013.c
fkhaidari: Addressed

src/target/riscv/riscv-013.c
JanMatCodasip: Please:

- either keep the new status and make use of it where suitable,
- or remove it for now, don't leave it unused. It can be added later in a separate PR as an improvement.

My intention was to give a very clear error message to the user: If there is no program buffer in a given target, it is better to say `no progbuf` than `insufficient progbuf`, since the latter could be misleading. The user could think that a progbuf is actually there but not large enough.

(I am sorry for my not very clear previous comment.)

src/target/riscv/riscv-013.c
```diff
@@ -2190,9 +2190,7 @@ static unsigned riscv013_data_bits(struct target *target)
 			 * take those into account as well.  For now we assume abstract commands
 			 * support XLEN-wide accesses. */
 			return riscv_xlen(target);
-		} else if (method == RISCV_MEM_ACCESS_UNSPECIFIED)
-			/* No further mem access method to try. */
-			break;
+		}
```
JanMatCodasip, line 2195: An optional detail (if you like):

```suggestion
		} else {
			assert(false);
		}
```

src/target/riscv/riscv-013.c
```diff
@@ -4263,9 +4343,31 @@ static int read_memory_progbuf(struct target *target, target_addr_t address,
...
+
+	int ret = ERROR_FAIL;
+	uint64_t sbver = get_field(get_info(target)->sbcs, DM_SBCS_SBVERSION);
+	if (sbver == 0)
+		ret = read_memory_bus_v0(target, address, size, count, buffer, increment);
+	else if (sbver == 1)
+		ret = read_memory_bus_v1(target, address, size, count, buffer, increment);
+	else
+		LOG_TARGET_ERROR(target,
+			"Unknown system bus version: %" PRIu64, sbver);
```
JanMatCodasip: Detail: Could you try whether `checkpatch` would permit braces in this if-else cascade? 

If so, it'd improve readability.

```suggestion
	if (sbver == 0) {
		ret = read_memory_bus_v0(target, address, size, count, buffer, increment);
	} else if (sbver == 1) {
		ret = read_memory_bus_v1(target, address, size, count, buffer, increment);
	} else  {
		LOG_TARGET_ERROR(target,
			"Unknown system bus version: %" PRIu64, sbver);
	}
```

src/target/riscv/riscv-013.c
```diff
@@ -4274,62 +4376,58 @@ static int read_memory(struct target *target, target_addr_t address,
...
+				return ERROR_FAIL;
+		}
 
-		log_mem_access_result(target, ret == ERROR_OK, method, true);
+		if (is_mem_access_failed(skip_reason[method]))
+			goto failure;
 
-		if (ret == ERROR_OK)
-			return ret;
+		bool success = (skip_reason[method] == MEM_ACCESS_OK);
```

```suggestion
		const bool success = (skip_reason[method] == MEM_ACCESS_OK);
```

src/target/riscv/riscv-013.c
```diff
@@ -4274,62 +4376,58 @@ static int read_memory(struct target *target, target_addr_t address,
 	if (count == 0)
 		return ERROR_OK;
 
-	if (size != 1 && size != 2 && size != 4 && size != 8 && size != 16) {
+	if (!IS_PWR_OF_2(size) || size > 16) {
```
JanMatCodasip: The macro `IS_PWR_OF_2` in align.h is buggy because it evaluates to true even for `0` which is not a power of 2.

I have submitted this little fix to upstream (https://review.openocd.org/c/openocd/+/8511) but it will take some time for it to propagate back to riscv-openocd.

For that reason, please, consider extending the check:

```suggestion
	if (!IS_PWR_OF_2(size) || size < 1 || size > 16) {
```

src/target/riscv/riscv-013.c
```diff
@@ -4810,73 +4908,96 @@ static int write_memory_progbuf(struct target *target, target_addr_t address,
...
-
-			ret = write_memory_progbuf(target, address, size, count, buffer);
+	uint64_t sbver = get_field(info->sbcs, DM_SBCS_SBVERSION);
+	if (sbver == 0)
+		ret = write_memory_bus_v0(target, address, size, count, buffer);
+	else if (sbver == 1)
+		ret = write_memory_bus_v1(target, address, size, count, buffer);
+	else
+		LOG_TARGET_ERROR(target,
+			"Unknown system bus version: %" PRIu64, sbver);
```
JanMatCodasip: 1) If checkpatch allows the braces here, it'd help readability.

2) Also a blank line before and after the `if-else` cascade would help (especially to visually separate it from the below `if`).

```suggestion

	if (sbver == 0) {
		ret = write_memory_bus_v0(target, address, size, count, buffer);
	} else if (sbver == 1) {
		ret = write_memory_bus_v1(target, address, size, count, buffer);
	} else {
		LOG_TARGET_ERROR(target,
			"Unknown system bus version: %" PRIu64, sbver);
	}
			
```

JanMatCodasip (commented):
Thank you for addressing the previous comments.

I have completed my review and am sending few last comments. I do not expect to have any more. If you can please check it out, thanks.

src/target/riscv/riscv-013.c
fkhaidari: Addressed

src/target/riscv/riscv-013.c
fkhaidari: Used new status

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
JanMatCodasip: Thank you

JanMatCodasip (approved):
LGTM, thank you.

en-sc (approved):

PR #1135, opened 2024-09-23
Move FD_SETSIZE to configure.ac

borneoa:
The commit from the pull request #644 sets the macro FD_SETSIZE to 128 for every host OS, while it is needed on Windows only. Note that this macro has a much higher value on other OS and after this commit it get reduced to 128 only.

Since in configure.ac is already present a hook to change the compile defines for Windows, move there also FD_SETSIZE.

I have pushed in upstream OpenOCD the patch https://review.openocd.org/c/openocd/+/8503 that, together with this pull request, aligns the two repositories on this topic.
I will wait for review and eventual merge of this pull request before merging the corresponding patch upstream. Of course, the review process is open upstream too.
Hope this helps the sync between the two repositories.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -37,15 +37,13 @@ DISTCLEANFILES += jimtcl/jsmn/jsmn.o
 endif
 
 # common flags used in openocd build
-AM_CFLAGS = $(GCC_WARNINGS)\
-			  -DFD_SETSIZE=128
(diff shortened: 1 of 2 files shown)


en-sc (approved):
Thanks for the patch! LGTM.
@JanMatCodasip, @MarekVCodasip, please take a look.

JanMatCodasip (approved):
Thank you, LGTM.

PR #1136, opened 2024-09-23
target/riscv: DMI logging improvements

### LINKED ISSUE #1043: Stop assuming dmi.address is not changed in `riscv_decode_dmi_scan()`
Related to #1025.

When a DMI read result is decoded, the current code assumes that the hardware provides the read address in the dmi.address field when returning the dmi.data.

That is however not required by the spec, and therefore not guaranteed.

Spike ISA simulator behaves like that but other RISC-V debug implementations may not.

e.g.
```
Debug: 409 19 riscv-013.c:420 dump_field(): 50b r 00000000 @10 -> + 00000000 @10; 3i
Debug: 419 20 riscv-013.c:420 dump_field(): 50b - 00000000 @10 -> + 04010001 @10; 3i
Debug: 420 20 riscv-013.c:434 dump_field(): read: dmcontrol=0x4010001 {...}
```

Here the decoding on the 3-rd line is based on the `dmi.address` field OpenOCD got in the result of the nop opreation on the 2-nd line.

en-sc:
Fixes #1043

There were multiple issuese with DMI logging:
1. Address was assumed to be the same (#1043).
2. Reported IDLE count was not affected by a reset of the delays.
3. VLA were used.

These issues are addressed in the commit.


en-sc:
Note to the reviewers -- this PR contains two commits.
* The first just moves `decode_dmi()` and `riscv_log_dmi_scan()` functions.
* The second addresses the issues outlined in the PR description.

src/target/riscv/batch.c
```diff
@@ -132,18 +132,22 @@ static void add_idle_before_batch(const struct riscv_batch *batch, size_t start_
 }
 
 static int get_delay(const struct riscv_batch *batch, size_t scan_idx,
-		const struct riscv_scan_delays *delays)
+		const struct riscv_scan_delays *delays, bool resets_delays,
+		size_t reset_delays_after)
```
JanMatCodasip, line 136: (Just a note, no change requested)

The items `reset_delays` and `reset_delays_after` are (almost) always passed along with the `delays`. For that reason, they could be moved directly to `struct riscv_scan_delays` which would simplify the API of multiple functions.

This would however require additional changes in `struct riscv_info` and `struct riscv013_info`. 

No need to do it now. It is just an idea for a possible future refactor.

src/target/riscv/batch.c
```diff
@@ -172,51 +177,85 @@ static unsigned int decode_dmi(const struct target *target, char *text, uint32_t
 	return 0;
```
JanMatCodasip, line 177: Please, could you help me check this part. It seems to me that we should return value `1`.

Reason: If no text is to be printed, we need to allocate a buffer of size 1 for the empty string - for the terminating `\0`.

src/target/riscv/batch.c
```diff
@@ -235,18 +275,16 @@ int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
...
 
-		const bool delays_were_reset = resets_delays
-			&& (i >= reset_delays_after);
-		const int delay = get_delay(batch, i, delays);
-
-		if (!delays_were_reset)
-			jtag_add_runtest(delay, TAP_IDLE);
+		delay = get_delay(batch, i, delays, resets_delays,
+				reset_delays_after);
+		jtag_add_runtest(delay, TAP_IDLE);
```
JanMatCodasip: (This relates to the original code, not to the actual changes.)

Detail:

I have noticed that it is not necessary to call jtag_add_runtest() at all when `delay == 0`. 

Even though `jtag_add_runtest(0, TAP_IDLE)` should be treated as no-op by any reasonable adapter [^1], I'd add an `if` check for robustness:

```
if (delay > 0)
    jtag_add_runtest(delay, TAP_IDLE);
```

Furthermore, readers might wonder if `jtag_add_runtest(0, TAP_IDLE)` actually has any meaning, which it does not have.

[^1]: I checked few of them: jlink, remote_bitbang, jtag_vpi

src/target/riscv/batch.c
```diff
@@ -172,51 +177,85 @@ static unsigned int decode_dmi(const struct target *target, char *text, uint32_t
...
-			 * required by the spec, and therefore not guaranteed.
-			 * See https://github.com/riscv-collab/riscv-openocd/issues/1043
-			 */
-			LOG_DEBUG("read: %s", in_decoded);
-		}
-	} else {
-		LOG_DEBUG("%db %s %08" PRIx32 " @%02" PRIx32 " -> ?; %di",
-				field->num_bits, op_string[out_op], out_data, out_address,
-				idle);
+	bool reads = false;
```
JanMatCodasip: Please, if you could give this variable a more descriptive name (to help code readers).

For instance:

`bool last_scan_was_read = false;`

(or something similar)

src/target/riscv/batch.c
```diff
@@ -172,51 +177,85 @@ static unsigned int decode_dmi(const struct target *target, char *text, uint32_t
...
+	uint32_t last_address = -1 /* to silence maybe-uninitialized */;
```
JanMatCodasip: This could be renamed to `last_scan_address` (or similar).

src/target/riscv/batch.c
```diff
@@ -131,22 +132,139 @@ static void add_idle_before_batch(const struct riscv_batch *batch, size_t start_
...
+static void log_batch(const struct riscv_batch *batch, size_t start_idx,
+		const struct riscv_scan_delays *delays, bool resets_delays,
+		size_t reset_delays_after)
+{
+	if (debug_level < LOG_LVL_DEBUG)
+		return;
+
+	bool reads = false;
+	uint32_t last_address = -1 /* to silence maybe-uninitialized */;
+	if (start_idx > 0) {
```
JanMatCodasip, line 215: This part assumes that the previous batch was the same as the current, right? That is, the caller does not mix multiple batches.

If that is correct understanding, maybe you can mention it in a comment.

JanMatCodasip, line 215: If you like: Blank line for readability + comment.

```suggestion

	/* Determine the "op" and "address" of the scan that preceded the first executed scan. */
	if (start_idx > 0) {
```

src/target/riscv/batch.c
```diff
@@ -131,22 +132,139 @@ static void add_idle_before_batch(const struct riscv_batch *batch, size_t start_
...
+		const struct scan_field * const field = &batch->fields[start_idx - 1];
+		assert(field->out_value);
+		reads = buf_get_u32(field->out_value, DTM_DMI_OP_OFFSET,
+				DTM_DMI_OP_LENGTH) == DTM_DMI_OP_READ;
+		const unsigned int abits = field->num_bits - DTM_DMI_OP_LENGTH
+			- DTM_DMI_DATA_LENGTH;
+		last_address = buf_get_u32(field->out_value,
+				DTM_DMI_ADDRESS_OFFSET, abits);
+	}
+	for (size_t i = start_idx; i < batch->used_scans; ++i) {
```
JanMatCodasip: If you like:

```suggestion
	}

	/* Decode and log every executed scan */
	for (size_t i = start_idx; i < batch->used_scans; ++i) {
```

src/target/riscv/batch.c
```diff
@@ -131,22 +132,139 @@ static void add_idle_before_batch(const struct riscv_batch *batch, size_t start_
...
+
+	bool reads = false;
+	uint32_t last_address = -1 /* to silence maybe-uninitialized */;
+	if (start_idx > 0) {
+		const struct scan_field * const field = &batch->fields[start_idx - 1];
+		assert(field->out_value);
+		reads = buf_get_u32(field->out_value, DTM_DMI_OP_OFFSET,
+				DTM_DMI_OP_LENGTH) == DTM_DMI_OP_READ;
+		const unsigned int abits = field->num_bits - DTM_DMI_OP_LENGTH
+			- DTM_DMI_DATA_LENGTH;
```
JanMatCodasip: I'd recommend to compute `abits` only once at the top of the function, since it is a constant for the given target.

```
const unsigned int abits = riscv_get_dmi_scan_length(batch->target) - DTM_DMI_OP_LENGTH - DTM_DMI_DATA_LENGTH;
```

Otherwise the reader might get confused why we recompute it every time, and wonder whether it could be different for each scan.

Optionally -- for additional safety -- you may also add an assertion for each of the scans:

```
assert(field->num_bits == riscv_get_dmi_scan_length(batch->target));
```

src/target/riscv/batch.c
JanMatCodasip: Alternatively, it would be possible to record `last_scan_was_read` and `last_scan_address` into `dm013_info_t`. But that is just an idea. No need to go that far in this PR.

JanMatCodasip (commented):
Thank you for this improvement! It looks very fine overall.

I have posted several smaller comments, if you can please take a look at them.

src/target/riscv/batch.c

src/target/riscv/batch.c

src/target/riscv/batch.c
en-sc: Addressed.

en-sc: Thanks for the catch! Added a `FIXME`.

src/target/riscv/batch.c

src/target/riscv/batch.c

src/target/riscv/batch.c

src/target/riscv/batch.c
en-sc: It's not the size of the buffer returned here, it's the length of the string (similar to `sprintf()`).
So for an empty string it's zero.

src/target/riscv/batch.c
en-sc: 1. `reset_delays` and `reset_delays_after`  are used for testing purposes only. IMHO, they should be removed one day, together with `riscv reset_delays` command.

src/target/riscv/batch.c
JanMatCodasip: You're right. I missed the `+ 1` at the end of line 183 below.

Thank you for explanation.

src/target/riscv/batch.c
JanMatCodasip: It sounds OK to me to remove this testing-only feature one day.

However, it would be good to replace it by a target that allows to test various busy states (in DMI access, abstract command and system bus access). I have been thinking about writing a simple target mock for that purpose, that would simulate all these cases, and would speak remote_bitbang protocol like Spike. However, I do not expect to find the time for that anytime soon :/

src/target/riscv/batch.c
```diff
@@ -131,22 +132,149 @@ static void add_idle_before_batch(const struct riscv_batch *batch, size_t start_
...
+}
+
+static void log_batch(const struct riscv_batch *batch, size_t start_idx,
+		const struct riscv_scan_delays *delays, bool resets_delays,
+		size_t reset_delays_after)
+{
+	if (debug_level < LOG_LVL_DEBUG)
+		return;
+
+	const unsigned int scan_bits = batch->fields->num_bits;
```
JanMatCodasip, line 201: Pedantic detail:

You can use `fields[0].num_bits` instead of `fields->num_bits` to make it clearer that `fields` is pointing to an array of items, not just a single item.

```suggestion
	assert(batch->used_scans > 0);
	const unsigned int scan_bits = batch->fields[0].num_bits;
```

JanMatCodasip (approved):
LGTM, thank you!

PR #1141, opened 2024-10-03
target/riscv: Consistent data type for trigger indexes

JanMatCodasip:
Cosmetic detail: Use `unsigned int` to index triggers (as opposed to `riscv_reg_t`), which corresponds to
the data type of `trigger_count` in `struct riscv_info_t`.

Change-Id: I83539abdffa41aec2060fbd0c81496ab9607c9ea

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -5023,9 +5023,9 @@ static COMMAND_HELPER(report_reserved_triggers, struct target *target)
 	if (riscv_enumerate_triggers(target) != ERROR_OK)
 		return ERROR_FAIL;
 	const char *separator = "";
-	for (riscv_reg_t t = 0; t < r->trigger_count; ++t) {
+	for (unsigned int t = 0; t < r->trigger_count; ++t) {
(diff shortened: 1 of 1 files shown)


en-sc (changes requested):
IMHO, it's better to use `riscv_reg_t`. Trigger index is the value of `tselect`. AFAIK, the spec allows for any number of triggers that should fit into `tselect` (up to 2^`XLEN`).

Current OpenOCD version limits the number of triggers, but this limit is completely arbitrary.

JanMatCodasip:
> en-sc: IMHO, it's better to use riscv_reg_t. [...]

Thanks for your reply. I agree and I am closing this merge request.

I have started looking at the data types more broadly with the hope that `-Wconversion` compiler warning could be enabled on the RISC-V target code, giving us more confidence when making changes.

I have submitted first two merge requests on that topic already:
- https://github.com/riscv-collab/riscv-openocd/pull/1202
- https://github.com/riscv/riscv-debug-spec/pull/1099

I plan to submit more related patches later.

PR #1144, opened 2024-10-09
target/riscv:Perform single step before resume if necessary

sunnyzhu-learning:
Reason for modification：
1. When `wfi` in starting, this scene does not require single step behavior;
2. When dpc equal to Trigger address, this scene will cause hart not halted.

@en-sc @JanMatCodasip please code review~

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -623,12 +623,12 @@ static int find_first_trigger_by_id(struct target *target, int unique_id)
 
 static unsigned int count_trailing_ones(riscv_reg_t reg)
 {
-	assert(sizeof(riscv_reg_t) * 8 == 64);
-	for (unsigned int i = 0; i < 64; i++) {
+	const unsigned int riscv_reg_bits = sizeof(riscv_reg_t) * CHAR_BIT;
+	for (unsigned int i = 0; i < riscv_reg_bits; i++) {
 		if ((1 & (reg >> i)) == 0)
 			return i;
 	}
-	return 64;
+	return riscv_reg_bits;
 }
 
 static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdata1, riscv_reg_t tdata2)
@@ -1561,21 +1561,75 @@ int riscv_remove_watchpoint(struct target *target,
 	return ERROR_OK;
 }
 
+typedef enum {
+	M6_HIT_ERROR,
+	M6_HIT_NOT_SUPPORTED,
+	M6_NOT_HIT,
+	M6_HIT_BEFORE,
+	M6_HIT_AFTER,
+	M6_HIT_IMM_AFTER
+} mctrl6hitstatus;
+
+static mctrl6hitstatus check_mcontrol6_hit_status(struct target *target,
+		riscv_reg_t tdata1, uint64_t hit_mask)
+{
+	const uint32_t hit0 = get_field(tdata1, CSR_MCONTROL6_HIT0);
+	const uint32_t hit1 = get_field(tdata1, CSR_MCONTROL6_HIT1);
+	const uint32_t hit_info = (hit1 << 1) | hit0;
+	if (hit_info == CSR_MCONTROL6_HIT0_BEFORE)
+		return M6_HIT_BEFORE;
+
+	if (hit_info == CSR_MCONTROL6_HIT0_AFTER)
+		return M6_HIT_AFTER;
+
+	if (hit_info == CSR_MCONTROL6_HIT0_IMMEDIATELY_AFTER)
+		return M6_HIT_IMM_AFTER;
+
+	if (hit_info == CSR_MCONTROL6_HIT0_FALSE) {
+		/* hit[1..0] equals 0, which can mean one of the following:
+		 * - "hit" bits are supported and this trigger has not fired
+		 * - "hit" bits are not supported on this trigger
+		 * To distinguish these two cases, try writing all non-zero bit
+		 * patterns to hit[1..0] to determine if the "hit" bits are supported:
+		 */
+		riscv_reg_t tdata1_tests[] = {
+			set_field(tdata1, CSR_MCONTROL6_HIT0, 1),
+			set_field(tdata1, CSR_MCONTROL6_HIT1, 1),
+			set_field(tdata1, CSR_MCONTROL6_HIT0, 1) | field_value(CSR_MCONTROL6_HIT1, 1)
+		};
+		riscv_reg_t tdata1_test_rb;
+		for (uint64_t i = 0; i < ARRAY_SIZE(tdata1_tests); ++i) {
+			if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_tests[i]) != ERROR_OK)
+				return M6_HIT_ERROR;
+			if (riscv_reg_get(target, &tdata1_test_rb, GDB_REGNO_TDATA1) != ERROR_OK)
+				return M6_HIT_ERROR;
+			if (tdata1_test_rb == tdata1_tests[i]) {
+				if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_test_rb & ~hit_mask) != ERROR_OK)
+					return M6_HIT_ERROR;
+				return M6_NOT_HIT;
+			}
+		}
+	}
+	return M6_HIT_NOT_SUPPORTED;
+}
+
 /**
  * Look at the trigger hit bits to find out which trigger is the reason we're
  * halted.  Sets *unique_id to the unique ID of that trigger. If *unique_id is
  * RISCV_TRIGGER_HIT_NOT_FOUND, no match was found.
  */
 
-static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_id)
+static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_id,
+		bool *need_single_step)
 {
 	/* FIXME: this function assumes that we have only one trigger that can
 	 * have hit bit set. Debug spec allows hit bit to bit set if a trigger has
 	 * matched but did not fire. Such targets will receive erroneous results.
 	 */
 
-	// FIXME: Add hit bits support detection and caching
 	RISCV_INFO(r);
+	assert(need_single_step);
+	*need_single_step = false;
 
 	riscv_reg_t tselect;
 	if (riscv_reg_get(target, &tselect, GDB_REGNO_TSELECT) != ERROR_OK)
@@ -1601,9 +1655,21 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
 				break;
 			case CSR_TDATA1_TYPE_MCONTROL:
 				hit_mask = CSR_MCONTROL_HIT;
+				*need_single_step = true;
 				break;
 			case CSR_TDATA1_TYPE_MCONTROL6:
 				hit_mask = CSR_MCONTROL6_HIT0 | CSR_MCONTROL6_HIT1;
+				if (r->tinfo_version == CSR_TINFO_VERSION_0) {
+					*need_single_step = true;
+				} else if (r->tinfo_version == RISCV_TINFO_VERSION_UNKNOWN
+					|| r->tinfo_version == CSR_TINFO_VERSION_1) {
+					mctrl6hitstatus hits_status = check_mcontrol6_hit_status(target,
+								tdata1, hit_mask);
+					if (hits_status == M6_HIT_ERROR)
+						return ERROR_FAIL;
+					if (hits_status == M6_HIT_BEFORE || hits_status == M6_HIT_NOT_SUPPORTED)
+						*need_single_step = true;
+				}
 				break;
 			case CSR_TDATA1_TYPE_ICOUNT:
 				hit_mask = CSR_ICOUNT_HIT;
@@ -1622,8 +1688,9 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
 		/* FIXME: this logic needs to be changed to ignore triggers that are not
 		 * the last one in the chain. */
 		if (tdata1 & hit_mask) {
-			LOG_TARGET_DEBUG(target, "Trigger %u (unique_id=%" PRIi64 ") has hit bit set.",
-				i, r->trigger_unique_id[i]);
+			LOG_TARGET_DEBUG(target, "Trigger %u (unique_id=%" PRIi64
+			") has hit bit set. (need_single_step=%s)",
+				i, r->trigger_unique_id[i], (*need_single_step) ? "yes" : "no");
 			if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1 & ~hit_mask) != ERROR_OK)
 				return ERROR_FAIL;
 
@@ -2285,13 +2352,15 @@ static int set_debug_reason(struct target *target, enum riscv_halt_reason halt_r
 {
 	RISCV_INFO(r);
 	r->trigger_hit = -1;
+	r->need_single_step = false;
 	switch (halt_reason) {
 		case RISCV_HALT_EBREAK:
 			target->debug_reason = DBG_REASON_BREAKPOINT;
 			break;
 		case RISCV_HALT_TRIGGER:
 			target->debug_reason = DBG_REASON_UNDEFINED;
-			if (riscv_trigger_detect_hit_bits(target, &r->trigger_hit) != ERROR_OK)
+			if (riscv_trigger_detect_hit_bits(target, &r->trigger_hit,
+					&r->need_single_step) != ERROR_OK)
 				return ERROR_FAIL;
 			// FIXME: handle multiple hit bits
 			if (r->trigger_hit != RISCV_TRIGGER_HIT_NOT_FOUND) {
@@ -2553,10 +2622,19 @@ static int resume_prep(struct target *target, int current,
 	if (handle_breakpoints) {
 		/* To be able to run off a trigger, we perform a step operation and then
 		 * resume. If handle_breakpoints is true then step temporarily disables
-		 * pending breakpoints so we can safely perform the step. */
-		if (old_or_new_riscv_step_impl(target, current, address, handle_breakpoints,
-				false /* callbacks are not called */) != ERROR_OK)
-			return ERROR_FAIL;
+		 * pending breakpoints so we can safely perform the step.
+		 *
+		 * Two cases where single step is needed before resuming:
+		 * 1. ebreak used in software breakpoint;
+		 * 2. a trigger that is taken just before the instruction that triggered it is retired.
+		 */
+		if (target->debug_reason == DBG_REASON_BREAKPOINT
+		    || (target->debug_reason == DBG_REASON_WATCHPOINT
+			&& r->need_single_step)) {
+			if (old_or_new_riscv_step_impl(target, current, address, handle_breakpoints,
+					false /* callbacks are not called */) != ERROR_OK)
+				return ERROR_FAIL;
+		}
 	}
 
 	if (r->get_hart_state) {
@@ -5780,6 +5858,19 @@ int riscv_enumerate_triggers(struct target *target)
 		free(r->reserved_triggers);
 		r->reserved_triggers = NULL;
 		return ERROR_OK;
+	}
+
+	/* Obtaining tinfo.version value once.
+	 * No need to enumerate per-trigger.
+	 * See https://github.com/riscv/riscv-debug-spec/pull/1081.
+	 */
+	riscv_reg_t tinfo;
+	if (riscv_reg_get(target, &tinfo, GDB_REGNO_TINFO) == ERROR_OK) {
+		r->tinfo_version = get_field(tinfo, CSR_TINFO_VERSION);
+		LOG_TARGET_DEBUG(target, "Trigger tinfo.version = %d.", r->tinfo_version);
+	} else {
+		r->tinfo_version = RISCV_TINFO_VERSION_UNKNOWN;
+		LOG_TARGET_DEBUG(target, "Trigger tinfo.version is unknown.");
 	}
 
 	unsigned int t = 0;
--- a/src/target/riscv/riscv.h
+++ b/src/target/riscv/riscv.h
@@ -112,6 +112,7 @@ typedef struct {
 } range_list_t;
 
 #define DTM_DTMCS_VERSION_UNKNOWN ((unsigned int)-1)
+#define RISCV_TINFO_VERSION_UNKNOWN (-1)
 
 struct reg_name_table {
 	unsigned int num_entries;
@@ -151,6 +152,17 @@ struct riscv_info {
 
 	/* record the tinfo of each trigger */
 	unsigned int trigger_tinfo[RISCV_MAX_TRIGGERS];
+
+	/* Version of the implemented Sdtrig extension */
+	int tinfo_version;
+
+	/* Record if single-step is needed prior to resuming
+	 * from a software breakpoint or trigger.
+	 * Single-step is needed if the instruction that
+	 * caused the halt was not retired. That is,
+	 * when we halted "before" that instruction.
+	 */
+	bool need_single_step;
 
 	/* For each physical trigger contains:
 	 * -1: the hwbp is available


en-sc:
@sunnyzhu-learning, AFAIU, the commit is misnamed.

src/target/riscv/riscv.c
```diff
@@ -668,6 +668,16 @@ static int set_trigger(struct target *target, unsigned int idx, riscv_reg_t tdat
...
 	const bool is_mcontrol = type == CSR_TDATA1_TYPE_MCONTROL;
 
+	if (type == CSR_TDATA1_TYPE_MCONTROL) {
+		if (get_field(tdata1_rb, CSR_MCONTROL_TIMING) == CSR_MCONTROL_TIMING_BEFORE)
+			r->halt_before_dpc = true;
+	} else if (type == CSR_TDATA1_TYPE_MCONTROL6) {
+		int hit0 = get_field(tdata1_rb, CSR_MCONTROL6_HIT0);
+		int hit1 = get_field(tdata1_rb, CSR_MCONTROL6_HIT1);
+		if (((hit1 << 1) | hit0) == CSR_MCONTROL6_HIT0_BEFORE)
+			r->halt_before_dpc = true;
```
en-sc: [5.7.12. Match Control Type 6 (`mcontrol6`, at 0x7a1), `hit0` field description]:
> The TM updates this field when the trigger fires. 

Therefore a zero is always read here and `r->halt_before_dpc` is always false.

Please, take a look at `riscv_trigger_detect_hit_bits()`.

src/target/riscv/riscv.h
```diff
@@ -152,6 +152,9 @@ struct riscv_info {
 	/* record the tinfo of each trigger */
 	unsigned int trigger_tinfo[RISCV_MAX_TRIGGERS];
 
+	/* record the dpc that triggered it is retired. */
+	bool halt_before_dpc;
```
en-sc: AFAIU, the field is never reset.

en-sc (changes requested):
@sunnyzhu-learning, thank you for the patch.

Please note that `hit` bits are optional and therefore the behavior of targets that don't support them should also be considered.

src/target/riscv/riscv.c
sunnyzhu-learning: addressedaddressed

sunnyzhu-learning: addressed

sunnyzhu-learning:
Q: Please note that `hit` bits are optional and therefore the behavior of targets that don't support them should also be considered.
A: internal discussion a lot, modified as above, considering `hit` bits are optional.

src/target/riscv/riscv.c
```diff
@@ -1601,18 +1602,29 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
...
+				int hit0 = get_field(tdata1, CSR_MCONTROL6_HIT0);
+				int hit1 = get_field(tdata1, CSR_MCONTROL6_HIT1);
+				int trigger_retired_info = (hit1 << 1) | hit0;
+				if (trigger_retired_info == CSR_MCONTROL6_HIT0_AFTER
+				    || trigger_retired_info == CSR_MCONTROL6_HIT0_IMMEDIATELY_AFTER)
+					r->need_single_step = false;
 				break;
 			case CSR_TDATA1_TYPE_ICOUNT:
 				hit_mask = CSR_ICOUNT_HIT;
+				r->need_single_step = false;
```
en-sc: `r->need_single_step` should be reset only if `tdata1 & hit_mask` is not zero.

With current implementation in case you have a `mcontrol` trigger at `tselect == 0` and `icount` trigger at `tselect == 1` and  `mcontrol` is hit but the timing is not `CSR_MCONTROL_TIMING_AFTER` `r->need_single_step` will be `false`.

en-sc (changes requested):

sunnyzhu-learning:
1. Single step is needed if any `mcontrol` trigger has `timing` `before`. This may cause unnecessary single step in case like trigger0(timing before) not fire and trigger1(timing after) fires, but it's a conservative handling to make sure single step is performed in timing before case. 
2. `hit` bits 0 stands for trigger not firing or `hit` not support by HW. Single step is needed if any `mcontrol6` trigger has `hit0[1]` `before`, or `hit` not support by HW, or even not firing if `hit` is supported. This may cause unnecessary single step, but we have to consider the case where `hit` is not support by HW.
3. this patch will benefit in cases like
a. all tirggers have `timing/hit0[1]` `after` for `mcontrol[6]` trigger.
b. trigger0 is `timing` `after` and `hit` is implenmented.

src/target/riscv/riscv.c
sunnyzhu-learning: yeah, next patch considering this scenario, avoiding `icount/itrigger/etrigger` overwrites `mcontrol.timing/mcontrol6.hit0[1]`  is `before` situation.

src/target/riscv/riscv.c
```diff
@@ -1601,9 +1602,17 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
 				break;
 			case CSR_TDATA1_TYPE_MCONTROL:
 				hit_mask = CSR_MCONTROL_HIT;
+				if (get_field(tdata1, CSR_MCONTROL_TIMING) == CSR_MCONTROL_TIMING_BEFORE)
+					r->need_single_step = true;
```
en-sc: Here it is only known that there is a trigger with type `mcontrol`. It is not yet known that this trigger is the one that was hit . To know that one needs to check `tdata1 & hit_mask`.
```suggestion
					if (tdata1 & hit_mask)
						r->need_single_step = true;
```

Also, the spec states [5.7.11. Match Control (mcontrol, at 0x7a1)] the `hit` field description:
> If this bit is implemented then it must become set when this trigger fires and may become set when this trigger matches.

Therefore one can be sure that this trigger with type `mcontrol` has fired only if it's the only trigger that has it's `hit` field set to a non-zero value and all other triggers support `hit` fields. The last part is the most problematic -- OpenOCD does not gather information of `hit` field support at the moment. Though it can be done by changing how the triggers are set: `hit` fields are WARL and therefore `hit` field support can be discovered by writing to it when setting a trigger.

There is no such issue for `mcontrol6`triggers since [5.7.12. Match Control Type 6 (mcontrol6, at 0x7a1)], `hit0` field description:
> The TM updates this field when the trigger fires.

I'd suggest adding a `TODO`/filing an issue about it and only setting `r->need_single_step = true` for `mcontrol6` triggers for now.

src/target/riscv/riscv.c
```diff
@@ -2547,16 +2556,29 @@ static int resume_prep(struct target *target, int current,
 	assert(target->state == TARGET_HALTED);
 	RISCV_INFO(r);
 
+	riscv_reg_t dcsr;
+	if (riscv_reg_get(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
```
en-sc: This read of `dcsr` is unnecessary.  You could avoid it by setting `r->need_single_step` when halt reason is investigated and the hart is halted due to an `ebreak` (look into `set_debug_reason()`).

en-sc (changes requested):

src/target/riscv/riscv.c
sunnyzhu-learning: addressed, and checking  `mcontrol.hit` not implemented scenario

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

sunnyzhu-learning:
I don't have a best solution to solve this question, above patch is a optimize solution, Please help with code review,thanks~
@en-sc

src/target/riscv/riscv.c
sunnyzhu-learning: Though it can be done by changing how the triggers are set: hit fields are WARL and therefore hit field support can be discovered by writing to it when setting a trigger.
——————————
now，this patch using this solution and checking `hit` bits whether implemented

src/target/riscv/riscv.c
```diff
@@ -1589,21 +1590,100 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
...
+						}
+						if (riscv_reg_set(target, GDB_REGNO_TDATA1,
+							tdata1_test & ~hit_mask) != ERROR_OK)
+							return ERROR_FAIL;
+					}
+				}
+done:
+				if (mcontrol6_hit_not_supported
+				    || trigger_retired_info == CSR_MCONTROL6_HIT0_BEFORE)
+					r->need_single_step = true;
```
en-sc: `step` is not needed if the trigger was not hit or of the trigger was hit after.
Therefore this can be simplified:
```suggestion
				r->need_single_step = true;
				if (trigger_retired_info == CSR_MCONTROL6_HIT0_FALSE) {
					/* Check timing "before" support. */
					const riscv_reg_t tdata1_test = set_field(tdata1, CSR_MCONTROL6_HIT0, CSR_MCONTROL6_HIT0_BEFORE);
					if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_test) != ERROR_OK)
						return ERROR_FAIL;
					riscv_reg_t tdata1_test_rb;
					if (riscv_reg_get(target, &tdata1_test_rb, GDB_REGNO_TDATA1) != ERROR_OK)
						return ERROR_FAIL;
					if (tdata1_test_rb == tdata1_test) {
						/* Timing "before" is supported. Therefore this trigger was not hit or was hit after and single step is not needed. */
						r->need_single_step = false;
					}
				} else if (trigger_retired_info != CSR_MCONTROL6_HIT0_BEFORE) {
					/* Trigger has triggered after the instruction. */
					r->need_single_step = false;
				}
```

src/target/riscv/riscv.c
```diff
@@ -1589,21 +1590,100 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
...
+					if (riscv_reg_get(target,
+					&tdata1_test_rb, GDB_REGNO_TDATA1) != ERROR_OK)
+						return ERROR_FAIL;
+					int tdata1_test_hit =
+					get_field(tdata1_test, CSR_MCONTROL_HIT);
+					int tdata1_test_rb_hit =
+					get_field(tdata1_test_rb, CSR_MCONTROL_HIT);
+					if (tdata1_test_hit != tdata1_test_rb_hit)
+						mcontrol_hit_not_supported = true;
+					if (riscv_reg_set(target, GDB_REGNO_TDATA1,
```
en-sc: This is not necessary. The function already resets hit bits after the `switch`.
https://github.com/riscv-collab/riscv-openocd/blob/0097f6b46f56551d632d87f87b551c9ff2088092/src/target/riscv/riscv.c#L1707-L1709

src/target/riscv/riscv.c
en-sc: Please, consider always setting `r->need_single_step` in `set_debug_reason()`.

sunnyzhu-learning: this patch (https://github.com/riscv-collab/riscv-openocd/pull/1144/commits/ec2818a6864b7d1e3b0613ca01813d05546871f2) consider `r->need_single_step` in `set_debug_reason()`，please code review，thanks~

src/target/riscv/riscv.c
sunnyzhu-learning: the code simplify leading to below scenario missed：
trigger0 is supported `mcontrol.timing=1`, `r->need_single_step = true`;
trigger1 is supported `mcontrol6.hit0[1] = 3`, `r->need_single_step = false`;
trigger1 `r->need_single_step` info will overwrite trigger0 `r->need_single_step` .

src/target/riscv/riscv.c
sunnyzhu-learning: this function must `hit` bit is implemented, but `hit` is not implemented in HW, reset register `tdata1` lost.

src/target/riscv/riscv.c
en-sc: The issue you have mentioned can be mitigated by changing
```
r->need_single_step = false
```
to
```
r->need_single_step |= false
```

src/target/riscv/riscv.c
en-sc: You are right. Thanks!

src/target/riscv/riscv.c
```diff
@@ -2278,6 +2279,67 @@ derive_debug_reason_without_hitbit(const struct target *target, riscv_reg_t dpc)
 	}
 	return DBG_REASON_WATCHPOINT;
 }
+
+static bool check_single_step_needed_in_hit_supported_case(struct target *target,
```
en-sc: Please, consider moving it back to the previous location inside `hit` detection. This will make the change/restore of `tselect` and the read of `tdata1` unnecessary.

src/target/riscv/riscv.c
```diff
@@ -2278,6 +2279,67 @@ derive_debug_reason_without_hitbit(const struct target *target, riscv_reg_t dpc)
...
+			r->need_single_step = true;
+	} else {
+		r->need_single_step = false;
+	}
+	if (riscv_reg_set(target, GDB_REGNO_TSELECT, tselect) != ERROR_OK)
+		return ERROR_FAIL;
+	return r->need_single_step;
+}
+
+static bool check_single_step_needed_in_hit_unsupported_case(struct target *target)
```
en-sc: Can this be avoided?
I'd suggest just conservatively setting `r->need_single_step = true;` if `hit` fields are not supported.
This will also probably be faster -- one step operation vs a bunch of register reads via JTAG.

en-sc (changes requested):

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: avoided, and addressed

src/target/riscv/riscv.c
sunnyzhu-learning: I thank `r->need_single_step |= false` cannot be avoided muilt type trigger issue, below patch(https://github.com/riscv-collab/riscv-openocd/pull/1144/commits/5f65be97a519cc9e55acda8d8a783265e85d054b) function `mcontrol6_hit_supported` soving overwrite phenomenon, Please code review~

sunnyzhu-learning:
update patch, internal code review done, @en-sc please code review(https://github.com/riscv-collab/riscv-openocd/pull/1144/commits/106240206e2ed388df78aed201fd61cc89ccba8f) thanks~

src/target/riscv/riscv.c
```diff
@@ -1561,6 +1561,100 @@ int riscv_remove_watchpoint(struct target *target,
...
+		if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_hit_imm_after_test) != ERROR_OK)
+			return M6_HIT_ERROR;
+		if (riscv_reg_get(target, &tdata1_test_rb, GDB_REGNO_TDATA1) != ERROR_OK)
+			return M6_HIT_ERROR;
+		if (tdata1_test_rb == tdata1_hit_imm_after_test) {
+			if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_test_rb & ~hit_mask) != ERROR_OK)
+				return M6_HIT_ERROR;
+			return M6_NOT_HIT;
+		}
+	}
```
en-sc, line 1612: If a trigger supports a subset of `hit` field values, at this point the value of the `hit` field may be not 0. (`tdata1_test_rb != tdata1_...test` does not mean `tdata1.hit` is zero in such case).

src/target/riscv/riscv.c
```diff
@@ -1561,6 +1561,100 @@ int riscv_remove_watchpoint(struct target *target,
...
+		riscv_reg_t tdata1_hit_imm_after_test = tdata1_hit_before_test | tdata1_hit_after_test;
+		if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_hit_imm_after_test) != ERROR_OK)
+			return M6_HIT_ERROR;
+		if (riscv_reg_get(target, &tdata1_test_rb, GDB_REGNO_TDATA1) != ERROR_OK)
+			return M6_HIT_ERROR;
+		if (tdata1_test_rb == tdata1_hit_imm_after_test) {
+			if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_test_rb & ~hit_mask) != ERROR_OK)
+				return M6_HIT_ERROR;
+			return M6_NOT_HIT;
+		}
```
en-sc: Please, consider looping over an array of `tdata1` values to check:
```suggestion
	if (hit_info == CSR_MCONTROL6_HIT0_FALSE) {
		riscv_reg_t tdata1_tests[] = {
			set_field(tdata1, CSR_MCONTROL6_HIT0, 1),
			set_field(tdata1, CSR_MCONTROL6_HIT1, 1),
			set_field(tdata1, CSR_MCONTROL6_HIT0, 1) | field_value(CSR_MCONTROL6_HIT0, 1)
		};
		for (int i = 0; i < ARRAY_SIZE(tdata1_tests); ++i) {
			if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_hit_before_test) != ERROR_OK)
				return M6_HIT_ERROR;
			riscv_reg_t tdata1_test_rb;
			if (riscv_reg_get(target, &tdata1_test_rb, GDB_REGNO_TDATA1) != ERROR_OK)
				return M6_HIT_ERROR;
			if (tdata1_test_rb == tdata1_hit_before_test) {
				if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_test_rb & ~hit_mask) != ERROR_OK)
					return M6_HIT_ERROR;
				return M6_NOT_HIT;
			}
		}
```

en-sc (changes requested):
I don't like this change in it's current state. The reason is:
1. There are too many operations to derive `hit` field support that are performed in the *most common* case -- a trigger is not hit.
2. Handling `mcontrol` type does not consider the fact that `mcontrol.hit` field can be set for a matched trigger that did not fire. Therefore there can be a matched `mcontrol` trigger and another trigger that fired with a different timing that lacks `hit` field support.

My suggestion is:
When looking into which trigger was hit (`riscv_trigger_detect_hit_bits()`):
* `r->need_single_step` is initializes to `false`.
* In case there is a `mcontrol` trigger (regardless if it has fired or not): `r->need_single_step = true` due to the fact that there can be another `mcontrol` or `mcontrol6` trigger that lacks `hit` field support and was hit `before`.
* In case there is a `mcontrol6` trigger: if `hit == after || hit == immed_after` `r->need_single_step |= false`. This condition can be detected without any additional reads from the target.
* In case of `itrigger`, `etrigger` or `tmexttrigger` trigger `r->need_single_step = true` for the same reason as `mcontrol`.
* In case there is a trigger of some other type (if I'm correct only `icount` is unaccounted for at this point) that has it's `hit` field set, `r->need_single_step = false`.

IMHO this will cover the most common cases without the need for excessive investigation of `hit` field support.

en-sc:
Please note: analyzing timings for `mcontrol6` depends on `tinfo.verison`.

src/target/riscv/riscv.c
JoshCafe: https://github.com/riscv-collab/riscv-openocd/commit/106240206e2ed388df78aed201fd61cc89ccba8f#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR1616
All checks are performed under conditon `if (hit_info == CSR_MCONTROL6_HIT0_FALSE)`

sunnyzhu-learning:
`Q`: 
1. There are too many operations to derive `hit` field support that are performed in the most common case -- a trigger is not hit.

When `halt_reason` is `RISCV_HALT_TRIGGER`, then enter `riscv_trigger_detect_hit_bits()`, I think this checking operations acceptably.

2. Handling `mcontrol` type does not consider the fact that `mcontrol.hit` field can be set for a matched trigger that did not fire. Therefore there can be a matched `mcontrol` trigger and another trigger that fired with a different timing that lacks `hit` field support.

Yeah，you are right，`mcontrol.hit` Description：
```
If this bit is implemented then it must become set 
when this trigger fires and may become set when this trigger matches.
```

your suggestion:

- In case there is a `mcontrol` trigger (regardless if it has fired or not): `r->need_single_step = true` due to the fact that there can be another `mcontrol` or `mcontrol6` trigger that lacks `hit` field support and was hit `before`.

The solution is too crude, but I agree your suggestion, 

example:
A CPU support two triggers, T0(mcontrol.hit=1[match but not fire], mcontrol.timing=after), then `r->need_single_step = false`, break, T1(mcontrol.hit=0[not support], mcontrol.timing=before) fire，now due to https://github.com/riscv-collab/riscv-openocd/blob/f9a129271642febac27f93673fd96481aff503ca/src/target/riscv/riscv.c#L1649 break, T1 is not needed single step.

next patch changes to:
```
case CSR_TDATA1_TYPE_MCONTROL:
	hit_mask = CSR_MCONTROL_HIT;
	r->need_single_step = true;
	break;
```


- In case there is a `mcontrol6` trigger: if `hit == after || hit == immed_after` `r->need_single_step |= false`. This condition can be detected without any additional reads from the target.

Internal discussion a lot, Disagree your suggestion, `hit == after || hit == immed_after` included in function `check_mcontrol6_hit_status` logic is ok，other releated function can call it directly, because it is fully `mcontrol6.hit0[1]` info.

- In case of `itrigger`, `etrigger` or `tmexttrigger` trigger `r->need_single_step = true` for the same reason as `mcontrol`.

I don't agree your suggestion, reference Debug spec 5.7.14 Description:
```
When the trigger fires, all CSRs are updated for the interrupt trap as defined by the Privileged Spec,
and the requested action is taken just before the first instruction of the trap handler is executed.
If the trigger fires with action =0 then zero is written to the tval CSR on the breakpoint trap (see
5.4)
```
this can ensure `itrigger`, `etrigger` or `tmexttrigger` trigger is `timing.after` ?  `r->need_single_step` should be used `false` by default when other type triggers not support.

- In case there is a trigger of some other type (if I'm correct only `icount` is unaccounted for at this point) that has it's `hit` field set, `r->need_single_step = false`.

I agree, use default is ok.

`Q`: If a trigger supports a subset of `hit` field values, at this point the value of the `hit` field may be not 0. (`tdata1_test_rb != tdata1_test` does not mean `tdata1.hit` is zero in such case).

this case first considering `hit_info == CSR_MCONTROL6_HIT0_FALSE`, I do not understand this comment.

`Q`: Please, consider looping over an array of tdata1 values to check:

I agree this change, next patch addressed it.

`Q`: analyzing timings for `mcontrol6` depends on `tinfo.verison`.

I agree this change, next patch addressed it.

Last, Thanks a lot your suggestions, Helped me a lot~

src/target/riscv/riscv.c
en-sc: Yes. The thing is: WARL is write-any-read-legal, not write-any-read-zero. Therefore it is incorrect to assume a WARL register field is zero after a write of an illegal value to it.

In this particular case:
1. `tdata1.hit` was 0
2. Write of illegal value to `tdata1.hit`
3. Here `tdata1.hit` is any legal value which may be zero but does not have to be zero.

en-sc:
> When halt_reason is RISCV_HALT_TRIGGER, then enter riscv_trigger_detect_hit_bits(), I think this checking operations acceptably.

The thing is, even when `riscv_halt_reason()` is `RISCV_HALT_TRIGGER` there can be plenty of triggers on the platform and only one of them is hit. This is what I mean by "too many operations to derive hit field support that are performed in the most common case".

> this case first considering hit_info == CSR_MCONTROL6_HIT0_FALSE, I do not understand this comment.

See the comment in the discussion: https://github.com/riscv-collab/riscv-openocd/pull/1144#discussion_r1824040455

> this can ensure itrigger, etrigger or tmexttrigger trigger is timing.after ? r->need_single_step should be used false by default when other type triggers not support.

Yes, you are right. My mistake.

src/target/riscv/riscv.c
JoshCafe: We didn't assume a WARL register field is zero after a write of an illegal value to it. `tdata1_test_rb` is matched against `tdata1_...test` and `tdata1_...test` is not zero.

JoshCafe: I may get your point, we need  to recover `tdata1.hit `to 0 in all cases.
Thansk for catching it.

src/target/riscv/riscv.c
```diff
@@ -1601,9 +1648,24 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
 				break;
 			case CSR_TDATA1_TYPE_MCONTROL:
 				hit_mask = CSR_MCONTROL_HIT;
+				r->need_single_step = true;
 				break;
 			case CSR_TDATA1_TYPE_MCONTROL6:
-				hit_mask = CSR_MCONTROL6_HIT0 | CSR_MCONTROL6_HIT1;
+				riscv_reg_t tinfo, tinfo_version;
```
sunnyzhu-learning, line 1654: I don't know build32 fail reason, value is defined. local regression pass......

src/target/riscv/riscv.c
```diff
@@ -1561,6 +1561,52 @@ int riscv_remove_watchpoint(struct target *target,
...
+	M6_HIT_AFTER,
+	M6_HIT_IMM_AFTER
+} mctrl6hitstatus;
+
+static mctrl6hitstatus check_mcontrol6_hit_status(struct target *target,
+		riscv_reg_t tdata1, uint64_t hit_mask)
+{
+	uint32_t hit0 = get_field(tdata1, CSR_MCONTROL6_HIT0);
+	uint32_t hit1 = get_field(tdata1, CSR_MCONTROL6_HIT1);
+	uint32_t hit_info = (hit1 << 1) | hit0;
```
JanMatCodasip: Pedantic detail:

The variables can be const. It will helps the code readers.

```suggestion
	const uint32_t hit0 = get_field(tdata1, CSR_MCONTROL6_HIT0);
	const uint32_t hit1 = get_field(tdata1, CSR_MCONTROL6_HIT1);
	const uint32_t hit_info = (hit1 << 1) | hit0;
```

src/target/riscv/riscv.c
```diff
@@ -1601,9 +1648,25 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
...
 				break;
 			case CSR_TDATA1_TYPE_MCONTROL6:
 				hit_mask = CSR_MCONTROL6_HIT0 | CSR_MCONTROL6_HIT1;
+				riscv_reg_t tinfo, tinfo_version;
+				int result = riscv_reg_get(target, &tinfo, GDB_REGNO_TINFO);
+				if (result == ERROR_OK) {
+					tinfo_version = get_field(tinfo, CSR_TINFO_VERSION);
+					if (tinfo_version == CSR_TINFO_VERSION_0)
+						r->need_single_step = true;
+				}
```
JanMatCodasip: Please, move the code that obtains `tinfo.version` to the trigger examination routine `riscv_enumerate_triggers()`. 

The `tinfo.version` should be read only once during the trigger enumeration , and the result stored into `struct riscv_info`.

src/target/riscv/riscv.c
```diff
@@ -1576,6 +1622,7 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
 
 	// FIXME: Add hit bits support detection and caching
 	RISCV_INFO(r);
+	r->need_single_step = false;
```
JanMatCodasip: The current code would return values from `riscv_trigger_detect_hit_bits` by three different channels:

- 1. By the return value
- 2. By the output parameter unique_id
- 3. By direct modification of the riscv_info structure

For the sake of code clarity, please don't introduce the iii.

Instead, please add another output parameter and return the `need_single_step` bool flag that way - in the same manner as it is done for `unique_id`.

Thanks.

src/target/riscv/riscv.c
```diff
@@ -1601,9 +1648,25 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
...
+					if (tinfo_version == CSR_TINFO_VERSION_0)
+						r->need_single_step = true;
+				}
+				if (result != ERROR_OK || tinfo_version == CSR_TINFO_VERSION_1) {
+					mctrl6hitstatus hits_status = check_mcontrol6_hit_status(target,
+								tdata1, hit_mask);
+					if (hits_status == M6_HIT_ERROR)
+						return ERROR_FAIL;
+					if (hits_status == M6_HIT_BEFORE || hits_status == M6_HIT_NOT_SUPPORTED)
+						r->need_single_step = true;
```
JanMatCodasip: On line 1688 below, please extend the debug print to show not only the `unique_id` value, but also the `need_single_step` value. 

This will help with log analysis and troubleshooting.

src/target/riscv/riscv.c
JanMatCodasip: @en-sc, my understanding is that the algorithm should be:

- All three combinations `0b01`, `0b10` and `0b11` are written to `tdata1.hit[1..0]` in a loop.
- If **any** of them results in `hit[1..0] != 0b00`, the result is `M6_NOT_HIT`.
- Otherwise the result is `M6_HIT_NOT_SUPPORTED`.

Is that please the same thing that you meant by [your comment](https://github.com/riscv-collab/riscv-openocd/pull/1144/files#r1824040455) above? Do we see it the same?

src/target/riscv/riscv.c
```diff
@@ -1561,6 +1561,52 @@ int riscv_remove_watchpoint(struct target *target,
...
+			if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_tests[i]) != ERROR_OK)
+				return M6_HIT_ERROR;
+			if (riscv_reg_get(target, &tdata1_test_rb, GDB_REGNO_TDATA1) != ERROR_OK)
+				return M6_HIT_ERROR;
+			if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_test_rb & ~hit_mask) != ERROR_OK)
+				return M6_HIT_ERROR;
+			if (tdata1_test_rb == tdata1_tests[i])
+				return M6_NOT_HIT;
+		}
+	}
```
JanMatCodasip: Once we learn for sure whether the `hit` bits are supported for a given trigger, we should cache that information in `struct riscv_info`, for example this way:

```c
yes_no_maybe_t mcontrol6_supports_hit[RISCV_MAX_TRIGGERS]; // initialized to YNM_MAYBE
```

This way the number of reads and writes of TDATA1 will be greatly reduced.

@sunnyzhu-learning if you decide to not implement this improvement in this merge request, please add at least a TODO comment into this function (and in that case we will also open an issue to track it).

src/target/riscv/riscv.c
```diff
@@ -1561,6 +1561,52 @@ int riscv_remove_watchpoint(struct target *target,
...
+	return M6_HIT_NOT_SUPPORTED;
+}
+
+
```
JanMatCodasip: Coding style detail:

No need to include two blank lines.

src/target/riscv/riscv.c
JanMatCodasip: This part of the algorithm was harder for me to grasp as a reader. 

Could you please add an explanatory comment. For example:

```suggestion
	if (hit_info == CSR_MCONTROL6_HIT0_FALSE) {
		/* hit[1..0] equals 0, which can mean one of the following:
		 * - "hit" bits are supported and this trigger has not fired
		 * - "hit" bits are not supported on this trigger
		 * To distinguish these two cases, try writing all non-zero bit 
		 * patterns to hit[1..0] to determine if the "hit" bits are supported:
		 */
		riscv_reg_t tdata1_tests[] = {
			set_field(tdata1, CSR_MCONTROL6_HIT0, 1),
			set_field(tdata1, CSR_MCONTROL6_HIT1, 1),
			set_field(tdata1, CSR_MCONTROL6_HIT0, 1) | field_value(CSR_MCONTROL6_HIT1, 1)
		};
		riscv_reg_t tdata1_test_rb;
		for (uint64_t i = 0; i < ARRAY_SIZE(tdata1_tests); ++i) {
			if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_tests[i]) != ERROR_OK)
				return M6_HIT_ERROR;
			if (riscv_reg_get(target, &tdata1_test_rb, GDB_REGNO_TDATA1) != ERROR_OK)
				return M6_HIT_ERROR;
			if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_test_rb & ~hit_mask) != ERROR_OK)
				return M6_HIT_ERROR;
			if (tdata1_test_rb == tdata1_tests[i])
				return M6_NOT_HIT;
		}
	}
```

src/target/riscv/riscv.h
```diff
@@ -152,6 +152,9 @@ struct riscv_info {
 	/* record the tinfo of each trigger */
 	unsigned int trigger_tinfo[RISCV_MAX_TRIGGERS];
 
+	/* record the dpc that triggered it is retired. */
```
JanMatCodasip: Please, make sure that comments are accurate and understandable to code readers.

Did you want to say this?

```suggestion
	/* Record if single-step is needed prior to resuming from a software breakpoint or trigger.
	 * Single-step is needed if the instruction that caused the halt was not retired. That is, 
	 * when we halted "before" that instruction.
	 */
```

JanMatCodasip (changes requested):
Overall I like the idea and thank you for this contribution.

Care needs to be taken that the feature is implemented correctly and that the code remains clear and understandable after the change.

Please, take a look at my comments.

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: Add TODO, addressed.

sunnyzhu-learning: thanks, addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.h
sunnyzhu-learning: addressed, thanks

src/target/riscv/riscv.c
en-sc: @JanMatCodasip, AFAIK the spec does not require for `hit` bit to be supported across all possible configurations of a trigger with a given type.
e.g. It is a Sdtrig-compliant system that supports `hit` field with `e` set (on breakpoints), but does not support `hit` field with `r` or `w` set (on watchpoints).
Therefore it is incorrect to cache support of `mcontrol6.hit` field just based on trigger index.

src/target/riscv/riscv.c
en-sc: @JanMatCodasip,

> Is that please the same thing that you meant by your comment above? Do we see it the same?


The algorithm you are describing is missing a step.
`hit` field should be set to zero after testing is done.

This is what my comment was about. Previously in the case `M6_HIT_NOT_SUPPORTED` it was not done. And looking at it once again I think **I was wrong**.
My motivation for adding the reset of `mcontrol.hit` to zero before returning `M6_HIT_NOT_SUPPORTED` was:
* The last written value to the field was illegal.
* Therefore it may contain any legal value and should be zeroed-out.

However, at this point it is known that the only legal value is zero, so `mcontrol6.hit` is zero.

@JoshCafe, @sunnyzhu-learning,  I'm sorry for the confusion. Please, consider moving
```
		if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_test_rb & ~hit_mask) != ERROR_OK)
				return M6_HIT_ERROR;
```
into the block protected by the `if`:
```
			if (tdata1_test_rb == tdata1_tests[i]) {
				if (riscv_reg_set(target, GDB_REGNO_TDATA1, tdata1_test_rb & ~hit_mask) != ERROR_OK)
					return M6_HIT_ERROR;
				return M6_NOT_HIT;
			}
```
(the same as it was before)

sunnyzhu-learning: addressed

JoshCafe: - M6_NOT_HIT is returned if any of `0b01`, `0b10 `, `0b11` can be readback after writing `tdata1.hit[1..0]`.
- M6_HIT_NOT_SUPPORTED is returned if none of `0b01`, `0b10 `, `0b11` can be readback after writing `tdata1.hit[1..0]`, then the only leagal value for `hit[1..0]` is zero.
@en-sc  Thanks for your detailed explanation.

src/target/riscv/riscv.c
JanMatCodasip: @en-sc, I believe you're right, thanks for the correction. It would be a bit extreme for hardware to support hit bits on one mcontrol6 configuration but not other, but still, the current spec permits it. I have opened a proposal to address this in a future version of the spec: https://github.com/riscv/riscv-debug-spec/issues/1080

@sunnyzhu-learning Per Evgeniy's comment, the TODO should be removed. We cannot cache the information. I'm sorry for this extra back and forth.

src/target/riscv/riscv.c
JanMatCodasip: The current version of the code looks all right to me. Thanks all.

src/target/riscv/riscv.h
```diff
@@ -152,6 +152,17 @@ struct riscv_info {
 	/* record the tinfo of each trigger */
 	unsigned int trigger_tinfo[RISCV_MAX_TRIGGERS];
 
+	/* record the tinfo_version of each trigger */
+	int64_t trigger_tinfo_version[RISCV_MAX_TRIGGERS];
```
JanMatCodasip: My interpretation of `tinfo.version`  has always been that it is a global constant, not a per-trigger constant. 

However, I am afraid that the [debug spec](https://github.com/riscv/riscv-debug-spec/releases/download/1.0.0-rc3/riscv-debug-specification.pdf) is not clear about that, either. On page 69, the spec says:
- `"The reset values listed here [in tinfo] apply to every underlying trigger."`
- `"[Field version] contains the version of the Sdtrig extension implemented."` (which I read as the "whole trigger implementation").

These seem to contradict each other.

@en-sc What is your opinion/interpretation of this part of the spec, please?

JanMatCodasip: `int64_t` looks excessive here to store the version which is no longer than 8 bits.

Please, use just an `int`.

Also `-1` can be defined as "unknown tinfo.version". See [my other comment](https://github.com/riscv-collab/riscv-openocd/pull/1144/files#r1832326783).

src/target/riscv/riscv.c
```diff
@@ -5698,19 +5779,21 @@ static int check_if_trigger_exists(struct target *target, unsigned int index)
...
 	if (riscv_reg_get(target, &tinfo, GDB_REGNO_TINFO) == ERROR_OK) {
+		*trigger_tinfo_version = get_field(tinfo, CSR_TINFO_VERSION);
 		/* tinfo.INFO == 1: trigger doesn’t exist
 		 * tinfo == 0 or tinfo.INFO != 1 and tinfo LSB is set: invalid tinfo */
 		if (tinfo == 0 || tinfo & 0x1)
 			return ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
 		*trigger_tinfo = tinfo;
 		return ERROR_OK;
 	}
+	*trigger_tinfo_version = ERROR_TARGET_RESOURCE_NOT_AVAILABLE;
```
JanMatCodasip: Please, don't reuse OpenOCD return codes to denote "trigger not available". Instead, make a dedicated constant for it, e.g. `#define RISCV_TINFO_VERSION_UNKNOWN (-1)`.

src/target/riscv/riscv.c
```diff
@@ -1622,8 +1691,9 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
 		/* FIXME: this logic needs to be changed to ignore triggers that are not
 		 * the last one in the chain. */
 		if (tdata1 & hit_mask) {
-			LOG_TARGET_DEBUG(target, "Trigger %u (unique_id=%" PRIi64 ") has hit bit set.",
-				i, r->trigger_unique_id[i]);
+			LOG_TARGET_DEBUG(target, "Trigger %u (unique_id=%" PRIi64 ""
+			", need_single_step=%" PRId32 ") has hit bit set.",
+				i, r->trigger_unique_id[i], r->need_single_step);
```
JanMatCodasip: Cosmetic suggestions:

- The message would be more clear if "need_single_step" is mentioned at the end, not right next to "trigger" and "unique_id".
- PRId32 should not be used for printing of a bool. Instead, you can print strings "yes", "no".

```suggestion
			LOG_TARGET_DEBUG(target, "Trigger %u (unique_id=%" PRIi64 
			") has hit bit set. (need_single_step=%s)",
				i, r->trigger_unique_id[i], r->need_single_step ? "yes" : "no");
```

src/target/riscv/riscv.c
JanMatCodasip: Please, double-check if your changes were pushed.

I can see that `riscv_trigger_detect_hit_bits()` still directly changes `r->need_single_step`.

JanMatCodasip (changes requested):

src/target/riscv/riscv.h
en-sc: @JanMatCodasip, I agree with you. My understanding also was that `tinfo.version` is the same regardless of `tselect` value.
However, it would be nice for the spec to be a bit clearer about it, especially considering  that the only other field in `timfo`, `tinfo.info` depends on the value of `tselect`.

src/target/riscv/riscv.c
sunnyzhu-learning: deleted TODO, addressed

src/target/riscv/riscv.h
sunnyzhu-learning: addressed

sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: checking again, Initialization `r->need_single_step` is deleted in `riscv_trigger_detect_hit_bits()`, please code review this patch(https://github.com/riscv-collab/riscv-openocd/commit/d57f3559767146c813bf2740aa471a7bdaea6955#diff-b4aa16f9e42cb8f0934baa7c8e0ec9c70a369bef98b99b26ae2e896c8aa95d6aR1623)

src/target/riscv/riscv.c
```diff
@@ -1601,9 +1654,21 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
 				break;
 			case CSR_TDATA1_TYPE_MCONTROL:
 				hit_mask = CSR_MCONTROL_HIT;
+				r->need_single_step = true;
```
```suggestion
				*need_single_step = true;
```

Please, add also these two lines to the top of the function. The variable `need_single_step` should always be assigned a valid value. For the sake of code clarity, we should not require the caller to initialize it to `false`:

```c
assert(need_single_step);
*need_single_step = false;
```

src/target/riscv/riscv.c
```diff
@@ -1601,9 +1654,21 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
...
 				break;
 			case CSR_TDATA1_TYPE_MCONTROL6:
 				hit_mask = CSR_MCONTROL6_HIT0 | CSR_MCONTROL6_HIT1;
+				if (r->trigger_tinfo_version[i] == CSR_TINFO_VERSION_0) {
+					r->need_single_step = true;
```
```suggestion
					*need_single_step = true;
```

src/target/riscv/riscv.c
```diff
@@ -1601,9 +1654,21 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
...
+				} else if (r->trigger_tinfo_version[i] == RISCV_TINFO_VERSION_UNKNOWN
+					|| r->trigger_tinfo_version[i] == CSR_TINFO_VERSION_1) {
+					mctrl6hitstatus hits_status = check_mcontrol6_hit_status(target,
+								tdata1, hit_mask);
+					if (hits_status == M6_HIT_ERROR)
+						return ERROR_FAIL;
+					if (hits_status == M6_HIT_BEFORE || hits_status == M6_HIT_NOT_SUPPORTED)
+						r->need_single_step = true;
```
```suggestion
						*need_single_step = true;
```

src/target/riscv/riscv.c
```diff
@@ -1622,8 +1687,9 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
 		/* FIXME: this logic needs to be changed to ignore triggers that are not
 		 * the last one in the chain. */
 		if (tdata1 & hit_mask) {
-			LOG_TARGET_DEBUG(target, "Trigger %u (unique_id=%" PRIi64 ") has hit bit set.",
-				i, r->trigger_unique_id[i]);
+			LOG_TARGET_DEBUG(target, "Trigger %u (unique_id=%" PRIi64
+			") has hit bit set. (need_single_step=%s)",
+				i, r->trigger_unique_id[i], r->need_single_step ? "yes" : "no");
```
```suggestion
				i, r->trigger_unique_id[i], (*need_single_step) ? "yes" : "no");
```

src/target/riscv/riscv.c
JanMatCodasip: I am sorry, I can still see multiple occurrences of `r->need_single_step = true;` in `riscv_trigger_detect_hit_bits()`.

I am closing this thread. Instead, I have opened threads on the concrete lines that should be fixed.

src/target/riscv/riscv.h
JanMatCodasip: @en-sc, thank you for your input. I have just submitted a proposal for the debug spec to clarify the description of tinfo.version: https://github.com/riscv/riscv-debug-spec/pull/1081. Feel free to comment on it.

@sunnyzhu-learning, please, update your code to read `tinfo.version` just once.

The tinfo in struct riscv_info should be just a scalar, not an array:

```
	/* Version of the implemented Sdtrig extension */
	int tinfo_version;
```

Please, also print the obtained tinfo.version value to the log (`LOG_TARGET_DEBUG`) so that it becomes easier to troubleshoot issues.

JanMatCodasip (changes requested):

src/target/riscv/riscv.h
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

sunnyzhu-learning:
> please, update your code to read tinfo.version just once.
Please, also print the obtained tinfo.version value to the log (LOG_TARGET_DEBUG) so that it becomes easier to troubleshoot issues.

addressed，move `tinfo.version` value to `add_trigger()` @JanMatCodasip Please code review,thanks~

src/target/riscv/riscv.h
JanMatCodasip: Please, the proper place for getting tinfo_version is not in `add_trigger()` but in `riscv_enumerate_triggers()`.

sunnyzhu-learning: addressed, getting tinfo_version move to `riscv_enumerate_triggers()`

src/target/riscv/riscv.c
```diff
@@ -5782,6 +5861,19 @@ int riscv_enumerate_triggers(struct target *target)
...
+	 */
+	int tinfo_version;
+	riscv_reg_t tinfo;
+	if (riscv_reg_get(target, &tinfo, GDB_REGNO_TINFO) == ERROR_OK)
+		tinfo_version = get_field(tinfo, CSR_TINFO_VERSION);
+	else
+		tinfo_version = RISCV_TINFO_VERSION_UNKNOWN;
+	r->tinfo_version = tinfo_version;
+	LOG_TARGET_DEBUG(target, "Trigger tinfo.version = %" PRId32 ".", r->tinfo_version);
+
```
JanMatCodasip: Two small things:

- When printing a value of type `int`, use `%d`. `PRId32` is strictly intended for `uint32_t`. (https://en.cppreference.com/w/c/io/fprintf, https://en.cppreference.com/w/c/types/integer)
- Please say `Trigger tinfo.version is unknown` instead of `Trigger tinfo.version = -1`.

```suggestion
	riscv_reg_t tinfo;
	if (riscv_reg_get(target, &tinfo, GDB_REGNO_TINFO) == ERROR_OK) {
		r->tinfo_version = get_field(tinfo, CSR_TINFO_VERSION);
		LOG_TARGET_DEBUG(target, "Trigger tinfo.version = %d.", r->tinfo_version);		
	} else {
		tinfo_version = RISCV_TINFO_VERSION_UNKNOWN;
		LOG_TARGET_DEBUG(target, "Trigger tinfo.version is unknown.");		
	}

```

src/target/riscv/riscv.c
```diff
@@ -1576,6 +1629,8 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
 
 	// FIXME: Add hit bits support detection and caching
```
JanMatCodasip: This TODO can now be removed. It would be confusing to leave it in the code because:

- Hit support detection you have just implemented.
- Caching cannot be implemented, as we have [already discussed](https://github.com/riscv-collab/riscv-openocd/pull/1144#discussion_r1832259614).

```suggestion
```

src/target/riscv/riscv.c
JoshCafe: We're assuming a vaid trigger is selected before reading `Trigger tinfo` , is that assumption correct? @JanMatCodasip

JanMatCodasip: My interpretation of the debug spec is that `tinfo.version` is a hardwired constant, independent on the value of `tselect`.

So your approach looks good to me. I believe there is no need to write anyting to `tselect` before reading `tinfo.version`.

src/target/riscv/riscv.c
sunnyzhu-learning: addressed

src/target/riscv/riscv.c
sunnyzhu-learning: addressed, thanks a lot @JanMatCodasip

JanMatCodasip (approved):
LGTM.

@sunnyzhu-learning Thank you for the contribution and for keeping up with the review suggestions.

src/target/riscv/riscv.c
en-sc: @JanMatCodasip, are you sure? As outlined [here](https://github.com/riscv/riscv-debug-spec/issues/1080#issuecomment-2462506319) in your spec proposal, OpenOCD could cache `hit` field support.

en-sc (approved):
LGTM

src/target/riscv/riscv.c
en-sc: Created an issue instead of this `TODO`: https://github.com/riscv-collab/riscv-openocd/issues/1169

JanMatCodasip: @en-sc, perhaps I initially misunderstood your comment. 

Did you mean it this way? For a given triplet of `tselect`, `tdata1` and `tdata2` values, the support of `hit` bit would be detected at the time the trigger is set using the procedure described in https://github.com/riscv/riscv-debug-spec/issues/1080#issuecomment-2462506319, and then that information:

- would be remembered for the whole time the trigger remains set
- and perhaps also remembered (cached) for the future, in case that same trigger (defined by `tselect`, `tdata1` and `tdata2`) becomes set again.

en-sc: Yes, that's the idea. You've put it much more clearly. Thanks!

src/target/riscv/riscv.c
```diff
@@ -1601,9 +1655,21 @@ static int riscv_trigger_detect_hit_bits(struct target *target, int64_t *unique_
...
 			case CSR_TDATA1_TYPE_MCONTROL:
 				hit_mask = CSR_MCONTROL_HIT;
+				*need_single_step = true;
 				break;
 			case CSR_TDATA1_TYPE_MCONTROL6:
 				hit_mask = CSR_MCONTROL6_HIT0 | CSR_MCONTROL6_HIT1;
+				if (r->tinfo_version == CSR_TINFO_VERSION_0) {
+					*need_single_step = true;
+				} else if (r->tinfo_version == RISCV_TINFO_VERSION_UNKNOWN
+					|| r->tinfo_version == CSR_TINFO_VERSION_1) {
```
en-sc, line 1665: @sunnyzhu-learning, I don't quite get it:
Why `RISCV_TINFO_VERSION_UNKNOWN` is handled the same way as `CSR_TINFO_VERSION_1` and not the `CSR_TINFO_VERSION_0`?

Shouldn't this be rewritten as:
```
				if (r->tinfo_version == CSR_TINFO_VERSION_0
					|| r->tinfo_version == RISCV_TINFO_VERSION_UNKNOWN) {
					*need_single_step = true;
				} else if (r->tinfo_version == CSR_TINFO_VERSION_1) {
```

sunnyzhu-learning: Considering `RISCV_TINFO_VERSION_UNKNOWN` equal to `tinfo` register is not implemented, and `mcontrol6` is implemented, so `RISCV_TINFO_VERSION_UNKNOWN` is handled the same way as `CSR_TINFO_VERSION_1`, this is my understand.

PR #1146, opened 2024-10-10
target/riscv: check other TAPs in `select_dmi()`

en-sc:
If some other TAP is not in BYPASS, an IR scan is needed to select BYPASS on that TAP.

Change-Id: Iae425a415109b1a853db3718762661877eea56e8

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -357,10 +357,31 @@ static void select_dmi(struct target *target)
 		select_dmi_via_bscan(target);
 		return;
 	}
-	if (buf_eq(target->tap->cur_instr, select_dbus.out_value,
-				target->tap->ir_length))
-		return;
-	jtag_add_ir_scan(target->tap, &select_dbus, TAP_IDLE);
+	if (!target->tap->enabled)
+		LOG_TARGET_ERROR(target, "BUG: Target's TAP '%s' is disabled!",
+				jtag_tap_name(target->tap));
+
+	bool need_ir_scan = false;
+	/* FIXME: make "tap" a const pointer. */
+	for (struct jtag_tap *tap = jtag_tap_next_enabled(NULL);
+			tap; tap = jtag_tap_next_enabled(tap)) {
+		if (tap != target->tap) {
+			/* Different TAP than ours - check if it is in bypass */
+			if (!tap->bypass) {
+				need_ir_scan = true;
+				break;
+			}
+		} else {
+			/* Our TAP - check if the correct instruction is already loaded */
+			if (!buf_eq(target->tap->cur_instr, select_dbus.out_value, target->tap->ir_length)) {
+				need_ir_scan = true;
+				break;
+			}
+		}
+	}
+
+	if (need_ir_scan)
+		jtag_add_ir_scan(target->tap, &select_dbus, TAP_IDLE);
 }
 
 static int increase_dmi_busy_delay(struct target *target)
--- a/src/jtag/jtag.h
+++ b/src/jtag/jtag.h
@@ -151,6 +151,7 @@ const char *jtag_tap_name(const struct jtag_tap *tap);
 struct jtag_tap *jtag_tap_by_string(const char *dotted_name);
 struct jtag_tap *jtag_tap_by_jim_obj(Jim_Interp *interp, Jim_Obj *obj);
 struct jtag_tap *jtag_tap_by_position(unsigned int abs_position);
+/* FIXME: "jtag_tap_next_enabled()" should accept a const pointer. */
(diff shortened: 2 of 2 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -357,10 +357,18 @@ static void select_dmi(struct target *target)
...
+				jtag_tap_name(target->tap));
+
+	/* FIXME: make "tap" a const pointer. */
+	for (struct jtag_tap *tap = jtag_tap_next_enabled(NULL);
+			tap; tap = jtag_tap_next_enabled(tap))
+		if ((tap != target->tap && !tap->bypass)
+				|| (tap == target->tap
+					&& !buf_eq(target->tap->cur_instr, select_dbus.out_value,
+						target->tap->ir_length)))
+			return jtag_add_ir_scan(target->tap, &select_dbus, TAP_IDLE);
```
JanMatCodasip: I would prefer to break down the complex `if` conditions for readability. 

When trying to simplify the code, I came up with the suggestion below. Please let me know if you like it.

```suggestion
	bool need_ir_scan = false;
	for (struct jtag_tap *tap = jtag_tap_next_enabled(NULL);
			tap; tap = jtag_tap_next_enabled(tap)) {
		if ((tap != target->tap) {
			/* Different TAP than ours - check if it is in bypass */
			if (!tap->bypass) {
				need_ir_scan = true;
				break;			
			}
		} else {
			/* Our TAP - check if the correct instruction is already loaded */		
			if (!buf_eq(target->tap->cur_instr, select_dbus.out_value, target->tap->ir_length)) {
				need_ir_scan = true;
				break;						
			}
		}
	}
	
	if (need_ir_scan)
		jtag_add_ir_scan(target->tap, &select_dbus, TAP_IDLE);
```

JanMatCodasip: jtag_add_ir_scan() returns void. I am a bit surprised that the compiler allows this statement: 

```
return jtag_add_ir_scan(...);
```

For clarity, I'd recommend to use this instead:

```
jtag_add_ir_scan(...); 
return;
```

(In any case, if you decide to adopt my suggestion above, the `return` won't be needed anymore.)

JanMatCodasip: Update: I have tested this with GCC 8.5.0 and it would only complain about this if `-pedantic` were used:

```
test.c:11:12: warning: ISO C forbids ‘return’ with expression, in function returning void [-Wpedantic]
```

Still, I'd prefer to not use `return <expr>;` in void functions.

src/target/riscv/riscv-013.c
```diff
@@ -357,10 +357,18 @@ static void select_dmi(struct target *target)
...
-		return;
-	jtag_add_ir_scan(target->tap, &select_dbus, TAP_IDLE);
+	if (!target->tap->enabled)
+		LOG_TARGET_ERROR(target, "BUG: Target's TAP '%s' is disabled!",
+				jtag_tap_name(target->tap));
+
+	/* FIXME: make "tap" a const pointer. */
+	for (struct jtag_tap *tap = jtag_tap_next_enabled(NULL);
+			tap; tap = jtag_tap_next_enabled(tap))
+		if ((tap != target->tap && !tap->bypass)
```
JanMatCodasip: Could you please help me understand in which case there can be more than one non-bypassed TAP.

Is this pull request part of preparation for [this upstream change](https://review.openocd.org/c/openocd/+/8504/1) that adds multi-TAP scans?

If so, please consider mentioning it in the commit message.

JanMatCodasip (commented):
Overall, this change looks all right to me. Could you please take a look at my comments and questions. I don't expect to have any more findings than what I have just posted.

Thank you.

src/target/riscv/riscv-013.c
en-sc: Thanks for the catch! I didn't know that `return <expr>;` in a `void` function is not ISO-compliant.

src/target/riscv/riscv-013.c
en-sc: You are right, this is related to multi-TAP scans, in particular to [`jtag execute ...` command](https://review.openocd.org/c/openocd/+/8505).

However, I believe this to be a valid change by it's own.

To avoid adding `jtag_add_ir_scan()` it is needed to verify that the whole chain is in the expected state (since `jtag_add_ir_scan()` modifies the state of the whole chain).

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
JanMatCodasip: Thank you for explanation :+1:

JanMatCodasip (approved):
LGTM, thanks!

PR #1148, opened 2024-10-12
[target/riscv]read each word individually in memory progbuf

likun2:
Why was this part of the code removed in this modification（ https://github.com/riscv-collab/riscv-openocd/commit/8d660ea98d08d542e8a00b59930a7c199b856552 ）
It is useful when the full read did not succeed. This will not be fast, but reading outside actual memory is a special case anyway.It will make the toolchain happier, especially Eclipse Memory View as it reads ahead.
@en-sc

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -4270,6 +4270,35 @@ read_memory_progbuf(struct target *target, target_addr_t address,
 		read_memory_progbuf_inner_one(target, access, mprven) :
 		read_memory_progbuf_inner(target, access, count, mprven);
 
+	if (result != ERROR_OK) {
+		/* The full read did not succeed, so we will try to read each word individually. */
+		/* This will not be fast, but reading outside actual memory is a special case anyway. */
+		/* It will make the toolchain happier, especially Eclipse Memory View as it reads ahead. */
+		target_addr_t address_i = access.target_address;
+		uint8_t *buffer_i = access.buffer_address;
+
+		for (uint32_t i = 0; i < count; i++, address_i += access.increment, buffer_i += access.element_size) {
+			/* TODO: This is much slower than it needs to be because we end up
+			 * writing the address to read for every word we read. */
+			const struct memory_access_info access_single = {
+			.target_address = address_i,
+			.increment = increment,
+			.buffer_address = buffer_i,
+			.element_size = size,
+			};
+			result = read_memory_progbuf_inner_one(target, access_single, mprven);
+
+			/* The read of a single word failed, so we will just return 0 for that instead */
+			if (result != ERROR_OK) {
+				LOG_DEBUG("error reading single word of %d bytes from 0x%" TARGET_PRIxADDR,
+						access_single.element_size, access_single.target_address);
+
+				buf_set_u32(access_single.buffer_address, 0, 8 * access_single.element_size, 0);
+			}
+		}
+		result = ERROR_OK;
+	}
+
 	if (mstatus != mstatus_old &&
 			register_write_direct(target, GDB_REGNO_MSTATUS, mstatus_old) != ERROR_OK)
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -4270,6 +4270,35 @@ read_memory_progbuf(struct target *target, target_addr_t address,
...
+			const struct memory_access_info access_single = {
+			.target_address = address_i,
+			.increment = increment,
+			.buffer_address = buffer_i,
+			.element_size = size,
+			};
+			result = read_memory_progbuf_inner_one(target, access_single, mprven);
+
+			/* The read of a single word failed, so we will just return 0 for that instead */
+			if (result != ERROR_OK) {
```
aap-sc, line 4292: I don't get it. If read fails - the error should be returned.

aap-sc:
> It is useful when the full read did not succeed. This will not be fast, but reading outside actual memory is a special case anyway.

I don't like this change at all. It does not solve any issue, moreover it lies to the user by returning "zero" in cases when memory fails. This is not the intention of these functions.

> It will make the toolchain happier, especially Eclipse Memory View as it reads ahead.

No. This is not how the problem is solved. Instead of hacking OpenOCD user of a debugger should configure memory regions, so these reads won't happen (https://sourceware.org/gdb/current/onlinedocs/gdb.html/Memory-Region-Attributes.html#Memory-Region-Attributes)

aap-sc:
> , but reading outside actual memory is a special case anyway.

Special case for what? If you debug with GDB/IDE, then maybe. However, this is not the only scenario when OpenOCD is used. Situations when user tries to access unavailable memory are pretty common.

TommyMurphyTM1234:
I agree with @aap-sc. Plus, in my experience, this is more of a problem with the Eclipse memory view than with OpenOCD:

> especially Eclipse Memory View as it reads ahead.

It should deal better with memory read errors or have some way of detecting and rendering "unimplemented" areas of memory. My recollection is that it displays `0xffffffff` and errors out instead causing the console rather than the memory view to become active?

src/target/riscv/riscv-013.c
likun2: If a section of memory read by openocd starts with an illegal address and ends with a legal address, openocd will directly return an error for the entire memory.The added code can return 0 for memory reading from illegal addresses and return the correct value for memory reading from legal addresses.

likun2:
The issue with Eclipse is that when a portion of the memory  starts with an illegal address and ends with a valid address, openocd will return 0 for the entire memory . The Eclipse Memory View displays all 0 when reading this memory because the value returned by openocd is incorrect.

aap-sc:
> The issue with Eclipse is that when a portion of the memory starts with an illegal address and ends with a valid address, openocd will return 0 for the entire memory . The Eclipse Memory View displays all 0 when reading this memory because the value returned by openocd is incorrect.

1. OpenOCD does not return  zero for the entire memory. OpenOCD should return an error via gdb-remote protocol.  You should debug why the error is not processing correctly either on GDB side, or GDB-MI protocol on Eclipse side.
2.
>  Instead of hacking OpenOCD user of a debugger should configure memory regions, so these reads won't happen (https://sourceware.org/gdb/current/onlinedocs/gdb.html/Memory-Region-Attributes.html#Memory-Region-Attributes). 

Have you tried that option?

likun2:
I did not set gdb_port_data_abort to enable in cfg file , so the return value of the entire memory is 0. When I set this option, the memory value is normal and there is no need for this modification. Thank you very much for your answer

en-sc:
> When I set this option, the memory value is normal and there is no need for this modification.

@likun2, am I understanding correctly that the issue is resolved by using `gdb report_data_abort disable` and there is no need for the change in OpenOCD codebase?

If so, please, close the PR,

PR #1149, opened 2024-10-12
riscv: fix read/write virtual memory across page boundaries

zqb-all:
When read/write virtual addresses cross page boundaries, the physical addresses are not necessarily contiguous and need to call virt2phys again.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3083,11 +3083,25 @@ static int riscv_read_phys_memory(struct target *target, target_addr_t phys_addr
 	return r->read_memory(target, phys_address, size, count, buffer, size);
 }
 
-static int riscv_read_memory(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, uint8_t *buffer)
+static int riscv_write_phys_memory(struct target *target, target_addr_t phys_address,
+			uint32_t size, uint32_t count, const uint8_t *buffer)
 {
+	struct target_type *tt = get_target_type(target);
+	if (!tt)
+		return ERROR_FAIL;
+	return tt->write_memory(target, phys_address, size, count, buffer);
+}
+
+static int riscv_rw_memory(struct target *target, target_addr_t address, uint32_t size,
+		uint32_t count, uint8_t *read_buffer, const uint8_t *write_buffer)
+{
+	/* Exactly one of the buffers must be set, the other must be NULL */
+	assert(!!read_buffer != !!write_buffer);
+
+	const bool is_write = write_buffer ? true : false;
 	if (count == 0) {
-		LOG_TARGET_WARNING(target, "0-length read from 0x%" TARGET_PRIxADDR, address);
+		LOG_TARGET_WARNING(target, "0-length %s 0x%" TARGET_PRIxADDR,
+				is_write ? "write to" : "read from", address);
 		return ERROR_OK;
 	}
 
@@ -3097,11 +3111,18 @@ static int riscv_read_memory(struct target *target, target_addr_t address,
 		return result;
 
 	RISCV_INFO(r);
+	struct target_type *tt = get_target_type(target);
+	if (!tt)
+		return ERROR_FAIL;
 
-	if (!mmu_enabled)
-		return r->read_memory(target, address, size, count, buffer, size);
+	if (!mmu_enabled) {
+		if (is_write)
+			return tt->write_memory(target, address, size, count, write_buffer);
+		else
+			return r->read_memory(target, address, size, count, read_buffer, size);
+	}
 
-	result = check_virt_memory_access(target, address, size, count, false);
+	result = check_virt_memory_access(target, address, size, count, is_write);
 	if (result != ERROR_OK)
 		return result;
 
@@ -3118,7 +3139,10 @@ static int riscv_read_memory(struct target *target, target_addr_t address,
 		 * which is  4 KiB. The algorithm can be improved to detect the real page size, and allow to use larger
 		 * memory transfers and avoid extra unnecessary virt2phys address translations. */
 		uint32_t chunk_count = MIN(count - current_count, (RISCV_PGSIZE - RISCV_PGOFFSET(address)) / size);
-		result = r->read_memory(target, physical_addr, size, chunk_count, buffer + current_count * size, size);
+		if (is_write)
+			result = tt->write_memory(target, physical_addr, size, chunk_count, write_buffer + current_count * size);
+		else
+			result = r->read_memory(target, physical_addr, size, chunk_count, read_buffer + current_count * size, size);
 		if (result != ERROR_OK)
 			return result;
 
@@ -3128,57 +3152,16 @@ static int riscv_read_memory(struct target *target, target_addr_t address,
 	return ERROR_OK;
 }
 
-static int riscv_write_phys_memory(struct target *target, target_addr_t phys_address,
-			uint32_t size, uint32_t count, const uint8_t *buffer)
+static int riscv_read_memory(struct target *target, target_addr_t address,
+		uint32_t size, uint32_t count, uint8_t *buffer)
 {
-	struct target_type *tt = get_target_type(target);
-	if (!tt)
-		return ERROR_FAIL;
-	return tt->write_memory(target, phys_address, size, count, buffer);
+	return riscv_rw_memory(target, address, size, count, buffer, NULL);
 }
 
 static int riscv_write_memory(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, const uint8_t *buffer)
 {
-	if (count == 0) {
-		LOG_TARGET_WARNING(target, "0-length write to 0x%" TARGET_PRIxADDR, address);
-		return ERROR_OK;
-	}
-
-	int mmu_enabled;
-	int result = riscv_mmu(target, &mmu_enabled);
-	if (result != ERROR_OK)
-		return result;
-
-	struct target_type *tt = get_target_type(target);
-	if (!tt)
-		return ERROR_FAIL;
-
-	if (!mmu_enabled)
-		return tt->write_memory(target, address, size, count, buffer);
-
-	result = check_virt_memory_access(target, address, size, count, true);
-	if (result != ERROR_OK)
-		return result;
-
-	uint32_t current_count = 0;
-	while (current_count < count) {
-		target_addr_t physical_addr;
-		result = target->type->virt2phys(target, address, &physical_addr);
-		if (result != ERROR_OK) {
-			LOG_TARGET_ERROR(target, "Address translation failed.");
-			return result;
-		}
-
-		uint32_t chunk_count = MIN(count - current_count, (RISCV_PGSIZE - RISCV_PGOFFSET(address)) / size);
-		result = tt->write_memory(target, physical_addr, size, chunk_count, buffer + current_count * size);
-		if (result != ERROR_OK)
-			return result;
-
-		current_count += chunk_count;
-		address += chunk_count * size;
-	}
-	return ERROR_OK;
+	return riscv_rw_memory(target, address, size, count, NULL, buffer);
 }
 
 static const char *riscv_get_gdb_arch(const struct target *target)
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3061,6 +3061,27 @@ static int riscv_virt2phys(struct target *target, target_addr_t virtual, target_
 			virtual, physical);
 }
 
+static int check_access_request(struct target *target, target_addr_t address,
+			uint32_t size, uint32_t count, bool is_write)
+{
+	const bool is_misaligned = address % size != 0;
+	// TODO: This assumes that size of each page is 4 KiB, which is not necessarily the case.
+	const bool crosses_page_boundary = RISCV_PGBASE(address + size * count - 1) != RISCV_PGBASE(address);
+	if (is_misaligned && crosses_page_boundary) {
+		int mmu_enabled;
+		int result = riscv_mmu(target, &mmu_enabled);
+		if (result != ERROR_OK)
+			return result;
+		if (mmu_enabled) {
+			LOG_TARGET_ERROR(target, "Mis-aligned memory %s (address=0x%" TARGET_PRIxADDR ", size=%d, count=%d)"
+				" would access an element across page boundary. This is not supported.",
+				 is_write ? "write" : "read", address, size, count);
+			return ERROR_FAIL;
+		}
+	}
+	return ERROR_OK;
+}
+
 static int riscv_read_phys_memory(struct target *target, target_addr_t phys_address,
 			uint32_t size, uint32_t count, uint8_t *buffer)
 {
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv.c
```diff
@@ -3071,20 +3071,38 @@ static int riscv_read_phys_memory(struct target *target, target_addr_t phys_addr
...
 		return ERROR_OK;
 	}
 
-	target_addr_t physical_addr;
-	int result = target->type->virt2phys(target, address, &physical_addr);
-	if (result != ERROR_OK) {
-		LOG_TARGET_ERROR(target, "Address translation failed.");
-		return result;
+	if (address % size) {
+		LOG_TARGET_ERROR(target, "Address 0x%" TARGET_PRIxADDR " not align to size %d", address, size);
```
TommyMurphyTM1234: "aligned" not "align"

src/target/riscv/riscv.c
```diff
@@ -3099,22 +3117,41 @@ static int riscv_write_phys_memory(struct target *target, target_addr_t phys_add
...
-		return result;
-	}
-
 	struct target_type *tt = get_target_type(target);
 	if (!tt)
 		return ERROR_FAIL;
-	return tt->write_memory(target, physical_addr, size, count, buffer);
+
+	if (address % size) {
+		LOG_TARGET_ERROR(target, "Address 0x%" TARGET_PRIxADDR " not align to size %d", address, size);
```
TommyMurphyTM1234: "aligned" not "align"

TommyMurphyTM1234 (commented):
Small typos

src/target/riscv/riscv.c
zqb-all: Thanks. Addressed.

src/target/riscv/riscv.c
zqb-all: Thanks. Addressed.

src/target/riscv/riscv.c
```diff
@@ -3099,22 +3117,41 @@ static int riscv_write_phys_memory(struct target *target, target_addr_t phys_add
...
-
 	struct target_type *tt = get_target_type(target);
 	if (!tt)
 		return ERROR_FAIL;
-	return tt->write_memory(target, physical_addr, size, count, buffer);
+
+	if (address % size) {
+		LOG_TARGET_ERROR(target, "Address 0x%" TARGET_PRIxADDR " not aligned to size %d", address, size);
+		return ERROR_FAIL;
+	}
```
en-sc: AFAIU, misaligned access can be supported and is supported in current OpenOCD.

However, IMHO, accessing an element across a page boundary can be addressed later.

I'd suggest something like:

```suggestion
        const target_addr_t page_offset_mask = RISCV_PGSIZE - 1;
	const bool crosses_page_boundary = (address + size * count - 1) & ~page_offset_mask != address & ~page_offset_mask;
	if (address % size && crosses_page_boundary) {
		int mmu_enabled;
		int res = riscv_mmu(target, &mmu_enabled);
		if (res != ERROR_OK)
			return res;
		if (mmu_enabled) {
			LOG_TARGET_ERROR(target, "Accessing an element across a page boundary is not supported.");
			return ERROR_FAIL;
		}
	}
```

src/target/riscv/riscv.c
```diff
@@ -3071,20 +3071,38 @@ static int riscv_read_phys_memory(struct target *target, target_addr_t phys_addr
 static int riscv_read_memory(struct target *target, target_addr_t address,
 		uint32_t size, uint32_t count, uint8_t *buffer)
 {
+	uint32_t current_count = 0;
```
en-sc: There is no need to declare variables at the top of a function. If you like, the declaration can be moved closer to the use of the variable.

en-sc (changes requested):
Thanks for the patch!

I have only one significant comment.

src/target/riscv/riscv.c
zqb-all: Thanks. Addressed.

src/target/riscv/riscv.c
zqb-all: Thanks. Addressed.

src/target/riscv/riscv.c
```diff
@@ -3076,15 +3076,39 @@ static int riscv_read_memory(struct target *target, target_addr_t address,
...
+		((address + size * count - 1) & ~page_offset_mask) != (address & ~page_offset_mask);
+	if (address % size && crosses_page_boundary) {
+		int mmu_enabled;
+		int result = riscv_mmu(target, &mmu_enabled);
+		if (result != ERROR_OK)
+			return result;
+		if (mmu_enabled) {
+			LOG_TARGET_ERROR(target, "Accessing an element across a page boundary is not supported.");
+			return ERROR_FAIL;
+		}
```
en-sc: This code is used in both `riscv_read_memory()` and `riscv_write_memory()`.
Maybe it can be separated into a function like `int detect_write_on_page_boundary(...)`?

zqb-all: Agree.
In fact, the entire function code is very similar. But I'm not sure if it's worth creating something like` riscv_rw_memory(..., bool is_write)` ， then both` riscv_read_memory()` and `riscv_write_memory()` can call it.

src/target/riscv/riscv.c
```diff
@@ -3076,15 +3093,29 @@ static int riscv_read_memory(struct target *target, target_addr_t address,
...
+	uint32_t current_count = 0;
+	while (current_count < count) {
+		target_addr_t physical_addr;
+		result = target->type->virt2phys(target, address, &physical_addr);
+		if (result != ERROR_OK) {
+			LOG_TARGET_ERROR(target, "Address translation failed.");
+			return result;
+		}
+
+		uint32_t chunk_count = MIN(count - current_count, (RISCV_PGSIZE - RISCV_PGOFFSET(address)) / size);
```
zqb-all: When mmu is not enabled, access does not need to consider crossing page boundaries.
Is it worth changing it to
```
if (mmu_enable)
  chunk_count = MIN(count - current_count, (RISCV_PGSIZE - RISCV_PGOFFSET(address)) / size);
else
  chunk_count = count;
```

zqb-all: Oh, thinking about this made me realize that there is a problem with misaligned access support here. 
`( RISCV_PGSIZE - RISCV_PGOFFSET(address))/size` sets `chunk_count` to 0, so the while loop will never end.

zqb-all: https://github.com/riscv-collab/riscv-openocd/pull/1149/commits/cd29d6a7219a5ac62cbed46f14762d378664c1d7

src/target/riscv/riscv.c
en-sc: >  But I'm not sure if it's worth creating something like `riscv_rw_memory(..., bool is_write)`

IMHO that is a great idea and will make the code much more maintainable.

en-sc (dismissed):
LGTM

en-sc:
@JanMatCodasip, @MarekVCodasip, please, take a look.

src/target/riscv/riscv.c
zqb-all: Done. At first I wanted to use ` uint8_t* buffer, bool is_write`, but since` riscv_write_memory` uses const buffer, I had to use two buffer parameters:  `uint8_t *read_buffer, const uint8_t *write_buffer`

src/target/riscv/riscv.c
```diff
@@ -3096,25 +3088,76 @@ static int riscv_write_phys_memory(struct target *target, target_addr_t phys_add
 	return tt->write_memory(target, phys_address, size, count, buffer);
 }
 
-static int riscv_write_memory(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, const uint8_t *buffer)
+static int riscv_rw_memory(struct target *target, target_addr_t address, uint32_t size,
+		uint32_t count, uint8_t *read_buffer, const uint8_t *write_buffer)
 {
+	if (read_buffer && write_buffer) {
+		LOG_TARGET_ERROR(target, "Can't handle both read and write buffer.");
```
JanMatCodasip: 1) Detail: OpenOCD upstream developers prefer to prefix messages like these with `BUG:`.

```
		LOG_TARGET_ERROR(target, "BUG: Can't have both read and write buffer specified.");
```

2) The case of `(!read_buffer && !write_buffer)` should be checked, too, for completeness.

My preference would be to replace both these checks with an assert, since these are programmer's invariants and not runtime error checks.

For example:

```c
/* Exactly one of the buffers must be set, the other must be NULL */
assert(!!read_buffer != !!write_buffer);
```

src/target/riscv/riscv.c
```diff
@@ -3096,25 +3088,76 @@ static int riscv_write_phys_memory(struct target *target, target_addr_t phys_add
...
+	uint32_t current_count = 0;
+	while (current_count < count) {
+		target_addr_t physical_addr;
+		result = target->type->virt2phys(target, address, &physical_addr);
+		if (result != ERROR_OK) {
+			LOG_TARGET_ERROR(target, "Address translation failed.");
+			return result;
+		}
+
+		uint32_t chunk_count = MIN(count - current_count, (RISCV_PGSIZE - RISCV_PGOFFSET(address)) / size);
```
JanMatCodasip, line 3141: This assumes the "worst case" - pages of size 4 KiB. The pages can however be larger than that (e.g. 2MiB).

I don't mind this limitation, though. Please, just make a TODO comment about that, so that it is very clear to future code readers.

For example:

```suggestion
		/* TODO: For simplicity, this algorithm assumes the worst case - the smallest possible page size,
		 * which is  4 KiB. The algorithm can be improved to detect the real page size, and allow to use larger
		 * memory transfers and avoid extra unnecessary virt2phys address translations. */
		uint32_t chunk_count = MIN(count - current_count, (RISCV_PGSIZE - RISCV_PGOFFSET(address)) / size);
```

src/target/riscv/riscv.c
```diff
@@ -3061,30 +3061,22 @@ static int riscv_virt2phys(struct target *target, target_addr_t virtual, target_
...
 
-static int riscv_read_phys_memory(struct target *target, target_addr_t phys_address,
-			uint32_t size, uint32_t count, uint8_t *buffer)
+static int check_virt_memory_access(struct target *target, target_addr_t address,
+			uint32_t size, uint32_t count)
 {
-	RISCV_INFO(r);
-	return r->read_memory(target, phys_address, size, count, buffer, size);
+	bool crosses_page_boundary = RISCV_PGBASE(address + size * count - 1) != RISCV_PGBASE(address);
+	if (address % size && crosses_page_boundary) {
```
JanMatCodasip: Detail for readability:

```suggestion
	const bool is_misaligned = address % size != 0;
	const bool crosses_page_boundary = RISCV_PGBASE(address + size * count - 1) != RISCV_PGBASE(address);
	if (is_misaligned && crosses_page_boundary) {
```

src/target/riscv/riscv.c
```diff
@@ -3061,30 +3061,22 @@ static int riscv_virt2phys(struct target *target, target_addr_t virtual, target_
...
+		LOG_TARGET_ERROR(target, "Accessing an element across a page boundary is not supported.");
```
JanMatCodasip: It was bit harder for me as a reader to grasp what has actually happened and why we need to check for this case. For that reason, could you please improve the error message. It could be phrased like this:

`Mis-aligned memory read|write (address=XX, size=XX, count=XX) would access an element across page boundary. This is not supported.`

Optional: The error print can also be moved one level up to riscv_rw_memory(), if you find it more convenient.

src/target/riscv/riscv.c
```diff
@@ -3096,25 +3088,76 @@ static int riscv_write_phys_memory(struct target *target, target_addr_t phys_add
...
-		uint32_t size, uint32_t count, const uint8_t *buffer)
+static int riscv_rw_memory(struct target *target, target_addr_t address, uint32_t size,
+		uint32_t count, uint8_t *read_buffer, const uint8_t *write_buffer)
 {
+	if (read_buffer && write_buffer) {
+		LOG_TARGET_ERROR(target, "Can't handle both read and write buffer.");
+		return ERROR_FAIL;
+	}
+
+	bool is_write = write_buffer ? true : false;
```
JanMatCodasip: Nitpick (it helps code readers a little):

```suggestion
	const bool is_write = write_buffer ? true : false;
```

src/target/riscv/riscv.c
```diff
@@ -3061,30 +3061,22 @@ static int riscv_virt2phys(struct target *target, target_addr_t virtual, target_
...
 }
 
-static int riscv_read_phys_memory(struct target *target, target_addr_t phys_address,
-			uint32_t size, uint32_t count, uint8_t *buffer)
+static int check_virt_memory_access(struct target *target, target_addr_t address,
+			uint32_t size, uint32_t count)
 {
-	RISCV_INFO(r);
-	return r->read_memory(target, phys_address, size, count, buffer, size);
+	bool crosses_page_boundary = RISCV_PGBASE(address + size * count - 1) != RISCV_PGBASE(address);
```
JanMatCodasip: This part also makes an assumption that that pages are 4 KiB large, so it would be good to make a TODO comment about it:

```suggestion
	// TODO: This assumes that size of each page is 4 KiB, which is not necessarily the case.
	bool crosses_page_boundary = RISCV_PGBASE(address + size * count - 1) != RISCV_PGBASE(address);
```

JanMatCodasip (changes requested):
@zqb-all Thank you for the contribution. In principle, the change looks fine to me.

I have just several code-related suggestions (code style, log prints, TODO comments and similar). Please if you can look at them.

src/target/riscv/riscv.c
zqb-all: Thanks, this assert code is pretty clean. Addressed.

src/target/riscv/riscv.c
zqb-all: Thanks. Addressed.

src/target/riscv/riscv.c
zqb-all: Thanks. Addressed.

src/target/riscv/riscv.c
zqb-all: Thanks. Addressed.

src/target/riscv/riscv.c
zqb-all: Thanks. Addressed.

src/target/riscv/riscv.c
zqb-all: Thanks. Addressed.

src/target/riscv/riscv.c
```diff
@@ -3061,32 +3061,28 @@ static int riscv_virt2phys(struct target *target, target_addr_t virtual, target_
...
+			uint32_t size, uint32_t count)
+{
+	const bool is_misaligned = address % size != 0;
+	// TODO: This assumes that size of each page is 4 KiB, which is not necessarily the case.
+	const bool crosses_page_boundary = RISCV_PGBASE(address + size * count - 1) != RISCV_PGBASE(address);
+	if (is_misaligned && crosses_page_boundary) {
+		LOG_TARGET_ERROR(target, "Mis-aligned memory read|write (address=0x%" TARGET_PRIxADDR ", size=%d, count=%d)"
+			" would access an element across page boundary. This is not supported.",
+			address, size, count);
+		return ERROR_FAIL;
```
JanMatCodasip: Here I meant that the actual type of the memory operation would be printed (`read` or `write`):

```suggestion
	if (is_misaligned && crosses_page_boundary) {
		LOG_TARGET_ERROR(target, "Mis-aligned memory %s (address=0x%" TARGET_PRIxADDR ", size=%d, count=%d)"
			" would access an element across page boundary. This is not supported.",
			is_write ? "write" : "read", address, size, count);
		return ERROR_FAIL;
```

src/target/riscv/riscv.c
```diff
@@ -3096,25 +3092,77 @@ static int riscv_write_phys_memory(struct target *target, target_addr_t phys_add
...
 		return result;
-	}
 
+	RISCV_INFO(r);
 	struct target_type *tt = get_target_type(target);
 	if (!tt)
 		return ERROR_FAIL;
-	return tt->write_memory(target, physical_addr, size, count, buffer);
+
+
```
JanMatCodasip: Detail: Please, don't use two blank lines.

```suggestion

```

JanMatCodasip (commented):
Thank you for addressing the comments so far.

I have two last small suggestions, otherwise it looks good.

src/target/riscv/riscv.c
zqb-all: Aha, now I understand that the `Optional: The error print can also be moved one level up to riscv_rw_memory(), if you find it more convenient.` in the original comment is because `is_write `is required.
Addressed.

src/target/riscv/riscv.c
zqb-all: Mistake. Addressed. Perhaps we can try to detect such problems in the `Check Code Style` phase in the future.

JanMatCodasip: The checkpatch tool apparently does not check for extra blank lines. 

I would also prefer to have a more robust tool for code style checking (e.g. clang-format) but it is uncertain whether:

- me or anyone else will find the time to prepare it
- upstream developers would be in favor of having it

JanMatCodasip (approved):
LGTM, thank you.

src/target/riscv/riscv.c
zqb-all: In fact, the checkpatch can check it, let's see if upstream developers are willing to turn it on:
https://review.openocd.org/c/openocd/+/8365

en-sc (approved):

PR #1150, opened 2024-10-13
target: riscv: convert 'unsigned' to 'unsigned int'

borneoa:
This is only for coding-style rework.

In upstream OpenOCD all 'unsigned' has already been converted to 'unsigned int' and there are two patches pending for merge to extend the conversion to the risc-v code:
https://review.openocd.org/c/openocd/+/8483
https://review.openocd.org/c/openocd/+/8484

This patch for the risc-v fork runs the same conversion, in the hope to simplify the alignment between the fork and the upstream.

Conversion done with
	checkpatch --fix-inplace -types UNSPECIFIED_INT

Change-Id: I62fad88dd33716c24154d44c5a23ae2c0f7c4a4c

### DIFF
--- a/src/target/riscv/opcodes.h
+++ b/src/target/riscv/opcodes.h
@@ -194,26 +194,26 @@ static uint32_t fld(unsigned int dest, unsigned int base, uint16_t offset)
 	return imm_i(offset) | inst_rs1(base) | inst_rd(dest) | MATCH_FLD;
 }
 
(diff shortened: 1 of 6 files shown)


en-sc (approved):
Thanks for the patch! LGTM.

JanMatCodasip (approved):
LGTM, thank you.

PR #1152, opened 2024-10-16
target/riscv: added translation drivers

fkhaidari:
Existing flags: 'enable_virtual' and 'enable_virt2phys' were replaced with explicit translation drivers. Motivation:

(1) Having 'enable_virtual' and 'enable_virt2phys' flags set simultaneously may cause double address translation which is unacceptable

(2) Flags were global for all targets which is wrong too

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -11356,16 +11356,18 @@ This command can be used to change the memory access methods if the default
 behavior is not suitable for a particular target.
 @end deffn
 
-@deffn {Command} {riscv set_enable_virtual} on|off
-When on, memory accesses are performed on physical or virtual memory depending
-on the current system configuration. When off (default), all memory accessses are performed
-on physical memory.
-@end deffn
-
-@deffn {Command} {riscv set_enable_virt2phys} on|off
-When on (default), memory accesses are performed on physical or virtual memory
-depending on the current satp configuration. When off, all memory accessses are
-performed on physical memory.
+@deffn {Command} {riscv virt2phys_mode} [@option{hw}|@option{sw}|@option{off}]
+Configure how OpenOCD translates virtual addresses to physical:
+@itemize @bullet
+@item @option{sw} - OpenOCD translates virtual addresses explicitly by
+traversing the page table entries (by performing physical memory accesses to
+read the respective entries). This is the default mode.
+@item @option{hw} - Virtual addresses are translated implicitly by hardware.
+(Virtual memory access will fail with an error if the hardware doesn't
+support the necessary functionality.)
+@item @option{off} - Virtual addresses are not translated (identity mapping is assumed).
+@end itemize
+Returns current translation mode if called without arguments.
 @end deffn
 
 @deffn {Command} {riscv resume_order} normal|reversed
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1974,11 +1974,11 @@ static int examine(struct target *target)
 				info->impebreak);
 	}
 
-	if (info->progbufsize < 4 && riscv_enable_virtual) {
-		LOG_TARGET_ERROR(target, "set_enable_virtual is not available on this target. It "
-				"requires a program buffer size of at least 4. (progbufsize=%d) "
-				"Use `riscv set_enable_virtual off` to continue."
-					, info->progbufsize);
+	if (info->progbufsize < 4 && riscv_virt2phys_mode_is_hw(target)) {
+		LOG_TARGET_ERROR(target, "software address translation "
+				"is not available on this target. It requires a "
+				"program buffer size of at least 4. (progbufsize=%d) "
+				"Use `riscv set_enable_virtual off` to continue.", info->progbufsize);
 	}
 
 	/* Don't call any riscv_* functions until after we've counted the number of
@@ -3054,7 +3054,8 @@ static int read_sbcs_nonbusy(struct target *target, uint32_t *sbcs)
 
 static int modify_privilege(struct target *target, uint64_t *mstatus, uint64_t *mstatus_old)
 {
-	if (riscv_enable_virtual && has_sufficient_progbuf(target, 5)) {
+	if (riscv_virt2phys_mode_is_hw(target)
+			&& has_sufficient_progbuf(target, 5)) {
 		/* Read DCSR */
 		uint64_t dcsr;
 		if (register_read_direct(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
@@ -4259,7 +4260,8 @@ read_memory_progbuf(struct target *target, target_addr_t address,
 	if (modify_privilege(target, &mstatus, &mstatus_old) != ERROR_OK)
 		return MEM_ACCESS_FAILED_PRIV_MOD_FAILED;
 
-	const bool mprven = riscv_enable_virtual && get_field(mstatus, MSTATUS_MPRV);
+	const bool mprven = riscv_virt2phys_mode_is_hw(target)
+			&& get_field(mstatus, MSTATUS_MPRV);
 	const struct memory_access_info access = {
 		.target_address = address,
 		.increment = increment,
@@ -4831,7 +4833,8 @@ write_memory_progbuf(struct target *target, target_addr_t address,
 	if (modify_privilege(target, &mstatus, &mstatus_old) != ERROR_OK)
 		return MEM_ACCESS_FAILED_PRIV_MOD_FAILED;
 
-	const bool mprven = riscv_enable_virtual && get_field(mstatus, MSTATUS_MPRV);
+	const bool mprven = riscv_virt2phys_mode_is_hw(target)
+			&& get_field(mstatus, MSTATUS_MPRV);
 
 	int result = write_memory_progbuf_inner(target, address, size, count, buffer, mprven);
 
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -139,6 +139,35 @@ struct tdata1_cache {
 	struct list_head elem_tdata1;
 };
 
+bool riscv_virt2phys_mode_is_hw(const struct target *target)
+{
+	assert(target);
+	RISCV_INFO(r);
+	return r->virt2phys_mode == RISCV_VIRT2PHYS_MODE_HW;
+}
+
+bool riscv_virt2phys_mode_is_sw(const struct target *target)
+{
+	assert(target);
+	RISCV_INFO(r);
+	return r->virt2phys_mode == RISCV_VIRT2PHYS_MODE_SW;
+}
+
+const char *riscv_virt2phys_mode_to_str(riscv_virt2phys_mode_t mode)
+{
+	assert(mode == RISCV_VIRT2PHYS_MODE_OFF
+			|| mode == RISCV_VIRT2PHYS_MODE_SW
+			|| mode == RISCV_VIRT2PHYS_MODE_HW);
+
+	static const char *const names[] = {
+		[RISCV_VIRT2PHYS_MODE_HW] = "hw",
+		[RISCV_VIRT2PHYS_MODE_SW] = "sw",
+		[RISCV_VIRT2PHYS_MODE_OFF] = "off",
+	};
+
+	return names[mode];
+}
+
 /* Wall-clock timeout for a command/access. Settable via RISC-V Target commands.*/
 static int riscv_command_timeout_sec_value = DEFAULT_COMMAND_TIMEOUT_SEC;
 
@@ -149,10 +178,6 @@ int riscv_get_command_timeout_sec(void)
 {
 	return MAX(riscv_command_timeout_sec_value, riscv_reset_timeout_sec);
 }
-
-static bool riscv_enable_virt2phys = true;
-
-bool riscv_enable_virtual;
 
 static enum {
 	RO_NORMAL,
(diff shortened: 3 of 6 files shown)


aap-sc:
@MarekVCodasip , @JanMatCodasip could you take a look, please?

en-sc (dismissed):
LGTM (reviewed internally).

src/target/riscv/riscv.c
```diff
@@ -5093,6 +5098,38 @@ COMMAND_HANDLER(handle_reserve_trigger)
...
+	if (CMD_ARGC == 0) {
+		riscv_translation_driver_t driver = info->translation_driver;
+		command_print(CMD, "%s", riscv_translation_driver_to_str(driver));
+		return ERROR_OK;
+	}
+
+	if (CMD_ARGC != 1) {
+		command_print(CMD, "Command takes 0 or 1 parameters");
+		return ERROR_COMMAND_ARGUMENT_INVALID;
+	}
```

IIRC, the upstream reviewers prefer to return different error code in this case, and to not display any error message. That's because OpenOCD will show the command's help automatically when "SYNTAX_ERROR" is returned.

```suggestion
	if (CMD_ARGC != 1)
		return ERROR_COMMAND_SYNTAX_ERROR;
```

src/target/riscv/riscv.c
```diff
@@ -5353,6 +5373,16 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.usage = "[index ('on'|'off')]",
 		.help = "Controls which RISC-V triggers shall not be touched by OpenOCD.",
 	},
+	{
+		.name = "mmu_translation_driver",
+		.handler = handle_riscv_mmu_translation_driver,
+		.mode = COMMAND_ANY,
+		.usage = "'sw'|'hw'|'off'",
```
```suggestion
		.usage = "['sw'|'hw'|'off']",
```

doc/openocd.texi
```diff
@@ -11356,16 +11356,18 @@ This command can be used to change the memory access methods if the default
...
-When on, memory accesses are performed on physical or virtual memory depending
-on the current system configuration. When off (default), all memory accessses are performed
-on physical memory.
-@end deffn
-
-@deffn {Command} {riscv set_enable_virt2phys} on|off
-When on (default), memory accesses are performed on physical or virtual memory
-depending on the current satp configuration. When off, all memory accessses are
-performed on physical memory.
+@deffn {Command} {riscv mmu_translation_driver} [@option{hw}|@option{sw}|@option{off}]
```
JanMatCodasip: I was thinking about a more descriptive name for the command. Does `riscv virt2phys_translation` or `riscv virt2phys_mode` sound good to you?

I'd prefer to not use the word "driver" and instead call it "mode".

```suggestion
@deffn {Command} {riscv virt2phys_translation} [@option{hw}|@option{sw}|@option{off}]
```

or:

```suggestion
@deffn {Command} {riscv virt2phys_mode} [@option{hw}|@option{sw}|@option{off}]
```

doc/openocd.texi
```diff
@@ -11356,16 +11356,18 @@ This command can be used to change the memory access methods if the default
...
+Configure address translation driver:
```
JanMatCodasip: Some reword suggestions for clarity:

```suggestion
Configure how OpenOCD translates virtual addresses to physical:
```

doc/openocd.texi
```diff
@@ -11356,16 +11356,18 @@ This command can be used to change the memory access methods if the default
...
+@itemize @bullet
+@item @option{sw} - virtual address is translated explicitly by OpenOCD using
+traversal of page table entries (by performing physical memory accesses to
+read the respected entries).
```
JanMatCodasip: Reword suggestions and corrections, cont'd:

```suggestion
@item @option{sw} - OpenOCD translates virtual addresses explicitly by
traversing the page table entries (by performing physical memory accesses to
read the respective entries). This is the default mode.
```

doc/openocd.texi
```diff
@@ -11356,16 +11356,18 @@ This command can be used to change the memory access methods if the default
...
+@item @option{hw} - virtual address is translated implicitly by hardware
+(virtual memory access fails with error if hardware doesn't
+support the necessary functionality).
```
JanMatCodasip:
```suggestion
@item @option{hw} - Virtual addresses are translated implicitly by hardware.
(Virtual memory access will fail with an error if the hardware doesn't
support the necessary functionality.)
```

doc/openocd.texi
```diff
@@ -11356,16 +11356,18 @@ This command can be used to change the memory access methods if the default
...
+@item @option{off} - virtual address is not translated (assume identity mapping).
```
JanMatCodasip:
```suggestion
@item @option{off} - Virtual addresses are not translated (identity mapping is assumed).
```

doc/openocd.texi
```diff
@@ -11356,16 +11356,18 @@ This command can be used to change the memory access methods if the default
...
+@end itemize
+Returns current translation driver if called without arguments.
```
JanMatCodasip:
```suggestion
Returns current translation mode if called without arguments.
```

src/target/riscv/riscv.c
```diff
@@ -139,6 +139,35 @@ struct tdata1_cache {
 	struct list_head elem_tdata1;
 };
 
+bool riscv_target_config_allows_hw_translation(const struct target *target)
```
JanMatCodasip: Please, consider shortening the name of the function.

You can for example call it `riscv_is_virt2phys_hw_mode()` which is shorter, readable and still very clear.

```suggestion
bool riscv_is_virt2phys_hw_mode(const struct target *target)
```

src/target/riscv/riscv.c
```diff
@@ -139,6 +139,35 @@ struct tdata1_cache {
...
+{
+	assert(target);
+	RISCV_INFO(r);
+	return r->translation_driver == RISCV_TRANSLATION_DRIVER_HW;
+}
+
+bool riscv_target_config_allows_sw_translation(const struct target *target)
```

```suggestion
bool riscv_is_virt2phys_sw_mode(const struct target *target)
```

src/target/riscv/riscv.c
```diff
@@ -139,6 +139,35 @@ struct tdata1_cache {
...
+{
+	assert(target);
+	RISCV_INFO(r);
+	return r->translation_driver == RISCV_TRANSLATION_DRIVER_SW;
+}
+
+const char *riscv_translation_driver_to_str(riscv_translation_driver_t driver)
```
```suggestion
const char *riscv_virt2phys_mode_to_str(riscv_translation_driver_t driver)
```

src/target/riscv/riscv.h
```diff
@@ -57,6 +57,14 @@ typedef enum riscv_mem_access_method {
 	RISCV_MEM_ACCESS_MAX_METHODS_NUM
 } riscv_mem_access_method_t;
 
+typedef enum riscv_translation_driver {
+	RISCV_TRANSLATION_DRIVER_HW,
+	RISCV_TRANSLATION_DRIVER_SW,
+	RISCV_TRANSLATION_DRIVER_OFF
+} riscv_translation_driver_t;
```
JanMatCodasip: Similarly to the other comments, this can be renamed. For example:

```suggestion
typedef enum riscv_virt2phys_mode {
	RISCV_VIRT2PHYS_MODE_HW,
	RISCV_VIRT2PHYS_MODE_SW,
	RISCV_VIRT2PHYS_MODE_OFF
} riscv_virt2phys_mode_t;
```

src/target/riscv/riscv.h
```diff
@@ -165,6 +173,9 @@ struct riscv_info {
 	 * most recent halt was not caused by a trigger, then this is -1. */
 	int64_t trigger_hit;
 
+	/* The way to translate virtual address to physical */
+	riscv_translation_driver_t translation_driver;
```
JanMatCodasip:
```suggestion
	/* The configured approach to translate virtual addresses to physical */
	riscv_virt2phys_mode_t virt2phys_mode;
```

src/target/riscv/startup.tcl
```diff
@@ -0,0 +1,51 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
```
JanMatCodasip, line 1: Just a comment - no change requested:

I have not reviewed the changes in this file in detail - I trust you on those.

The new type of tests that @en-sc has proposed [here](https://review.openocd.org/c/openocd/+/8507) would be very good for testing changes like these. Perhaps you are using such tests already internally?

src/target/riscv/riscv.c
```diff
@@ -5353,6 +5373,16 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.usage = "[index ('on'|'off')]",
 		.help = "Controls which RISC-V triggers shall not be touched by OpenOCD.",
 	},
+	{
+		.name = "mmu_translation_driver",
+		.handler = handle_riscv_mmu_translation_driver,
+		.mode = COMMAND_ANY,
+		.usage = "'sw'|'hw'|'off'",
+		.help = "Configure address translation way: "
```
```suggestion
		.help = "Configure the virtual address translation mode: "
```

or

```suggestion
		.help = "Configure the virtual address translation mechanism: "
```

or similar

src/target/riscv/riscv-013.c
```diff
@@ -1974,7 +1974,7 @@ static int examine(struct target *target)
 				info->impebreak);
 	}
 
-	if (info->progbufsize < 4 && riscv_enable_virtual) {
+	if (info->progbufsize < 4 && riscv_target_config_allows_hw_translation(target)) {
 		LOG_TARGET_ERROR(target, "set_enable_virtual is not available on this target. It "
```
JanMatCodasip: The old command name (`set_enable_virtual`) in the message should be replaced by the new name.

src/target/riscv/riscv-013.c
```diff
@@ -1974,7 +1974,7 @@ static int examine(struct target *target)
 				info->impebreak);
 	}
 
-	if (info->progbufsize < 4 && riscv_enable_virtual) {
```
JanMatCodasip, line 1977: The program buffer size check `info->progbufsize < 4` does not look right because it does not account for `impebreak`. Can you please check it.

I believe `has_sufficient_progbuf()` should be used instead.

JanMatCodasip (commented):
This change looks very good to me and makes the TCL interface for virt2phys configuration much clearer than it is now.

I have only minor comments, mostly just suggestions for renaming or reword in documentation. Please if you can check them out.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

doc/openocd.texi
fkhaidari: Addressed

doc/openocd.texi
fkhaidari: Addressed

doc/openocd.texi
fkhaidari: Addressed

doc/openocd.texi
fkhaidari: Addressed

doc/openocd.texi
fkhaidari: Addressed

doc/openocd.texi
fkhaidari: Addressed

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.h
fkhaidari: Addressed

src/target/riscv/riscv.h
fkhaidari: Addressed

src/target/riscv/riscv.c

src/target/riscv/riscv-013.c
fkhaidari: I will address this in future MR

fkhaidari:
@JanMatCodasip thank you for the review. All threads were addressed

src/target/riscv/riscv-013.c
JanMatCodasip: Ok, thank you.

src/target/riscv/startup.tcl
EccoTheDolphin: ...

src/target/riscv/riscv.c
```diff
@@ -5353,6 +5371,16 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
 		.help = "Controls which RISC-V triggers shall not be touched by OpenOCD.",
 	},
+	{
+		.name = "virt2phys_mode",
+		.handler = handle_riscv_virt2phys_mode,
+		.mode = COMMAND_ANY,
+		.usage = "['sw'|'hw'|'off']",
+		.help = "Configure the virtual address translation mode: "
+				"sw - translate vaddr to paddr by manually traversing page tables, "
+				"hw - translate vaddr to paddr by hardware,"
```
JanMatCodasip: Really a detail: Added trailing space character:

```suggestion
				"hw - translate vaddr to paddr by hardware, "
```

JanMatCodasip (dismissed):
LGTM, thank you.

src/target/riscv/startup.tcl
aap-sc: Answered from the wrong account, sorry for spam.

> The new type of tests that @en-sc has proposed [here](https://review.openocd.org/c/openocd/+/8507) would be very good for testing changes like these. Perhaps you are using such tests already internally?

1. These types of tests would be useful, though integrating it to that infrastructure will be difficult, since these specific tests require a specific response from the target (dummy adapter just  guarantees that some results are returned, not a valid one). 
2. This commit is tested internally against using internal testsuite that I'm still in the process of opensourcing :(. Those test just generate MMU pages on the fly, flip the switch for translation mode and try to do memory reads/writes  comparing results with the expected one.

src/target/riscv/riscv.c

JanMatCodasip (approved):

PR #1154, opened 2024-10-24
target/riscv: avoid unnecessary `dcsr.ebreak*` update on reset

en-sc:
There is no need to change if `dcsr.ebreak*` fields after a reset if a user requested a configuration that will result `dcsr.ebreak*` field values equal to reset values.

Change-Id: I2844d30aef8f735c7b37394ee422e9b3f04a2e3b

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2837,6 +2837,12 @@ static int assert_reset(struct target *target)
 	return riscv013_invalidate_cached_progbuf(target);
 }
 
+static bool dcsr_ebreak_config_equals_reset_value(const struct target *target)
+{
+	RISCV_INFO(r);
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -2837,6 +2837,12 @@ static int assert_reset(struct target *target)
 	return riscv013_invalidate_cached_progbuf(target);
 }
 
+static bool dcsr_ebreak_config_equals_reset_value(const struct target *target)
+{
+	RISCV_INFO(r);
+	return !(r->riscv_ebreakm || r->riscv_ebreaks || r->riscv_ebreaks);
```
zqb-all: Duplicate `riscv_ebreaks`, should be `riscv_ebreaku`.

en-sc: Huge thanks for the catch!

JanMatCodasip (approved):
LGTM, thank you.

PR #1157, opened 2024-10-29
target/riscv: support disable auto fence

zqb-all:
Support disable automatic fence, it's useful for debug some cache related issue.

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -11631,6 +11631,19 @@ riscv exec_progbuf 0x0330000f 0x0000100f
 riscv exec_progbuf 0x94a20405
 @end example
 
+@deffn {Command} {riscv autofence} [on|off]
+When on (default), OpenOCD will automatically execute RISC-V fence instructions
+(@var{fence.i} and @var{fence rw, rw}) in these cases:
+@itemize @bullet
+@item before step or resume,
+@item before memory read via the Program Buffer,
+@item after memory write via the Program Buffer.
+@end itemize
+When off, users need to take care of memory coherency themselves, for example
+using the @var{riscv exec_progbuf} command to execute fences or CMO instructions
+(RISC-V Cache Management Operations).
+@end deffn
+
 @section ARC Architecture
 @cindex ARC
 
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2913,10 +2913,14 @@ static int deassert_reset(struct target *target)
 	return dm_write(target, DM_DMCONTROL, control);
 }
 
-static int execute_fence(struct target *target)
+static int execute_autofence(struct target *target)
 {
 	if (dm013_select_target(target) != ERROR_OK)
 		return ERROR_FAIL;
+
+	RISCV_INFO(r);
+	if (!r->autofence)
+		return ERROR_OK;
 
 	/* FIXME: For non-coherent systems we need to flush the caches right
 	 * here, but there's no ISA-defined way of doing that. */
@@ -2939,8 +2943,9 @@ static int execute_fence(struct target *target)
 				LOG_TARGET_ERROR(target, "Unexpected error during fence execution");
 				return ERROR_FAIL;
 			}
-			LOG_TARGET_DEBUG(target, "Unable to execute fence");
+			LOG_TARGET_DEBUG(target, "Unable to execute fence.i and fence rw, rw");
 		}
+		LOG_TARGET_DEBUG(target, "Successfully executed fence.i and fence rw, rw");
 		return ERROR_OK;
 	}
 
@@ -2954,6 +2959,7 @@ static int execute_fence(struct target *target)
 			}
 			LOG_TARGET_DEBUG(target, "Unable to execute fence.i");
 		}
+		LOG_TARGET_DEBUG(target, "Successfully executed fence.i");
 
 		riscv_program_init(&program, target);
 		riscv_program_fence_rw_rw(&program);
@@ -2964,6 +2970,7 @@ static int execute_fence(struct target *target)
 			}
 			LOG_TARGET_DEBUG(target, "Unable to execute fence rw, rw");
 		}
+		LOG_TARGET_DEBUG(target, "Successfully executed fence rw, rw");
 		return ERROR_OK;
 	}
 
@@ -4273,7 +4280,7 @@ read_memory_progbuf(struct target *target, target_addr_t address,
 
 	memset(buffer, 0, count*size);
 
-	if (execute_fence(target) != ERROR_OK)
+	if (execute_autofence(target) != ERROR_OK)
 		return MEM_ACCESS_SKIPPED_FENCE_EXEC_FAILED;
 
 	uint64_t mstatus = 0;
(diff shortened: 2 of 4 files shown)


aap-sc:
@zqb-all I'm not sure we need the functionality of `riscv fence` command, since we already have capabilities to execute instructions from program buffer directly. Have you tried `riscv exec_progbuf` command?

aap-sc:
as for `auto fence` - it looks reasonable

zqb-all:
> @zqb-all I'm not sure we need the functionality of `riscv fence` command, since we already have capabilities to execute instructions from program buffer directly. Have you tried `riscv exec_progbuf` command?

Thanks, I didn't notice this command before. With this command, I agree that a separate fence command is not needed

src/target/riscv/riscv-013.c
```diff
@@ -2918,6 +2918,10 @@ static int execute_fence(struct target *target)
 	if (dm013_select_target(target) != ERROR_OK)
```
JanMatCodasip, line 2918: Could the function be renamed to `execute_autofence()` for improved clarity?

There is also function `maybe_execute_fence_i() ` which just calls `execute_autofence()`. Please, could you remove that function. Instead, call `execute_autofence()` directly.

src/target/riscv/riscv.c
```diff
@@ -4521,6 +4521,23 @@ COMMAND_HANDLER(riscv_set_maskisr)
...
+	if (CMD_ARGC != 1) {
+		command_print(CMD, "autofence enabled: %s", r->autofence ? "on" : "off");
+		return ERROR_OK;
+	} else if (CMD_ARGC == 1) {
+		COMMAND_PARSE_ON_OFF(CMD_ARGV[0], r->autofence);
+		return ERROR_OK;
+	}
+
+	LOG_ERROR("Command takes 0 or 1 parameters");
+	return ERROR_COMMAND_SYNTAX_ERROR;
```
JanMatCodasip: 1) There seems to be mistake in this function - the last two lines of the function are currently dead code.

2) Per current OpenOCD conventions, LOG_ERROR should not be printed when ERROR_COMMAND_SYNTAX_ERROR is returned. OpenOCD will print the command syntax to the user on its own.

3) It is more natural to say simply `autofence: off` instead of `autofence enabled: off`. 

```suggestion
	if (CMD_ARGC == 0) {
		command_print(CMD, "autofence: %s", r->autofence ? "on" : "off");
		return ERROR_OK;
	} else if (CMD_ARGC == 1) {
		COMMAND_PARSE_ON_OFF(CMD_ARGV[0], r->autofence);
		return ERROR_OK;
	}
	return ERROR_COMMAND_SYNTAX_ERROR;
```

doc/openocd.texi
```diff
@@ -11631,6 +11631,12 @@ riscv exec_progbuf 0x0330000f 0x0000100f
 riscv exec_progbuf 0x94a20405
 @end example
 
+@deffn {Command} {riscv autofence} [on|off]
+When on (default), some situations, such as stepi, will automatically execute fence.
+When off, user needs to take care of the cache themselves, for example, using the
+riscv exec_progbuf to execute fence or cmo command.
+@end deffn
+
```
JanMatCodasip: Please, could you write the documentation in more accurate manner:

- What instructions are actually executed.
- What exactly is "some situations".
- There is nothing in OpenOCD called `stepi`.
- `cmo` should be spelled as `CMO` and explained to the reader what it stands for
- `the cache` --> `memory coherency`

For example:

```suggestion
@deffn {Command} {riscv autofence} [on|off]
When on (default), OpenOCD will automatically execute RISC-V fence instructions
(@var{fence.i} and @var{fence rw, rw}) in these cases:

@itemize @bullet

@item before step or resume,
@item before memory read via the Program Buffer,
@item after memory write via the Program Buffer.

@end itemize

When off, users needs to take care of memory coherency themselves, for example
using the @var{riscv exec_progbuf} command to execute fences or CMO instructions
(RISC-V Cache Management Operations).
@end deffn

```

src/target/riscv/riscv-013.c
```diff
@@ -2941,6 +2945,7 @@ static int execute_fence(struct target *target)
 			}
 			LOG_TARGET_DEBUG(target, "Unable to execute fence");
 		}
+		LOG_TARGET_DEBUG(target, "Successfully executed fence");
```
```suggestion
		LOG_TARGET_DEBUG(target, "Successfully executed fence.i and fence rw, rw");
```

src/target/riscv/riscv.c
```diff
@@ -5428,6 +5445,15 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
 				"off - no address translation."
 	},
+	{
+		.name = "autofence",
+		.handler = riscv_set_autofence,
+		.mode = COMMAND_ANY,
+		.usage = "[on|off]",
+		.help = "When on (default), some situations, such as stepi, will automatically execute fence. "
+			"When off, user needs to take care of the cache themselves, for example, using the "
+			"riscv exec_progbuf to execute fence or cmo command."
```
JanMatCodasip: Similar updates as in the documentation should be made to the help message.

For example:

```suggestion
		.help = "When on (default), OpenOCD will automatically execute fence instructions in some situations. "
			"When off, users needs to take care of memory coherency themselves, for example by using "
			"`riscv exec_progbuf` to execute fence or CMO instructions."
```

JanMatCodasip (changes requested):
@zqb-all Overall this is a nice improvement. However, some parts of the code and documentation should be improved. Please take a look at my comments. 

Thank you.

src/target/riscv/riscv-013.c
zqb-all: Thanks, addressed

src/target/riscv/riscv.c
zqb-all: My mistake, addressed

doc/openocd.texi
zqb-all: Thanks,  this doc is much better than before

src/target/riscv/riscv.c
zqb-all: Thanks, addressed

src/target/riscv/riscv-013.c
zqb-all: Addressed, also modified another log in the previous two lines

src/target/riscv/riscv.c
zqb-all: > LOG_ERROR should not be printed when ERROR_COMMAND_SYNTAX_ERROR is returned.

Another [PR](https://github.com/riscv-collab/riscv-openocd/pull/1160 ) is created to remove unnecessary LOG_ERROR

JanMatCodasip (commented):
@zqb-all Thank you for addressing the comments so far. The code looks all right to me now, however the CI is failing. Could you please look into why it is so?

zqb-all:
> @zqb-all Thank you for addressing the comments so far. The code looks all right to me now, however the CI is failing. Could you please look into why it is so?

A bug was introduced when removing `maybe_execute_fence_i`, which has now been fixed

JanMatCodasip (approved):
LGTM, thank you.

aap-sc (approved):
LGTM too. @en-sc ?

en-sc (approved):

PR #1160, opened 2024-11-05
[NFC] target/riscv: remove LOG_ERROR when COMMAND check ARGC fail

zqb-all:
Per current OpenOCD conventions, LOG_ERROR should not be printed when ERROR_COMMAND_SYNTAX_ERROR is returned. OpenOCD will print the command syntax to the user on its own.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3927,10 +3927,9 @@ int riscv_openocd_step(struct target *target, int current,
 /* Command Handlers */
 COMMAND_HANDLER(riscv_set_command_timeout_sec)
 {
-	if (CMD_ARGC != 1) {
-		LOG_ERROR("Command takes exactly 1 parameter.");
+	if (CMD_ARGC != 1)
 		return ERROR_COMMAND_SYNTAX_ERROR;
-	}
+
 	int timeout = atoi(CMD_ARGV[0]);
 	if (timeout <= 0) {
 		LOG_ERROR("%s is not a valid integer argument for command.", CMD_ARGV[0]);
@@ -3945,10 +3944,9 @@ COMMAND_HANDLER(riscv_set_command_timeout_sec)
 COMMAND_HANDLER(riscv_set_reset_timeout_sec)
 {
 	LOG_WARNING("The command 'riscv set_reset_timeout_sec' is deprecated! Please, use 'riscv set_command_timeout_sec'.");
-	if (CMD_ARGC != 1) {
-		LOG_ERROR("Command takes exactly 1 parameter.");
+	if (CMD_ARGC != 1)
 		return ERROR_COMMAND_SYNTAX_ERROR;
-	}
+
 	int timeout = atoi(CMD_ARGV[0]);
 	if (timeout <= 0) {
 		LOG_ERROR("%s is not a valid integer argument for command.", CMD_ARGV[0]);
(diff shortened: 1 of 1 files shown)


JanMatCodasip (dismissed):
LGTM, thank you.

src/target/riscv/riscv.c
```diff
@@ -3927,10 +3927,9 @@ int riscv_openocd_step(struct target *target, int current,
 /* Command Handlers */
 COMMAND_HANDLER(riscv_set_command_timeout_sec)
 {
-	if (CMD_ARGC != 1) {
-		LOG_ERROR("Command takes exactly 1 parameter.");
+	if (CMD_ARGC != 1)
```
en-sc, line 3930: Please, fix the usage for this command (`sec` is not optional):
https://github.com/riscv-collab/riscv-openocd/blob/7f054ce0cf1a1dff5b69ce4b5965331d04a8673b/src/target/riscv/riscv.c#L5172

src/target/riscv/riscv.c
```diff
@@ -3945,10 +3944,9 @@ COMMAND_HANDLER(riscv_set_command_timeout_sec)
 COMMAND_HANDLER(riscv_set_reset_timeout_sec)
 {
 	LOG_WARNING("The command 'riscv set_reset_timeout_sec' is deprecated! Please, use 'riscv set_command_timeout_sec'.");
-	if (CMD_ARGC != 1) {
-		LOG_ERROR("Command takes exactly 1 parameter.");
+	if (CMD_ARGC != 1)
```
en-sc, line 3947: Please, fix the usage for this command (`sec` is not optional):
https://github.com/riscv-collab/riscv-openocd/blob/7f054ce0cf1a1dff5b69ce4b5965331d04a8673b/src/target/riscv/riscv.c#L5179

src/target/riscv/riscv.c
```diff
@@ -4409,10 +4397,8 @@ COMMAND_HANDLER(riscv_reset_delays)
 
 COMMAND_HANDLER(riscv_set_ir)
 {
-	if (CMD_ARGC != 2) {
-		LOG_ERROR("Command takes exactly 2 arguments");
+	if (CMD_ARGC != 2)
```
en-sc, line 4400: Please, fix the usage for this command (the first argument is not optional):
https://github.com/riscv-collab/riscv-openocd/blob/7f054ce0cf1a1dff5b69ce4b5965331d04a8673b/src/target/riscv/riscv.c#L5288

src/target/riscv/riscv.c
```diff
@@ -4483,12 +4467,10 @@ COMMAND_HANDLER(riscv_set_bscan_tunnel_ir)
 {
 	int ir_id = 0;
 
-	if (CMD_ARGC > 1) {
-		LOG_ERROR("Command takes at most one arguments");
+	if (CMD_ARGC > 1)
 		return ERROR_COMMAND_SYNTAX_ERROR;
-	} else if (CMD_ARGC == 1) {
+	else if (CMD_ARGC == 1)
```
en-sc: Please, fix the usage (`value` is optional):
https://github.com/riscv-collab/riscv-openocd/blob/7f054ce0cf1a1dff5b69ce4b5965331d04a8673b/src/target/riscv/riscv.c#L5302

Also consider using a plain `if` instead of an `else if` after the `return`.

en-sc (changes requested):

src/target/riscv/riscv.c
zqb-all: Thanks, addressed

src/target/riscv/riscv.c
zqb-all: Thanks, addressed

src/target/riscv/riscv.c
zqb-all: Thanks, addressed

src/target/riscv/riscv.c
zqb-all: Thanks, addressed

en-sc (approved):
LGTM

JanMatCodasip (approved):

PR #1161, opened 2024-11-08
target/riscv: avoid updating `target` if `ackhavereset` fails

en-sc:
`target`'s `state` and `debug_reason` should not be updated in `deassert_reset` if sending reset acknowledgment fails.

Change-Id: I86136fe829e7a7c6b69f718f0cf32322e40341b0


JanMatCodasip (approved):
LGTM, thank you.

PR #1162, opened 2024-11-09
target/riscv: allow hexadecimal values to expose_csr-like commands

aap-sc:


src/target/riscv/riscv.c
```diff
@@ -4010,74 +4010,85 @@ COMMAND_HANDLER(riscv_set_mem_access)
...
-	char *arg = strtok(args, ",");
-	while (arg) {
+static bool parse_csr_address(const char* reg_address_str, unsigned int *reg_addr) {
+	*reg_addr = -1;
+	// skip initial spaces
+	while(isspace(reg_address_str[0])) {
+		++reg_address_str;
+	}
+	// try to detect if string starts with 0x
+	bool is_hex_address = strncmp(reg_address_str, "0x", 2) == 0 ||
```
TommyMurphyTM1234, line 4019: Why not simply a case sensitive comparison: `strnicmp(reg_address_str, "0x", 2)`?

src/target/riscv/riscv.c
```diff
@@ -4010,74 +4010,85 @@ COMMAND_HANDLER(riscv_set_mem_access)
...
+		strncmp(reg_address_str, "0X", 2) == 0;
+
+	unsigned chars_scanned;
```
TommyMurphyTM1234: Could use `unsigned char` instead of `unsigned`  here but maybe coding style doesn't like that?

TommyMurphyTM1234 (commented):
Small comments

src/target/riscv/riscv.c
aap-sc: I could. I just forgot that this function existed. Will address

src/target/riscv/riscv.c
aap-sc: this should be an `unsigned int` , actually. Writing `unsigned` is just an old habit of mine.

src/target/riscv/riscv.c
aap-sc: actually no! This function does not exist in C language. This is Microsoft-specific extension.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
TommyMurphyTM1234: > actually no! This function does not exist in C language. This is Microsoft-specific extension.

What about `strncasecmp()`?

- https://linux.die.net/man/3/strncasecmp

aap-sc: strcasecmp is not in the C or C++ standard. It's defined by POSIX.1-2001 and 4.4BSD.   I would prefer not to use it.

en-sc (dismissed):
LGTM (reviewed internally).

src/target/riscv/riscv.c
aap-sc: @TommyMurphyTM1234 , given that there is a reaction from you under this thread, can I assume that this one is addressed from your point of view? 

(Typically I would expect reviewer to resolve his threads, so I want to double-check that our expectations are aligned).

JanMatCodasip:
I will be able to review this change by the end of the week.

src/target/riscv/riscv.c
```diff
@@ -4010,74 +4010,85 @@ COMMAND_HANDLER(riscv_set_mem_access)
...
-static int parse_ranges(struct list_head *ranges, const char *tcl_arg, const char *reg_type, unsigned int max_val)
-{
-	char *args = strdup(tcl_arg);
-	if (!args)
-		return ERROR_FAIL;
 
-	/* For backward compatibility, allow multiple parameters within one TCL argument, separated by ',' */
-	char *arg = strtok(args, ",");
-	while (arg) {
+static bool parse_csr_address(const char *reg_address_str, unsigned int *reg_addr)
```
JanMatCodasip, line 4012: Could this function be replaced by existing `parse_uint()` from command.c?

src/target/riscv/riscv.c
```diff
@@ -4010,74 +4010,85 @@ COMMAND_HANDLER(riscv_set_mem_access)
...
+	}
+	return scanned_chars == strlen(reg_address_str);
+}
+
+static int parse_reg_ranges_impl(struct list_head *ranges, char *args,
+		const char *reg_type, unsigned int max_val, char ** const name_buffer)
+{
+	/* For backward compatibility, allow multiple parameters within one TCL
+	 * argument, separated by ',' */
+	for (char *arg = strtok(args, ","); arg; arg = strtok(NULL, ",")) {
```
JanMatCodasip, line 4047: (Just a comment, no change requested)

In future, it would be my preference to simplify the syntax of `expose_csrs` and `expose_custom`. I'd deprecate (and eventually remove) the syntax with comma-separated list of registers.

JanMatCodasip (dismissed):
LGTM, thanks.

I am sending one optional suggestion and one general comment.

src/target/riscv/riscv.c
aap-sc: If i the description of those commands right - they still accept octal numbers. Given the nature of what we are dealing with here - I would forbid usage of octals to reduce risk of errors...

Though, now that I've said it - I think it would not hurt to add an actual check that the number does not start with zero.

JanMatCodasip: Good point about the octals. Better to leave these unsupported to avoid user mistakes.

> I think it would not hurt to add an actual check that the number does not start with zero.

Sounds good.

src/target/riscv/riscv.c
TommyMurphyTM1234: > strcasecmp is not in the C or C++ standard. It's defined by POSIX.1-2001 and 4.4BSD. I would prefer not to use it.

OK - that makes sense - apologies for the distraction!

zqb-all:
LGTM, I wanted this feature too but haven't really put in the time to implement it yet, thank you for implementing it

src/target/riscv/riscv.c
aap-sc: addressed

aap-sc:
Note to reviewers:

the following checks were added in most recent update:

```
...
		/* If we are here and register address string starts with zero, this is
		 * an indication that most likely user has an incorrect input because:
		 * - decimal numbers typically do not start with "0"
		 * - octals are not supported by our interface
		 * - hexadecimal numbers should have "0x" prefix
		 * Thus such input is rejected. */
		if (strncmp(reg_address_str, "0", 1) == 0 && strlen(reg_address_str) > 1)
			return false;
...
```

Would you kindly take a look?

TommyMurphyTM1234:
> Note to reviewers:
> 
> the following checks were added in most recent update:
> 
> ```
> ...
> 		/* If we are here and register address string starts with zero, this is
> 		 * an indication that most likely user has an incorrect input because:

This comment seems a bit misleading because in the code there is a prior explicit check for `0x` prefix, so  a register number starting with `0` IS valid, if the next character is `x`/`X`.

On a more general point, why not just allow decimal numbers that start with `0`(s)? Might not somebody legitimately use a Tcl script that generates register numbers of a similar form as generated by C's `printf("04%u", x);` padded with leading `0`s? Why go to the bother of explicitly excluding such number representation? Do many people actually try to use octal representation in such circumstances?

src/target/riscv/riscv.c
```diff
@@ -4008,74 +4008,93 @@ COMMAND_HANDLER(riscv_set_mem_access)
...
 				LOG_ERROR("Failed to parse single register number from '%s'.", arg);
-				free(args);
 				return ERROR_COMMAND_SYNTAX_ERROR;
 			}
-			if (sscanf(dash, "%u%n", &high, &pos) != 1 || pos != strlen(dash)) {
-				LOG_ERROR("Failed to parse single register number from '%s'.", dash);
-				free(args);
+			const char *high_num_in = dash + 1;
+			if (!parse_csr_address(high_num_in, &high)) {
+				LOG_ERROR("Failed to parse single register number from '%s'.", high_num_in);
```
JanMatCodasip: To help users, I'd recommend to rephrase the error message and say what number formats are expected.

There are three occurrences of this message.

```suggestion
				LOG_ERROR("Failed to parse '%s' - not a valid decimal or hexadecimal number.", high_num_in);
```

aap-sc:
> This comment seems a bit misleading because in the code there is a prior explicit check for `0x` prefix, so a register number starting with `0` IS valid, if the next character is `x`/`X`.

But there is  a precondition: "*If we are here*". Meaning we've already checked for hexadecimal numbers a few lines above.

> On a more general point, why not just allow decimal numbers that start with 0(s)?

Because it suspiciously looks like someone copy-pasted a hexadecimal number from some documentation and forgot to add a leading "0x". 

>  Might not somebody legitimately use a Tcl script that generates register numbers of a similar form as generated by C's printf("04%u", x); padded with leading 0s?

Same point as above.

> Do many people actually try to use octal representation in such circumstances?

People may copy-paste a hexadecimal number.

src/target/riscv/riscv.c
```diff
@@ -4008,74 +4008,93 @@ COMMAND_HANDLER(riscv_set_mem_access)
...
+			return false;
+	} else {
+		/* If we are here and register address string starts with zero, this is
+		 * an indication that most likely user has an incorrect input because:
+		 * - decimal numbers typically do not start with "0"
+		 * - octals are not supported by our interface
+		 * - hexadecimal numbers should have "0x" prefix
+		 * Thus such input is rejected. */
+		if (strncmp(reg_address_str, "0", 1) == 0 && strlen(reg_address_str) > 1)
+			return false;
```
JanMatCodasip: Optional cosmetic suggestion:

```suggestion
		if (reg_address_str[0] == '0' && strlen(reg_address_str) > 1)
			return false;
```

src/target/riscv/riscv.c
```diff
@@ -4008,74 +4008,93 @@ COMMAND_HANDLER(riscv_set_mem_access)
...
-
-			if (sscanf(equals, "%[_a-zA-Z0-9]%n", name + strlen(reg_type) + 1, &pos) != 1 || pos != strlen(equals)) {
-				LOG_ERROR("Failed to parse register name from '%s'.", equals);
-				free(args);
-				free(name);
+			unsigned int scanned_chars;
+			char *scan_dst = name + strlen(reg_type) + 1;
+			if (sscanf(reg_name_in, "%[_a-zA-Z0-9]%n", scan_dst, &scanned_chars) != 1 ||
+				scanned_chars != strlen(reg_name_in)) {
+				LOG_ERROR("Failed to parse register name from '%s'.", reg_name_in);
```

```suggestion
				LOG_ERROR("Invalid characters in register name '%s'.", reg_name_in);
```

src/target/riscv/riscv.c
```diff
@@ -4107,32 +4125,42 @@ static int parse_ranges(struct list_head *ranges, const char *tcl_arg, const cha
...
-	free(args);
 	return ERROR_OK;
 }
 
+static int parse_reg_ranges(struct list_head *ranges, const char *tcl_arg,
+		const char *reg_type, unsigned int max_val)
+{
+	char *args = strdup(tcl_arg);
+	if (!args) {
+		LOG_ERROR("Failed to allocate memory for reg range string copy");
```
JanMatCodasip: Cosmetic detail: 

I'd simplify the error message so that the readers are not bothered with implementation details and instead get clear idea of what is going on:

```suggestion
		LOG_ERROR("Out of memory");
```

Or if you like:

```suggestion
		LOG_ERROR("Out of memory in " __func__);
```

src/target/riscv/riscv.c
```diff
@@ -4087,9 +4106,8 @@ static int parse_ranges(struct list_head *ranges, const char *tcl_arg, const cha
 		high = MAX(high, low);
 
 		if (high > max_val) {
-			LOG_ERROR("Cannot expose %s register number %u, maximum allowed value is %u.", reg_type, high, max_val);
-			free(name);
-			free(args);
+			LOG_ERROR("Cannot expose %s register number %u, maximum allowed value is %u.",
+				reg_type, high, max_val);
```
JanMatCodasip: Now that hexadecimal literals are accepted, you may consider using hexadecimal literals also in the messages:

```suggestion
			LOG_ERROR("Cannot expose %s register number 0x%x, maximum allowed value is 0x%x.",
				reg_type, high, max_val);
```

JanMatCodasip (commented):
I have submitted few last minor comments and I do not expect to have any more. Other than that the change looks fine.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

JanMatCodasip (approved):
LGTM, thank you.

en-sc (approved):

EccoTheDolphin:
> > This comment seems a bit misleading because in the code there is a prior explicit check for `0x` prefix, so a register number starting with `0` IS valid, if the next character is `x`/`X`.
> 
> But there is a precondition: "_If we are here_". Meaning we've already checked for hexadecimal numbers a few lines above.
> 
> > On a more general point, why not just allow decimal numbers that start with 0(s)?
> 
> Because it suspiciously looks like someone copy-pasted a hexadecimal number from some documentation and forgot to add a leading "0x".
> 
> > Might not somebody legitimately use a Tcl script that generates register numbers of a similar form as generated by C's printf("04%u", x); padded with leading 0s?
> 
> Same point as above.
> 
> > Do many people actually try to use octal representation in such circumstances?
> 
> People may copy-paste a hexadecimal number.

@TommyMurphyTM1234 could you please clarify is my reasoning is fine with you? Or do you still have objections or concerns?

aap-sc:
> > This comment seems a bit misleading because in the code there is a prior explicit check for `0x` prefix, so a register number starting with `0` IS valid, if the next character is `x`/`X`.
> 
> But there is a precondition: "_If we are here_". Meaning we've already checked for hexadecimal numbers a few lines above.
> 
> > On a more general point, why not just allow decimal numbers that start with 0(s)?
> 
> Because it suspiciously looks like someone copy-pasted a hexadecimal number from some documentation and forgot to add a leading "0x".
> 
> > Might not somebody legitimately use a Tcl script that generates register numbers of a similar form as generated by C's printf("04%u", x); padded with leading 0s?
> 
> Same point as above.
> 
> > Do many people actually try to use octal representation in such circumstances?
> 
> People may copy-paste a hexadecimal number.

@TommyMurphyTM1234 could you please clarify is my reasoning is fine with you? Or do you still have objections or concerns?

aap-sc:
> @TommyMurphyTM1234 could you please clarify is my reasoning is fine with you? Or do you still have objections or concerns?

@TommyMurphyTM1234 ping)

aap-sc:
Well, if no additional objections follow, I'm going to merge this by the end of week (around Friday).

TommyMurphyTM1234:
> > @TommyMurphyTM1234 could you please clarify is my reasoning is fine with you? Or do you still have objections or concerns?
> 
> @TommyMurphyTM1234 ping)

Apologies for the delay - I have no strong objections - definitely none that should hold this up further.

src/target/riscv/riscv.c
```diff
@@ -4008,74 +4008,96 @@ COMMAND_HANDLER(riscv_set_mem_access)
...
 			if (!name) {
-				LOG_ERROR("Failed to allocate register name.");
-				free(args);
+				LOG_ERROR("Out of memory");
 				return ERROR_FAIL;
 			}
 
-			/* Register prefix: "csr_" or "custom_" */
-			strcpy(name, reg_type);
-			name[strlen(reg_type)] = '_';
```
aap-sc, line 4073: @JanMatCodasip , @en-sc 

This MR introduced a regression since these lines were deleted. It seems that I've made a mistake while my changes were rebased (our internal change reviewed by @en-sc had this code). The regression was detected by our internal testsuite.

I'll post a patch shortly and will try to extend riscv-tests.

aap-sc: MR with the fix: https://github.com/riscv-collab/riscv-openocd/pull/1176

PR #1165, opened 2024-11-13
target/riscv: detailed error messages for cases when resume operation fails

aap-sc:
This change aims to provide more context in case if resume operation fails. Before the change messages were quite confusing.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -5385,7 +5385,9 @@ static int riscv013_step_or_resume_current_hart(struct target *target,
 		LOG_TARGET_ERROR(target, "Hart is not halted!");
 		return ERROR_FAIL;
 	}
-	LOG_TARGET_DEBUG(target, "resuming (for step?=%d)", step);
+
+	LOG_TARGET_DEBUG(target, "resuming (operation=%s)",
+		step ? "single-step" : "resume");
 
 	if (riscv_reg_flush_all(target) != ERROR_OK)
 		return ERROR_FAIL;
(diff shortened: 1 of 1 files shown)


aap-sc:
@JanMatCodasip could you take a look, please?

src/target/riscv/riscv-013.c
```diff
@@ -5418,16 +5418,22 @@ static int riscv013_step_or_resume_current_hart(struct target *target,
 		return ERROR_OK;
 	}
 
+	LOG_TARGET_ERROR(target, "unable to %s. dmstatus=0x%08x",
```
JanMatCodasip: Detail, if you like:

`Failed` is slightly more descriptive than `unable`. It makes it more clear that we tried and it did not go well.

```suggestion
	LOG_TARGET_ERROR(target, "Failed to %s. dmstatus=0x%08x",
```

src/target/riscv/riscv-013.c
```diff
@@ -5418,16 +5418,22 @@ static int riscv013_step_or_resume_current_hart(struct target *target,
 		return ERROR_OK;
 	}
```
JanMatCodasip, line 5422: Optionally, If you like, the debug print on line 5388 can be improved for clarity, too:

```
LOG_TARGET_DEBUG(target, "resuming (operation=%s)", step ? "single-step" : "resume");
```

src/target/riscv/riscv-013.c
```diff
@@ -5418,16 +5418,22 @@ static int riscv013_step_or_resume_current_hart(struct target *target,
...
 
+	LOG_TARGET_ERROR(target, "unable to %s. dmstatus=0x%08x",
+		step ? "single-step" : "resume", dmstatus);
+
 	dm_write(target, DM_DMCONTROL, dmcontrol);
+	LOG_TARGET_ERROR(target, "  cancelling RESUMEREQ");
```
JanMatCodasip: (Just a comment - no code change requested)

After reading the code of this function, this part looks questionable. I am referring to the original code, not to your changes:

`dmcontrol.resumereq` bit is specified as W1 and that means only writing 1 has an effect. Writing 0 to this bit should not have any effect (pointless operation). This is true for both 0.13 and 1.0 versions of the debug spec.

This should not block this merge request, though. I just wanted to mention it. We may file an issue for it, if you like.

```suggestion
	LOG_TARGET_ERROR(target, "  cancelling the resume request (dmcontrol.resumereq <- 0)");
```

src/target/riscv/riscv-013.c
```diff
@@ -5418,16 +5418,22 @@ static int riscv013_step_or_resume_current_hart(struct target *target,
...
 	if (dmstatus_read(target, &dmstatus, true) != ERROR_OK)
 		return ERROR_FAIL;
-	LOG_TARGET_ERROR(target, "  dmstatus=0x%08x", dmstatus);
+
+	LOG_TARGET_ERROR(target, "  dmstatus after cancellation=0x%08x", dmstatus);
 
 	if (step) {
-		LOG_TARGET_ERROR(target, "  was stepping, halting");
+		LOG_TARGET_ERROR(target,
+			"  trying to restore from failed single-step, by requesting throw-away halt");
```
JanMatCodasip: I have proposed minor reword for clarity.

What have you please meant by "throw-away" halt? For me the sentence makes sense without it.

```suggestion
		LOG_TARGET_ERROR(target,
			"  trying to recover from a failed single-step by requesting halt");
```

JanMatCodasip (commented):
The added error messages look all right to me overall.

I have made few smaller comments.

src/target/riscv/riscv-013.c
aap-sc: > What have you please meant by "throw-away" halt? 

By "throw-away" halt I mean a halt request without any error processing. Here I wanted to emphasize that we have no guarantees that the halt actually happens.

src/target/riscv/riscv-013.c
aap-sc: Yeah... Thanks for the catch. 
This whole section where we try to restore the state is a mess. 

I'll double check and will probably remove this in a subsequent commit.

For now I just only want to cleanup error messages.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
aap-sc: addressed

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
aap-sc: anyway. I've adopted you suggestion and just added an additional printout that checks for halt result. Could you please take another look?

JanMatCodasip (approved):
LGTM, thank you.

en-sc (approved):
LGTM (reviewed internally).

PR #1167, opened 2024-11-14
target/riscv: pass memory access info in struct, move write_memory pointer

fkhaidari:
This changes will allow to unite read_memory/write_memory functions to one access function

(1) Replaced read/write functions arguments with one structure
(2) Unified read_memory/write_memory function pointers to be stored in same structure

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1972,9 +1972,16 @@ static int deassert_reset(struct target *target)
 		return wait_for_state(target, TARGET_RUNNING);
 }
 
-static int read_memory(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, uint8_t *buffer, uint32_t increment)
+static int read_memory(struct target *target, const riscv_mem_access_args_t args)
 {
+	assert(riscv_mem_access_is_read(args));
+
+	const target_addr_t address = args.address;
+	const uint32_t size = args.size;
+	const uint32_t count = args.count;
+	const uint32_t increment = args.increment;
+	uint8_t * const buffer = args.read_buffer;
+
 	if (increment != size) {
 		LOG_ERROR("read_memory with custom increment not implemented");
 		return ERROR_NOT_IMPLEMENTED;
@@ -2142,9 +2149,20 @@ static int setup_write_memory(struct target *target, uint32_t size)
 	return ERROR_OK;
 }
 
-static int write_memory(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, const uint8_t *buffer)
+static int write_memory(struct target *target, const riscv_mem_access_args_t args)
 {
+	assert(riscv_mem_access_is_write(args));
+
+	if (args.increment != args.size) {
+		LOG_TARGET_ERROR(target, "Write increment size has to be equal to element size");
+		return ERROR_NOT_IMPLEMENTED;
+	}
+
+	const target_addr_t address = args.address;
+	const uint32_t size = args.size;
+	const uint32_t count = args.count;
+	const uint8_t * const buffer = args.write_buffer;
+
 	riscv011_info_t *info = get_info(target);
 	jtag_add_ir_scan(target->tap, &select_dbus, TAP_IDLE);
 
@@ -2371,7 +2389,9 @@ static int init_target(struct command_context *cmd_ctx,
 {
 	LOG_DEBUG("init");
 	RISCV_INFO(generic_info);
+	/* TODO: replace read and write with single access function*/
 	generic_info->read_memory = read_memory;
+	generic_info->write_memory = write_memory;
 	generic_info->authdata_read = &riscv011_authdata_read;
 	generic_info->authdata_write = &riscv011_authdata_write;
 	generic_info->print_info = &riscv011_print_info;
@@ -2405,8 +2425,6 @@ struct target_type riscv011_target = {
 
 	.assert_reset = assert_reset,
 	.deassert_reset = deassert_reset,
-
-	.write_memory = write_memory,
 
 	.arch_state = arch_state,
 };
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -6,6 +6,7 @@
  */
 
 #include <assert.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <time.h>
 
@@ -64,10 +65,8 @@ static int register_read_direct(struct target *target, riscv_reg_t *value,
 		enum gdb_regno number);
 static int register_write_direct(struct target *target, enum gdb_regno number,
 		riscv_reg_t value);
-static int read_memory(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, uint8_t *buffer, uint32_t increment);
-static int write_memory(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, const uint8_t *buffer);
+static int read_memory(struct target *target, const riscv_mem_access_args_t args);
+static int write_memory(struct target *target, const riscv_mem_access_args_t args);
 static bool riscv013_get_impebreak(const struct target *target);
 static unsigned int riscv013_get_progbufsize(const struct target *target);
 
@@ -1201,7 +1200,14 @@ static int scratch_read64(struct target *target, scratch_mem_t *scratch,
 		case SPACE_DMI_RAM:
 			{
 				uint8_t buffer[8] = {0};
-				if (read_memory(target, scratch->debug_address, 4, 2, buffer, 4) != ERROR_OK)
+				const riscv_mem_access_args_t args = {
+					.address = scratch->debug_address,
+					.read_buffer = buffer,
+					.size = 4,
+					.count = 2,
+					.increment = 4,
+				};
+				if (read_memory(target, args) != ERROR_OK)
 					return ERROR_FAIL;
 				*value = buffer[0] |
 					(((uint64_t) buffer[1]) << 8) |
@@ -1242,7 +1248,14 @@ static int scratch_write64(struct target *target, scratch_mem_t *scratch,
 					value >> 48,
 					value >> 56
 				};
-				if (write_memory(target, scratch->debug_address, 4, 2, buffer) != ERROR_OK)
+				const riscv_mem_access_args_t args = {
+					.address = scratch->debug_address,
+					.write_buffer = buffer,
+					.size = 4,
+					.count = 2,
+					.increment = 4,
+				};
+				if (write_memory(target, args) != ERROR_OK)
 					return ERROR_FAIL;
 			}
 			break;
@@ -2763,7 +2776,9 @@ static int init_target(struct command_context *cmd_ctx,
 	generic_info->dmi_read = &dmi_read;
 	generic_info->dmi_write = &dmi_write;
 	generic_info->get_dmi_address = &riscv013_get_dmi_address;
+	/* TODO: replace read and write with single access function*/
 	generic_info->read_memory = read_memory;
+	generic_info->write_memory = write_memory;
 	generic_info->data_bits = &riscv013_data_bits;
 	generic_info->print_info = &riscv013_print_info;
 	generic_info->get_impebreak = &riscv013_get_impebreak;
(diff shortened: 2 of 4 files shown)


en-sc (approved):
LGTM (reviewed internally).

en-sc (changes requested):
Seems like the tests have failed. @fk-sc please take a look

Tako-San:
> Seems like the tests have failed. @fk-sc please take a look

Fixed

en-sc (dismissed):

fkhaidari:
> Seems like the tests have failed. @fk-sc please take a look

Fixed

fkhaidari:

src/target/riscv/riscv.h
```diff
@@ -129,6 +129,22 @@ struct reg_name_table {
 	char **reg_names;
 };
 
+typedef struct riscv_mem_access_info {
+	target_addr_t target_address;
+
+	union {
+		uint8_t *read;
+		const uint8_t *write;
+	} buffer;
```
JanMatCodasip: My preference would be to not use unions because it does not really prevent the code writers from assigning to `buffer.read` but reading back `buffer.write` (or vice versa).

Instead, I'd like to suggest:
- Having separate `uint8* read_buffer` and `const uint8_t* write_buffer`
- Using assertions at places where the data from `riscv_mem_access_info_t` is actually used

This would provide bit more robustness and safety than the current solution.

```suggestion
	uint8_t *read_buffer;
	const uint8_t *write_buffer;
```

```c
// For read functions:
assert(mem_args->is_read);
assert(mem_args->read_buffer);
assert(!mem_args->write_buffer);

// Analogously for write functions.
```

src/target/riscv/riscv.c
```diff
@@ -2867,8 +2867,16 @@ static int riscv_address_translate(struct target *target,
 
 		uint8_t buffer[8];
 		assert(info->pte_shift <= 3);
-		int retval = r->read_memory(target, pte_address,
-				4, (1 << info->pte_shift) / 4, buffer, 4);
+		riscv_mem_access_info_t read_info = {
```
JanMatCodasip: Calling this `info` is IMO not bad but I was thinking if I can come up with a more fitting name.

Maybe `args` (memory access arguments)?

I'll leave it up to you whether to use `args` or keep `info`.

```suggestion
		riscv_mem_access_args_t read_args = {
```

src/target/riscv/riscv.c
```diff
@@ -3104,29 +3112,42 @@ static int check_virt_memory_access(struct target *target, target_addr_t address
...
+		.is_virtual = false,
+	};
+
+	RISCV_INFO(r);
+	return r->write_memory(target, write_info);
 }
 
-static int riscv_rw_memory(struct target *target, target_addr_t address, uint32_t size,
-		uint32_t count, uint8_t *read_buffer, const uint8_t *write_buffer)
+static int riscv_rw_memory(struct target *target, riscv_mem_access_info_t access)
```
JanMatCodasip: Could the `riscv_mem_access_info_t` arguments be `const` (everywhere)?

```suggestion
static int riscv_rw_memory(struct target *target, const riscv_mem_access_info_t access)
```

src/target/riscv/riscv-011.c
```diff
@@ -2371,7 +2385,9 @@ static int init_target(struct command_context *cmd_ctx,
 {
 	LOG_DEBUG("init");
 	RISCV_INFO(generic_info);
+	/* TODO: replace read and write with single access function*/
 	generic_info->read_memory = read_memory;
+	generic_info->write_memory = write_memory;
```
JanMatCodasip, line 2394: (Just a note, no change requested)

1) As already mentioned in a past merge request, I'd like to eventually remove these function pointers and instead use direct calls to 0.11 or 0.13 versions of the functions. (But struggle to find the time for that.)

2) I do not have preference whether to have a single function for reading/writing or a combined one.

src/target/riscv/riscv-013.c
```diff
@@ -3114,20 +3132,19 @@ static int modify_privilege(struct target *target, uint64_t *mstatus, uint64_t *
 	return ERROR_OK;
 }
 
-static int read_memory_bus_v0(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, uint8_t *buffer, uint32_t increment)
+static int read_memory_bus_v0(struct target *target, riscv_mem_access_info_t access)
```
JanMatCodasip: If you name the structure with arguments `args`, you will not need to rename the `access` variable in the function.

```suggestion
static int read_memory_bus_v0(struct target *target, riscv_mem_access_args_t args)
```

Side note, no change requested:

I am afraid that we do not have a good way of testing the functions for system bus v0 (similarly to other features like the 0.11 version of the debug spec or the BSCAN bridge).

src/target/riscv/riscv.h
```diff
@@ -129,6 +129,22 @@ struct reg_name_table {
 	char **reg_names;
 };
 
+typedef struct riscv_mem_access_info {
+	target_addr_t target_address;
```
JanMatCodasip: Since there is no other address in this structure, could it be called just `address` for simplicity?

One might also argue that in case of read operations, it is actually a `source_address`. :)

```suggestion
	target_addr_t address;
```

src/target/riscv/riscv-013.c
```diff
@@ -3114,20 +3132,19 @@ static int modify_privilege(struct target *target, uint64_t *mstatus, uint64_t *
...
-			TARGET_PRIxADDR, size, count, address);
-	uint8_t *t_buffer = buffer;
-	riscv_addr_t cur_addr = address;
-	riscv_addr_t fin_addr = address + (count * size);
-	uint32_t access = 0;
+			TARGET_PRIxADDR, access.element_size, access.count, access.target_address);
+	uint8_t *t_buffer = access.buffer.read;
+	riscv_addr_t cur_addr = access.target_address;
+	riscv_addr_t fin_addr = access.target_address + (access.count * access.element_size);
+	uint32_t bus_access = 0;
```
JanMatCodasip: If you end up in position when you still need to rename this variable, I'd call it `sbcs` (to match the name of the DM register):

```suggestion
	uint32_t sbcs = 0;
```

src/target/riscv/riscv-013.c
```diff
@@ -3203,35 +3220,34 @@ static int read_memory_bus_v0(struct target *target, target_addr_t address,
...
 		LOG_TARGET_ERROR(target, "sba v1 reads only support increment of size or 0");
 		return ERROR_NOT_IMPLEMENTED;
 	}
 
-	assert(size <= 16);
-	assert(IS_PWR_OF_2(size));
+	assert(access.element_size <= 16);
+	assert(IS_PWR_OF_2(access.element_size));
 
 	dm013_info_t *dm = get_dm(target);
```
JanMatCodasip: The code of `read_memory_bus_v1` is already quite complicated (and bugs were discovered here in the past).

For that reason, I'd recommend to make every effort to keep this part readable. Please consider unpacking the structure to individual variables, as it is done in 0.11 versions of read_memory and write_memory. That way the code will become somewhat simpler to read (the reader would not be bothered by all those `access.` prefixes).

src/target/riscv/riscv.h
```diff
@@ -129,6 +129,22 @@ struct reg_name_table {
...
+
+	union {
+		uint8_t *read;
+		const uint8_t *write;
+	} buffer;
+
+	uint32_t element_size;
```
JanMatCodasip: Could this be called just `size` for simplicity?

The name `size` is well-established in memory read/write functions across the whole OpenOCD, so I expect no confusion here.

JanMatCodasip (commented):
Overall I like this idea very much. 

I have managed a partial review only and am sending some coding suggestions. 

I will come back to finish the review in the next 1-2 days.

src/target/riscv/riscv.c
fkhaidari: Thanks. Addressed

src/target/riscv/riscv.c

src/target/riscv/riscv.h
fkhaidari: Addressed

src/target/riscv/riscv-013.c

src/target/riscv/riscv.h
fkhaidari: Addressed

src/target/riscv/riscv-013.c
fkhaidari: Structure renamed

src/target/riscv/riscv-013.c

src/target/riscv/riscv-011.c
fkhaidari: Thanks! I got you. One of my next MRs replaces high-level read/write functions to a combined one. It will decrease amount of copy-paste. I think it also will be able to be called directly

src/target/riscv/riscv.h
en-sc: > My preference would be to not use unions because it does not really prevent the code writers from assigning to `buffer.read` but reading back `buffer.write` (or vice versa).
> 
> Instead, I'd like to suggest:
> 
> * Having separate `uint8* read_buffer` and `const uint8_t* write_buffer`
> * Using assertions at places where the data from `riscv_mem_access_info_t` is actually used

@JanMatCodasip, I'd like to argue with that.
AFAIU, the approach you suggest will not fix the issue you are talking about.
IMHO, an assertion at the start of the function does not prevent the user from reading the wrong field.
Sure, probably static analysis tools will be more likely to catch the error, but my opinion is: `union` creates less boilerplate (only need to write one field) and therefore an error is less likely.

Anyway, either solution seems fine by me.

Though, if the union is dropped I'd also suggest dropping `is_write` field and creating a function that takes a `riscv_mem_access_info_t` and returns whether it's a write or a read based on buffer pointers (with the necessary assertions).

src/target/riscv/riscv-011.c
```diff
@@ -1972,9 +1972,14 @@ static int deassert_reset(struct target *target)
...
 
-static int read_memory(struct target *target, target_addr_t address,
-		uint32_t size, uint32_t count, uint8_t *buffer, uint32_t increment)
+static int read_memory(struct target *target, const riscv_mem_access_args_t args)
 {
+	target_addr_t address = args.address;
+	uint32_t size = args.size;
+	uint32_t count = args.count;
+	uint8_t *buffer = args.buffer.read;
+	uint32_t increment = args.increment;
```
JanMatCodasip: Detail for readability: The variables can be made const, if you like.

```suggestion
	const target_addr_t address = args.address;
	const uint32_t size = args.size;
	const uint32_t count = args.count;
	uint8_t *buffer = args.buffer.read;
	const uint32_t increment = args.increment;
```

src/target/riscv/riscv-011.c
```diff
@@ -2142,9 +2147,18 @@ static int setup_write_memory(struct target *target, uint32_t size)
...
 {
+	if (args.increment != args.size) {
+		LOG_TARGET_ERROR(target, "Write increment size has to be equal to element size");
+		return ERROR_NOT_IMPLEMENTED;
+	}
+
+	target_addr_t address = args.address;
+	uint32_t size = args.size;
+	uint32_t count = args.count;
+	const uint8_t *buffer = args.buffer.write;
```
JanMatCodasip: (same as previous)

```suggestion
	const target_addr_t address = args.address;
	const uint32_t size = args.size;
	const uint32_t count = args.count;
	const uint8_t *buffer = args.buffer.write;
```

src/target/riscv/riscv.h
```diff
@@ -129,6 +129,22 @@ struct reg_name_table {
...
+	union {
+		uint8_t *read;
+		const uint8_t *write;
+	} buffer;
+
+	uint32_t size;
+	uint32_t increment;
+	uint32_t count;
+
+	bool is_virtual;
```
JanMatCodasip: If I read the code correctly, there is no function that would read the  `is_virtual` item.

Also it looks to me the `is_virtual` attribute is not relevant on all hierarchy levels of the memory r/w functions.

For that reason, I'd recommend to remove it for now:

```suggestion
```

If "is_virtual" becomes relevant in a later merge request, then it can be added. But perhaps not as a member of this structure, but more likely as an additional function argument, and only on those functions for which it really matters:

```c
int some_read_write_func(struct target *target, const riscv_mem_access_args_t args, bool is_virtual);
```

src/target/riscv/riscv.h
JanMatCodasip: An assertion would however ensure that exactly one of the values is non-NULL. And if the programmer makes a mistake and picks the wrong one, the use of NULL would become apparent very soon.

> if the union is dropped I'd also suggest dropping is_write field and creating a function that takes a riscv_mem_access_info_t and returns whether it's a write or a read based on buffer pointers

Sounds good :)

JanMatCodasip (commented):
I have managed to move forward with the review but haven't reached the end.

I will return to this PR tomorrow.

JanMatCodasip (commented):
Overall this change looks fine - thanks.

I have finished my review and do not expect to have any further comments.

src/target/riscv/riscv.h
fkhaidari: Union removed

src/target/riscv/riscv-011.c
fkhaidari: Addressed

src/target/riscv/riscv-011.c
fkhaidari: Addressed

src/target/riscv/riscv.h
fkhaidari: Addressed

JanMatCodasip (approved):
LGTM, thanks.

en-sc (approved):

PR #1168, opened 2024-11-14
target/riscv: new `ebreak` controls

en-sc:
* Deprecate `riscv set_ebreak*` commands.
* Introduce RISC-V-sepecific `configure` parameter `-ebreak` instead.
* Separate controls for VS and VU modes.

Change-Id: I0ff88318dcb52af6923eb9f20f9d0c056ad09cf0

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1074,9 +1074,18 @@ static int maybe_write_tselect(struct target *target)
 	return ERROR_OK;
 }
 
+static uint64_t set_ebreakx_fields(uint64_t dcsr, const struct target *target)
+{
+	const struct riscv_private_config * const config = riscv_private_config(target);
+	dcsr = set_field(dcsr, DCSR_EBREAKM, config->dcsr_ebreak_fields[RISCV_MODE_M]);
+	dcsr = set_field(dcsr, DCSR_EBREAKS, config->dcsr_ebreak_fields[RISCV_MODE_S]);
+	dcsr = set_field(dcsr, DCSR_EBREAKU, config->dcsr_ebreak_fields[RISCV_MODE_U]);
+	dcsr = set_field(dcsr, DCSR_EBREAKH, 1);
+	return dcsr;
+}
+
 static int execute_resume(struct target *target, bool step)
 {
-	RISCV_INFO(r);
 	riscv011_info_t *info = get_info(target);
 
 	LOG_DEBUG("step=%d", step);
@@ -1108,10 +1117,7 @@ static int execute_resume(struct target *target, bool step)
 		}
 	}
 
-	info->dcsr = set_field(info->dcsr, DCSR_EBREAKM, r->riscv_ebreakm);
-	info->dcsr = set_field(info->dcsr, DCSR_EBREAKS, r->riscv_ebreaks);
-	info->dcsr = set_field(info->dcsr, DCSR_EBREAKU, r->riscv_ebreaku);
-	info->dcsr = set_field(info->dcsr, DCSR_EBREAKH, 1);
+	info->dcsr = set_ebreakx_fields(info->dcsr, target);
 	info->dcsr &= ~DCSR_HALT;
 
 	if (step)
@@ -1928,7 +1934,6 @@ static int riscv011_resume(struct target *target, int current,
 
 static int assert_reset(struct target *target)
 {
-	RISCV_INFO(r);
 	riscv011_info_t *info = get_info(target);
 	/* TODO: Maybe what I implemented here is more like soft_reset_halt()? */
 
@@ -1942,10 +1947,7 @@ static int assert_reset(struct target *target)
 
 	/* Not sure what we should do when there are multiple cores.
 	 * Here just reset the single hart we're talking to. */
-	info->dcsr = set_field(info->dcsr, DCSR_EBREAKM, r->riscv_ebreakm);
-	info->dcsr = set_field(info->dcsr, DCSR_EBREAKS, r->riscv_ebreaks);
-	info->dcsr = set_field(info->dcsr, DCSR_EBREAKU, r->riscv_ebreaku);
-	info->dcsr = set_field(info->dcsr, DCSR_EBREAKH, 1);
+	info->dcsr = set_ebreakx_fields(info->dcsr, target);
 	info->dcsr |= DCSR_HALT;
 	if (target->reset_halt)
 		info->dcsr |= DCSR_NDRESET;
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -467,9 +467,30 @@ static struct target_type *get_target_type(struct target *target)
 	}
 }
 
+static struct riscv_private_config *alloc_default_riscv_private_config(void)
+{
+	struct riscv_private_config * const config = malloc(sizeof(*config));
+	if (!config) {
+		LOG_ERROR("Out of memory!");
+		return NULL;
+	}
+
+	for (unsigned int i = 0; i < ARRAY_SIZE(config->dcsr_ebreak_fields); ++i)
+		config->dcsr_ebreak_fields[i] = true;
+
+	return config;
+}
+
 static int riscv_create_target(struct target *target, Jim_Interp *interp)
 {
 	LOG_TARGET_DEBUG(target, "riscv_create_target()");
+	struct riscv_private_config *config = target->private_config;
+	if (!config) {
+		config = alloc_default_riscv_private_config();
+		if (!config)
+			return ERROR_FAIL;
+		target->private_config = config;
+	}
 	target->arch_info = calloc(1, sizeof(struct riscv_info));
 	if (!target->arch_info) {
 		LOG_TARGET_ERROR(target, "Failed to allocate RISC-V target structure.");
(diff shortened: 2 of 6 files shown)


en-sc:
@JanMatCodasip, @MarekVCodasip ping

src/target/riscv/riscv.h
```diff
@@ -326,17 +326,33 @@ struct riscv_info {
...
+	RISCV_DCSR_EBREAK_S_CTL,
+	RISCV_DCSR_EBREAK_U_CTL,
+	RISCV_DCSR_EBREAK_VS_CTL,
+	RISCV_DCSR_EBREAK_VU_CTL,
+	N_RISCV_DCSR_EBREAK_CTL
+};
+
+struct riscv_private_config {
+	bool dcsr_ebreak_ctls[N_RISCV_DCSR_EBREAK_CTL];
+};
```
JanMatCodasip: Why have you please decided to create an additional structure in addition to `struct riscv_info`?

Is the goal to move user-configurable items to this new structure and keep the rest of the information (everything that is not user-configurable) in `struct riscv_info`?

Or do you have something different in mind?

src/target/riscv/riscv.h
```diff
@@ -326,17 +326,33 @@ struct riscv_info {
...
 };
 
+enum riscv_ebreak_ctl {
+	RISCV_DCSR_EBREAK_M_CTL,
+	RISCV_DCSR_EBREAK_S_CTL,
+	RISCV_DCSR_EBREAK_U_CTL,
+	RISCV_DCSR_EBREAK_VS_CTL,
+	RISCV_DCSR_EBREAK_VU_CTL,
+	N_RISCV_DCSR_EBREAK_CTL
+};
```
JanMatCodasip: This enum is in fact a list of RISC-V privilege modes, so I would recommend to rename it in that fashion.

Once renamed, it will also become possible to reuse the enum elsewhere else.

Example:

```suggestion
enum riscv_priv_mode {
	RISCV_MODE_M,
	RISCV_MODE_S,
	RISCV_MODE_U,
	RISCV_MODE_VS,
	RISCV_MODE_VU,
	N_RISCV_MODE
};
```

src/target/riscv/riscv.h
```diff
@@ -326,17 +326,33 @@ struct riscv_info {
...
+
+struct riscv_private_config {
+	bool dcsr_ebreak_ctls[N_RISCV_DCSR_EBREAK_CTL];
```
JanMatCodasip: For clarity, I'd recommend to rename the field to `_bits` or `_fields` as opposed to `_ctls`.

(`ctrls` (controls) is a very general term that does not help code readers or reviewers much.)

```suggestion
	bool dcsr_ebreak_bits[N_RISCV_MODE];
```
or

```suggestion
	bool dcsr_ebreak_fields[N_RISCV_MODE];
```

src/target/riscv/riscv.c
```diff
@@ -4598,52 +4747,69 @@ COMMAND_HANDLER(riscv_set_autofence)
...
 }
 
-COMMAND_HANDLER(riscv_set_ebreakm)
+COMMAND_HELPER(ebreakx_deprecation_helper, enum riscv_ebreak_ctl mode)
 {
-	struct target *target = get_current_target(CMD_CTX);
-	RISCV_INFO(r);
-
+	struct target * const target = get_current_target(CMD_CTX);
+	struct riscv_private_config * const pc = riscv_private_config(target);
```
JanMatCodasip: Please, choose a different identifier than `pc` here (and in all other places). The reason is that `pc` commonly stands for program counter.

I noticed that some other targets do use `pc` as an ID for `private_config`, but it would IMO be better to avoid this confusion.

For example:

```suggestion
	struct riscv_private_config * const config = riscv_private_config(target);
```

src/target/riscv/riscv.c
```diff
@@ -5411,24 +5577,21 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.handler = riscv_set_ebreakm,
 		.mode = COMMAND_ANY,
 		.usage = "[on|off]",
-		.help = "Control dcsr.ebreakm. When off, M-mode ebreak instructions "
-			"don't trap to OpenOCD. Defaults to on."
+		.help = "DEPRECATED! use '<target_name> configure/cget -ebreak'"
```
JanMatCodasip: If the help message is expanded like these, it will hopefully be more clear to users:

```suggestion
		.help = "DEPRECATED! use '<target_name> configure -ebreak' or '<target_name> cget -ebreak'"
```

src/target/riscv/riscv.c
```diff
@@ -480,10 +480,157 @@ static int riscv_create_target(struct target *target, Jim_Interp *interp)
...
+
+static struct riscv_private_config *default_riscv_private_config(void)
+{
+	struct riscv_private_config * const pc = malloc(sizeof(*pc));
+	if (!pc) {
+		LOG_ERROR("Out of memory!");
+		return NULL;
+	}
+
+	memset(pc->dcsr_ebreak_ctls, true, sizeof(pc->dcsr_ebreak_ctls));
```
JanMatCodasip: (Detail, but important for readers and bool value consistency, and also helpful for static analysis tools.)

Since `config->dcsr_ebreak_ctls` is not an array of bytes (but an array of bools which can be larger than 1 B), please initialize it to non-zero values by normal assignments, not by memset:

```suggestion
	for (unsigned int i = 0; i < N_RISCV_MODE; i++)
		config->dcsr_ebreak_bits[i] = true;
```

src/target/riscv/riscv.c
```diff
@@ -480,10 +480,157 @@ static int riscv_create_target(struct target *target, Jim_Interp *interp)
...
+		return JIM_ERR;
+	}
+	struct jim_nvp *mode_nvp;
+	if (jim_getopt_nvp(goi, nvp_ebreak_mode_opts, &mode_nvp) != JIM_OK) {
+		jim_getopt_nvp_unknown(goi, nvp_ebreak_mode_opts, /*hadprefix*/ true);
+		return JIM_ERR;
+	}
+	pc->dcsr_ebreak_ctls[ctrl_nvp->value] = mode_nvp->value;
+	return JIM_OK;
+}
```
JanMatCodasip: I believe it would be good to check the number of arguments in this function (`goi->argc`).

Here is an example how to do it: https://github.com/riscv-collab/riscv-openocd/blob/bfc12522395af86ba634aa8e085a8051ca6fd43c/src/target/arm_dap.c#L348-L351

src/target/riscv/riscv.c
```diff
@@ -480,10 +480,157 @@ static int riscv_create_target(struct target *target, Jim_Interp *interp)
...
+
+static int str_ebreak_config(struct target *target, char *buffer)
```
JanMatCodasip: 1) Would the name `ebreak_config_to_tcl_dict()` sound good to you?

2) A little comment would help the reader

```suggestion
/**
 * Obtain dcsr.ebreak* configuration as a Tcl dictionary.
 * Print the resulting string to the "buffer" and return the string length.
 * The "buffer" can be NULL, in which case only the length is computed but nothing is written.
 */
static int ebreak_config_to_tcl_dict(struct target *target, char *buffer)
```

src/target/riscv/riscv.c
```diff
@@ -480,10 +480,157 @@ static int riscv_create_target(struct target *target, Jim_Interp *interp)
...
+	char *str = malloc(len + 1);
+	if (!str) {
+		LOG_ERROR("Unable to allocate a string of %d bytes.", len + 1);
+		return JIM_ERR;
+	}
+	str_ebreak_config(target, str);
+	Jim_SetResultString(interp, str, len);
+	free(str);
+	return JIM_OK;
+}
```
JanMatCodasip: The number of arguments (`goi->argc`) could be checked in this function, too.

JanMatCodasip (commented):
Overall the change looks all right - thank you.

I have finished my review and do not expect to have any further comments.

src/target/riscv/riscv.h
en-sc: Initially, I've planned to use `riscv_info`, however this is not possible with how `target create` is currently handled:
* `-...` parameters that can be passed to `<target name> configure` can also be passed directly to `target create ...`.
* However, in the handler of `target create` these parameters are handled (in part via the call to `target_type->target_jim_configure`) before `target->arch_info` is allocated by the call to `target_type->target_create`.
* Therefore there needs to be a separate location for saving this configuration that is `target->private_config`.

For additional info please see: https://review.openocd.org/c/openocd/+/8434

IMHO, it's a viable solution, though perhaps not the one I would go with.
As a side-effect it may provide nice separation:
* `riscv_private_config` for all things that need configuration.
* `riscv_info` for all things that need discovery.

src/target/riscv/riscv.c
en-sc: Is this necessary?
This handler does not require arguments (`goi` is not passed to this function).

src/target/riscv/riscv.h
JanMatCodasip: Thank you for explanation.

src/target/riscv/riscv.c
JanMatCodasip: ~~I take my comment back. By trying it out, I have just found out that OpenOCD already provides proper validation~~ for the arguments of `<target_name> cget`:

```
> my_target cget -nonsense     
Unknown param: -nonsense, try one of: -type, -event, -work-area-virt, -work-area-phys, -work-area-size, -work-area-backup, -endian, -coreid, -chain-position, -dbgbase, -rtos, -defer-examine, -gdb-port, or -gdb-max-connections
> my_target cget -ebreak -nonsense
Unknown param: -nonsense, try one of: -type, -event, -work-area-virt, -work-area-phys, -work-area-size, -work-area-backup, -endian, -coreid, -chain-position, -dbgbase, -rtos, -defer-examine, -gdb-port, or -gdb-max-connections
> my_target cget -ebreak -event   
wrong # args: should be "-event ?event-name? ..."
> my_target cget               
wrong # args: should be "missing: -option ..."
```

As a general rule, my preference would be to perform strict validation of any user's input. Of course only in case the generic OpenOCD code does not perform the validation for us already.

JanMatCodasip: I have to correct myself. OpenOCD itself provides only limited validation for `cget`. Try e.g. this:

```
> my_target cget -ebreak -rtos  
```

The actual argument validation is done in the command handlers itself, e.g. here: https://github.com/riscv-collab/riscv-openocd/blob/cabb6000dfd96a72858dfd8165a5e758b9c0b14b/src/target/target.c#L5114-L5117

So considering that other `cget` handlers in OpenOCD perform this validation on their own, it would be my preference to do the same thing here, too.

src/target/riscv/riscv.c
```diff
@@ -480,10 +480,157 @@ static int riscv_create_target(struct target *target, Jim_Interp *interp)
...
+
+	return pc;
+}
+
+static int riscv_jim_configure(struct target *target,
+		struct jim_getopt_info *goi)
+{
+	struct riscv_private_config *pc = target->private_config;
+	if (!pc) {
+		pc = default_riscv_private_config();
```
en-sc: There is an issue with `private_config` deallocation in case of an error.
Please, see https://review.openocd.org/c/openocd/+/8637/comments/512687b3_e6274fb6

en-sc: Should be fixed by https://review.openocd.org/c/openocd/+/8642

src/target/riscv/riscv.c
en-sc: Addressed. Thanks for the catch!
Though I believe `Jim_WrongNumArgs()` is used incorrectly throughout the codebase. Please see https://review.openocd.org/c/openocd/+/8643 for details.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.h

src/target/riscv/riscv.h

src/target/riscv/riscv.c

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: The issue is somewhat minor and is not specific to this patch.
IMHO, the patch can be merged without the fix.

JanMatCodasip: I agree - no problem with merging this now.

src/target/riscv/riscv-011.c
```diff
@@ -1074,9 +1074,18 @@ static int maybe_write_tselect(struct target *target)
...
 
+static uint64_t set_ebreakx_fields(uint64_t dcsr, const struct target *target)
+{
+	const struct riscv_private_config * const config = riscv_private_config(target);
+	dcsr = set_field(dcsr, DCSR_EBREAKM, config->dcsr_ebreak_fields[RISCV_MODE_M]);
+	dcsr = set_field(dcsr, DCSR_EBREAKS, config->dcsr_ebreak_fields[RISCV_MODE_S]);
+	dcsr = set_field(dcsr, DCSR_EBREAKU, config->dcsr_ebreak_fields[RISCV_MODE_U]);
+	dcsr = set_field(dcsr, DCSR_EBREAKH, 1);
+	return dcsr;
+}
```
JanMatCodasip, line 1085: There is no `ebreakh` field in register `dcsr` anymore, so this looks like outdated code.

I believe that the other variant of the dcsr.ebreakX manipulation -- which is on [line 1615](https://github.com/riscv-collab/riscv-openocd/pull/1168/files#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63R1615) -- is the correct one.

I don't mind if this part is fixed in a separate pull request.

src/target/riscv/riscv.c
```diff
@@ -480,10 +480,179 @@ static int riscv_create_target(struct target *target, Jim_Interp *interp)
...
+			++ebreak_ctl_i) {
+		const char * const format = "%s%s %s";
+		const char * const ctl = jim_nvp_value2name_simple(nvp_ebreak_config_opts, ebreak_ctl_i)->name;
+		const struct riscv_private_config * const config = riscv_private_config(target);
+		const char * const mode = jim_nvp_value2name_simple(nvp_ebreak_mode_opts,
+				config->dcsr_ebreak_fields[ebreak_ctl_i])->name;
+		if (!buffer)
+			len += snprintf(NULL, 0, format, separator, ctl, mode);
+		else
+			len += sprintf(buffer + len, format, separator, ctl, mode);
```
JanMatCodasip: Detail but important for readability:

- `ctl` should be named something like `priv_mode` (because it is the RISC-V privilege mode)
- `mode` should be named something like `ebreak_config`, `ebreak_behavior`, or similar

```suggestion
		if (!buffer)
			len += snprintf(NULL, 0, format, separator, priv_mode, ebreak_config);
		else
			len += sprintf(buffer + len, format, separator, priv_mode, ebreak_config);
```

src/target/riscv/riscv.c
```diff
@@ -480,10 +480,179 @@ static int riscv_create_target(struct target *target, Jim_Interp *interp)
...
+	RISCV_CFG_EBREAK,
+	RISCV_CFG_INVALID = -1
+};
+
+static struct jim_nvp nvp_config_opts[] = {
+	{ .name = "-ebreak", .value = RISCV_CFG_EBREAK },
+	{ .name = NULL, .value = RISCV_CFG_INVALID }
+};
+
+static struct riscv_private_config *default_riscv_private_config(void)
```

```suggestion
static struct riscv_private_config *alloc_default_riscv_private_config(void)
```

JanMatCodasip (commented):
Thank you for addressing the suggestions so far.

I have noticed three more items, and shouldn't have more comments than that.

src/target/riscv/riscv-011.c
en-sc: I'm not sure I understand what you mean.
This is a change to 0.11.
[RISC-V Debug Specification v0.11](https://static.dev.sifive.com/riscv-debug-spec-0.11nov12.pdf) defines in [8.6.1 Debug Control and Status] `ebreakh` field but does not define `ebreakvs`/`ebreakvu`.

src/target/riscv/riscv.c

src/target/riscv/riscv.c

en-sc:
Please, note: In the last change I didn't only address review comments.
I've also moved creation of `private_config` to target create and replaced a pointer to `target` with a pointer to `private_config` where applicable.

src/target/riscv/riscv-011.c
JanMatCodasip: You're right - I misread the code and thought this was for 0.13+ version of the spec.

JanMatCodasip (dismissed):
LGTM, thank you.

src/target/riscv/riscv.c
```diff
@@ -480,6 +501,155 @@ static int riscv_create_target(struct target *target, Jim_Interp *interp)
...
+	int e = jim_nvp_name2value_obj(goi->interp, nvp_config_opts,
+				goi->argv[0], &n);
+	if (e != JIM_OK)
+		return JIM_CONTINUE;
+
+	e = jim_getopt_obj(goi, NULL);
+	if (e != JIM_OK)
+		return e;
+
+	if (!goi->isconfigure && goi->argc > 0) {
```
en-sc: Needs to be adjusted after #1186 is merged due to 8c739a45a042cf7a9b5bba48f0e20cbc3168ee73

en-sc: Done

en-sc:
@JanMatCodasip, can you please double-check and re-approve the patch. I'd like to merge it.

JanMatCodasip:
@en-sc If I read the changes since the last review correctly, this patch has only been rebased (without further changes). Is that correct?

If so, it LGTM :)

JanMatCodasip (approved):

src/target/riscv/riscv.c
en-sc: d4b3b4ea82ba6d34b050a1cc068e0b105533e2f2 fixes this in #1210

PR #1171, opened 2024-11-20
target/riscv: Fix SMP group is in inconsistent state

lz-bro:
If all current halted states are due to a halt group, then a new "triggered a halt" event has occurred.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -37,6 +37,8 @@
 
 #define RISCV_TRIGGER_HIT_NOT_FOUND ((int64_t)-1)
 
+#define RISCV_HALT_GROUP_REPOLL_LIMIT 5
+
 static uint8_t ir_dtmcontrol[4] = {DTMCONTROL};
 struct scan_field select_dtmcontrol = {
 	.in_value = NULL,
@@ -3722,6 +3724,8 @@ int riscv_openocd_poll(struct target *target)
 {
 	LOG_TARGET_DEBUG(target, "Polling all harts.");
 
+	struct riscv_info *i = riscv_info(target);
+
 	struct list_head *targets;
 
 	LIST_HEAD(single_target_list);
@@ -3743,6 +3747,7 @@ int riscv_openocd_poll(struct target *target)
 	unsigned int should_resume = 0;
 	unsigned int halted = 0;
 	unsigned int running = 0;
+	unsigned int cause_groups = 0;
 	struct target_list *entry;
 	foreach_smp_target(entry, targets) {
 		struct target *t = entry->target;
@@ -3790,6 +3795,59 @@ int riscv_openocd_poll(struct target *target)
 		LOG_TARGET_DEBUG(target, "resume all");
 		riscv_resume(target, true, 0, 0, 0, false);
 	} else if (halted && running) {
+		LOG_TARGET_DEBUG(target, "SMP group is in inconsistent state: %u halted, %u running",
+					halted, running);
+
+		/* The SMP group is in an inconsistent state - some harts in the group have halted
+		 * whereas others are running. The reasons for that (and corresponding
+		 * OpenOCD actions) could be:
+		 * 1) The targets are in the process of halting due to halt groups
+		 *    but not all of them halted --> poll again so that the halt reason of every
+		 *    hart can be accurately determined (e.g. semihosting).
+		 * 2) The targets do not support halt groups --> OpenOCD must halt
+		 *    the remaining harts by a standard halt request.
+		 * 3) The hart states got out of sync for some other unknown reason (problem?). -->
+		 *    Same as previous - try to halt the harts by a standard halt request
+		 *    to get them back in sync. */
+
+		/* Detect if the harts are just in the process of halting due to a halt group */
+		foreach_smp_target(entry, targets)
+		{
+			struct target *t = entry->target;
+			if (t->state == TARGET_HALTED) {
+				riscv_reg_t dcsr;
+				if (riscv_reg_get(t, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
+					return ERROR_FAIL;
+				if (get_field(dcsr, CSR_DCSR_CAUSE) == CSR_DCSR_CAUSE_GROUP)
+					cause_groups++;
+				else
+					/* This hart has halted due to something else than a halt group.
+					 * Don't continue checking the rest - exit early. */
+					break;
+			}
+		}
+		/* Condition: halted == cause_groups
+		 *
+		 * This condition indicates a paradox where:
+		 * - All currently halted harts show CSR_DCSR_CAUSE_GROUP
+		 * - However, no individual hart can be identified as the actual initiator of the halt condition
+		 *
+		 * Poll again so that the true halt reason can be discovered (e.g. CSR_DCSR_CAUSE_EBREAK) */
+		if (halted == cause_groups) {
+			LOG_TARGET_DEBUG(target, "The harts appear to just be in the process of halting due to a halt group.");
+			if (i->halt_group_repoll_count < RISCV_HALT_GROUP_REPOLL_LIMIT) {
+				/* Wait a little, then re-poll. */
+				i->halt_group_repoll_count++;
+				alive_sleep(10);
+				LOG_TARGET_DEBUG(target, "Re-polling the state of the SMP group.");
+				return riscv_openocd_poll(target);
+			}
+			/* We have already re-polled multiple times but the halt group is still inconsistent. */
+			LOG_TARGET_DEBUG(target, "Re-polled the SMP group %d times it is still not in a consistent state.",
+					RISCV_HALT_GROUP_REPOLL_LIMIT);
+		}
+
+		/* Halting the whole SMP group to bring it in sync. */
 		LOG_TARGET_DEBUG(target, "halt all; halted=%d",
 			halted);
 		riscv_halt(target);
@@ -3806,6 +3864,8 @@ int riscv_openocd_poll(struct target *target)
 			}
 		}
 	}
+
+	i->halt_group_repoll_count = 0;
 
 	/* Call tick() for every hart. What happens in tick() is opaque to this
 	 * layer. The reason it's outside the previous loop is that at this point
--- a/src/target/riscv/riscv.h
+++ b/src/target/riscv/riscv.h
@@ -193,6 +193,7 @@ struct riscv_info {
 	/* Used by riscv_openocd_poll(). */
 	bool halted_needs_event_callback;
 	enum target_event halted_callback_event;
+	unsigned int halt_group_repoll_count;
 
 	enum riscv_isrmasking_mode isrmask_mode;
 


src/target/riscv/riscv.c
```diff
@@ -3790,6 +3791,22 @@ int riscv_openocd_poll(struct target *target)
...
+			if (t->state == TARGET_HALTED) {
+				riscv_reg_t dcsr;
+				if (riscv_reg_get(t, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
+					return ERROR_FAIL;
+				if (get_field(dcsr, CSR_DCSR_CAUSE) == CSR_DCSR_CAUSE_GROUP)
+					cause_groups++;
+				else
+					break;
+			}
+		}
```
aap-sc, line 3828: This loop seems weird. What is it's purpose?  DCSR is writable, so we can occasionally trick debugger into wrong conclusion.

 halt groups are an optional feature and I'm quite confused that we don't check for it.

Could you please provide a test scenario to reproduce your issue? Is it possible to use spike to model it? Or do you need a specific HW ?

lz-bro: This bug was discovered when I was testing [Semihosting](https://github.com/riscv-software-src/riscv-tests/blob/0494f954a3d8d2ca9e4972da7a01e94b6a909bce/debug/gdbserver.py#L1026) on our hardware and `smp` is enable. It fails when the following happens:
If all current halted states are due to a halt group, and other harts state was running. In fact, there was a hart  halted,  which caused the other harts to halt because of the hart group.
https://github.com/riscv-collab/riscv-openocd/blob/f51900b4a292ddbdd8de21728503e927f6e95cf9/src/target/riscv/riscv.c#L3792-L3796
If there is such a halted hart，but the record status is running，it would not process `riscv_semihosting`.
https://github.com/riscv-collab/riscv-openocd/blob/f51900b4a292ddbdd8de21728503e927f6e95cf9/src/target/riscv/riscv.c#L3605-L3623

lz-bro: @aap-sc I think this is a bug, would you provide some suggestions?

aap-sc: @lz-bro I'm still trying to understand your reasoning and what the issue is exactly (the situation is still not quite obvious to me). It will take a couple of days - I'll ask additional question if necessary.

JanMatCodasip:
@lz-bro I am afraid I have not understood what case this merge request addresses; not even after reading the commit description and the discussion so far.

Please, could you provide a very clear description in the commit message. Doing so will help:

- Code reviewers to understand your changes (and have them merged)
- Other users of OpenOCD source code - both current and future.

Thank you.

zqb-all:
Let me try to explain this issue.
Suppose we have two cores, core0 and core1, belong to one smp group.
Openocd constantly calls `riscv_openocd_poll` to check the status.

When openocd calls `riscv_openocd_poll` on core0:

A. if no hardware state change occurs，sequence is:
A.1. Check that core0 belongs to smp and start checking the status of all smp cores.
A.2. `riscv_poll_hart` checks core0. core0's status is running without change. `next_action=RPH_NONE, running++`
A.3. `riscv_poll_hart` checks core1. core1's status is running without change. . `next_action=RPH_NONE, running++`
A.4. Finally, `should_remain_halt=0, should_remain_resume=0, halted=0, and running=2`, nothing happen

B. If core0 hit soft breakpoints on hardware, one possible sequence is
B.1. Check that core0 belongs to smp and start to check the status of all smp cores.
B.2. `riscv_poll_hart` checks core0. core0's status is running without change. `next_action=RPH_NONE, running++`
[Between B.2 and B.3, core0 hits the breakpoint and becomes halted, and core1 also halted immediately due to hw haltgroup]
B.3. `riscv_poll_hart` checks core1 and finds that core1's status has changed from running to halted. The reason is `RISCV_HALTED_GROUP`. This results in `next_action=RPH_NONE, halted++`
4. Finally, `should_remain_halt=0, should_remain_resume=0, halted=1, running=1`. Enter [else if (halted && running)](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3870-L3873), and call `riscv_halt` to halt all cores in the smp group. In the process core0's `target->status` will also be corrected to halted, and next time `riscv_openocd_poll` will consider core0's status unchanged.
There is no problem in this case.

Let's re-assume that core0/core1 are both running and consider case C

C. If core0 hit semihosting ebeak on hardware, one possible sequence is:
C.1. Check that core0 belongs to smp and start to check the status of all smp cores.
C.2. `riscv_poll_hart` checks core0.  core0's status is running without change. ` next_action=RPH_NONE, running++`
[Between C.2 and C.3, core0 hit semihosting ebreak and halted, and core1 also halted immediately due to hw haltgroup]
C.3. `riscv_poll_hart` checks core1 and finds that core1 status has changed from running to halted. The reason is `RISCV_HALT_GROUP`. This results in `next_action=RPH_NONE, halted++`
C.4. Finally, `should_remain_halt=0, should_remain_resume=0, halted=1, running=1`. Enter [else if (halted && running)](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3870-L3873), and call `riscv_halt` to halt all cores in the smp group. In the process core0's `target->status` will also be corrected to halted, and next time `riscv_openocd_poll` will consider core0's status unchanged.
**Now here's the problem: as status not change, poll will not enter [this if](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3670) , cannot realize that core0 is actually semihosting ebreak and will not [handle it](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3683-L3686). core0 and core1 will remain halted.**

Let's re-assume that core0/core1 are both running and consider case D

D. If core0 hit semihosting ebeak on the hardware, but the timing was earlier than in case C, one possible sequence is:
D.1. Check that core0 belongs to smp and start checking the status of all smp cores.
[Between D.1 and D.2, core0 hit semihosting ebreak and halted, and core1 also halted immediately due to hw haltgroup]
D.2. `riscv_poll_hart` checks core0 and finds that core0 has changed from running to halted. Because halt_reason is `RISCV_HALT_EBREAK`, semihosting will be further [checked and processed](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3683-L3686), after that one possible result is  `next_action=RPH_RESUME, should_resume++`
D.3. `riscv_poll_hart` checks core1 and finds that core1 status has changed from running to halted. The reason is `RISCV_HALTED_GROUP`. This results in `next_action=RPH_NONE, halted++`
D.4. Finally, `should_remain_halt=0, should_remain_resume=1, halted=1, running=0.` Enter the [else if (should_resume)](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3867-L3869), then call `riscv_resume`, and eventually core0 and core1 will return to the running state.
There is no problem in this case.

Thank you.

zqb-all:
Let me try to explain this issue.
Suppose we have two cores, core0 and core1, belong to one smp group.
Openocd constantly calls `riscv_openocd_poll` to check the status.

When openocd calls `riscv_openocd_poll` on core0:

A. if no hardware state change occurs，sequence is:
A.1. Check that core0 belongs to smp and start checking the status of all smp cores.
A.2. `riscv_poll_hart` checks core0. core0's status is running without change. `next_action=RPH_NONE, running ++`
A.3. `riscv_poll_hart` checks core1. core1's status is running without change. . `next_action=RPH_NONE, running ++`
A.4. Finally, `should_remain_halt=0, should_remain_resume=0, halted=0, and running=2`, nothing happen

B. If core0 hit soft breakpoints on hardware, one possible sequence is
B.1. Check that core0 belongs to smp and start to check the status of all smp cores.
B.2. `riscv_poll_hart` checks core0. core0's status is running without change. `next_action=RPH_NONE, running ++`
[Between B.2 and B.3, core0 hits the breakpoint and becomes halted, and core1 also halted immediately due to hw haltgroup]
B.3. `riscv_poll_hart` checks core1 and finds that core1's status has changed from running to halted. The reason is `RISCV_HALTED_GROUP`. This results in `next_action=RPH_NONE, halted++`
4. Finally, `should_remain_halt=0, should_remain_resume=0, halted=1, running=1`. Enter [else if (halted && running)](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3870-L3873), and call `riscv_halt` to halt all cores in the smp group. In the process core0's `target->status` will also be corrected to halted, and next time `riscv_openocd_poll` will consider core0's status unchanged.
There is no problem in this case.



Let's re-assume that core0/core1 are both running and consider case C

C. If core0 hit semihosting ebeak on hardware, one possible sequence is:
C.1. Check that core0 belongs to smp and start to check the status of all smp cores.
C.2. `riscv_poll_hart` checks core0.  core0's status is running without change. ` next_action=RPH_NONE, running++`
[Between C.2 and C.3, core0 hit semihosting ebreak and halted, and core1 also halted immediately due to hw haltgroup]
C.3. `riscv_poll_hart` checks core1 and finds that core1 status has changed from running to halted. The reason is `RISCV_HALT_GROUP`. This results in `next_action=RPH_NONE, halted++`
C.4. Finally, `should_remain_halt=0, should_remain_resume=0, halted=1, running=1`. Enter [else if (halted && running)](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3870-L3873), and call `riscv_halt` to halt all cores in the smp group. In the process core0's `target->status` will also be corrected to halted, and next time `riscv_openocd_poll` will consider core0's status unchanged.
**Now here's the problem: poll cannot realize that core0 is actually semihosting ebreak and will not [handle it](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3683-L3686). core0 and core1 will remain halted.**

Let's re-assume that core0/core1 are both running and consider case D

D. If core0 hit semihosting ebeak on the hardware, but the timing was earlier than in case C, one possible sequence is:
D.1. Check that core0 belongs to smp and start checking the status of all smp cores.
[Between D.1 and D.2, core0 hit semihosting ebreak and halted, and core1 also halted immediately due to hw haltgroup]
D.2. `riscv_poll_hart` checks core0 and finds that core0 has changed from running to halted. Because halt_reason is `RISCV_HALT_EBREAK`, semihosting will be further [checked and processed](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3683-L3686), after that one possible result is  `next_action=RPH_RESUME, should_resume++`
D.3. `riscv_poll_hart` checks core1 and finds that core1 status has changed from running to halted. The reason is `RISCV_HALTED_GROUP`. This results in `next_action=RPH_NONE, halted++`
D.4. Finally, `should_remain_halt=0, should_remain_resume=1, halted=1, running=0.` Enter the [else if (should_resume)](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3867-L3869), then call `riscv_resume`, and eventually core0 and core1 will return to the running state.
There is no problem in this case.
 
Please correct me if I have misunderstood something. Thank you

zqb-all:
Things are a bit complicated. 
The poll function of the software takes time to run, and hart may hit ebreak (breakpoint or semihosting) at any time, then harts in the same smp group,  may be halted immediately (by hardware haltgroup) or may continue running temporarily (for example, hart is controlled by another DM and no hardware supports synchronous halt). 
We expect the poll function to eventually adjust the smp group to [halted](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3870-L3873) or [running](https://github.com/riscv-collab/riscv-openocd/blob/ca8092015716dbc9ae69dd8240197b00b5900b0a/src/target/riscv/riscv.c#L3867-L3869). But before the software completes processing, these temporarily  running harts may also hit ebreak, make the situation more complicated.

zqb-all:
@JanMatCodasip @aap-sc   Does my description of the issue help you understand what the issue is ?

JanMatCodasip:
@zqb-all Thank you for describing the situation in more detail. It will take me some time to get back to it.

src/target/riscv/riscv.c
```diff
@@ -3790,6 +3791,22 @@ int riscv_openocd_poll(struct target *target)
 		LOG_TARGET_DEBUG(target, "resume all");
 		riscv_resume(target, true, 0, 0, 0, false);
 	} else if (halted && running) {
+		foreach_smp_target(entry, targets)
```
JanMatCodasip, line 3814:
```suggestion
		LOG_TARGET_DEBUG(target, "SMP group is in inconsistent state: %u halted, %u running", halted, running);

		/* The SMP group is in an inconsistent state - some harts in the group have halted
		 * whereas others are running. The reasons for that (and corresponding 
		 * OpenOCD actions) could be:
		 * 1) The targets are in the process of halting due to halt groups 
		 *    but not all of them halted --> wait a moment and then poll again so that
		 *    the halt reason of every hart can be accurately determined (e.g. semihosting).
		 * 2) The targets do not support halt groups --> OpenOCD must halt 
		 *    the remaining harts by a standard halt request.
		 * 3) The hart states got out of sync for some other unknown reason (problem?). --> 
		 *    Same as previous - try to halt the harts by a standard halt request
		 *    to get them back in sync.
		 
		/* Detect if the harts are just in the process of halting due to a halt group */ 
		foreach_smp_target(entry, targets)
```

src/target/riscv/riscv.c
```diff
@@ -3790,6 +3791,22 @@ int riscv_openocd_poll(struct target *target)
...
+		{
+			struct target *t = entry->target;
+			if (t->state == TARGET_HALTED) {
+				riscv_reg_t dcsr;
+				if (riscv_reg_get(t, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
+					return ERROR_FAIL;
+				if (get_field(dcsr, CSR_DCSR_CAUSE) == CSR_DCSR_CAUSE_GROUP)
+					cause_groups++;
+				else
+					break;
```
JanMatCodasip, line 3826:
```suggestion
					/* This hart has halted due to something else than a halt group. 
					 * Don't continue checking the rest - exit early. */
					break;
```

src/target/riscv/riscv.c
```diff
@@ -3790,6 +3791,22 @@ int riscv_openocd_poll(struct target *target)
...
+			}
+		}
+		if (halted == cause_groups)
+			return ERROR_OK;
```
```suggestion
		if (halted == cause_groups) {
			LOG_TARGET_DEBUG(target, "The harts appear to just in the process of halting due to halt group. Giving them more time - will poll their state later.");
			return ERROR_OK;
		}
```

src/target/riscv/riscv.c
JanMatCodasip, line 3814: I wonder if this loop can be skipped if `haltgroup_supported` is equal to false.

src/target/riscv/riscv.c
```diff
@@ -3790,6 +3791,22 @@ int riscv_openocd_poll(struct target *target)
...
+
 		LOG_TARGET_DEBUG(target, "halt all; halted=%d",
 			halted);
```
JanMatCodasip, line 3852:
```suggestion
		/* Halting the whole SMP group to bring it in sync. */
		LOG_TARGET_DEBUG(target, "halt all; halted=%d",
			halted);
```

src/target/riscv/riscv.c
JanMatCodasip: I believe I have figured out the code thanks to the [detailed description](https://github.com/riscv-collab/riscv-openocd/pull/1171#issuecomment-2508162875) that @zqb-all has provided:

If the harts are just in the process of halting due to being members of the halt group, we should wait until they finish halting, so that the true halt reason can be discovered - for instance semihosting request - and then handled correctly.

src/target/riscv/riscv.c
JanMatCodasip: (1) I would recommend to keep a counter of how many times this situation has occurred. If it reaches a certain limit, OpenOCD should:
- stop waiting for the harts to halt, 
- print an error to the user, and 
- try to halt all the harts directly (riscv_halt()).

(2) Instead of returning from `riscv_openocd_poll()` and waiting for the next poll interval, it would perhaps be better to create a loop inside this function. This would allow to immediately re-poll the hart state and react quickly. 
 - @aap-sc @zqb-all What is your opinion on that? 
 - If you are in favor of it but don't feel like implementing it in this PR, consider adding a TODO comment.

JanMatCodasip (commented):
Hi @zqb-all, thank you for the [clear and accurate description](https://github.com/riscv-collab/riscv-openocd/pull/1171#issuecomment-2508162875) of the situation that you're trying to address. It helped me to understand the code change. I am also sorry for replying late.

Overall this fix looks good and I have posted some comments.

src/target/riscv/riscv.c
zqb-all: > create a loop inside this function. This would allow to immediately re-poll the hart state and react quickly.

Sounds good, this helps smp harts finish state adjustment faster.

src/target/riscv/riscv.c
lz-bro: Addressed

src/target/riscv/riscv.c
lz-bro: Addressed

src/target/riscv/riscv.c
lz-bro: Addressed

src/target/riscv/riscv.c
JanMatCodasip: After thinking about it more, the `haltgroup_supported` can have different value for each hart.

So it is IMO better to leave the code as is, and don't try to skip the whole loop (don't try to over-optimize).

Marking this thread as resolved.

src/target/riscv/riscv.c
```diff
@@ -3790,6 +3791,44 @@ int riscv_openocd_poll(struct target *target)
...
+				if (get_field(dcsr, CSR_DCSR_CAUSE) == CSR_DCSR_CAUSE_GROUP)
+					cause_groups++;
+				else
+					/* This hart has halted due to something else than a halt group.
+					 * Don't continue checking the rest - exit early. */
+					break;
+			}
+		}
+		if (halted == cause_groups) {
+			LOG_TARGET_DEBUG(target, "The harts appear to just in the process of halting due to halt group. "
```
JanMatCodasip: Just a detail:

```suggestion
			LOG_TARGET_DEBUG(target, "The harts appear to just be in the process of halting due to halt group. "
```

src/target/riscv/riscv.c
```diff
@@ -3790,6 +3791,44 @@ int riscv_openocd_poll(struct target *target)
...
+						"re-poll the hart state.");
+			return riscv_openocd_poll(target);
```
JanMatCodasip: This direct recursion looks dangerous. Please, could you constrain it so that the recursion is guaranteed to happen only a limited number of times:

1) add a little delay before recursing
2) add a safety counter to prevent infinite recursion (by design)

The code could look like this (or similar):

```c
#define RISCV_HALT_GROUP_REPOLL_LIMIT 5

/* ... */

		if (halted == cause_groups) {
			LOG_TARGET_DEBUG(target, "The harts appear to just be in the process of halting due to a halt group. ");
			if (info->halt_group_repoll_count < RISCV_HALT_GROUP_REPOLL_LIMIT) {
				/* Wait a little, then re-poll. */
				info->halt_group_repoll_count++;
				alive_sleep(10);
				LOG_TARGET_DEBUG(target, "Re-polling the state of the SMP group.");
				return riscv_openocd_poll(target);
			} else {
				/* We have already re-polled multiple times but the halt group is still inconsistent. */
				LOG_TARGET_DEBUG(target, "Re-polled the SMP group %d times it is still not "
						"in a consistent state.", RISCV_HALT_GROUP_REPOLL_LIMIT);
			}
		}
		/* Halting the whole SMP group to bring it in sync. */
		LOG_TARGET_DEBUG(target, "halt all; halted=%d", halted);

/* ... */

info->halt_group_repoll_count = 0;
```

JanMatCodasip (changes requested):

src/target/riscv/riscv.c

JanMatCodasip (dismissed):

JanMatCodasip:
LGTM, thank you.

aap-sc:
> @JanMatCodasip @aap-sc Does my description of the issue help you understand what the issue is ?

@zqb-all 

Sorry for the delay. I finally took the time and courage and thoroughly read @zqb-all `s explanation. Yep I managed to understand what the problem is.

src/target/riscv/riscv.h
```diff
@@ -193,6 +193,7 @@ struct riscv_info {
 	/* Used by riscv_openocd_poll(). */
 	bool halted_needs_event_callback;
 	enum target_event halted_callback_event;
+	unsigned int halt_group_repoll_count;
```
aap-sc, line 196: If looks like that `halt_group_repoll_count` should not be part of target state. It looks like we use it only for the purposes of riscv_openocd_poll. Maybe we could introduce some kind of wrapper function?

This can be implemented in separate MR.

src/target/riscv/riscv.c
```diff
@@ -3790,6 +3795,52 @@ int riscv_openocd_poll(struct target *target)
...
+					return ERROR_FAIL;
+				if (get_field(dcsr, CSR_DCSR_CAUSE) == CSR_DCSR_CAUSE_GROUP)
+					cause_groups++;
+				else
+					/* This hart has halted due to something else than a halt group.
+					 * Don't continue checking the rest - exit early. */
+					break;
+			}
+		}
+		if (halted == cause_groups) {
```
aap-sc, line 3836: Can you add a detailed comment describing this condition?

This condition effectively means that we still could not find a hart that causes CSR_DCSR_CAUSE_GROUP reason. It took a while for me to understand.

aap-sc (dismissed):
LGTM to me, though I have few minor comments.

src/target/riscv/riscv.c
lz-bro: Addressed

aap-sc (approved):

JanMatCodasip (approved):

PR #1176, opened 2024-11-27
fix incorrect parsing of names for custom csr registers

aap-sc:
this commit fixes a regression introduced in
https://github.com/riscv-collab/riscv-openocd/commit/ba8c1eef5a32193d9b5dc80d681a4f18d1d94f11.

The regression was caused by removal of these lines:

```
-                       /* Register prefix: "csr_" or "custom_" */
-                       strcpy(name, reg_type);
-                       name[strlen(reg_type)] = '_';
```

causing all CSR names with custom names to be parsed as empty strings.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -4164,12 +4164,16 @@ static int parse_reg_ranges_impl(struct list_head *ranges, char *args,
 			}
 
 			const char * const reg_name_in = equals + 1;
-			*name_buffer = calloc(1, strlen(reg_name_in) + strlen(reg_type) + 2);
+			const size_t reg_type_len = strlen(reg_type);
+			/* format is: <reg_type>_<reg_name_in>\0 */
+			*name_buffer = calloc(1, strlen(reg_name_in) + reg_type_len + 2);
 			name = *name_buffer;
 			if (!name) {
 				LOG_ERROR("Out of memory");
 				return ERROR_FAIL;
 			}
+			strcpy(name, reg_type);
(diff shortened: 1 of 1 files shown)


aap-sc:
@JanMatCodasip , @en-sc take a look please. This commit should fix a regression, please see commit description

JanMatCodasip (approved):
LGTM, thank you.

So it looks like there is no test in `riscv-tests/debug` that would cover custom CSRs (or full-custom registers), is that correct?

aap-sc:
> So it looks like there is no test in `riscv-tests/debug` that would cover custom CSRs (or full-custom registers), is that correct?

@JanMatCodasip to my surprise there are tests that should catch this issue. I don't understand why these tests passed. The regression is definitely there. Looks like something is wrong with testing - I'm trying to figure things out.

aap-sc:
> > So it looks like there is no test in `riscv-tests/debug` that would cover custom CSRs (or full-custom registers), is that correct?
> 
> @JanMatCodasip to my surprise there are tests that should catch this issue. I don't understand why these tests passed. The regression is definitely there. Looks like something is wrong with testing - I'm trying to figure things out.

Ok, I figured things out.

No, there are not such tests. To trigger the issue you have to have something like this in your configuration file:

```
riscv expose_csrs 2288, 2289=csr_name
```

That is you have to specify a name for the exposed register. We have no such constructs in the configuration files used in riscv-tests.

I'll add those shortly (once this change is merged in)

en-sc (approved):
LGTM

Issue #902, opened 2023-08-07
Incorrect work of examine() in case of halt groups


kr-sc:
I launch `spike` simulator with 4 harts:

```
spike \
  --rbb-port=7777 \
  --isa=rv64gc \
  --dm-progsize=6 \
  --dm-sba=0 \
  --dm-abstract-rti=0 \
  -p4 \
  -m0x80000000:0xa00000 \
  <some_elf_file>
```

then the following script:
```
for LOG in {1..2}; \
  do openocd \
    -c "adapter driver remote_bitbang" \
    -c "remote_bitbang host 127.0.0.1" \
    -c "remote_bitbang port 7777" \
    -c 'gdb_port disabled' \
    -f spike_smp4_rtoshw4.cfg \
    -c "log_output ${LOG}.log" \
    -c init \
    -c targets \
    -c shutdown -d3; \
done;
```

<details>
  <summary> spike_smp4_rtoshw4.cfg </summary>

```
proc init_targets {} {
	set _CHIPNAME riscv
	jtag newtap $_CHIPNAME cpu -irlen 5 -expected-id 0x10e31913

	set _TARGETNAME_0 $_CHIPNAME.cpu0
	set _TARGETNAME_1 $_CHIPNAME.cpu1
	set _TARGETNAME_2 $_CHIPNAME.cpu2
	set _TARGETNAME_3 $_CHIPNAME.cpu3

	target create $_TARGETNAME_0 riscv -chain-position $_CHIPNAME.cpu -coreid 0 -rtos hwthread
	target create $_TARGETNAME_1 riscv -chain-position $_CHIPNAME.cpu -coreid 1 -rtos hwthread
	target create $_TARGETNAME_2 riscv -chain-position $_CHIPNAME.cpu -coreid 2 -rtos hwthread
	target create $_TARGETNAME_3 riscv -chain-position $_CHIPNAME.cpu -coreid 3 -rtos hwthread

	target smp $_TARGETNAME_0 $_TARGETNAME_1 $_TARGETNAME_2 $_TARGETNAME_3
}
```

</details>

Result of the first iteration:
```
TargetName         Type       Endian TapName            State
--  ------------------ ---------- ------ ------------------ ------------
 0  riscv.cpu0         riscv      little riscv.cpu          running
 1  riscv.cpu1         riscv      little riscv.cpu          running
 2  riscv.cpu2         riscv      little riscv.cpu          running
[...]

PR #1179, opened 2024-12-02
target/riscv: Add support for external triggers

rbradford:
Add support for associating a halt group with an external trigger via a newly exposed configuration option "riscv set_external_trigger".

Change-Id: If10c67d2e14d8bc7cd6d59011b3215fda4ff4b02

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -11476,6 +11476,11 @@ The second argument configures how OpenOCD should use the selected trigger featu
 With no parameters, prints current trigger features configuration.
 @end deffn
 
+@deffn {Command} {riscv set_external_trigger} value
+Associate the supplied external trigger with the halt group for the harts. When
+the external trigger fires the harts in the halt group will be halted.
+@end deffn
+
 @subsection RISC-V Authentication Commands
 
 The following commands can be used to authenticate to a RISC-V system. Eg.  a
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1703,6 +1703,32 @@ static void deinit_target(struct target *target)
 	info->version_specific = NULL;
 }
 
+static int set_external_trigger(struct target *target, unsigned int group,
+		grouptype_t grouptype, unsigned int external_trigger)
+{
+	uint32_t write_val = DM_DMCS2_HGWRITE | DM_DMCS2_HGSELECT;
+	assert(group <= 31);
+	assert(external_trigger < 16);
+	write_val = set_field(write_val, DM_DMCS2_GROUP, group);
+	write_val = set_field(write_val, DM_DMCS2_GROUPTYPE, (grouptype == HALT_GROUP) ? 0 : 1);
+	write_val = set_field(write_val, DM_DMCS2_DMEXTTRIGGER, external_trigger);
+	if (dm_write(target, DM_DMCS2, write_val) != ERROR_OK)
+		return ERROR_FAIL;
+	uint32_t read_val;
+	if (dm_read(target, &read_val, DM_DMCS2) != ERROR_OK)
+		return ERROR_FAIL;
+	if (get_field(read_val, DM_DMCS2_GROUP) == group &&
+		get_field(read_val, DM_DMCS2_DMEXTTRIGGER) == external_trigger &&
+		get_field(read_val, DM_DMCS2_HGSELECT) == 1) {
+		LOG_TARGET_INFO(target, "External trigger %d added to group %d", external_trigger,
+				group);
+	} else {
+		LOG_TARGET_ERROR(target, "External trigger %d not supported", external_trigger);
+	}
+
+	return ERROR_OK;
+}
+
 static int set_group(struct target *target, bool *supported, unsigned int group,
 		grouptype_t grouptype)
 {
@@ -2051,6 +2077,9 @@ static int examine(struct target *target)
 		else
 			LOG_TARGET_INFO(target, "Core %d could not be made part of halt group %d.",
 					info->index, target->smp);
+		if (r->external_trigger)
+			if (set_external_trigger(target, target->smp, HALT_GROUP, r->external_trigger) != ERROR_OK)
+				return ERROR_FAIL;
 	}
 
 	/* Some regression suites rely on seeing 'Examined RISC-V core' to know
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -5262,6 +5262,26 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
 	return ERROR_OK;
 }
 
+COMMAND_HANDLER(riscv_set_external_trigger)
+{
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	if (CMD_ARGC != 1) {
+		LOG_ERROR("Command takes exactly 1 parameter.");
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+	int value = atoi(CMD_ARGV[0]);
+	if (value <= 0 || value > 16) {
+		LOG_ERROR("%s is not a valid integer argument for command.", CMD_ARGV[0]);
+		return ERROR_FAIL;
+	}
+
+	r->external_trigger = value;
+
+	return ERROR_OK;
+}
+
 static const struct command_registration riscv_exec_command_handlers[] = {
 	{
 		.name = "dump_sample_buf",
@@ -5523,6 +5543,13 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.help = "When on (default), OpenOCD will automatically execute fence instructions in some situations. "
 			"When off, users need to take care of memory coherency themselves, for example by using "
 			"`riscv exec_progbuf` to execute fence or CMO instructions."
+	},
+	{
+		.name = "set_external_trigger",
+		.handler = riscv_set_external_trigger,
+		.mode = COMMAND_CONFIG,
+		.usage = "value",
+		.help = "Add the given external trigger to the halt group"
 	},
 	COMMAND_REGISTRATION_DONE
 };
--- a/src/target/riscv/riscv.h
+++ b/src/target/riscv/riscv.h
@@ -191,6 +191,9 @@ struct riscv_info {
 	/* The configured approach to translate virtual addresses to physical */
 	riscv_virt2phys_mode_t virt2phys_mode;
 
+	/* Halt group may be associated with an external trigger */
+	unsigned int external_trigger;
+
 	bool triggers_enumerated;
 
 	/* Decremented every scan, and when it reaches 0 we clear the learned


doc/openocd.texi
```diff
@@ -11476,6 +11476,11 @@ The second argument configures how OpenOCD should use the selected trigger featu
 With no parameters, prints current trigger features configuration.
 @end deffn
 
+@deffn {Command} {riscv set_external_trigger} value
+Associate the supplied external trigger with the SMP halt group for the harts. When the external trigger
```
aap-sc: I believe that external trigger does not necessarily implies and SMP group.

rbradford: Thank you - i've dropped the word "SMP" from the documentation.

src/target/riscv/riscv-013.c
```diff
@@ -2051,6 +2077,9 @@ static int examine(struct target *target)
 		else
 			LOG_TARGET_INFO(target, "Core %d could not be made part of halt group %d.",
 					info->index, target->smp);
+		if (r->external_trigger)
```
aap-sc, line 2080: 1. it's not quite clear why this code is under `target->smp` condition. We can control an external triggger without creating an SMP group.
2. @en-sc I remember you wanted to implement an interface to debug module. Could you please assess if we can implement the functionality in question without this interface?
3. `if (r->external_trigger)` condition is dubious, since even the debug spec implies that external trigger
have numbers starting from "0":

```
By convention external triggers 0-7 are bidirectional,
triggers 8-11 are input-only, and triggers 12-15 are output-only but this is not required.
```

4. what about trigger directions?
5. do you have plans to add tests targeting this functionality?

doc/openocd.texi
```diff
@@ -11476,6 +11476,11 @@ The second argument configures how OpenOCD should use the selected trigger featu
 With no parameters, prints current trigger features configuration.
 @end deffn
 
+@deffn {Command} {riscv set_external_trigger} value
```
aap-sc, line 11479: RISC-V debug spec allows external triggers to be associated with halt and resume groups. I assume you are trying to add support for halt groups only. Looks like this command needs to be renamed, since resume groups are out of scope, or you should extend the interface to support resume groups too.

aap-sc:
@rbradford I've left some additional comments. Though, I believe we are far from done :(. 

The more I look at this patch and the the **current codebase** - the more concerns I have. 

Please, keep in mind that the current implementation of halt groups in OpenOCD is very questionable leading to issues like https://github.com/riscv-collab/riscv-openocd/issues/902 .

I would go as far as to say that the concept of halt groups is not implemented properly. The more I look at the relevant codebase - the more it seems to me that proper support for halt/resume groups should be implemented first. Once that is done we could adopt something like this patch. Others could have a different opinion, though - @JanMatCodasip , @en-sc do you have any comments regarding the matter?

JanMatCodasip:
I will be able to take a closer look at the end of this week.

doc/openocd.texi
zqb-all, line 11479: My initial idea was to create external triggers separately in the cfg file and list them in the `target smp` command, something like this:
```
# when no dmexttrigger
target create hart0 riscv -dbgbase 0
target create hart1 riscv -dbgbase 0x400
target smp hart0 hart1

# when support dmexttrigger, type0 halt, type1 resume
target create dm0extrigger0 riscv-dmexttrigger -index 0 -type 0  -dbgbase 0
target create dm0extrigger1 riscv-dmexttrigger  -index 1 -type 0  -dbgbase 0
target create hart0 riscv -dbgbase 0
target create dm1extrigger0 riscv-dmexttrigger -index 0 -type 0  -dbgbase 0x400
target create dm1extrigger1 riscv-dmexttrigger -index 1 -type 0  -dbgbase 0x400
target create hart1 riscv -dbgbase 0x400
target smp hart0 hart1 dm0extrigger0  dm0extrigger1  dm1extrigger0 dm1extrigger1
```

But there are a lot of changes required, so currently associating external trigger with a target seems like an easy enough way to start supporting it.

Suppose we support it this way, then my suggestion is let this cmd support multiple external triggers instead of just one. 
For example, if the external trigger is not bidirectional, we need at least two triggers, one input-only and one output-only, to support smp with cores on other DM.
There are more complex situations that may require more triggers.

src/target/riscv/riscv.c
```diff
@@ -5262,6 +5262,26 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
...
+		LOG_ERROR("Command takes exactly 1 parameter.");
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+	int value = atoi(CMD_ARGV[0]);
+	if (value <= 0 || value > 16) {
+		LOG_ERROR("%s is not a valid integer argument for command.", CMD_ARGV[0]);
+		return ERROR_FAIL;
+	}
+
+	r->external_trigger = value;
```
zqb-all, line 5280: It would be safer to add some checks to prevent user from associating one trigger to multiple harts on same DM.

src/target/riscv/riscv-013.c
rbradford: @aap-sc Thanks for taking the time to reply - some specific answers to your points

1. I was using `target->smp` as that was a condition to have access to a halt group
3. Well spotted - indeed trigger 0 is valid - this is wrong (I used external trigger 1 in my Spike test version for this change which is why I didn't spot this.) I will address this
4. I'm not sure trigger direction is something OpenOCD can handle since it's just a convention in the numbering
5. I hacked Spike to group some harts together separately and then "joined" them with an external trigger.

src/target/riscv/riscv.c
rbradford: Indeed this is a good point - the external trigger is logically associated with a halt group which will span multiple targets (assuming the one target per hart configuration method.)

doc/openocd.texi
rbradford: Indeed - my initial focus was only halt groups - this should be renamed if we go ahead with this very limited solution.

rbradford: The design of this syntax looks really nice - i'm very happy to wait on this or if you prefer take a go at implementing that style myself?

JanMatCodasip:
> aap-sc: The more I look at this patch and the the current codebase - the more concerns I have. [...]
> JanMatCodasip, en-sc do you have any comments regarding the matter?

I will take a close look at the SMP / halt-resume groups implementation to form an opinion on this matter (which I wanted to do anyway). It just may take me a day or two to get to this.

doc/openocd.texi
zqb-all: I haven't started looking at the code yet, just a basic idea that needs further thought.
I think we can move forward with your current implementation and make the external trigger available to those who need it first.

JanMatCodasip:
Hello all, I have checked the overall state of SMP and halt/resume groups in riscv-openocd.

To summarize the current state:

- Riscv-openocd currently implements halt groups only. Resume groups are not implemented.
- Halt groups are used only in relationship with the SMP mode.
- External triggers for halt/resume groups are not implemented. 

Comments, questions:

1) Halt groups are implemented only in relationship with SMP. Do we wish to keep that approach or allow placing also singular (non-SMP) targets into halt groups? 

2) As an alternative, the user could create SMP group with just a single target (hart).

3) Riscv-openocd auto-selects the halt group number for each SMP group (sequentially: 1, 2, 3, ...). Do we wish to keep it that way or give the user control over the halt group numbers?

4) External triggers for halt groups seem OK to be implemented. We just need to agree on the solution, which is already happening in this PR :)

5) Resume groups alone (without external triggers) do not seem to bring any benefit. Near-simultaneous resume of multiple harts is already implemented by selecting the harts and then resuming.

6) Resume groups with external **output** triggers could be eventually implemented into riscv-openocd, if desired. I don't see any principal problem there.

7) Resume groups with external **input** triggers look to be very problematic. Neither OpenOCD nor GDB is ready for the case when a target suddenly resumes (due to the trigger), without an explicit resume request and at arbitrary moment in time.

doc/openocd.texi
JanMatCodasip: @zqb-all Thank you for your [idea](https://github.com/riscv-collab/riscv-openocd/pull/1179#discussion_r1868560553) above.

Still, I'd recommend to not abuse `target create` in order to assign external triggers to SMP groups. "Target" has a well-defined meaning in OpenOCD: it is a debuggable entity with its state, registers, target API operations (target_type.h), ... External trigger is not a target. So I'd prefer to find a different syntax to assign external triggers to halt goups.

doc/openocd.texi
JanMatCodasip: If we decide to keep the status quo -- keep using halt groups for SMP groups -- then a simple Tcl command like this looks all right to me. 

Few suggestions:

- One halt group can have multiple external triggers (1:N relationship). So I'd call the command `add` rather than `set`.
- Since the command is only used in the configuration phase (COMMAND_CONFIG), it looks unnecessary to have a `remove` command to remove the trigger from the group.
- If the current relationship between halt groups and SMP stays in place, then the command name can contain `smp` -- as an additional hint for the user that it will apply to the whole SMP group.
- The command should contain `halt` or `halt_group` somewhere it its name to indicate it is for halt groups, not for resume groups.

For example:

```tcl
[target_name] riscv smp_add_ext_trigger halt_group <ext_trigger_number>
# halt_group would remain the only supported trigger type for the time being
```

src/target/riscv/riscv.h
```diff
@@ -191,6 +191,9 @@ struct riscv_info {
 	/* The configured approach to translate virtual addresses to physical */
 	riscv_virt2phys_mode_t virt2phys_mode;
 
+	/* Halt group may be associated with an external trigger */
+	unsigned int external_trigger;
```
JanMatCodasip, line 195: As pointed out in another comment, there can be multiple external triggers in a halt (or resume) group.

The code, on the other hand, seems to force 1:1 relationship.

doc/openocd.texi
zqb-all: > Still, I'd recommend to not abuse `target create` in order to assign external triggers to SMP groups. 

Agreed. Some concepts in debug spec are not explicitly represented in OpenOCD cfg file now. For example, `DM` is only specified with `dbgbase` when `target create`, and `halt group` is only implicitly used for `target smp`, so it is not easy to create a suitable syntax to configure a dmextrigger to a specific group now.
When I have a more mature idea, I will create another PR to discuss it. 
Thank you for your feedback on this idea.

lz-bro:
@rbradford @JanMatCodasip @aap-sc @en-sc @zqb-all  Hi all, is there a plan to continue the implementation of the external triggers for halt/resume groups ?

JanMatCodasip:
@lz-bro has continued the development of this feature in https://github.com/riscv-collab/riscv-openocd/pull/1243, which got later moved to https://review.openocd.org/c/openocd/+/9224.

@lz-bro, @rbradford, can I close this PR, please?

PR #1181, opened 2024-12-06
target/riscv: clean-up register invalidation

en-sc:
* Registers were not invalidated if the hart became unavailable.
* Improved logging in the case register invalidation involves loss of information.

Change-Id: Icfb5e190dd6dcb1a97e4d314d802466cab7a01e4

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2728,6 +2728,13 @@ static int handle_became_unavailable(struct target *target,
 		enum riscv_hart_state previous_riscv_state)
 {
 	RISCV013_INFO(info);
+
+	if (riscv_reg_cache_any_dirty(target, LOG_LVL_WARNING))
+		LOG_TARGET_WARNING(target, "Discarding values of dirty registers "
+				"(due to target becoming unavailable).");
+
+	riscv_reg_cache_invalidate_all(target);
+
 	info->dcsr_ebreak_is_set = false;
 	return ERROR_OK;
 }
@@ -5434,6 +5441,8 @@ static int riscv013_step_or_resume_current_hart(struct target *target,
 
 	if (riscv_reg_flush_all(target) != ERROR_OK)
 		return ERROR_FAIL;
+
+	riscv_reg_cache_invalidate_all(target);
 
 	dm013_info_t *dm = get_dm(target);
 	/* Issue the resume command, and then wait for the current hart to resume. */
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -290,7 +290,6 @@ static const virt2phys_info_t sv57x4 = {
 
 static enum riscv_halt_reason riscv_halt_reason(struct target *target);
 static void riscv_info_init(struct target *target, struct riscv_info *r);
-static void riscv_invalidate_register_cache(struct target *target);
 static int riscv_step_rtos_hart(struct target *target);
 
 static void riscv_sample_buf_maybe_add_timestamp(struct target *target, bool before)
@@ -2485,10 +2484,15 @@ static int riscv_halt_go_all_harts(struct target *target)
 				return ERROR_FAIL;
 		}
 	} else {
+		// Safety check:
+		if (riscv_reg_cache_any_dirty(target, LOG_LVL_ERROR))
+			LOG_TARGET_INFO(target, "BUG: Registers should not be dirty while "
+					"the target is not halted!");
+
+		riscv_reg_cache_invalidate_all(target);
+
 		if (r->halt_go(target) != ERROR_OK)
 			return ERROR_FAIL;
-
-		riscv_invalidate_register_cache(target);
 	}
 
 	return ERROR_OK;
@@ -2572,7 +2576,11 @@ static int riscv_assert_reset(struct target *target)
 	struct target_type *tt = get_target_type(target);
 	if (!tt)
 		return ERROR_FAIL;
-	riscv_invalidate_register_cache(target);
+
+	if (riscv_reg_cache_any_dirty(target, LOG_LVL_INFO))
+		LOG_TARGET_INFO(target, "Discarding values of dirty registers.");
+
+	riscv_reg_cache_invalidate_all(target);
 	return tt->assert_reset(target);
 }
 
@@ -2699,7 +2707,15 @@ static int resume_go(struct target *target, int current,
 static int resume_finish(struct target *target, int debug_execution)
 {
 	assert(target->state == TARGET_HALTED);
-	register_cache_invalidate(target->reg_cache);
+	if (riscv_reg_cache_any_dirty(target, LOG_LVL_ERROR)) {
+		/* If this happens, it means there is a bug in the previous
+		 * register-flushing algorithm: not all registers were flushed
+		 * back to the target in preparation for the resume.*/
+		LOG_TARGET_ERROR(target,
+				"BUG: registers should have been flushed by this point.");
+	}
+
+	riscv_reg_cache_invalidate_all(target);
 
 	target->state = debug_execution ? TARGET_DEBUG_RUNNING : TARGET_RUNNING;
 	target->debug_reason = DBG_REASON_NOTHALTED;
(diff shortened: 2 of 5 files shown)


src/target/riscv/riscv_reg.h
```diff
@@ -21,6 +21,16 @@ void riscv_reg_free_all(struct target *target);
...
 /** Write all dirty registers to the target. */
 int riscv_reg_flush_all(struct target *target);
+/**
+ * Checks whether there are dirty registers, reprorting dirty registers usign
+ * the supplied "log_level".
+ */
+bool riscv_reg_any_dirty(const struct target *target, int log_level);
+/**
+ * Invalidate all registers.
+ * WRNING: If a register was dirty it's walue will be silently lost!
```
JanMatCodasip: Detail - typos:

```suggestion
 * WARNING: If a register was dirty, its value will be silently lost!
```

src/target/riscv/riscv_reg.h
```diff
@@ -21,6 +21,16 @@ void riscv_reg_free_all(struct target *target);
 
 /** Write all dirty registers to the target. */
 int riscv_reg_flush_all(struct target *target);
+/**
+ * Checks whether there are dirty registers, reprorting dirty registers usign
+ * the supplied "log_level".
+ */
```
JanMatCodasip: Detail - typo fix and a minor reword to improve the clarity:

```suggestion
/**
 * Check whether there are any dirty registers in the OpenOCD's register cache.
 * In addition, all dirty registers will be reported to the log using the supplied "log_level".
 */
```

src/target/riscv/riscv_reg.h
```diff
@@ -21,6 +21,16 @@ void riscv_reg_free_all(struct target *target);
...
+bool riscv_reg_any_dirty(const struct target *target, int log_level);
+/**
+ * Invalidate all registers.
```
JanMatCodasip: Detail: Minor reword for clarity.

I wanted to stress that this deals with register cache (not with register values in the target).

```suggestion
 * Invalidate all registers - forget their cached register values.
```

src/target/riscv/riscv-013.c
```diff
@@ -2728,6 +2728,12 @@ static int handle_became_unavailable(struct target *target,
 		enum riscv_hart_state previous_riscv_state)
 {
 	RISCV013_INFO(info);
+
+	if (riscv_reg_any_dirty(target, LOG_LVL_WARNING))
+		LOG_TARGET_WARNING(target, "Discarding values of dirty registers.");
```

```suggestion
		LOG_TARGET_WARNING(target, "Discarding values of dirty registers (due to target becoming unavailable).");
```

src/target/riscv/riscv_reg.c
```diff
@@ -888,6 +888,34 @@ static int riscv_set_or_write_register(struct target *target,
 	return ERROR_OK;
 }
 
+bool riscv_reg_any_dirty(const struct target *target, int log_level)
```
JanMatCodasip: Optionally, if you like, the function could be renamed to contain the word `cache`, to make it very clear to the reader what is going on.

```suggestion
bool riscv_reg_cache_any_dirty(const struct target *target, int log_level)
```

src/target/riscv/riscv_reg.c
```diff
@@ -888,6 +888,34 @@ static int riscv_set_or_write_register(struct target *target,
...
+		if (reg->dirty) {
+			log_printf_lf(log_level, __FILE__, __LINE__, __func__,
+					"[%s] Register %s is dirty!", target_name(target), reg->name);
+			any_dirty = true;
+		}
+	}
+	return any_dirty;
+}
+
+void riscv_reg_invalidate_all(struct target *target)
```
JanMatCodasip: (same as previous)

```suggestion
void riscv_reg_cache_invalidate_all(struct target *target)
```

src/target/riscv/riscv.c
```diff
@@ -4001,7 +4012,11 @@ static int riscv_openocd_step_impl(struct target *target, int current,
 		LOG_TARGET_ERROR(target, "Unable to step rtos hart.");
 	}
 
-	register_cache_invalidate(target->reg_cache);
+	if (riscv_reg_any_dirty(target, LOG_LVL_ERROR))
+		LOG_TARGET_ERROR(target,
+				"BUG: registers should have been flushed by this point.");
```
JanMatCodasip, line 4025: If you don't mind, I would make it very explicit to the code reader what this error means:

```suggestion
		// If this happens, it means there is a bug in the previous register-flushing algorithm:
		// not all registers were flushed back to the target prior to single-step.
		LOG_TARGET_ERROR(target,
				"BUG: registers should have been flushed by this point.");
```

src/target/riscv/riscv.c
```diff
@@ -2699,7 +2706,11 @@ static int resume_go(struct target *target, int current,
 static int resume_finish(struct target *target, int debug_execution)
 {
 	assert(target->state == TARGET_HALTED);
-	register_cache_invalidate(target->reg_cache);
+	if (riscv_reg_any_dirty(target, LOG_LVL_ERROR))
+		LOG_TARGET_ERROR(target,
```
JanMatCodasip, line 2714:
```suggestion
		// If this happens, it means there is a bug in the previous register-flushing algorithm:
		// not all registers were flushed back to the target prior to resume.
		LOG_TARGET_ERROR(target,
```

src/target/riscv/riscv-013.c
```diff
@@ -5435,6 +5441,8 @@ static int riscv013_step_or_resume_current_hart(struct target *target,
 	if (riscv_reg_flush_all(target) != ERROR_OK)
```
JanMatCodasip, line 5442: If you like:

```suggestion
	if (riscv_reg_cache_flush_all(target) != ERROR_OK)
```

src/target/riscv/riscv.c
```diff
@@ -2485,10 +2484,14 @@ static int riscv_halt_go_all_harts(struct target *target)
 				return ERROR_FAIL;
 		}
 	} else {
+		if (riscv_reg_any_dirty(target, LOG_LVL_ERROR))
```
```suggestion
		// Safety check:
		if (riscv_reg_any_dirty(target, LOG_LVL_ERROR))
```

JanMatCodasip (dismissed):
Overall, this looks all right to me. Thank you for the additional hardening of the code.

I am sending few optional suggestions about function naming and comments.

src/target/riscv/riscv-013.c
en-sc: I'd like to leave the `flush` as-is. The reason is: while invalidation and `_any_dirty` are only concerned with the cache and do not touch the target, "flush" does.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
en-sc: Addressed.

src/target/riscv/riscv.c
en-sc: Addressed.

src/target/riscv/riscv_reg.c
en-sc: Addressed.

src/target/riscv/riscv_reg.c
en-sc: Addressed.

src/target/riscv/riscv-013.c

src/target/riscv/riscv_reg.h
en-sc: Addressed.

src/target/riscv/riscv_reg.h
en-sc: Addressed.

src/target/riscv/riscv_reg.h
en-sc: Thanks! Addressed.

src/target/riscv/riscv.c
en-sc: Addressed. The wording used is a bit different. Please take a look.

src/target/riscv/riscv-013.c
JanMatCodasip: Ok, sure.

JanMatCodasip (approved):

PR #1187, opened 2024-12-20
target/riscv: access registers via `reg->type`

en-sc:
* `int riscv_reg_get()` and `int riscv_reg_set()` are implemented in
  terms of `reg->type->get/set` instead of the other way around. This
  makes it easier to support custom behavior for some registers.
* Cacheability is determined by `reg->type` instead of
  `riscv_reg_impl_gdb_regno_cacheable()`.
* Issues with redirection of `priv` -> `dcsr` and `pc` -> `dpc` are
  addressed at the "topmost" level.
    - `priv` and `pc` are always invalid.
    - Fixed some issues, e.g. the first `pc` write printed-out an
      uninitialized value:
```
> reg pc 0
pc (/64): 0x000075da6b33db20
```

### DIFF
--- a/src/target/riscv/program.h
+++ b/src/target/riscv/program.h
@@ -39,11 +39,10 @@ int riscv_program_init(struct riscv_program *p, struct target *t);
 /* Write the program to the program buffer. */
 int riscv_program_write(struct riscv_program *program);
 
-/* Executes a program, returning 0 if the program successfully executed.  Note
- * that this may cause registers to be saved or restored, which could result to
- * calls to things like riscv013_reg_save which itself could require a
- * program to execute.  That's OK, just make sure this eventually terminates.
- * */
+/* Executes the RISC-V Program Buffer and returns ERROR_OK if the program
+ * buffer got successfully executed. In case of failure, more detailed error reason
+ * can be found in p->execution_result.
+ */
 int riscv_program_exec(struct riscv_program *p, struct target *t);
 
 /* A lower level interface, you shouldn't use this unless you have a reason. */
--- a/src/target/riscv/riscv-013_reg.c
+++ b/src/target/riscv/riscv-013_reg.c
@@ -13,73 +13,455 @@
 #include "debug_defines.h"
 #include <helper/time_support.h>
 
-static int riscv013_reg_get(struct reg *reg)
+/* Start of register fetch/send definitions.
+ * - "reg_fetcher" (named "fetch_*") -- loads the value from target into reg
+ *   cache entry. Modifies only "reg->value".
+ * - "reg_sender" (named "send_*") -- stores the value from "buf" to the
+ *   target. Does not modify "reg" at all.
+ */
+typedef int (*reg_fetcher)(struct reg *reg);
+typedef int (*reg_sender)(const struct reg *reg, const uint8_t *buf);
+
+/* TODO: Introduce separate "fetch_*" and "send_*" functions for register
+ * classes (xregs, fregs, csrs).
+ */
+static int fetch_riscv013_reg(struct reg *reg)
 {
-	struct target *target = riscv_reg_impl_get_target(reg);
+	riscv_reg_t value;
+	int res = riscv013_get_register(riscv_reg_impl_get_target(reg),
+			&value, reg->number);
+	if (res != ERROR_OK)
+		return res;
+	buf_set_u64(reg->value, 0, reg->size, value);
+	return res;
+}
 
-	/* TODO: Hack to deal with gdb that thinks these registers still exist. */
-	if (reg->number > GDB_REGNO_XPR15 && reg->number <= GDB_REGNO_XPR31 &&
-			riscv_supports_extension(target, 'E')) {
-		buf_set_u64(reg->value, 0, reg->size, 0);
-		return ERROR_OK;
+static int send_riscv013_reg(const struct reg *reg, const uint8_t *buf)
+{
+	assert(reg->size <= 64);
+	const riscv_reg_t value = buf_get_u64(buf, 0, reg->size);
+	return riscv013_set_register(riscv_reg_impl_get_target(reg),
+			reg->number, value);
+}
+
+/* TODO: Inline "riscv013_*_register_buf" into "vreg" accessors.
+ */
+static int fetch_vreg(struct reg *reg)
+{
+	return riscv013_get_register_buf(riscv_reg_impl_get_target(reg),
+			reg->value, reg->number);
+}
+
+static int send_vreg(const struct reg *reg, const uint8_t *buf)
+{
+	return riscv013_set_register_buf(riscv_reg_impl_get_target(reg),
+			reg->number, reg->value);
+}
+
+/* End of register fetch/send definitions. */
+/* Start of cache entry utils. */
+
+static void set_cache_value(struct reg *reg, const uint8_t *buf)
+{
+	assert(riscv_reg_impl_is_existing(reg));
+	buf_cpy(buf, reg->value, reg->size);
+}
+
+/**
+ * Note: There are only 3 reachable states for an item in the register cache:
+ * |: state   :|: "reg->valid" :|: "reg->dirty" :|
+ * | "invalid" | false          | false          |
+ * | "valid"   | true           | false          |
+ * | "dirty"   | true           | true           |
+ *
+ * The state "reg->valid == dirty && reg->valid == false" should be unreacheable.
+ * To achieve this, "reg->valid" and "reg->dirty" are not assigned directly
+ * outside of "mark_<state>()" functions.
+ */
+static void mark_invalid(struct reg *reg)
+{
+	assert(riscv_reg_impl_is_existing(reg));
+	reg->valid = false;
+	reg->dirty = false;
+}
+
+static void mark_clean(struct reg *reg)
+{
+	assert(riscv_reg_impl_is_existing(reg));
+	assert(riscv_reg_impl_get_target(reg)->state == TARGET_HALTED);
+	reg->valid = true;
+	reg->dirty = false;
+}
+
+static void mark_dirty(struct reg *reg)
+{
+	assert(riscv_reg_impl_is_existing(reg));
+	assert(riscv_reg_impl_get_target(reg)->state == TARGET_HALTED);
+	reg->valid = true;
+	reg->dirty = true;
+}
+
+static int access_prologue(const struct reg *reg, bool write)
+{
+	assert(riscv_reg_impl_is_initialized(reg));
+	const struct target * const target = riscv_reg_impl_get_target(reg);
+	if (!reg->exist) {
+		LOG_TARGET_DEBUG(target, "Register %s does not exist",
+				reg->name);
+		return ERROR_FAIL;
 	}
-
-	if (reg->number >= GDB_REGNO_V0 && reg->number <= GDB_REGNO_V31) {
-		if (riscv013_get_register_buf(target, reg->value, reg->number) != ERROR_OK)
-			return ERROR_FAIL;
-
-		reg->valid = riscv_reg_impl_gdb_regno_cacheable(reg->number, /* is write? */ false);
-	} else {
-		uint64_t value;
-		int result = riscv_reg_get(target, &value, reg->number);
-		if (result != ERROR_OK)
-			return result;
-		buf_set_u64(reg->value, 0, reg->size, value);
-	}
-	char *str = buf_to_hex_str(reg->value, reg->size);
-	LOG_TARGET_DEBUG(target, "Read 0x%s from %s (valid=%d).", str, reg->name,
-			reg->valid);
-	free(str);
+	LOG_TARGET_DEBUG(target, "%s %s (valid=%s, dirty=%s)",
+			write ? "Writing" : "Reading",
+			reg->name,
+			reg->valid ? "true" : "false",
+			reg->dirty ? "true" : "false");
 	return ERROR_OK;
 }
 
-static int riscv013_reg_set(struct reg *reg, uint8_t *buf)
+static void access_epilogue(const struct reg *reg, bool write)
 {
-	struct target *target = riscv_reg_impl_get_target(reg);
+	assert(riscv_reg_impl_is_existing(reg));
+	if (debug_level < LOG_LVL_DEBUG)
+		return;
+	char *value_hex = buf_to_hex_str(reg->value, reg->size);
+	LOG_TARGET_DEBUG(riscv_reg_impl_get_target(reg),
+			"%s 0x%s %s %s (valid=%s, dirty=%s)",
+			write ? "Wrote" : "Read",
+			value_hex,
+			write ? "to" : "from",
+			reg->name,
+			reg->valid ? "true" : "false",
+			reg->dirty ? "true" : "false");
+	free(value_hex);
+}
 
-	char *str = buf_to_hex_str(buf, reg->size);
-	LOG_TARGET_DEBUG(target, "Write 0x%s to %s (valid=%d).", str, reg->name,
-			reg->valid);
-	free(str);
-
-	/* TODO: Hack to deal with gdb that thinks these registers still exist. */
-	if (reg->number > GDB_REGNO_XPR15 && reg->number <= GDB_REGNO_XPR31 &&
-			riscv_supports_extension(target, 'E') &&
-			buf_get_u64(buf, 0, reg->size) == 0)
-		return ERROR_OK;
-
-	if (reg->number >= GDB_REGNO_V0 && reg->number <= GDB_REGNO_V31) {
-		if (riscv013_set_register_buf(target, reg->number, buf) != ERROR_OK)
-			return ERROR_FAIL;
-
-		memcpy(reg->value, buf, DIV_ROUND_UP(reg->size, 8));
-		reg->valid = riscv_reg_impl_gdb_regno_cacheable(reg->number, /* is write? */ true);
-	} else {
-		const riscv_reg_t value = buf_get_u64(buf, 0, reg->size);
-		if (riscv_reg_set(target, reg->number, value) != ERROR_OK)
-			return ERROR_FAIL;
-	}
-
+static int noncaching_get(struct reg *reg, reg_fetcher fetch_reg)
+{
+	assert(riscv_reg_impl_is_initialized(reg));
+	int res = access_prologue(reg, /*write*/ false);
+	if (res != ERROR_OK)
+		return res;
+	res = fetch_reg(reg);
+	mark_invalid(reg);
+	if (res != ERROR_OK)
+		return res;
+	access_epilogue(reg, /*write*/ false);
 	return ERROR_OK;
 }
+
+static int caching_get(struct reg *reg, reg_fetcher fetch_reg)
+{
+	assert(riscv_reg_impl_is_initialized(reg));
+	int res = access_prologue(reg, /*write*/ false);
+	if (res != ERROR_OK)
+		return res;
+	if (reg->valid) {
+		LOG_TARGET_DEBUG(riscv_reg_impl_get_target(reg),
+				"Reading %s from cache", reg->name);
+		return ERROR_OK;
+	}
+	res = fetch_reg(reg);
+	if (res != ERROR_OK) {
+		mark_invalid(reg);
+		return res;
+	}
+	if (riscv_reg_impl_get_target(reg)->state == TARGET_HALTED)
+		mark_clean(reg);
+	else
+		mark_invalid(reg);
+	access_epilogue(reg, /*write*/ false);
+	return ERROR_OK;
+}
+
+static int noncaching_set(struct reg *reg, const uint8_t *buf,
+		reg_sender send_reg)
+{
+	int res = access_prologue(reg, /*write*/ true);
+	if (res != ERROR_OK)
+		return res;
+	res = send_reg(reg, buf);
+	mark_invalid(reg);
+	if (res != ERROR_OK)
+		return res;
+	set_cache_value(reg, buf);
+	access_epilogue(reg, /*write*/ true);
+	return ERROR_OK;
+}
+
+static int caching_set(struct reg *reg, const uint8_t *buf,
+		reg_sender send_reg)
+{
+	assert(riscv_reg_impl_is_initialized(reg));
+	const struct target * const target = riscv_reg_impl_get_target(reg);
+	if (target->state != TARGET_HALTED) {
+		LOG_TARGET_DEBUG(target, "Not caching the write to %s", reg->name);
+		return noncaching_set(reg, buf, send_reg);
+	}
+	int res = access_prologue(reg, /*write*/ true);
+	if (res != ERROR_OK)
+		return res;
+	if (reg->valid && buf_eq(reg->value, buf, reg->size)) {
+		LOG_TARGET_DEBUG(riscv_reg_impl_get_target(reg),
+				"Writing the same value to %s", reg->name);
+	} else {
+		LOG_TARGET_DEBUG(target, "Caching the write to %s", reg->name);
+		set_cache_value(reg, buf);
+		mark_dirty(reg);
+	}
+	access_epilogue(reg, /*write*/ true);
+	return ERROR_OK;
+}
+
+static int flush_reg(struct reg *reg, reg_sender send_reg)
+{
+	if (!reg->dirty)
+		return ERROR_OK;
+	int res = access_prologue(reg, /*write*/ true);
+	if (res != ERROR_OK)
+		return res;
+	assert(reg->dirty);
+	assert(reg->valid);
+	const struct target * const target = riscv_reg_impl_get_target(reg);
+	if (target->state != TARGET_HALTED) {
+		LOG_TARGET_ERROR(target,
+				"BUG: register %s is dirty while the target is not halted",
+				reg->name);
+		return ERROR_TARGET_NOT_HALTED;
+	}
+	res = send_reg(reg, reg->value);
+	if (res != ERROR_OK) {
+		/* Register is not marked "invalid" here (like it's done on failure in
+		 * "caching_set") since it is "dirty", i.e. the most recent value
+		 * is currently in the cache. */
+		return res;
+	}
+	mark_clean(reg);
+	access_epilogue(reg, /*write*/ true);
+	return res;
+}
+
+/* End of cache entry utils. */
+/* Start of "reg_arch_type" method definitions. */
+
+static int riscv013_noncaching_get(struct reg *reg)
+{
+	return noncaching_get(reg, fetch_riscv013_reg);
+}
+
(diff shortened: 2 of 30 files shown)


en-sc:
Please note: this PR cherry-picks https://review.openocd.org/c/openocd/+/8070

src/target/target.c
```diff
@@ -4781,18 +4890,29 @@ static int target_jim_get_reg(Jim_Interp *interp, int argc,
...
 COMMAND_HANDLER(handle_set_reg_command)
 {
-	if (CMD_ARGC != 1)
+	if (CMD_ARGC < 1 || CMD_ARGC > 2)
+		return ERROR_COMMAND_SYNTAX_ERROR;
+
+	bool flush = false;
+	const char *force_arg = NULL;
+
+	if (CMD_ARGC == 3)
```
JanMatCodasip: There seems to be a typo.

Note that in upstream I have already corrected this. (https://review.openocd.org/c/openocd/+/8070)

```suggestion
	if (CMD_ARGC == 2)
```

JanMatCodasip (commented):
@en-sc Thank you for preparing this change.

I have not yet managed a proper review, but I have revived and rebased the underlying [upstream patch](https://review.openocd.org/c/openocd/+/8070).

My preference would be to have that upstream patch merged first, then import the change to riscv-openocd via a periodic merge from upstream, and then finally merge this PR. However, it is difficult to estimate when/if the upstream patch gets a review. 

To anyone from the riscv-openocd audience: If you could help with the [review in OpenOCD upstream](https://review.openocd.org/c/openocd/+/8070), that would be very appreciated!

src/target/target.c
en-sc: You are right, thanks!

src/target/riscv/riscv-013_reg.c
```diff
@@ -358,30 +737,21 @@ int riscv013_reg_examine_all(struct target *target)
  */
 int riscv013_reg_save(struct target *target, enum gdb_regno regid)
```
JanMatCodasip: If you like, this could be named `riscv013_reg_save_gpr` to make it very clear that the function operates with GPRs only.

```suggestion
int riscv013_reg_save_gpr(struct target *target, enum gdb_regno regid)
```

The comment on top of the function could also be updated to reflect that:

```
/**
 * This function is used to save the value of a general purpose register (GPR)
 * into the cache. The register is marked as dirty, and writeback is delayed
 * for as long as possible.
 */
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -358,30 +737,21 @@ int riscv013_reg_examine_all(struct target *target)
...
-
 	LOG_TARGET_DEBUG(target, "Saving %s", reg->name);
-	if (riscv_reg_get(target, &value, regid) != ERROR_OK)
-		return ERROR_FAIL;
+	int res = reg->type->get(reg);
+	if (res != ERROR_OK)
+		return res;
 
 	assert(reg->valid &&
 			"The register is cacheable, so the cache entry must be valid now.");
```
JanMatCodasip, line 749: Optional:

The comment below could be enhanced to make it clearer to the reader what it means "dirty".

```
	/* Mark the register as dirty so that its value is written back to the target
	 * during the next register cache flush. We assume that this function is called
	 * because the caller is about to mess with the underlying value of the
	 * register, and wants to make sure the value gets restored later. */
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
+static void set_cache_value(struct reg *reg, const uint8_t *buf)
+{
+	assert(reg);
+	assert(reg->value);
+	assert(reg->size);
+	buf_cpy(buf, reg->value, reg->size);
+}
+
+/**
+ * Note: there are only 3 states a cahce entry can be in:
```
JanMatCodasip: I like the three functions and the comments that clearly describes the cache states.

Details:

1) Typo: cache

2) Very minor reword

```suggestion
 * Note: There are only 3 valid states for an item in the register cache:
```

src/target/riscv/riscv_reg.c
```diff
@@ -795,97 +782,17 @@ int riscv_reg_flush_all(struct target *target)
...
-				reg->name, value, cacheable ? "true" : "false",
-				reg->valid ? "true" : "false",
-				reg->dirty ? "true" : "false");
-		if (riscv013_set_register(target, regid, value) != ERROR_OK)
-			return ERROR_FAIL;
-
-		reg->dirty = false;
-	} else {
-		reg->dirty = need_to_write;
+		if (reg->type->flush(reg) != ERROR_OK)
```
JanMatCodasip, line 780: If you like:

```suggestion
		assert(reg->type->flush);
		if (reg->type->flush(reg) != ERROR_OK)
```

src/target/riscv/riscv_reg.c
```diff
@@ -795,97 +782,17 @@ int riscv_reg_flush_all(struct target *target)
...
-	LOG_TARGET_DEBUG(target,
-			"Wrote 0x%" PRIx64 " to %s (cacheable=%s, valid=%s, dirty=%s)",
-			value, reg->name, cacheable ? "true" : "false",
-			reg->valid ? "true" : "false",
-			reg->dirty ? "true" : "false");
-	return ERROR_OK;
+	if (res == ERROR_OK)
+		LOG_TARGET_DEBUG(target, "Flush of register cache completed successfully.");
+	else
+		LOG_TARGET_DEBUG(target, "Flush of register cache failed.");
```
JanMatCodasip: 1) Should this be an error instead of debug print?

2) It turns out that the OpenOCD upstream code style does not like punctuation at the end of log messages. I have created an easy issue to fix all the messages in `src/target/riscv`: https://github.com/riscv-collab/riscv-openocd/issues/1212

```suggestion
		LOG_TARGET_ERROR(target, "Flush of register cache failed");
```

src/target/riscv/riscv_reg.c
JanMatCodasip, line 780: If you like:

```suggestion
		assert(reg->type->flush);
		if (reg->type->flush(reg) != ERROR_OK)
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
 #include "debug_defines.h"
 #include <helper/time_support.h>
 
-static int riscv013_reg_get(struct reg *reg)
+/* Start of register fetch/send definitions.
+ * - "reg_fetcher" (named "fetch_*") -- loads the value from target into reg
+ *   cache entry. Modifyes only "reg->value".
+ * - "reg_sender" (named "send_*") -- stores the value from "buf" to the
```
JanMatCodasip: I'd recommend to use the name "setter" or "writter" instead of "sender". It sounds more natural in this context.

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
-		uint64_t value;
-		int result = riscv_reg_get(target, &value, reg->number);
-		if (result != ERROR_OK)
-			return result;
-		buf_set_u64(reg->value, 0, reg->size, value);
+static int send_vreg(const struct reg *reg, const uint8_t *buf)
+{
+	return riscv013_set_register_buf(riscv_reg_impl_get_target(reg),
+			reg->number, reg->value);
+}
```
JanMatCodasip: Would it make sense to put assertions here that the `reg` is one of the vector registers?

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
+
+/* End of register fetch/send definitions. */
+/* Start of cache entry utils. */
+
+static void set_cache_value(struct reg *reg, const uint8_t *buf)
+{
+	assert(reg);
+	assert(reg->value);
+	assert(reg->size);
```
JanMatCodasip: Could the check for an existing register be improved like this? Would it make sense to you?

```
static inline bool riscv_reg_impl_is_existing(const struct reg *reg)
{
	assert(riscv_reg_impl_is_initialized(reg));
	assert(reg->exist);
	assert(reg->size > 0);
	assert(reg->value);
}
```

```suggestion
	assert(riscv_reg_impl_is_existing(reg));
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
+static void mark_dirty(struct reg *reg)
+{
+	assert(riscv_reg_impl_is_initialized(reg));
+	assert(reg->exist);
+	assert(riscv_reg_impl_get_target(reg)->state == TARGET_HALTED);
+	reg->valid = true;
+	reg->dirty = true;
+}
+
+static int access_prelude(const struct reg *reg, bool write)
```
JanMatCodasip: Pedantic detail: Since the other function is called `epilogue`, this one can be called `prologue` (for naming symmetry).

- prelude vs. postlude
- prologue vs. epilogue

```suggestion
static int access_prologue(const struct reg *reg, bool write)
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
-static int riscv013_reg_set(struct reg *reg, uint8_t *buf)
+static void access_epilogue(const struct reg *reg, bool write)
 {
-	struct target *target = riscv_reg_impl_get_target(reg);
-
-	char *str = buf_to_hex_str(buf, reg->size);
-	LOG_TARGET_DEBUG(target, "Write 0x%s to %s (valid=%d).", str, reg->name,
-			reg->valid);
+	assert(riscv_reg_impl_is_initialized(reg));
+	assert(reg->exist);
```
JanMatCodasip: The predicate `riscv_reg_impl_is_existing()`, suggested in another comment, could perhaps be used here, too:

```suggestion
	assert(riscv_reg_impl_is_existing(reg));
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
+}
+
+static int riscv013_reg_flush(struct reg *reg)
+{
+	return flush_reg(reg, send_riscv013_reg);
+}
+
+static int flush_uncacheable(struct reg *reg)
+{
+	assert(!reg->dirty);
```
JanMatCodasip: If you like:

```suggestion
	/* Non-cacheable register must never get dirty */
	assert(!reg->dirty);
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
+{
+	return caching_set(reg, buf, send_riscv013_reg);
+}
+
+static int riscv013_reg_flush(struct reg *reg)
+{
+	return flush_reg(reg, send_riscv013_reg);
+}
+
+static int flush_uncacheable(struct reg *reg)
```
JanMatCodasip: Detail: Saying "non-cacheable" is more common than "uncacheable".

Please, if you could update also other occurrences, if any.

```suggestion
static int flush_noncacheable(struct reg *reg)
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
+	case GDB_REGNO_PC:
+		return &pc_type;
+	case GDB_REGNO_PRIV:
+		return &priv_type;
+	case GDB_REGNO_VLENB:
+	/* "vlenb" is read-only, but there is nothig wrong with attempting to
+	 * write it (e.g. for the purpose of HW testing). */
+	case GDB_REGNO_DPC:
+	case GDB_REGNO_VSTART:
+	case GDB_REGNO_VXSAT:
```
JanMatCodasip, line 444: Could you please double-check if DPC should have the same handlers as PC?

IIUC, they are aliases of each other, so they should be handled the same way (they should be symmetrical).

```suggestion
	case GDB_REGNO_ZERO:
		return &zero_type;
	case GDB_REGNO_PC:
	case GDB_REGNO_DPC:
		return &pc_type;
	case GDB_REGNO_PRIV:
		return &priv_type;
	case GDB_REGNO_VLENB:
	/* "vlenb" is read-only, but there is nothig wrong with attempting to
	 * write it (e.g. for the purpose of HW testing). */
	case GDB_REGNO_VSTART:
	case GDB_REGNO_VXSAT:
```

The handlers would then look something like this:

```c
static int set_pc(struct reg *pc_or_dpc, uint8_t *buf)
{
	assert(pc_or_dpc->number == GDB_REGNO_PC || pc_or_dpc->number == GDB_REGNO_DPC);

	struct target *target = riscv_reg_impl_get_target(pc_or_dpc);
	assert(target);

	/* Write the value to DPC */
	struct reg * const reg_dpc =
		riscv_reg_impl_cache_entry(target, GDB_REGNO_DPC);
	assert(!reg_dpc->valid);
	assert(!reg_dpc->dirty);

	int res = dpc->type->set(reg_dpc, buf);
	if (res == ERROR_OK) {
		/* Keep the "pc" item in sync, because it is an alias of "dpc". */
		struct reg * const reg_pc =
			riscv_reg_impl_cache_entry(target, GDB_REGNO_PC);
		assert(!reg_pc->valid);
		assert(!reg_pc->dirty);
		set_cache_value(reg_pc, dpc->value);
	}

	return res;
}
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
+{
+	assert(pc->number == GDB_REGNO_PC);
+	assert(!pc->valid);
+	assert(!pc->dirty);
+	struct reg * const dpc =
+		riscv_reg_impl_cache_entry(riscv_reg_impl_get_target(pc),
+				GDB_REGNO_DPC);
+	int res = dpc->type->get(dpc);
+	if (res == ERROR_OK)
+		set_cache_value(pc, dpc->value);
```
JanMatCodasip: Did I get the meaning of this operation right?

```suggestion
	if (res == ERROR_OK)
		/* There are two aliases for DPC in the register cache - "pc" and "dpc".
		 * Keep them in sync. */
		set_cache_value(pc, dpc->value);
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
+	assert(!pc->valid);
+	assert(!pc->dirty);
+	struct reg * const dpc =
+		riscv_reg_impl_cache_entry(riscv_reg_impl_get_target(pc),
+			GDB_REGNO_DPC);
+	int res = dpc->type->set(dpc, buf);
+	if (res == ERROR_OK)
+		set_cache_value(pc, dpc->value);
+	return res;
+}
```
JanMatCodasip: Should get_pc() and set_pc() make calls to access_prologue() and access_epilogue(), too? Otherwise the debug prints about the register access won't be in the logs, right?

JanMatCodasip (commented):
Note: In the review I focused on the second commit only: `target/riscv: access registers via reg->type`. Any comments to the other underlying upstream commit should be posted directly [to the upstream review](https://review.openocd.org/c/openocd/+/8070).

Overall, I like this change very much. I have some concerns over get_pc() and set_pc(), and then trivial suggestions about comments and naming. Other than that, the code looks all right - thank you.

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,452 @@
...
+	case GDB_REGNO_VXSAT:
+	case GDB_REGNO_VXRM:
+	case GDB_REGNO_VL:
+	case GDB_REGNO_VTYPE:
+	case GDB_REGNO_MISA:
+	case GDB_REGNO_DCSR:
+	case GDB_REGNO_DSCRATCH0:
+	case GDB_REGNO_MSTATUS:
+	case GDB_REGNO_MEPC:
+	case GDB_REGNO_MCAUSE:
```
en-sc: Shouldn't cache `mstatus` and `mcause` for now. See #1217, #1219

aap-sc:
@en-sc minor comment regarding commit description:

```
priv and pc are alvais invalid.
```

alvais -> always :)

src/target/riscv/riscv-013_reg.c
en-sc: I'd like to avoid the repetition: there are already `reg->set` and `riscv_reg_write`.

src/target/riscv/riscv-013_reg.c
en-sc: `riscv013_get_register_buf()` already has such assertion. I intend on inlining it here in a follow-up commit.

src/target/riscv/riscv-013_reg.c
en-sc: That's one way to look at it. From my point of view, `pc` is just an alias and the access actually happens through `dpc` (a bit similar to `virt` and `dcsr`). Therefore `pc` is never cached and always redirects to `dpc`. Please see `get_pc()/set_pc()`. IMHO it's a bit cleaner.

src/target/target.c
en-sc: Addressed (Updated the cherry-pick to the latest patch https://review.openocd.org/c/openocd/+/8070/21)

src/target/riscv/riscv-013_reg.c
en-sc: Addressed

src/target/riscv/riscv-013_reg.c
en-sc: Addressed

src/target/riscv/riscv-013_reg.c
en-sc: Addressed. Although replaced `valid states` with `reachable states` since `valid` is one of the states.

src/target/riscv/riscv_reg.c
en-sc: Addressed.

src/target/riscv/riscv_reg.c
en-sc: (1) Addressed. Thanks for the catch!
(2) Looked through the changes -- should be fixed.

src/target/riscv/riscv_reg.c
en-sc: Addressed

src/target/riscv/riscv-013_reg.c
en-sc: Great suggestion! Thanks! Addressed

src/target/riscv/riscv-013_reg.c
en-sc: Addressed. Thanks!

src/target/riscv/riscv-013_reg.c
en-sc: Addressed

src/target/riscv/riscv-013_reg.c
en-sc: Addressed. Although I've moved the message into the assertion.

src/target/riscv/riscv-013_reg.c
en-sc: Addressed

src/target/riscv/riscv-013_reg.c
en-sc: Yeah, I'll add the comment.

src/target/riscv/riscv-013_reg.c
en-sc: I was going for as thin of a wrapper as possible. IMHO it's not a big deal since what is actually happening is an access to `pc`/`dcsr`.

src/target/riscv/riscv-013_reg.c
en-sc: Addressed

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,451 @@
...
+		return;
 	char *str = buf_to_hex_str(reg->value, reg->size);
-	LOG_TARGET_DEBUG(target, "Read 0x%s from %s (valid=%d).", str, reg->name,
-			reg->valid);
+	LOG_TARGET_DEBUG(riscv_reg_impl_get_target(reg),
+			"%s 0x%s %s %s (valid=%s, dirty=%s)",
+			write ? "Wrote" : "Read", str,
+			write ? "to" : "from", reg->name,
+			reg->valid ? "true" : "false",
+			reg->dirty ? "true" : "false");
```
JanMatCodasip: Detail for readability: 

1) It helps if each of the arguments is on a separate line.

2) `str` --> `value_hex` (or similar)

```
LOG_TARGET_DEBUG(riscv_reg_impl_get_target(reg),
			"%s 0x%s %s %s (valid=%s, dirty=%s)",
			write ? "Wrote" : "Read",
			value_hex,
			write ? "to" : "from",
			reg->name,
			reg->valid ? "true" : "false",
			reg->dirty ? "true" : "false");
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,451 @@
...
+	const struct target * const target = riscv_reg_impl_get_target(reg);
+	if (!reg->exist) {
+		LOG_TARGET_DEBUG(target, "Register %s does not exist",
+				reg->name);
+		return ERROR_FAIL;
 	}
+	LOG_TARGET_DEBUG(target, "%s %s (valid=%s, dirty=%s)",
+			write ? "Writing" : "Reading", reg->name,
+			reg->valid ? "true" : "false",
+			reg->dirty ? "true" : "false");
```
JanMatCodasip: Same as in another comment - it helps if each of the arguments is on its own line:

```
LOG_TARGET_DEBUG(target, "%s %s (valid=%s, dirty=%s)",
			write ? "Writing" : "Reading",
			reg->name,
			reg->valid ? "true" : "false",
			reg->dirty ? "true" : "false");
```

src/target/riscv/riscv-013_reg.c
JanMatCodasip: Ok, understood, thank you for explanation. Let's keep your solution.

Perhaps the only little downside is that when the user types the `reg` command, the value of `pc` will not be displayed in the register list (even though it is known). I personally don't mind that.

src/target/riscv/riscv_reg.c
```diff
@@ -795,97 +783,19 @@ int riscv_reg_flush_all(struct target *target)
...
-				reg->valid ? "true" : "false",
-				reg->dirty ? "true" : "false");
-		if (riscv013_set_register(target, regid, value) != ERROR_OK)
-			return ERROR_FAIL;
-
-		reg->dirty = false;
-	} else {
-		reg->dirty = need_to_write;
+		assert(reg->type->flush
+				&& "RISC-V targets must define the 'flush' method");
```
JanMatCodasip: Minor wording detail - if you like:

```
assert(reg->type->flush
		&& "RISC-V targets must define the 'flush' method on all registers");

// or

assert(reg->type->flush
		&& "All RISC-V registers must have the 'flush' method defined");

// or ...
```

src/target/riscv/riscv-013_reg.c
JanMatCodasip: Ok, let's keep your names.

src/target/riscv/riscv-013_reg.c
JanMatCodasip: Ok, understood, thanks for explanation.

JanMatCodasip (commented):
Due to the size (and importance) of this PR, I would like to go through it once more.

Anyway, it looks very good overall. Also thanks for addressing my suggestions so far.

src/target/riscv/riscv-013_reg.c
en-sc: Addressed

src/target/riscv/riscv-013_reg.c
en-sc: Addressed

src/target/riscv/riscv_reg.c
en-sc: Addressed

src/target/riscv/program.h
```diff
@@ -41,7 +41,7 @@ int riscv_program_write(struct riscv_program *program);
 
 /* Executes a program, returning 0 if the program successfully executed.  Note
  * that this may cause registers to be saved or restored, which could result to
- * calls to things like riscv013_reg_save which itself could require a
+ * calls to things like riscv013_reg_save_gpr which itself could require a
  * program to execute.  That's OK, just make sure this eventually terminates.
  * */
```
JanMatCodasip: (This is a detail, but helpful to code readers.)

This comment is outdated - `riscv_program_exec()` no longer attempts to save and restore registers.

Furthermore the error reporting has been improved.

The comment can be adjusted to account for that:

```
/* Executes the RISC-V Program Buffer and returns ERROR_OK if the program 
 * buffer got successfully executed. In case of failure, more detailed error reason
 * can be found in p->execution_result.
 */
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,454 @@
 #include "debug_defines.h"
 #include <helper/time_support.h>
 
-static int riscv013_reg_get(struct reg *reg)
+/* Start of register fetch/send definitions.
+ * - "reg_fetcher" (named "fetch_*") -- loads the value from target into reg
+ *   cache entry. Modifyes only "reg->value".
```
JanMatCodasip: Pedantic detail - typo

```suggestion
 *   cache entry. Modifies only "reg->value".
```

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,454 @@
...
+	case GDB_REGNO_MISA:
+	case GDB_REGNO_DCSR:
+	case GDB_REGNO_DSCRATCH0:
+	case GDB_REGNO_MEPC:
+	case GDB_REGNO_SATP:
+		return &warl_csr_type;
+	/* TODO: https://github.com/riscv-collab/riscv-openocd/issues/1219 */
+	case GDB_REGNO_MSTATUS:
+	case GDB_REGNO_MCAUSE:
+	/* TODO: Sdtrig register should be WARL. */
```
JanMatCodasip: IIUC, any cached values of `tdata*` and `tinfo` must get invalidated when `tselect` changes.

This can be mentioned in the TODO:

```suggestion
	/* TODO: Sdtrig register should be WARL. 
	 * Also cached values of "tdata*" and "tinfo" must get invalidated when tselect is changed. */
```

JanMatCodasip (commented):
Overall, this change looks fine.

I have posted few last minor comments and do not expect to have any more.

Also there is a merge conflict at the moment.

src/target/riscv/riscv-013_reg.c

src/target/riscv/riscv-013_reg.c

src/target/riscv/program.h
en-sc: Thanks! Addressed.

JanMatCodasip (approved):

en-sc:
@JanMatCodasip, I've closed this one by mistake, reopened as #1269.

PR #1190, opened 2024-12-24
target/riscv: move the dcsr modification out of program buffer

### LINKED ISSUE #1188: How to enable hardware translation when progbufsize too small
Hardware translation requires a program buffer size of at least 4. Can we use a `register_write_direct` function instead of two instructions `csrrsi zero, dcsr, dcsr_mprven` and `csrrci zero, dcsr, dcsr_mprven` ?
https://github.com/riscv-collab/riscv-openocd/blob/f82c5a7c048eb70fdc4dff6f53002fa1d3a1bda5/src/target/riscv/riscv-013.c#L2008-L2013

https://github.com/riscv-collab/riscv-openocd/blob/f82c5a7c048eb70fdc4dff6f53002fa1d3a1bda5/src/target/riscv/riscv-013.c#L3120-L3126

https://github.com/riscv-collab/riscv-openocd/blob/f82c5a7c048eb70fdc4dff6f53002fa1d3a1bda5/src/target/riscv/riscv-013.c#L4191-L4206

lz-bro:
When riscv virt2phys_mode is hw, reduce the use of unnecessary program buffers. See #1188  for the original issue.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2005,13 +2005,6 @@ static int examine(struct target *target)
 				info->impebreak);
 	}
 
-	if (info->progbufsize < 4 && riscv_virt2phys_mode_is_hw(target)) {
-		LOG_TARGET_ERROR(target, "software address translation "
-				"is not available on this target. It requires a "
-				"program buffer size of at least 4. (progbufsize=%d) "
-				"Use `riscv set_enable_virtual off` to continue.", info->progbufsize);
-	}
-
 	/* Don't call any riscv_* functions until after we've counted the number of
 	 * cores and initialized registers. */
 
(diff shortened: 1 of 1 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -3146,10 +3130,33 @@ static int modify_privilege(struct target *target, uint64_t *mstatus, uint64_t *
...
+	if (mstatus != mstatus_old &&
+		    register_write_direct(target, GDB_REGNO_MSTATUS, mstatus_old) != ERROR_OK)
+		return ERROR_FAIL;
+
+	/* Read DCSR */
+	riscv_reg_t dcsr;
+	if (register_read_direct(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
+		return ERROR_FAIL;
+
+	/* dcsr.mprven <- 0 */
```
aap-sc: why do you set the filed to `zero` unconditionally ?  should not we preserve the original value?

lz-bro: `riscv_program_load_mprv()` set the `dcsr.mprven` to zero unconditionally, and I followed its logic. If I set `riscv virt2phys_mode` back to sw, `dcsr.mprven` should set to zero?

src/target/riscv/riscv-013.c
```diff
@@ -3146,10 +3130,33 @@ static int modify_privilege(struct target *target, uint64_t *mstatus, uint64_t *
...
+	return register_write_direct(target, GDB_REGNO_DCSR,
```
en-sc: Please, use the operations that work with the cache (`riscv_reg_get/riscv_reg_set`).
There is no need to bypass the cache here.
Moreover, current implementation may introduce cache inconsistency. Consider the following:
```
> reg dcsr <some value with "mprven" set>
> reg dcsr # After this the "dcsr" value with "mprven" set is cached.
<some value with "mprven" set>
> mww <some addr> #After this "dcsr" on target has "mprven" cleared.
> reg dcsr
<some value with "mprven" set>
> reg dcsr -force
<some value with "mprven" cleared>
```
This is not a big deal since IMHO it's okay for the debugger to change "dcsr" however it likes but this is a peculiar behavior that may be easily avoided by using `riscv_reg_get/riscv_reg_set`.

As you have mentioned in #1188 `register_read/write_direct()` are used throughout `riscv-013.c`. However, most of the time this is not necessary and may introduce cache inconsistency. So I plan to phase out  `register_read/write_direct()`.

src/target/riscv/riscv-013.c
```diff
@@ -3146,10 +3130,33 @@ static int modify_privilege(struct target *target, uint64_t *mstatus, uint64_t *
...
 
-	return register_write_direct(target, GDB_REGNO_MSTATUS, *mstatus);
+	/* Restore MSTATUS */
+	if (mstatus != mstatus_old &&
+		    register_write_direct(target, GDB_REGNO_MSTATUS, mstatus_old) != ERROR_OK)
+		return ERROR_FAIL;
+
+	/* Read DCSR */
+	riscv_reg_t dcsr;
+	if (register_read_direct(target, &dcsr, GDB_REGNO_DCSR) != ERROR_OK)
```
en-sc: Please, consider adding `dcsr_old` (similar to `mstatus_old`) to avoid an extra read.

src/target/riscv/riscv-013.c
```diff
@@ -3146,10 +3130,33 @@ static int modify_privilege(struct target *target, uint64_t *mstatus, uint64_t *
 	*mstatus = set_field(*mstatus, MSTATUS_MPRV, 1);
 
 	/* Write MSTATUS */
-	if (*mstatus == *mstatus_old)
+	if (*mstatus != *mstatus_old &&
+			register_write_direct(target, GDB_REGNO_MSTATUS, *mstatus) != ERROR_OK)
```
en-sc: Consider also using `register_read/write_direct()` in `modify_privilege()`.
The issue is: what if the write of `mstatus` succeeds but the write of `dcsr` fails? This will change the value of `mstatus` on the target without invalidating the cached value.

en-sc (changes requested):
@lz-bro, thank you for the patch!

Please take a look at my suggestions.

src/target/riscv/riscv-013.c
en-sc: > should not we preserve the original value?

Strictly speaking, I don't think we should -- `dcsr` value is managed by the debugger exclusively. Though either approach should work just fine.

src/target/riscv/riscv-013.c
lz-bro: address

src/target/riscv/riscv-013.c
lz-bro: address

src/target/riscv/riscv-013.c
lz-bro: address

src/target/riscv/riscv-013.c
```diff
@@ -4365,17 +4358,16 @@ read_memory_progbuf(struct target *target, const riscv_mem_access_args_t args)
 
 	uint64_t mstatus = 0;
 	uint64_t mstatus_old = 0;
-	if (modify_privilege(target, &mstatus, &mstatus_old) != ERROR_OK)
+	uint64_t dcsr = 0;
```
en-sc: Please, use `riscv_reg_t` instead of `uint64_t` for `mstatus` and `dcsr` values.

en-sc (dismissed):
LGTM.

Only a minor nitpick.

src/target/riscv/riscv-013.c
lz-bro: address

en-sc (dismissed):

en-sc (dismissed):
Oh, great catch!

LGTM

src/target/riscv/riscv-013.c
```diff
@@ -3109,47 +3102,69 @@ static int read_sbcs_nonbusy(struct target *target, uint32_t *sbcs)
 }
 
 /* TODO: return mem_access_result_t */
-static int modify_privilege(struct target *target, uint64_t *mstatus, uint64_t *mstatus_old)
+static int modify_privilege(struct target *target, riscv_reg_t *mstatus, riscv_reg_t *mstatus_old,
```
```suggestion
static int modify_privilege_for_virt2phys_mode(struct target *target, riscv_reg_t *mstatus, riscv_reg_t *mstatus_old,
```

Up to your consideration, I recommend renaming this function to reflect what it does more. `modify_privilege()` feels too generic to me. Or maybe you have a better idea for a name.

src/target/riscv/riscv-013.c
```diff
@@ -3109,47 +3102,69 @@ static int read_sbcs_nonbusy(struct target *target, uint32_t *sbcs)
...
+
+	/* Write DCSR */
+	if (*dcsr != *dcsr_old &&
+			riscv_reg_set(target, GDB_REGNO_DCSR, *dcsr) != ERROR_OK)
+		return ERROR_FAIL;
+
+	return ERROR_OK;
+}
+
+static int restore_privilege(struct target *target, riscv_reg_t mstatus, riscv_reg_t mstatus_old,
```
```suggestion
static int restore_privilege_from_virt2phys_mode(struct target *target, riscv_reg_t mstatus, riscv_reg_t mstatus_old,
```

If you change the name of `modify_privilege()`, then change it here as well.

MarekVCodasip (dismissed):
LGTM, I do have a suggestion, up to you.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

MarekVCodasip: I feel like `restore_privilege_for_virt2phys_mode()` sounds wrong as we are restoring it from it. I would either remove the `for` word or replace it with `from`.

Thanks

MarekVCodasip (dismissed):
I still have one more thing left.

src/target/riscv/riscv-013.c

en-sc (approved):

MarekVCodasip (approved):

PR #1193, opened 2024-12-25
target/riscv: clear `abstract_cmd_maybe_busy` after commands

en-sc:
If a sufficient delay was used before reading `abstractcs` during the batch execution, `dm->abstract_cmd_maybe_busy` was not cleared and the following call to `wait_for_idle_if_needed()` (e.g. on `resume`), would result in a call to `wait_for_idle()` performing a redundant read of `abstractcs`.
While this is not a bug, it impedes the performance.

Change-Id: I9d234ef6d53af96c60892d71247c10e631dfcc3b

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -630,6 +630,15 @@ static int abstract_cmd_batch_check_and_clear_cmderr(struct target *target,
 		if (res != ERROR_OK)
 			goto clear_cmderr;
 	}
+
+	dm013_info_t * const dm = get_dm(target);
+	if (!dm) {
+		LOG_ERROR("BUG: Target %s is not assigned to any RISC-V debug module",
+				target_name(target));
+		return ERROR_FAIL;
+	}
+	dm->abstract_cmd_maybe_busy = false;
+
 	*cmderr = get_field32(abstractcs, DM_ABSTRACTCS_CMDERR);
 	if (*cmderr == CMDERR_NONE)
 		return ERROR_OK;


src/target/riscv/riscv-013.c
```diff
@@ -630,6 +630,15 @@ static int abstract_cmd_batch_check_and_clear_cmderr(struct target *target,
 		if (res != ERROR_OK)
 			goto clear_cmderr;
 	}
+
+	dm013_info_t * const dm = get_dm(target);
+	if (!dm) {
+		LOG_ERROR("BUG: Target %s is not assigned to any RISC-V debug module",
+				target_name(target));
```
MarekVCodasip, line 637:
```suggestion
		LOG_TARGET_ERROR(target, "BUG: Target is not assigned to any RISC-V debug module");
```

MarekVCodasip (commented):
I have a small suggestion, but otherwise it looks fine.

MarekVCodasip (approved):

src/target/riscv/riscv-013.c
en-sc: I'd like to leave it as-is.

IMHO this is an error message and the resulting text should be as human readable as possible.

`[<target_name>]: BUG: Target` seems less readable then `BUG: Target <target_name>`.

Moreover, the same message is used in `wait_for_idle()`:
https://github.com/riscv-collab/riscv-openocd/blob/83dc8a6446ca8a8604dbd706373c1904d4004b97/src/target/riscv/riscv-013.c#L561-L563

MarekVCodasip: Alright. LGTM, thanks for doing this.

PR #1198, opened 2025-01-07
RISC-V Semihosting 1 of 3: Remove dead code

JanMatCodasip:
Variable `target->semihosting` is always initialized in riscv_init_target() to a non-NULL value.

For that reason, checks like `if (target->semihosting)` are redundant (dead code).

Remove them to not confuse code readers. Replace them by assertions.

Change-Id: I85ef52300e240cfcb0119db6169993bc4767de8f


JanMatCodasip:
Note: This MR is built on top of https://github.com/riscv-collab/riscv-openocd/pull/1198.

MarekVCodasip (approved):
LGTM

en-sc (approved):
LGTM

PR #1199, opened 2025-01-07
RISC-V Semihosting 2 of 3: Refactor magic sequence detection

JanMatCodasip:
Refactor (clean up) the code in riscv_semihosting.c by moving
the magic sequence detection to its own function.

### DIFF
--- a/src/target/riscv/riscv_semihosting.c
+++ b/src/target/riscv/riscv_semihosting.c
@@ -37,6 +37,57 @@
 static int riscv_semihosting_setup(struct target *target, int enable);
 static int riscv_semihosting_post_result(struct target *target);
 
+static int riscv_semihosting_detect_magic_sequence(struct target *target,
+	const target_addr_t pc, bool *sequence_found)
+{
+	assert(sequence_found);
+
+	/* The semihosting "magic" sequence must be the exact three instructions
+	 * listed below. All these instructions, including the ebreak, must be
+	 * uncompressed (4 bytes long). */
+	const uint32_t magic[] = {
+		0x01f01013,	/* slli    zero,zero,0x1f */
+		0x00100073,	/* ebreak */
+		0x40705013	/* srai    zero,zero,0x7 */
+	};
+
+	LOG_TARGET_DEBUG(target, "Checking for RISC-V semihosting sequence "
+		"at PC = 0x%" TARGET_PRIxADDR, pc);
+
+	/* Read three uncompressed instructions:
+	 * The previous, the current one (pointed to by PC) and the next one. */
+	const target_addr_t sequence_start_address = pc - 4;
+	for (int i = 0; i < 3; i++) {
+		uint8_t buf[4];
+
+		/* Instruction memories may not support arbitrary read size.
+		 * Use any size that will work. */
+		const target_addr_t address = sequence_start_address + (4 * i);
+		int result = riscv_read_by_any_size(target, address, 4, buf);
+		if (result != ERROR_OK) {
+			*sequence_found = false;
+			return result;
+		}
+
+		/* RISC-V instruction layout in memory is always little endian,
+		 * regardless of the endianness of the whole system. */
+		const uint32_t value = le_to_h_u32(buf);
+
+		LOG_TARGET_DEBUG(target, "compare 0x%08" PRIx32 " from 0x%" PRIx64 " against 0x%08" PRIx32,
+			value, address, magic[i]);
+		if (value != magic[i]) {
+			LOG_TARGET_DEBUG(target, "Not a RISC-V semihosting sequence");
+			*sequence_found = false;
+			return ERROR_OK;
+		}
+	}
+
+	LOG_TARGET_DEBUG(target, "RISC-V semihosting sequence found "
+		"at PC = 0x%" TARGET_PRIxADDR, pc);
+	*sequence_found = true;
+	return ERROR_OK;
+}
+
 /**
  * Initialize RISC-V semihosting. Use common ARM code.
  */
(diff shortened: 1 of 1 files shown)


JanMatCodasip:
Note: This MR is built on top of https://github.com/riscv-collab/riscv-openocd/pull/1198.

src/target/riscv/riscv_semihosting.c
```diff
@@ -72,33 +115,19 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
...
-			return SEMIHOSTING_ERROR;
-		uint32_t value = target_buffer_get_u32(target, buf);
-		LOG_TARGET_DEBUG(target, "compare 0x%08x from 0x%" PRIx64 " against 0x%08x",
-			value, address, magic[i]);
-		if (value != magic[i]) {
-			LOG_TARGET_DEBUG(target, "   -> NONE (no magic)");
-			return SEMIHOSTING_NONE;
-		}
+	if (!sequence_found) {
+		LOG_TARGET_DEBUG(target, "   -> NONE (no magic)");
```
aap-sc: maybe move this printout inside riscv_semihosting_detect_magic_sequence?

```
		if (result != ERROR_OK) {
			*sequence_found = false;
			return result;
		}
```

Having context-dependent printout like this is somewhat confusing.

aap-sc (changes requested):

src/target/riscv/riscv_semihosting.c
JanMatCodasip: Addressed, but differently.

My understanding is that the prints like `   -> NONE (no magic)` were supposed to print the final outcome of `riscv_semihosting()` function into the debug log for easier troubleshooting.

I agree that the prints looked confusing.

For consistency and clarity, I changed all of them to this form `Semihosting outcome: SOMETHING (extra explanation)`. Please, let me know if this form looks OK to you.

src/target/riscv/riscv_semihosting.c
```diff
@@ -37,6 +37,52 @@
...
 static int riscv_semihosting_post_result(struct target *target);
 
+static int riscv_semihosting_detect_magic_sequence(struct target *target,
+	const target_addr_t pc, bool *sequence_found)
+{
+	assert(sequence_found);
+
+	const uint32_t magic[] = {
+		0x01f01013,	/* slli    zero,zero,0x1f */
+		0x00100073,	/* ebreak */
```
en-sc, line 50: Is there a need to read the memory at the PC? The fact that the hart was halted due to `ebreak` execution is known from reading `dcsr.cause`.
Please note, this is just a question and I don't think it should be addressed in this commit.

src/target/riscv/riscv_semihosting.c
```diff
@@ -37,6 +37,52 @@
...
+		0x40705013	/* srai    zero,zero,0x7 */
+	};
+
+	LOG_TARGET_DEBUG(target, "Checking for RISC-V semihosting sequence "
+		"at PC = 0x%" TARGET_PRIxADDR, pc);
+
+	/* Read three uncompressed instructions:
+	 * The previous, the current one (pointed to by PC) and the next one. */
+	const target_addr_t sequence_start_address = pc - 4;
+	for (int i = 0; i < 3; i++) {
```
en-sc, line 60: Wouldn't it be easier and faster to read 12 bytes at once using `target_read_buffer()` and compare it with the magic sequence from a buffer filled by `buf_set_u32()`?
Please note, this is just a question and I don't think it should be addressed in this commit.

src/target/riscv/riscv_semihosting.c
```diff
@@ -37,6 +37,52 @@
...
+		/* Instruction memories may not support arbitrary read size.
+		 * Use any size that will work. */
+		const target_addr_t address = sequence_start_address + (4 * i);
+		int result = riscv_read_by_any_size(target, address, 4, buf);
+		if (result != ERROR_OK) {
+			*sequence_found = false;
+			return result;
+		}
+
+		const uint32_t value = target_buffer_get_u32(target, buf);
```
en-sc: `target_buffer_get_u32()` converts the endianness, therefore it is not correct to use it here, since:
1. Instruction encoding does not depend on endianness.
2. `riscv_read_by_any_size()` fills the `buf` by reads of some size. I hope the endianness of the elements is converted to little before storing to the buffer. Alternatively it is a bug in `target_buffer_get_u32()`.

en-sc (changes requested):

src/target/riscv/riscv_semihosting.c
en-sc: In general, I believe  `riscv_read_by_any_size()` is redundant. See #1205

JanMatCodasip: Actually, the `riscv_read_by_any_size()` and `target_read_buffer_default()` work differently. I have tried to explain it under https://github.com/riscv-collab/riscv-openocd/issues/1205.

src/target/riscv/riscv_semihosting.c
JanMatCodasip: I think it is appropriate to read the whole magic sequence, including the ebreak. If we looked at `dcsr.cause` only and not at the actual instruction, we couldn't distinguish `ebreak` and `c.ebreak` (and c.ebreak means that it is not the semihosting sequence).

I have put a comment about that into the code.

src/target/riscv/riscv_semihosting.c
JanMatCodasip: Addressed.

I have double-checked it and you're right: The RISC-V instructions are always laid out as little endian, and for that reason `le_to_h_u32()` should be used instead of `target_buffer_get_u32()`. I have fixed that and added an explanatory comment.

src/target/riscv/riscv_semihosting.c
en-sc: You are right. Thanks for the explanation!

src/target/riscv/riscv_semihosting.c
en-sc: I do understand that they work differently now, I don't get why they should.
Anyway, I believe it's better to continue this discussion in #1205.

en-sc (approved):

src/target/riscv/riscv_semihosting.c
aap-sc: it looks good, thank you!

aap-sc (approved):

PR #1200, opened 2025-01-07
RISC-V Semihosting 3 of 3: Warn if encountered but disabled

JanMatCodasip:
If semihosting is disabled but there is a semihosting request encountered in the program, provide a clear hint to the user what happened and what can be done about it.

### DIFF
--- a/src/target/riscv/riscv_semihosting.c
+++ b/src/target/riscv/riscv_semihosting.c
@@ -110,11 +110,6 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
 	struct semihosting *semihosting = target->semihosting;
 	assert(semihosting);
 
-	if (!semihosting->is_active) {
-		LOG_TARGET_DEBUG(target, "Semihosting outcome: NONE (semihosting not enabled)");
-		return SEMIHOSTING_NONE;
-	}
-
 	riscv_reg_t pc;
 	int result = riscv_reg_get(target, &pc, GDB_REGNO_PC);
 	if (result != ERROR_OK) {
@@ -122,11 +117,28 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
 		return SEMIHOSTING_ERROR;
 	}
 
-	bool sequence_found;
+	bool sequence_found = false;
 	*retval = riscv_semihosting_detect_magic_sequence(target, pc, &sequence_found);
 	if (*retval != ERROR_OK) {
 		LOG_TARGET_DEBUG(target, "Semihosting outcome: ERROR (during magic seq. detection)");
 		return SEMIHOSTING_ERROR;
+	}
+
+	if (!semihosting->is_active) {
+		if (sequence_found) {
+			// If semihositing is encountered but disabled, provide an additional hint to the user.
+			LOG_TARGET_WARNING(target, "RISC-V semihosting call encountered in the program "
+				"but semihosting is disabled!");
+			LOG_TARGET_WARNING(target, "The target will remain halted (PC = 0x%" TARGET_PRIxADDR ").", pc);
+			LOG_TARGET_WARNING(target, "Hint: Restart your debug session and enable semihosting "
+				"by command 'arm semihosting enable'.");
+			// TODO: This can be improved: The ebreak halt cause detection and riscv_semihosting() call
+			// can be added also to "arm semihosting enable", which would allow the user to continue
+			// without restart of the debug session.
+		}
+
+		LOG_TARGET_DEBUG(target, "Semihosting outcome: NONE (semihosting not enabled)");
+		return SEMIHOSTING_NONE;
 	}
 
 	if (!sequence_found) {
@@ -135,7 +147,7 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
 	}
 
 	/* Otherwise we have a semihosting call (and semihosting is enabled).
-	 * Proceed with the semihosting. */
+	 * Proceed with the handling of semihosting. */
 
 	/*
 	 * Perform semihosting call if we are not waiting on a fileio


JanMatCodasip:
Note: This MR is built on top of https://github.com/riscv-collab/riscv-openocd/pull/1198 and https://github.com/riscv-collab/riscv-openocd/pull/1199.

src/target/riscv/riscv_semihosting.c
```diff
@@ -105,28 +105,37 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
...
 	if (*retval != ERROR_OK)
 		return SEMIHOSTING_ERROR;
 
+	if (!semihosting->is_active) {
+		if (sequence_found) {
+			// If semihositing is encountered but disabled, provide an additional hint to the user.
+			LOG_TARGET_WARNING(target, "RISC-V semihosting call encountered in the program "
+				"but semihosting is disabled!");
+			LOG_TARGET_WARNING(target, "The target will remain halted (PC = 0x%" TARGET_PRIxADDR ").", pc);
+			LOG_TARGET_WARNING(target, "Hint: Restart your debug session and enable semihosting "
```
en-sc, line 133: AFAIU, it is possible to process the request without requiring a restart e.g. by calling `riscv_semihosting()` in the handler of `arm semihosting enable` on RISC-V targets (with a check on `dcsr.cause`).
So maybe `Restart your debug session` part should be dropped and a `TODO` added instead?

JanMatCodasip: Partially addressed.

I have added the TODO comment but would like to keep the warning (hint) so that the user is properly informed about what needs to be done - that just enabling the semihosting and resuming is not enough in the current version of OpenOCD.

Would you like me to open an issue for this improvement, or do you think that the TODO note in the code is enough?

en-sc: IMHO a `TODO` is enough.

en-sc (dismissed):

JanMatCodasip:
Note: I have rebased this PR after because the dependency https://github.com/riscv-collab/riscv-openocd/pull/1199 has been merged.

en-sc (approved):

aap-sc:
@JanMatCodasip I would like to discuss the consequences of this patch.

1. This results in situations where unexpected memory reads occur (unintended for the user) in cases where the user is not concerned about semihosting execution. This may affect the stability of the debug session.
2. Moreover, in some edge cases, the implementation may even read from non-existent memory.
3. OpenOCD is often used in environments where the hardware itself is unstable. These additional actions can further complicate analysis.

In my opinion, even the first point in the list should be enough to justify disabling this behavior. The convenience for the user in certain cases (arguably not the primary use case, depending on the environment) should not compromise the stability and predictability of the debug session—after all, OpenOCD is a low-level debugger. In my view, checks like these could instead be implemented in a user configuration file via callback mechanisms, allowing for greater flexibility without affecting core functionality.

@JanMatCodasip do you have an opinion regarding this? Do you agree with the assessment above?

src/target/riscv/riscv_semihosting.c
```diff
@@ -110,32 +110,44 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
...
+	bool sequence_found = false;
 	*retval = riscv_semihosting_detect_magic_sequence(target, pc, &sequence_found);
 	if (*retval != ERROR_OK) {
 		LOG_TARGET_DEBUG(target, "Semihosting outcome: ERROR (during magic seq. detection)");
 		return SEMIHOSTING_ERROR;
 	}
 
+	if (!semihosting->is_active) {
+		if (sequence_found) {
+			// If semihositing is encountered but disabled, provide an additional hint to the user.
```
TommyMurphyTM1234, line 129: Typo: "semihositing".

src/target/riscv/riscv_semihosting.c
```diff
@@ -110,32 +110,44 @@ enum semihosting_result riscv_semihosting(struct target *target, int *retval)
...
+			LOG_TARGET_WARNING(target, "RISC-V semihosting call encountered in the program "
+				"but semihosting is disabled!");
+			LOG_TARGET_WARNING(target, "The target will remain halted (PC = 0x%" TARGET_PRIxADDR ").", pc);
+			LOG_TARGET_WARNING(target, "Hint: Restart your debug session and enable semihosting "
+				"by command 'arm semihosting enable'.");
```
TommyMurphyTM1234, line 134: Maybe "using" rather than "by"?

src/target/riscv/riscv_semihosting.c
aap-sc: @TommyMurphyTM1234 this is already merged, so it's kind of late for review. No my suggestion to @JanMatCodasip is to consider reverting this patch if my arguments are considered good enough.

aap-sc:
> @JanMatCodasip I would like to discuss the consequences of this patch.
> 
> 1. This results in situations where unexpected memory reads occur (unintended for the user) in cases where the user is not concerned about semihosting execution. This may affect the stability of the debug session.
> 2. Moreover, in some edge cases, the implementation may even read from non-existent memory.
> 3. OpenOCD is often used in environments where the hardware itself is unstable. These additional actions can further complicate analysis.
> 
> In my opinion, even the first point in the list should be enough to justify disabling this behavior. The convenience for the user in certain cases (arguably not the primary use case, depending on the environment) should not compromise the stability and predictability of the debug session—after all, OpenOCD is a low-level debugger. In my view, checks like these could instead be implemented in a user configuration file via callback mechanisms, allowing for greater flexibility without affecting core functionality.
> 
> @JanMatCodasip do you have an opinion regarding this? Do you agree with the assessment above?

@JanMatCodasip gentle ping (since looks like your are back)

PR #1202, opened 2025-01-08
Fix data types around batch.{c,h}

JanMatCodasip:
Check that the value of dtmcs.abits is in the expected range
during the target examination.

Make corrections of data types in batch.{c,h} and in related code.
Some of the issues were found by activating "-Wconversion" in GCC,
others by inspecting the code manually.

This is an initial step towards being able to use "-Wconversion" on
RISC-V target code, which will give us bit more confidence around
data types when refactoring or merging new patches.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -10,12 +10,23 @@
 #include "riscv.h"
 #include "field_helpers.h"
 
+// TODO: DTM_DMI_MAX_ADDRESS_LENGTH should be reduced to 32 (per the debug spec)
 #define DTM_DMI_MAX_ADDRESS_LENGTH	((1<<DTM_DTMCS_ABITS_LENGTH)-1)
 #define DMI_SCAN_MAX_BIT_LENGTH (DTM_DMI_MAX_ADDRESS_LENGTH + DTM_DMI_DATA_LENGTH + DTM_DMI_OP_LENGTH)
+
 #define DMI_SCAN_BUF_SIZE (DIV_ROUND_UP(DMI_SCAN_MAX_BIT_LENGTH, 8))
 
 /* Reserve extra room in the batch (needed for the last NOP operation) */
 #define BATCH_RESERVED_SCANS 1
+
+static unsigned int get_dmi_scan_length(const struct target *target)
+{
+	const unsigned int abits = riscv_get_dmi_address_bits(target);
+	assert(abits > 0);
+	assert(abits <= DTM_DMI_MAX_ADDRESS_LENGTH);
+
+	return abits + DTM_DMI_DATA_LENGTH + DTM_DMI_OP_LENGTH;
+}
 
 struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans)
 {
@@ -127,11 +138,10 @@ static void add_idle_before_batch(const struct riscv_batch *batch, size_t start_
 	const unsigned int idle_change = new_delay - batch->last_scan_delay;
 	LOG_TARGET_DEBUG(batch->target, "Adding %u idle cycles before the batch.",
 			idle_change);
-	assert(idle_change <= INT_MAX);
 	jtag_add_runtest(idle_change, TAP_IDLE);
 }
 
-static int get_delay(const struct riscv_batch *batch, size_t scan_idx,
+static unsigned int get_delay(const struct riscv_batch *batch, size_t scan_idx,
 		const struct riscv_scan_delays *delays, bool resets_delays,
 		size_t reset_delays_after)
 {
@@ -142,7 +152,6 @@ static int get_delay(const struct riscv_batch *batch, size_t scan_idx,
 	const enum riscv_scan_delay_class delay_class =
 		batch->delay_classes[scan_idx];
 	const unsigned int delay = riscv_scan_get_delay(delays, delay_class);
-	assert(delay <= INT_MAX);
 	return delays_were_reset ? 0 : delay;
 }
 
@@ -198,10 +207,7 @@ static void log_batch(const struct riscv_batch *batch, size_t start_idx,
 	if (debug_level < LOG_LVL_DEBUG)
 		return;
 
-	const unsigned int scan_bits = batch->fields->num_bits;
-	assert(scan_bits == (unsigned int)riscv_get_dmi_scan_length(batch->target));
-	const unsigned int abits = scan_bits - DTM_DMI_OP_LENGTH
-			- DTM_DMI_DATA_LENGTH;
+	const unsigned int abits = riscv_get_dmi_address_bits(batch->target);
 
 	/* Determine the "op" and "address" of the scan that preceded the first
 	 * executed scan.
@@ -211,7 +217,7 @@ static void log_batch(const struct riscv_batch *batch, size_t start_idx,
 	 * would be a more robust solution.
 	 */
 	bool last_scan_was_read = false;
-	uint32_t last_scan_address = -1 /* to silence maybe-uninitialized */;
+	uint32_t last_scan_address = (uint32_t)(-1) /* to silence maybe-uninitialized */;
 	if (start_idx > 0) {
 		const struct scan_field * const field = &batch->fields[start_idx - 1];
 		assert(field->out_value);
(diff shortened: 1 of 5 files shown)


src/target/riscv/batch.c
```diff
@@ -199,9 +199,10 @@ static void log_batch(const struct riscv_batch *batch, size_t start_idx,
 		return;
 
 	const unsigned int scan_bits = batch->fields->num_bits;
-	assert(scan_bits == (unsigned int)riscv_get_dmi_scan_length(batch->target));
-	const unsigned int abits = scan_bits - DTM_DMI_OP_LENGTH
-			- DTM_DMI_DATA_LENGTH;
+	assert(scan_bits == riscv_get_dmi_scan_length(batch->target));
```
en-sc: I'd suggest `riscv_get_dmi_scan_length()` -> `riscv_get_dmi_abits()` to avoid the awkward conversions.
Maybe a `TODO`?

src/target/riscv/riscv-013.c
```diff
@@ -56,9 +56,9 @@ static riscv_insn_t riscv013_read_progbuf(struct target *target, unsigned int
 		index);
 static int riscv013_invalidate_cached_progbuf(struct target *target);
 static int riscv013_execute_progbuf(struct target *target, uint32_t *cmderr);
-static void riscv013_fill_dmi_write(struct target *target, char *buf, uint64_t a, uint32_t d);
-static void riscv013_fill_dmi_read(struct target *target, char *buf, uint64_t a);
-static int riscv013_get_dmi_scan_length(struct target *target);
+static void riscv013_fill_dmi_write(struct target *target, char *buf, uint32_t a, uint32_t d);
+static void riscv013_fill_dmi_read(struct target *target, char *buf, uint32_t a);
+static unsigned int riscv013_get_dmi_scan_length(struct target *target);
```
```suggestion
static void riscv013_fill_dmi_write(const struct target *target, uint8_t *buf, uint32_t a, uint32_t d);
static void riscv013_fill_dmi_read(const struct target *target, uint8_t *buf, uint32_t a);
static unsigned int riscv013_get_dmi_scan_length(const struct target *target);
```

en-sc (dismissed):
LGTM

A couple of minor nitpicks.

src/target/riscv/batch.c
JanMatCodasip: Addressed:

I have replaced `riscv_get_dmi_scan_length()` by `riscv_get_dmi_address_bits()` and made the corresponding changes across the RISC-V code. Thanks for the idea.

src/target/riscv/batch.c
```diff
@@ -382,11 +396,12 @@ void riscv_batch_add_nop(struct riscv_batch *batch)
...
 	assert(batch->used_scans < batch->allocated_scans);
 	struct scan_field *field = batch->fields + batch->used_scans;
-	field->num_bits = riscv_get_dmi_scan_length(batch->target);
+	field->num_bits = get_dmi_scan_length(batch->target);
+	assert(field->num_bits <= DMI_SCAN_MAX_BIT_LENGTH);
 	field->out_value = (void *)(batch->data_out + batch->used_scans * DMI_SCAN_BUF_SIZE);
 	field->in_value  = (void *)(batch->data_in  + batch->used_scans * DMI_SCAN_BUF_SIZE);
-	riscv_fill_dm_nop(batch->target, (char *)field->out_value);
-	riscv_fill_dm_nop(batch->target, (char *)field->in_value);
+	riscv_fill_dm_nop(batch->target, (uint8_t *)(field->out_value));
```
JanMatCodasip: Note to reviewers: Strangely, `field->out_value` is defined as `const` (in `struct scan_field` in jtag.h). So in order to write to it, I have to keep the cast here to get rid of this `const`.

src/target/riscv/riscv-013.c
JanMatCodasip: Addressed:

I have updated the datatypes of arguments for all the three `riscv013_fill_*` functions.

src/target/riscv/riscv-013.c
```diff
@@ -5390,34 +5413,34 @@ static int riscv013_execute_progbuf(struct target *target, uint32_t *cmderr)
...
 }
 
-static void riscv013_fill_dmi_write(struct target *target, char *buf, uint64_t a, uint32_t d)
+static void riscv013_fill_dmi_write(const struct target *target, uint8_t *buf, uint32_t a, uint32_t d)
 {
 	RISCV013_INFO(info);
-	buf_set_u64((unsigned char *)buf, DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH, DMI_OP_WRITE);
-	buf_set_u64((unsigned char *)buf, DTM_DMI_DATA_OFFSET, DTM_DMI_DATA_LENGTH, d);
-	buf_set_u64((unsigned char *)buf, DTM_DMI_ADDRESS_OFFSET, info->abits, a);
+	buf_set_u32((unsigned char *)buf, DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH, DMI_OP_WRITE);
```
en-sc: [here and around]:
Please, drop the cast to `unsigned char`:
```suggestion
	buf_set_u32(buf, DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH, DMI_OP_WRITE);
```

src/target/riscv/batch.c
```diff
@@ -340,16 +350,20 @@ void riscv_batch_add_dmi_write(struct riscv_batch *batch, uint64_t address, uint
...
+	// TODO: Check that the bit width of "address" is no more than dtmcs.abits,
+	// otherwise return an error (during batch creation or when the batch is executed).
+
 	assert(batch->used_scans < batch->allocated_scans);
 	struct scan_field *field = batch->fields + batch->used_scans;
-	field->num_bits = riscv_get_dmi_scan_length(batch->target);
+	field->num_bits = get_dmi_scan_length(batch->target);
+	assert(field->num_bits <= DMI_SCAN_MAX_BIT_LENGTH);
 	field->out_value = (void *)(batch->data_out + batch->used_scans * DMI_SCAN_BUF_SIZE);
 	field->in_value  = (void *)(batch->data_in  + batch->used_scans * DMI_SCAN_BUF_SIZE);
```
en-sc: These casts also seem redundant:
```suggestion
	field->out_value = batch->data_out + batch->used_scans * DMI_SCAN_BUF_SIZE;
	field->in_value  = batch->data_in  + batch->used_scans * DMI_SCAN_BUF_SIZE;
```

src/target/riscv/batch.c
en-sc: > `field->out_value` is defined as `const`

This seems logical to me. JTAG queue execution and other JTAG-related operations should not change what is sent to the target therefore this field is read-only. Moreover, in many cases static data is passed here (e.g. https://github.com/riscv-collab/riscv-openocd/blob/22108977c0714ea0d3a0c103028d3b2269a909b1/src/target/riscv/riscv.c#L40)

I'd suggest filling by a reference to the `batch->data_out`:
```
uint8_t * const field_out = batch->data_out + batch->used_scans * DMI_SCAN_BUF_SIZE;
riscv_fill_dm_nop(batch->target, field_out);
field->out_value = field_out;
```
IMHO, it's a bit cleaner.

en-sc (dismissed):
Thanks for the patch!
IMHO it can be merged as-is, although I have a couple suggestion. Please take a look.

src/target/riscv/batch.c
JanMatCodasip: Addressed: `(void *)` casts removed.

src/target/riscv/riscv-013.c
JanMatCodasip: Addressed: extra casts removed.

src/target/riscv/batch.c
JanMatCodasip: > en-sc: This seems logical to me. JTAG queue execution and [...]

Understood, thanks for explanation.

> en-sc: I'd suggest filling by a reference to the batch->data_out

Addressed: I have used non-const pointers (as extra variables) for both in_value and out_value to make the code consistent and easy to read.

en-sc (approved):
LGTM. Thanks a lot for this cleanup!

PR #1204, opened 2025-01-13
Fix esp RISC-V configs

sobuch:
Seems to be broken since 6a27d9fbc, with the following error:
```
tcl//target/esp_common.cfg:69: Error: Command/target: riscv Exists
Traceback (most recent call last):
  File "tcl//target/esp32c6.cfg", line 142, in script
    create_esp_target riscv
  File "tcl//target/esp_common.cfg", line 79, in create_esp_target
    create_openocd_targets
  File "tcl//target/esp_common.cfg", line 69, in create_openocd_targets
    {target create} riscv riscv -chain-position riscv.cpu -coreid 0 -rtos hwthread
```

### DIFF
--- a/tcl/target/esp32c2.cfg
+++ b/tcl/target/esp32c2.cfg
@@ -5,7 +5,7 @@
 source [find target/esp_common.cfg]
 
 # Target specific global variables
-set _CHIPNAME 					"riscv"
+set _CHIPNAME 					"esp32c2"
 set _CPUTAPID 					0x0000cc25
 set _ESP_ARCH 					"riscv"
 set _ONLYCPU					1
@@ -14,7 +14,7 @@ set _ESP_SMP_BREAK 				0
 set _ESP_EFUSE_MAC_ADDR_REG  	0x60008840
 
 # Target specific functions should be implemented for each riscv chips.
-proc riscv_wdt_disable { } {
+proc esp32c2_wdt_disable { } {
     # Halt event can occur during config phase (before "init" is done).
     # Ignore it since mww commands don't work at that time.
     if { [string compare [command mode] config] == 0 } {
@@ -32,7 +32,7 @@ proc riscv_wdt_disable { } {
     mww 0x600080A0 0x84B00000
 }
 
-proc riscv_soc_reset { } {
+proc esp32c2_soc_reset { } {
 	global _RISCV_DMCONTROL
 
     # This procedure does "digital system reset", i.e. resets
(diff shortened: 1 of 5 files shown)


aap-sc:
@sobuch  are you sure that https://github.com/riscv-collab/riscv-openocd/commit/6a27d9fbc0aca57b3c4790f5e0bcfcc597a3b5cb is the right commit? Looking at the diff alone it does not look like this configuration is affected.

aap-sc:
> Looking at the diff alone it does not look like this configuration is affected.

actually... Maybe this line is to blame: 

```
# .../riscv/startup.tcl
proc riscv {cmd args} {
	tailcall "riscv $cmd" {*}$args
}
```

@en-sc , given that you are the author of this specific code  - could you take a look please?

MarekVCodasip (approved):
Checked visually only and looks good. (I don't have the hardware to test it.)

TommyMurphyTM1234:
> > Looking at the diff alone it does not look like this configuration is affected.
> 
> actually... Maybe this line is to blame:
> 
> ```
> # .../riscv/startup.tcl
> proc riscv {cmd args} {
> 	tailcall "riscv $cmd" {*}$args
> }
> ```
> 
> @en-sc , given that you are the author of this specific code - could you take a look please?

@en-sc, any update on this? Why is the procedure named `riscv`? Doesn't seem that descriptive? What is the rationale for it?

sobuch:
@TommyMurphyTM1234 startup.tcl can be probably discussed separately from this PR (though I didnt observe any issues when I tried removing this procedure), I believe its better in any case to decouple the target type from its name in our configs. The targets were only named "riscv" here to use the riscv target instead of custom per-chip targets (that we have in our repo).

sobuch:
Hi @MarekVCodasip, if you would like some esp hardware for testing OpenOCD, it probably can be arranged

en-sc (approved):
@sobuch, sorry for taking so long to reply.

IMHO this is a valuable patch regardless of `startup.tcl`.

The reason is -- by convention target name includes a dot.
https://openocd.org/doc-release/html/CPU-Configuration.html#Target-Configuration

Moreover, naming the target `riscv` results in commands like `riscv riscv ...`, which seem strange.

@TommyMurphyTM1234, @aap-sc, as for the issue with `startup.tcl`:
`proc riscv` was added to support deprecation of RISC-V-specific command in TCL.
Please see https://sourceforge.net/p/openocd/tickets/434/

However, there is an issue with applying this approach to `riscv ...` commands: many such commands can be used like `<target_name> riscv ...`. This use case is not covered by `proc riscv`.

en-sc:
See also #1213

aap-sc (approved):

JanMatCodasip:
> sobuch: Hi MarekVCodasip, if you would like some esp hardware for testing OpenOCD, it probably can be arranged

Thank you for this kind offer. 

We might start testing OpenOCD on 3rd party RISC-V hardware in future, but will stick to commercially available dev boards and standard distribution channels.

PR #1206, opened 2025-01-16
Remove target/riscv/asm.h

JanMatCodasip:
Remove the whole `asm.h` file and inline the two functions directly into `riscv-011.c`, which is the only place where they are used.

This is a code cleanup PR without any change of functionality.

Change-Id: Ifa4b2b87ab9f3f50c78a75361003ce478bfd9d5f


en-sc (changes requested):
Thanks for the patch!

The file also needs to be removed from `Makefile.am`:
https://github.com/riscv-collab/riscv-openocd/blob/eb9ba216e199a3a93f976246823c7a9653ad5064/src/target/riscv/Makefile.am#L5

JanMatCodasip:
> en-sc: The file also needs to be removed from Makefile.am

Addressed. Thanks.

en-sc (approved):
LGTM, thanks!

PR #1207, opened 2025-01-16
target/riscv: set VLENB/MTOPI/MTOPEI existence on 0.11 targets

en-sc:
commit 5f45b5bd73566028ee36c146803232b3dce77c52 ("target/riscv: reg cache entry is initialized before access") introduced an assertion in `riscv_reg_impl_gdb_regno_exist()`.
Link: https://github.com/riscv-collab/riscv-openocd/blob/f82c5a7c048eb70fdc4dff6f53002fa1d3a1bda5/src/target/riscv/riscv_reg.c#L385-L389 This assertion fails on RISC-V Debug Spec. 0.11 targets. The commit is intended to fix this.

Change-Id: I20b56df1517f4071f4b6e39c83178a29a9cf95b0

### DIFF
--- a/src/target/riscv/riscv-011_reg.c
+++ b/src/target/riscv/riscv-011_reg.c
@@ -36,23 +36,50 @@ static const struct reg_arch_type *riscv011_gdb_regno_reg_type(uint32_t regno)
 	return &riscv011_reg_type;
 }
 
-static int riscv011_init_reg(struct target *target, uint32_t regno)
-{
-	return riscv_reg_impl_init_cache_entry(target, regno,
-			riscv_reg_impl_gdb_regno_exist(target, regno),
-			riscv011_gdb_regno_reg_type(regno));
-}
 
 int riscv011_reg_init_all(struct target *target)
 {
-	if (riscv_reg_impl_init_cache(target) != ERROR_OK)
-		return ERROR_FAIL;
+	int res = riscv_reg_impl_init_cache(target);
+	if (res != ERROR_OK)
+		return res;
 
 	init_shared_reg_info(target);
 
-	for (uint32_t regno = 0; regno < target->reg_cache->num_regs; ++regno)
-		if (riscv011_init_reg(target, regno) != ERROR_OK)
-			return ERROR_FAIL;
+	RISCV_INFO(r);
+	assert(!r->vlenb
+			&& "VLENB discovery is not supported on RISC-V 0.11 targets");
+	assert(!r->mtopi_readable
+			&& "MTOPI discovery is not supported on RISC-V 0.11 targets");
+	assert(!r->mtopei_readable
+			&& "MTOPEI discovery is not supported on RISC-V 0.11 targets");
+	/* Existence of some registers depends on others.
+	 * E.g. the presence of "v0-31" registers is infered from "vlenb" being
+	 * non-zero.
+	 * Currently, discovery of the following registers is not supported on
+	 * RISC-V 0.11 targets. */
+	uint32_t non_discoverable_regs[] = {
+		GDB_REGNO_VLENB,
+		GDB_REGNO_MTOPI,
+		GDB_REGNO_MTOPEI
+	};
+	for (unsigned int i = 0; i < ARRAY_SIZE(non_discoverable_regs); ++i) {
+		const uint32_t regno = non_discoverable_regs[i];
+		res = riscv_reg_impl_init_cache_entry(target, regno,
+				/*exist*/ false, riscv011_gdb_regno_reg_type(regno));
+		if (res != ERROR_OK)
+			return res;
+	}
+
+	for (uint32_t regno = 0; regno < target->reg_cache->num_regs; ++regno) {
+		const struct reg * const reg = riscv_reg_impl_cache_entry(target, regno);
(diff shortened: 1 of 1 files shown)


aap-sc (dismissed):
LGTM, review internally.

MarekVCodasip:
Ran riscv-tests on the hifive1-reva board we have. These are the results:
```
:::::::::::::::::::::::::::[ ran 73 tests in 220s ]:::::::::::::::::::::::::::
19 tests returned not_applicable
42 tests returned pass
4 tests returned exception
   DebugFunctionCall > logs/20250123-175903-HiFive1-DebugFunctionCall.log
   IcountTest > logs/20250123-175928-HiFive1-IcountTest.log
   MemTestReadInvalid > logs/20250123-180059-HiFive1-MemTestReadInvalid.log
   RepeatReadTest > logs/20250123-180132-HiFive1-RepeatReadTest.log
8 tests returned fail
   EtriggerTest > logs/20250123-175917-HiFive1-EtriggerTest.log
   InstantHaltTest > logs/20250123-180013-HiFive1-InstantHaltTest.log
   InterruptTest > logs/20250123-180015-HiFive1-InterruptTest.log
   ItriggerTest > logs/20250123-180041-HiFive1-ItriggerTest.log
   MemorySampleMixed > logs/20250123-180101-HiFive1-MemorySampleMixed.log
   MemorySampleSingle > logs/20250123-180109-HiFive1-MemorySampleSingle.log
   Semihosting > logs/20250123-180134-HiFive1-Semihosting.log
   SemihostingFileio > logs/20250123-180137-HiFive1-SemihostingFileio.log
```

src/target/riscv/riscv-011_reg.c
```diff
@@ -36,23 +36,45 @@ static const struct reg_arch_type *riscv011_gdb_regno_reg_type(uint32_t regno)
...
-			riscv011_gdb_regno_reg_type(regno));
-}
 
 int riscv011_reg_init_all(struct target *target)
 {
-	if (riscv_reg_impl_init_cache(target) != ERROR_OK)
-		return ERROR_FAIL;
+	int res = riscv_reg_impl_init_cache(target);
+	if (res != ERROR_OK)
+		return res;
```
MarekVCodasip, line 44:
```suggestion
        // This is a workaround an assert in the regular riscv013 target, the cache is never actually used in the 0.11 target
	int res = riscv_reg_impl_init_cache(target);
	if (res != ERROR_OK)
		return res;
```

src/target/riscv/riscv-011_reg.c
```diff
@@ -36,23 +36,45 @@ static const struct reg_arch_type *riscv011_gdb_regno_reg_type(uint32_t regno)
...
+			&& "VLENB discovery is not supported on RISC-V 0.11 targets");
+	assert(!r->mtopi_readable
+			&& "MTOPI discovery is not supported on RISC-V 0.11 targets");
+	assert(!r->mtopei_readable
+			&& "MTOPEI discovery is not supported on RISC-V 0.11 targets");
+	uint32_t non_discoverable_regs[] = {
+		GDB_REGNO_VLENB,
+		GDB_REGNO_MTOPI,
+		GDB_REGNO_MTOPEI
+	};
```
MarekVCodasip:
```suggestion
	// riscv_reg_impl_gdb_regno_exist() has an assert on these three registers that their existence is determined elsewhere, in riscv011, they never exist.
	assert(!r->vlenb
			&& "VLENB discovery is not supported on RISC-V 0.11 targets");
	assert(!r->mtopi_readable
			&& "MTOPI discovery is not supported on RISC-V 0.11 targets");
	assert(!r->mtopei_readable
			&& "MTOPEI discovery is not supported on RISC-V 0.11 targets");
	uint32_t non_discoverable_regs[] = {
		GDB_REGNO_VLENB,
		GDB_REGNO_MTOPI,
		GDB_REGNO_MTOPEI
	};
```

MarekVCodasip (commented):
Not a huge fan of this solution, but I don't see any other way which does not interfere in the regular 0.13 target.

Please, add the comments I suggested to ensure that future readers know.

src/target/riscv/riscv-011_reg.c
en-sc: > This is a workaround an assert in the regular riscv013 target, the cache is never actually used in the 0.11 target

This is not true: the cache is still used to pass the value. Meaning TCL `reg`  command will read the value from `reg->value` (after `reg->get()` fills it). Therefore it is still necessary to initialize the cache entry.
On RISC-V 0.11 targets registers are never cached in the cache, meaning `reg->valid` is never set to true.

src/target/riscv/riscv-011_reg.c
en-sc: Addressed. Though I've worded the comment a bit differently. Please, take a look.

src/target/riscv/riscv-011_reg.c
MarekVCodasip: Yes, I should have worded it differently, I meant to say that the target registers are not in the cache on 0.11 targets. I thought a bit about it and I don't think the comment is needed here.

MarekVCodasip (approved):

PR #1208, opened 2025-01-17
target/riscv: pass `jtag_tap` instead of `target`

en-sc:
For some functions, passing `target` is excessive. The corresponding `tap` provides all the necessary data.

Change-Id: Ie5836024a15222bda7c2b727f5dbaac38f459b3c

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -310,18 +310,18 @@ static void riscv_sample_buf_maybe_add_timestamp(struct target *target, bool bef
 
 static int riscv_resume_go_all_harts(struct target *target);
 
-void select_dmi_via_bscan(struct target *target)
+void select_dmi_via_bscan(struct jtag_tap *tap)
 {
-	jtag_add_ir_scan(target->tap, &select_user4, TAP_IDLE);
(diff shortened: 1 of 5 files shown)


src/target/riscv/riscv.c
```diff
@@ -407,40 +407,42 @@ int dtmcontrol_scan_via_bscan(struct target *target, uint32_t out, uint32_t *in_
 }
 
 /* TODO: rename "dtmcontrol"-> "dtmcs" */
```
MarekVCodasip, line 409:
```suggestion
```

I think this TODO can be deleted now.

MarekVCodasip (commented):
I also noticed that this can be easily done with riscv_add_bscan_tunneled_scan() (riscv.c) Can you do that?

Thanks.

src/target/riscv/riscv.c
en-sc: There is still a bunch of mentions of `dtmcontrol` in `riscv.c/h` and `riscv-013.c`. I'd like to leave this TODO.

MarekVCodasip: Alright then

MarekVCodasip (approved):

PR #1209, opened 2025-01-17
target/riscv: drop `mtopi_readable/mtopei_readable`  `riscv_info` fields

en-sc:
These fields duplicate the info in the corresponding register cache entries.

### DIFF
--- a/src/target/riscv/riscv-011_reg.c
+++ b/src/target/riscv/riscv-011_reg.c
@@ -48,10 +48,6 @@ int riscv011_reg_init_all(struct target *target)
 	RISCV_INFO(r);
 	assert(!r->vlenb
 			&& "VLENB discovery is not supported on RISC-V 0.11 targets");
-	assert(!r->mtopi_readable
-			&& "MTOPI discovery is not supported on RISC-V 0.11 targets");
-	assert(!r->mtopei_readable
-			&& "MTOPEI discovery is not supported on RISC-V 0.11 targets");
 	/* Existence of some registers depends on others.
 	 * E.g. the presence of "v0-31" registers is infered from "vlenb" being
 	 * non-zero.
--- a/src/target/riscv/riscv_reg.c
+++ b/src/target/riscv/riscv_reg.c
@@ -346,22 +346,11 @@ static bool vlenb_exists(const struct target *target)
 	return riscv_vlenb(target) != 0;
 }
 
-static bool mtopi_exists(const struct target *target)
+static bool reg_exists(const struct target *target, uint32_t regno)
 {
-	RISCV_INFO(info)
-	/* TODO: The naming is quite unfortunate here. `mtopi_readable` refers
-	 * to how the fact that `mtopi` exists was deduced during examine.
-	 */
-	return info->mtopi_readable;
-}
-
-static bool mtopei_exists(const struct target *target)
-{
-	RISCV_INFO(info)
-	/* TODO: The naming is quite unfortunate here. `mtopei_readable` refers
-	 * to how the fact that `mtopei` exists was deduced during examine.
-	 */
-	return info->mtopei_readable;
+	const struct reg * const reg = riscv_reg_impl_cache_entry(target, regno);
+	assert(riscv_reg_impl_is_initialized(reg));
+	return reg->exist;
 }
 
 static bool is_known_standard_csr(unsigned int csr_num)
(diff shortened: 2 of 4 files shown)


en-sc:
Depends on https://github.com/riscv-collab/riscv-openocd/pull/1207

src/target/riscv/riscv_reg.c
```diff
@@ -346,22 +346,11 @@ static bool vlenb_exists(const struct target *target)
 	return riscv_vlenb(target) != 0;
 }
 
-static bool mtopi_exists(const struct target *target)
+static bool query_reg_exist(const struct target *target, uint32_t regno)
```
JanMatCodasip: Naming detail: If you like, you can call the function `reg_exists()` or `register_exists()` which is bit more natural, also simple and clear. The word `query` tends to make the reader think that an actual register access to the target could be performed inside this function.

src/target/riscv/riscv-011_reg.c
```diff
@@ -36,23 +36,41 @@ static const struct reg_arch_type *riscv011_gdb_regno_reg_type(uint32_t regno)
...
 }
 
-static int riscv011_init_reg(struct target *target, uint32_t regno)
-{
-	return riscv_reg_impl_init_cache_entry(target, regno,
-			riscv_reg_impl_gdb_regno_exist(target, regno),
-			riscv011_gdb_regno_reg_type(regno));
-}
 
 int riscv011_reg_init_all(struct target *target)
```
JanMatCodasip: @en-sc @aap-sc 

Some general thoughts/questions about the RISC-V 0.11 support:

- I wonder how many users actually use the legacy 0.11 RISC-V debug hardware nowadays. The only generally available board that I am aware of is [HiFive1](https://www.sifive.com/boards/hifive1) (the long-discontinued original revision of that board, not the Rev B).
- The ability to test the 0.11 support is limited and we do not even have a clean testsuite at the moment (see [here](https://github.com/riscv-collab/riscv-openocd/pull/1207#issuecomment-2610438739)). 
- Should we perhaps make an inquiry about how many users of 0.11 is there these days? (Ask on openocd.org mailing list and here on Github.)
- Depending on the outcome, we may start thinking about deprecation and later removal of the 0.11 code. 
- Alternatively, a snapshot of the 0.11 could be taken, separated from the 0.13 code and further maintained as a separate target, likely with minimum changes. This would allow the 0.13 code to develop independently and without interfering with the legacy 0.11 support (but at the cost of some code duplication).

What is your opinion on this matter? Thank you.

JanMatCodasip (dismissed):
Checked visually, LGTM. Thank you.

src/target/riscv/riscv-011_reg.c
en-sc: IMHO it's not that big of a deal.

AFAIK, you and @TommyMurphyTM1234 have 0.11 HW. We have been able to build [SiFive Freedom E300](https://github.com/sifive/freedom/tree/freedome300artydevkit-v1p0p2) for Arty FPGA, so we are also able to test 0.11 support. We have not set up the internal CI on it yet though, but this activity is planned and we will do it eventually.

Moreover, I believe supporting two non-compatible versions of debug interface makes RISC-V OpenOCD better prepared for future backward-incompatible changes.

As for the testsuite, I believe it won't be too much effort to debug the "expected" failures and I hope to do this before the merge back into mainline OpenOCD is started.

TommyMurphyTM1234: Hi @en-sc & @JanMatCodasip 

Yes, I still have a HiFive1 Rev A board so can assist with any testing/investigations as needed. I had mentioned that I'd look further into some odd behaviour (e.g. possibly unexpected failure/exception cases) with the tests but unfortunately never got around to that. Apologies, but if it's still considered useful I will try to have a look again? 

Another thing to consider here may be soft IP/FPGA implementations of the 0.11 debug block which do exist although I'm not sure how actively used they are at this stage? And I'm not sure how the responsive the providers of such IP would be if polled for information on same? 

I may have the wherewithal to implement and exercise some of these if it proves necessary/useful but, by default, the primary 0.11 target for OpenOCD/test suite testing purposes has always been the HiFive1 and that probably won't change fundamentally?

Regards
Tommy

src/target/riscv/riscv_reg.c
en-sc: Addressed.

src/target/riscv/riscv-011_reg.c
JanMatCodasip: Thank you for your replies.

> en-sc: I believe supporting two non-compatible versions of debug interface makes RISC-V OpenOCD better prepared for future backward-incompatible changes.

Still, my preference would be to separate the 0.11 and 0.13 implementations as much as possible/practical. Ideally, the version would be determined early in the examination phase and from that point onwards, the rest would be handled by two (mostly) separate implementations - sharing as little code as possible.

In another words, where there is a possibility to separate 0.11 and 0.13 code even more, I would recommend to take advantage of it.

JanMatCodasip (approved):

src/target/riscv/riscv-011_reg.c
JanMatCodasip: Also would you know if the 0.11 version of the debug specification can be obtained from somewhere? (I have not yet searched for that document.) Thanks.

en-sc: I've used this link https://static.dev.sifive.com/riscv-debug-spec-0.11nov12.pdf
Not sure whether it's the right one.

PR #1217, opened 2025-01-27
target/riscv: dont set mcause and mstatus as cachable

sobuch:
With CLIC extension (smclic), mcause and mstatus CSRs share mirrored fields for mpp and mpie. Therefore, neither can be assumed cacheable.

CLIC spec available - https://github.com/riscv/riscv-fast-interrupt

Implemented e.g. in esp32c5 or Codasip's L110

### DIFF
--- a/src/target/riscv/riscv_reg_impl.h
+++ b/src/target/riscv/riscv_reg_impl.h
@@ -204,9 +204,7 @@ static inline bool riscv_reg_impl_gdb_regno_cacheable(enum gdb_regno regno,
 		case GDB_REGNO_MISA:
 		case GDB_REGNO_DCSR:
 		case GDB_REGNO_DSCRATCH0:
-		case GDB_REGNO_MSTATUS:
 		case GDB_REGNO_MEPC:
-		case GDB_REGNO_MCAUSE:
 		case GDB_REGNO_SATP:
 			/*
 			 * WARL registers might not contain the value we just wrote, but


JanMatCodasip (commented):
@sobuch The change looks all right to me, thanks.

I just wanted to check I understand the situation right - 

Is this the case you are trying to fix: If the debugger writes to either `mstatus` or `mcause` register, that could cause the cached value of the other register to become invalid (bogus) due to the mirrored (shared) fields between the two registers. Marking both the registers as non-cacheable avoids this. Is that correct understanding?

@en-sc It looks like an equivalent fix should be applied to your PR https://github.com/riscv-collab/riscv-openocd/pull/1187.

sobuch:
@JanMatCodasip yes, for example after the sequence bellow, mcause is actually 0x0, but the value is not re-read when assumed cacheable:

```
> reg mstatus
mstatus (/32): 0x00001800
> reg mcause
mcause (/32): 0x30000000
> reg mstatus 0
mstatus (/32): 0x00000000
> reg mcause
mcause (/32): 0x30000000
```

JanMatCodasip (approved):

sobuch:
FYI just occurred to me that since mstatus also shares fields with the other *status registers, this patch solves similar issue for targets implementing S, H, or N extensions.

en-sc (commented):
LGTM

However, I believe there needs to be a better solution. Filed #1219

en-sc (approved):

PR #1218, opened 2025-01-29
target/riscv: refactored memory access result codes

fkhaidari:
Slightly refactored memory access result codes:
* Changed enum formatting
* Changed status handlers to decrease boilerplate

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3466,40 +3466,39 @@ typedef enum {
 	MEM_ACCESS_RESULT_TYPE_OK,
 	MEM_ACCESS_RESULT_TYPE_DISABLED,
 	MEM_ACCESS_RESULT_TYPE_SKIPPED,
-	MEM_ACCESS_RESULT_TYPE_FAILED
+	MEM_ACCESS_RESULT_TYPE_FAILED,
+	MEM_ACCESS_RESULT_TYPE_ENUM_SIZE,
 } mem_access_result_type_t;
 
-#define LIST_OF_MEM_ACCESS_RESULTS                                                    \
-	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_OK,                                          \
-			MEM_ACCESS_RESULT_TYPE_OK, "ok")                                          \
-	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_DISABLED,                                    \
-			MEM_ACCESS_RESULT_TYPE_DISABLED, "disabled")                              \
-	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_SKIPPED_ABSTRACT_ACCESS_CMDERR,              \
-			MEM_ACCESS_RESULT_TYPE_SKIPPED, "skipped (abstract access cmderr)")       \
-	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_SKIPPED_PROGBUF_NOT_PRESENT,                 \
-			MEM_ACCESS_RESULT_TYPE_SKIPPED, "skipped (progbuf not present)")          \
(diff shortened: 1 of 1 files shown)


fkhaidari:
This commit precedes the next commit, which introduces more precise/correct return codes.

fkhaidari:
@JanMatCodasip, @MarekVCodasip can you please take a look?

src/target/riscv/riscv-013.c
```diff
@@ -3415,43 +3415,61 @@ static void log_mem_access_result(struct target *target, bool success, riscv_mem
 
 typedef enum {
 	MEM_ACCESS_RESULT_TYPE_OK,
-	MEM_ACCESS_RESULT_TYPE_DISABLED,
+	MEM_ACCESS_RESULT_TYPE_COMMON,
```
JanMatCodasip: Instead of `COMMON`, it looks like you can directly say `DISABLED`. That name should be more clear to the readers.

```suggestion
	MEM_ACCESS_RESULT_TYPE_DISABLED,
```

src/target/riscv/riscv-013.c
```diff
@@ -3478,14 +3495,13 @@ bool is_mem_access_failed(mem_access_result_t status)
 bool is_mem_access_skipped(mem_access_result_t status)
 {
 	#define MEM_ACCESS_RESULT_HANDLER(name, kind, msg) \
-		case name: return kind == MEM_ACCESS_RESULT_TYPE_SKIPPED;
+		case name: return MEM_ACCESS_RESULT_TYPE_##kind \
+					== MEM_ACCESS_RESULT_TYPE_FAILED;
```
JanMatCodasip: This looks like a copy-paste typo (`FAILED -> SKIPPED`): 

```suggestion
		case name: return MEM_ACCESS_RESULT_TYPE_##kind \
					== MEM_ACCESS_RESULT_TYPE_SKIPPED;
```

JanMatCodasip (changes requested):

src/target/riscv/riscv-013.c
fkhaidari: Oh, my bad. Addressed, thank you!

src/target/riscv/riscv-013.c

JanMatCodasip (dismissed):

JanMatCodasip:
LGTM, thank you.

src/target/riscv/riscv-013.c
```diff
@@ -3417,41 +3417,59 @@ typedef enum {
...
+			FAILED, "failed (register write failed)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_FAILED_REG_READ_FAILED, \
+			FAILED, "failed (register read failed)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_FAILED_PROGBUF_STARTUP_FAILED, \
+			FAILED, "failed (progbuf startup failed)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_FAILED_EXECUTE_ABSTRACT_FAILED, \
+			FAILED, "failed (execute abstract failed)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_FAILED_NO_FORWARD_PROGRESS, \
+			FAILED, "failed (no forward progress)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_RESULT_ENUM_SIZE, \
```
en-sc: I'd suggest dropping this value.
1. It's meaningless in `is_mem_access_*()` handlers (and is currently mismanaged there).
2. It's unused and there are no casts from integer to this enum (and I believe there should be no such casts).

en-sc (approved):

JanMatCodasip (approved):

PR #1220, opened 2025-01-29
target/riscv: support deprecated user interrupt registers

sobuch:
Adds manually-written encoding header with the defines for these registers, as the extension was dropped and the defines removed from the generated encoding.h header.

Only affects targets with N extension bit in the misa register.

### DIFF
--- a/src/target/riscv/encoding_extras.h
+++ b/src/target/riscv/encoding_extras.h
@@ -1,0 +1,31 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+
+/*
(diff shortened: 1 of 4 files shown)


en-sc:
Isn't `expose_csr` TCL command a better fit here?

https://github.com/riscv-collab/riscv-openocd/blob/e870c5f2dece31d67bb1d69384cbb05d20a543ab/doc/openocd.texi#L11256-L11261

aap-sc:
@sobuch I believe we should fix it like this: https://github.com/riscv-collab/riscv-openocd/pull/1221

PR #1221, opened 2025-01-30
fix expose_csr for CSR with address "0"

aap-sc:
This change is a quick-and-dirty workaround for the problem when user
wants to expose CSR with address "0" and instead of user-specified name
"csr0" was used.

The problem looks as follows:

riscv013_reg_examine_all eventually calls init_cache_entry for all CSRs.
init_cache_entry eventually results in a call to riscv_reg_gdb_regno_name.

Then in case of non-standard CSRs we have the following logic:

```
// NULL when regno == 0, since names are not generated yet
if (info->reg_names[regno])
    return info->reg_names[regno]
...
if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095) {
    // generate names for all exposed CSRs (the function
    // lazy-initializes all the required names)
    init_custom_csr_names(target);
    // And here we have an error, since we overwrite the name generated
    // by init_custom_csr_names
    info->reg_names[regno] = init_reg_name_with_prefix("csr", regno - GDB_REGNO_CSR0);
...
```

The error happens because when initially this function is called with
regno = 0, the first condition false, so we have to go and generate all
the names.

### DIFF
--- a/src/target/riscv/riscv_reg.c
+++ b/src/target/riscv/riscv_reg.c
@@ -56,7 +56,6 @@ static const char * const default_reg_names[GDB_REGNO_COUNT] = {
 	[GDB_REGNO_T5] = "t5",
 	[GDB_REGNO_T6] = "t6",
 	[GDB_REGNO_PC] = "pc",
-	[GDB_REGNO_CSR0] = "csr0",
 	[GDB_REGNO_PRIV] = "priv",
 	[GDB_REGNO_FT0] = "ft0",
 	[GDB_REGNO_FT1] = "ft1",
@@ -196,7 +195,8 @@ const char *riscv_reg_gdb_regno_name(const struct target *target, enum gdb_regno
 	}
 	if (regno >= GDB_REGNO_CSR0 && regno <= GDB_REGNO_CSR4095) {
 		init_custom_csr_names(target);
-		info->reg_names[regno] = init_reg_name_with_prefix("csr", regno - GDB_REGNO_CSR0);
+		if (!info->reg_names[regno])
+			info->reg_names[regno] = init_reg_name_with_prefix("csr", regno - GDB_REGNO_CSR0);
 		return info->reg_names[regno];
 	}
 	assert(!"Encountered uninitialized entry in reg_names table");


aap-sc:
Note to reviewers and @sobuch: I consider this as a "quick fix" change. Probably it would be better to refactor the underlying code, however I think the quick fix would not hurt.

sobuch:
Thanks @aap-sc , I can confirm this patch works for me. My understanding of the issue is a little different from your description though.

I believe the issue is specifically only with ustatus (csr num 0), I dont see the same issue e.g. for csr uie (csr num 4). As far as I can tell, all of the CSRs, whether they are exposed or not, are initialized from here - https://github.com/riscv-collab/riscv-openocd/blob/riscv/src/target/riscv/riscv-013_reg.c#L341 So for all CSRs except the first one, the function init_custom_csr_names was called before it could drop to that condition.

It seems that init_custom_csr_names still gets called for every CSR that was not exposed or was not given a name when exposing, so you are right some refactoring would be useful here, but this fix definitely solves the issue for ustatus.

en-sc (approved):
LGTM

aap-sc:
> I believe the issue is specifically only with ustatus (csr num 0), I dont see the same issue e.g. for csr uie (csr num 4). As far as I can tell, all of the CSRs, whether they are exposed or not, are initialized from here - https://github.com/riscv-collab/riscv-openocd/blob/riscv/src/target/riscv/riscv-013_reg.c#L341 So for all CSRs except the first one, the function init_custom_csr_names was called before it could drop to that condition.

@sobuch yes. I think you are correct with your analysis. I'll adjust the commit message to be more precise with the description of the issue.

MarekVCodasip (commented):
Code looks fine to me. But the aforementioned commit message and preferably the PR title should be changed accordingly to the previous comments.

JanMatCodasip:
Please, if you could change the commit message / title to `Fix expose of CSR with address 0` (or similar), as Marek has suggested. 

Otherwise LGTM, thank you.

JanMatCodasip (approved):

aap-sc:
> Please, if you could change the commit message / title to `Fix expose of CSR with address 0` (or similar), as Marek has suggested.
> 
> Otherwise LGTM, thank you.

addressed

aap-sc:
> Code looks fine to me. But the aforementioned commit message and preferably the PR title should be changed accordingly to the previous comments.

addressed

aap-sc:
> Thanks @aap-sc , I can confirm this patch works for me. My understanding of the issue is a little different from your description though.
> 
> I believe the issue is specifically only with ustatus (csr num 0), I dont see the same issue e.g. for csr uie (csr num 4). As far as I can tell, all of the CSRs, whether they are exposed or not, are initialized from here - https://github.com/riscv-collab/riscv-openocd/blob/riscv/src/target/riscv/riscv-013_reg.c#L341 So for all CSRs except the first one, the function init_custom_csr_names was called before it could drop to that condition.
> 
> It seems that init_custom_csr_names still gets called for every CSR that was not exposed or was not given a name when exposing, so you are right some refactoring would be useful here, but this fix definitely solves the issue for ustatus.

@sobuch thanks once again for the analysis. I've fixed the description.

aap-sc:
@JanMatCodasip , @sobuch , @MarekVCodasip could you please take a look once again?

JanMatCodasip:
@aap-sc LGTM. Thank you.

PR #1222, opened 2025-02-05
target/riscv: only update mstatus.*ie bits with set_maskisr steponly

sobuch:
With "set_maskisr steponly" setting, OpenOCD can alter the program behaviour when stepping over instructions modifying mstatus. The original value of mstatus is restored after step, discarding any changes that were made. This patch limits the differences to the interrupt enable bits.

To eliminate any influence of OpenOCD on the program state, we would need to decode the stepped instruction and also detect exceptions. Therefore the patch for that is very extensive, but even this small improvement should help in many cases.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -294,6 +294,10 @@ static enum riscv_halt_reason riscv_halt_reason(struct target *target);
 static void riscv_info_init(struct target *target, struct riscv_info *r);
 static int riscv_step_rtos_hart(struct target *target);
 
+static const riscv_reg_t mstatus_ie_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
+static int riscv_interrupts_disable(struct target *target, riscv_reg_t *old_mstatus);
+static int riscv_interrupts_restore(struct target *target, riscv_reg_t old_mstatus);
+
 static void riscv_sample_buf_maybe_add_timestamp(struct target *target, bool before)
 {
 	RISCV_INFO(r);
@@ -3645,9 +3649,8 @@ static int riscv_run_algorithm(struct target *target, int num_mem_params,
 	}
 
 	/* Disable Interrupts before attempting to run the algorithm. */
-	uint64_t current_mstatus;
-	uint64_t irq_disabled_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
-	if (riscv_interrupts_disable(target, irq_disabled_mask, &current_mstatus) != ERROR_OK)
+	riscv_reg_t current_mstatus;
+	if (riscv_interrupts_disable(target, &current_mstatus) != ERROR_OK)
 		return ERROR_FAIL;
 
 	/* Run algorithm */
@@ -4227,14 +4230,12 @@ static int riscv_openocd_step_impl(struct target *target, int current,
 	}
 
 	bool success = true;
-	uint64_t current_mstatus;
+	riscv_reg_t current_mstatus;
 	RISCV_INFO(info);
 
 	if (info->isrmask_mode == RISCV_ISRMASK_STEPONLY) {
 		/* Disable Interrupts before stepping. */
-		uint64_t irq_disabled_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
-		if (riscv_interrupts_disable(target, irq_disabled_mask,
-				&current_mstatus) != ERROR_OK) {
+		if (riscv_interrupts_disable(target, &current_mstatus) != ERROR_OK) {
 			success = false;
 			LOG_TARGET_ERROR(target, "Unable to disable interrupts.");
 			goto _exit;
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv.c
```diff
@@ -4199,11 +4199,24 @@ static int riscv_openocd_step_impl(struct target *target, int current,
...
+		if (riscv_reg_get(target, &new_mstatus, GDB_REGNO_MSTATUS) != ERROR_OK) {
+			success = false;
+			LOG_TARGET_ERROR(target, "Unable to read mstatus after step");
+			goto _exit;
+		}
+		if (new_mstatus != (current_mstatus & ~irq_disabled_mask)) {
+			LOG_TARGET_DEBUG(target, "mstatus value changed while stepping, "
+					"only restoring interrupt enable bits.");
+			current_mstatus = new_mstatus | (current_mstatus & irq_disabled_mask);
+		}
```
en-sc: TBH, I'm confused by the current implementation of `riscv_interrupts_restore()`.
Looking at the name I would expect it only touches certain bits of `mstatus`, but it is actually equivalent to `riscv_reg_set(target, GDB_REGNO_MSTATUS, current_mstatus)`.

My suggestion is to either:
1. Move the read of `mstatus` and the masking-out of irrelevant bits into `riscv_interrupts_restore()`.
2. Replace `riscv_interrupts_restore()` with `riscv_reg_set()`.

I slightly prefer the first option since there is the `riscv_interrupts_disable()` function.
This will introduce a change in behavior since there is a use of `riscv_interrupts_disable/restore` in `riscv_run_algorithm()`, and if an algorithm clobbers `mstatus` current code restores it fully, but I don't think it is the intended behavior.

Note: if you decide to touch `riscv_interrupts_disable/restore()`, please make them `static`.

en-sc (changes requested):

src/target/riscv/riscv.c
sobuch: Yeah, I didnt want to affect the code in `riscv_run_algorithm`, both to keep this minimal and because I dont have a deep enough understanding of it.

Could you please elaborate on what you think the behavior should be, when restoring registers after algorithm run? Why would we not want to restore `mstatus` fully?

I am quite curious about this, in the espressif fork we use a different implementation, and actually save/restore a lot more registers (all writable registers as far as I understand).

MarekVCodasip: I am leaning on the second option. As together with that, we could log a warning if any of the interrupt enable bits have changed.

JanMatCodasip: Similarly to what @en-sc says above, I would prefer to:
- Move the mstatus change detection into `riscv_interrupt_restore()`. 
- If mstatus change is detected inside `riscv_interrupt_restore()`, display a warning, as @MarekVCodasip prefers it.
- Change the code in `riscv_interrupt_restore()` to update only the IRQ bits, and keep the others intact.

I am not concerned about `riscv_run_algorithm()` since sensible algorithms (like CRC calculation) will only clobber DPC and general-purpose registers. Restoration of DPC and GPRs is already handled inside `riscv_run_algorithm()` as of now. It is IMO unlikely that someone would come up with an algorithm that would modify mstatus. And if that happened, the mstatus save & restore would need to be added into `riscv_run_algorithm()`.

sobuch: Thank you all for the suggestions guys, please see the updated version

src/target/riscv/riscv.c
```diff
@@ -5941,50 +5940,34 @@ static int riscv_resume_go_all_harts(struct target *target)
...
-
-	retval = reg_mstatus->type->set(reg_mstatus, mstatus_bytes);
-	if (retval != ERROR_OK)
-		return retval;
-
 	if (old_mstatus)
 		*old_mstatus = current_mstatus;
-
-	return ERROR_OK;
+	riscv_reg_t ie_mask = MSTATUS_MIE | MSTATUS_HIE | MSTATUS_SIE | MSTATUS_UIE;
```
en-sc: I would prefer for this to be a static constant to avoid re-definition and protect against accidental mistakes.

src/target/riscv/riscv.c
```diff
@@ -5941,50 +5940,34 @@ static int riscv_resume_go_all_harts(struct target *target)
...
 	LOG_TARGET_DEBUG(target, "Disabling interrupts.");
-	struct reg *reg_mstatus = register_get_by_name(target->reg_cache,
-			"mstatus", true);
-	if (!reg_mstatus) {
-		LOG_TARGET_ERROR(target, "Couldn't find mstatus!");
+	riscv_reg_t current_mstatus;
+	int ret = riscv_reg_get(target, &current_mstatus, GDB_REGNO_MSTATUS);
+	if (ret != ERROR_OK) {
+		LOG_TARGET_ERROR(target, "Failed to read mstatus!");
 		return ERROR_FAIL;
```
```suggestion
		return ret;
```
Alternatively just drop the `ret`.

en-sc (dismissed):
Two minor suggestions. Other than that LGTM

src/target/riscv/riscv.c
sobuch: here I prefer to keep the  `ret`, seems more consistent as return value from `riscv_reg_set` is also passed on unchanged

en-sc (dismissed):
LGTM

src/target/riscv/riscv.c
```diff
@@ -5941,50 +5942,32 @@ static int riscv_resume_go_all_harts(struct target *target)
...
+		LOG_TARGET_ERROR(target, "Failed to read mstatus!");
+		return ret;
 	}
-
-	RISCV_INFO(info);
-	uint8_t mstatus_bytes[8];
-	buf_set_u64(mstatus_bytes, 0, info->xlen, old_mstatus);
-	return reg_mstatus->type->set(reg_mstatus, mstatus_bytes);
+	if ((current_mstatus & mstatus_ie_mask) != 0)
+		LOG_TARGET_WARNING(target, "Interrupt enable bits in mstatus changed!");
```
JanMatCodasip: Would you like this extended warning + hint? 

It should hopefully make more clear to the user what has happened.

(Feel free to suggest alternate text if you can think of better wording.)

```suggestion
	if ((current_mstatus & mstatus_ie_mask) != 0) {
		LOG_TARGET_WARNING(target, "Interrupt enable bits in mstatus changed during single-step.");
		LOG_TARGET_INFO(target, "Hint: OpenOCD might have affected the program when it restored the interrupt bits after single-step.");
		LOG_TARGET_INFO(target, "Use 'riscv set_maskisr off' to prevent OpenOCD from touching mstatus during single-step.");
	}
```

JanMatCodasip (commented):
I am sorry for a delayed review.

Overall this looks good. Please, just consider expanding the warning so that it becomes more clear to users what has happened.

src/target/riscv/riscv.c
sobuch: Done, only changed slightly to align with similar messages in https://github.com/riscv-collab/riscv-openocd/blob/riscv/src/target/riscv/riscv_semihosting.c#L130-L134

src/target/riscv/riscv.c
JanMatCodasip: Thank you.

JanMatCodasip (approved):
LGTM.

en-sc (approved):

PR #1225, opened 2025-02-09
Fix data types around opcodes.h and program.{c,h}

JanMatCodasip:
- Fix and clean-up data types in `opcodes.h` and `program.{c,h}`. Some of the changes were prompted by `-Wconversion`, others come from manual code inspection.

- Remove commented code from `opcodes.h` (unused for very long time).

- Add assertions to `opcodes.h` to check that instruction immediates (and other fields) don't exceed their range.

Change-Id: I7fc7c30ac2fdb00a93158d63d1379e7f16b1d168

### DIFF
--- a/src/target/riscv/opcodes.h
+++ b/src/target/riscv/opcodes.h
@@ -4,15 +4,36 @@
 #define OPENOCD_TARGET_RISCV_OPCODES_H
 
 #include "encoding.h"
+#include <assert.h>
+#include <stdbool.h>
+#include <stdint.h>
 
 #define ZERO	0
 #define T0      5
 #define S0      8
 #define S1      9
 
+#define MAX_GPR_NUM 31
+#define MAX_FPR_NUM 31
+#define MAX_VREG_NUM 31
+#define MAX_CSR_NUM 4095
+
+#define MIN_INT12 ((int16_t)(-0x800))
+#define MAX_INT12 ((int16_t)0x7ff)
+
+#define MIN_INT13 ((int16_t)(-0x1000))
+#define MAX_INT13 ((int16_t)0xfff)
+
+#define MIN_INT21 ((int16_t)(-0x100000))
+#define MAX_INT21 ((int16_t)0xfffff)
+
+#define MAX_UINT5 ((uint8_t)0x1f)
+#define MAX_UINT11 ((uint16_t)0x7ff)
+#define MAX_UINT12 ((uint16_t)0xfff)
+
 static uint32_t bits(uint32_t value, unsigned int hi, unsigned int lo)
 {
-	return (value >> lo) & ((1 << (hi+1-lo)) - 1);
+	return (value >> lo) & (((uint32_t)1 << (hi + 1 - lo)) - 1);
 }
 
 static uint32_t bit(uint32_t value, unsigned int b)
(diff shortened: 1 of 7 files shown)


src/target/riscv/opcodes.h
```diff
@@ -4,15 +4,36 @@
...
+#define MAX_CSR_NUM 4095
+
+#define MIN_INT12 ((int16_t)(-0x800))
+#define MAX_INT12 ((int16_t)0x7ff)
+
+#define MIN_INT13 ((int16_t)(-0x1000))
+#define MAX_INT13 ((int16_t)0xfff)
+
+#define MIN_INT21 ((int16_t)(-0x100000))
+#define MAX_INT21 ((int16_t)0xfffff)
```
en-sc:
```suggestion
#define MIN_INT21 ((int32_t)(-0x100000))
#define MAX_INT21 ((int32_t)0xfffff)
```

en-sc (changes requested):
Thanks for the patch!

AFAIU, there is a typo. Please, take a look. Otherwise LGTM

src/target/riscv/opcodes.h
JanMatCodasip: Thanks for the catch!

I have addressed that, but differently. I have decided to remove all the casts in the `MIN_*` and `MAX_*` macros - in the same manner as it is in `stdint.h`.

Note that I have made more changes than that: I have noticed that the calls to `load()`, `store()` and `jal()` from riscv-011.c had to be updated, too.

src/target/riscv/program.c
```diff
@@ -48,7 +48,7 @@ int riscv_program_exec(struct riscv_program *p, struct target *t)
 	if (riscv_program_ebreak(p) != ERROR_OK) {
 		LOG_TARGET_ERROR(t, "Unable to insert ebreak into program buffer");
 		for (size_t i = 0; i < riscv_progbuf_size(p->target); ++i)
-			LOG_TARGET_ERROR(t, "ram[%02x]: DASM(0x%08" PRIx32 ") [0x%08" PRIx32 "]",
+			LOG_TARGET_ERROR(t, "progbuf[%02x]: DASM(0x%08" PRIx32 ") [0x%08" PRIx32 "]",
 					(int)i, p->progbuf[i], p->progbuf[i]);
```
en-sc:
```suggestion
			LOG_TARGET_ERROR(t, "progbuf[%02zx]: DASM(0x%08" PRIx32 ") [0x%08" PRIx32 "]",
					i, p->progbuf[i], p->progbuf[i]);
```
Though IMHO `riscv_progbuf_size()` should return `unsigned int`, not `size_t`
1. There is a hard cap of 16 (for some weird reason `RISCV_MAX_PROGBUF_SIZE` is 32, I've thought this is from 0.11 spec but `riscv_program_exec` is not implemented by 0.11 targets).
2. `size_t` is form host context, `riscv_progbuf_size()` is from target context. To be completely correct, when referring to a size of a memory area something like `target_addr_t` should be used instead (IMHO `target_size_t` would be even better but there is none currently).

JanMatCodasip: Thanks for the feedback.

> en-sc: for some weird reason RISCV_MAX_PROGBUF_SIZE is 32 [...]

Addressed:

- Constant renamed to `RISCV013_MAX_PROGBUF_SIZE` and value changed to `16`.
- Constants `RISCV_REGISTER_COUNT` and `RISCV_DSCRATCH_COUNT` were removed - they are not used anywhere.

> en-sc: riscv_progbuf_size() should return unsigned int, not size_t

Addressed:

- Return type of riscv_progbuf_size() changed `size_t` -> `unsigned int`
- The type of `instruction_count` in `struct riscv_program` changed the same way
- Few statements in program.c updated accordingly

en-sc (approved):
LGTM

PR #1224, opened 2025-02-07
target/riscv: set appropriate memory access result codes

fkhaidari:
Set appropriate memory access result codes

PR #1194, opened 2024-12-25
target/riscv: merged read/write functions to one access function

fkhaidari:
Commit merges `read`/`write` functions to `access` function. It allows to decrease amount of code duplication

PR #1227, opened 2025-02-13
target/riscv: make mem_access_result_t enum type safe

fkhaidari:
Make mem_access_result_t enum type safe and fix related problems

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -3490,11 +3490,39 @@ typedef enum {
 			SKIPPED, "skipped (dm target select failed)") \
 	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_SKIPPED_FENCE_EXEC_FAILED, \
 			SKIPPED, "skipped (fence execution failed)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_SKIPPED_SYSBUS_ACCESS_FAILED, \
+			SKIPPED, "skipped (sysbus access failed)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_SKIPPED_REG_SAVE_FAILED, \
+			SKIPPED, "skipped (register save failed)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_SKIPPED_UNKNOWN_SYSBUS_VERSION, \
+			SKIPPED, "skipped (unknown sysbus version)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_SKIPPED_PROGRAM_WRITE_FAILED, \
+			SKIPPED, "skipped (program write failed)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_SKIPPED_PROGBUF_FILL_FAILED, \
+			SKIPPED, "skipped (progbuf fill failed)") \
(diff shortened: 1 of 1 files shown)


fkhaidari:
This MR is based on #1194, #1218 and #1224. So it has to be merged after them

src/target/riscv/riscv-013.c
```diff
@@ -3466,76 +3463,121 @@ typedef enum {
...
 typedef enum {
 	LIST_OF_MEM_ACCESS_RESULTS
-} mem_access_result_t;
+} mem_access_result_enum_t;
 #undef MEM_ACCESS_RESULT_HANDLER
 
-bool is_mem_access_failed(mem_access_result_t status)
+typedef struct mem_access_result {
+	mem_access_result_enum_t value;
+} mem_access_result_t;
```
MarekVCodasip: Did I understand it correctly?
```suggestion
/* Structure is intentionally used to contain the memory access result, for type safety - to avoid implicit conversions to integers. */
typedef struct mem_access_result {
	mem_access_result_enum_t value;
} mem_access_result_t;
```
Or similar comment, up to you.

src/target/riscv/riscv-013.c
```diff
@@ -3466,76 +3463,121 @@ typedef enum {
...
+			FAILED, "failed (no forward progress)") \
+	MEM_ACCESS_RESULT_HANDLER(MEM_ACCESS_FAILED_FENCE_EXEC_FAILED, \
+			FAILED, "failed (fence execution failed)") \
 
 #define MEM_ACCESS_RESULT_HANDLER(name, kind, msg) name,
 typedef enum {
 	LIST_OF_MEM_ACCESS_RESULTS
-} mem_access_result_t;
+} mem_access_result_enum_t;
 #undef MEM_ACCESS_RESULT_HANDLER
```
MarekVCodasip: It looks like the prevalent style in OpenOCD for enums is to not create typedefs. Can you consider adjusting it?

```
enum mem_access_result {
	LIST_OF_MEM_ACCESS_RESULTS
};
```

src/target/riscv/riscv-013.c
MarekVCodasip: Similarly to previous comment, it appears that `typedef` is not common in the OpenOCD codebase for structs.

Consider adjusting it this way:

```
struct mem_access_result {
	mem_access_result_enum_t value;
};
```

MarekVCodasip (commented):
This style of error handling seems to be a bit unusual compared to the rest of OpenOCD codebase.

However, I understand why you did it like that and I don't mind this different style.

That being said, please could you add few comments, that explain why it is done this way. So that future readers do not have to wonder. Thanks.

src/target/riscv/riscv-013.c
en-sc: Moreover, AFAIU recently Checkpatch started to check that there are no cosmetic typedefs in a commit. 
https://review.openocd.org/c/openocd/+/8710

src/target/riscv/riscv-013.c
fkhaidari: Thanks! Addressed

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

MarekVCodasip (dismissed):

MarekVCodasip (approved):

en-sc (approved):

en-sc:
If nobody objects, I'll merge this one instead of #1224, since this one includes the commit from #1224.

PR #1228, opened 2025-02-14
riscv-011: Don't trigger semihosting before the target is examined

JanMatCodasip:
In riscv-011 target, the `handle_halt()` function, and thus also `riscv_semihosting()`, can get called from within `examine()` before the examination is actually complete.

The chain of the function calls is:

- examine() -> riscv011_poll() -> poll_target() -> handle_halt() -> riscv_semihosting()

If the target is already halted due to a breakpoint (dcsr.cause = SWBP) at the time OpenOCD connects to it, semihosting will be attempted before completing the examination, and the examination will fail.

This issue was observed on HiFive1 Rev A01.

The fix is to handle semihosting only after the target got successfully examined.

Change-Id: Iccfa0da35d47a430d8674131ebd2eb8e5e2922c0

### DIFF
--- a/src/target/riscv/riscv-011.c
+++ b/src/target/riscv/riscv-011.c
@@ -1889,8 +1889,16 @@ static int handle_halt(struct target *target, bool announce)
 
 	if (target->debug_reason == DBG_REASON_BREAKPOINT) {
 		int retval;
-		if (riscv_semihosting(target, &retval) != 0)
-			return retval;
+		/* Hotfix: Don't try to handle semihosting before the target is marked as examined. */
+		/* TODO: The code should be rearranged so that:
+		 * - Semihosting is not attempted before the target is examined.
+		 * - When the target is already halted on a semihosting magic sequence
+		 *   at the time when OpenOCD connects to it, this semihosting attempt
+		 *   gets handled right after the examination.
+		 */
+		if (target_was_examined(target))
+			if (riscv_semihosting(target, &retval) != SEMIHOSTING_NONE)
+				return retval;
 	}
 
 	if (announce)


aap-sc (dismissed):

src/target/riscv/riscv-011.c
```diff
@@ -1884,8 +1884,11 @@ static int handle_halt(struct target *target, bool announce)
 
 	if (target->debug_reason == DBG_REASON_BREAKPOINT) {
 		int retval;
-		if (riscv_semihosting(target, &retval) != 0)
-			return retval;
+		/* Don't try to handle semihosting before the target
+		 * gets successfully examined. */
```
en-sc: Is this right?
What about the following case:
1. Debugger connects to the target and sets up semihosting, the target software uses semihosting.
3. Debugger disconnects for some reason.
4. Debugger reconnects. It sees that the hart is halted (this is evidently from the SW trying to call a semihosting function). My believe is we should strive to handle this call.

I'm fine with merging this patch (the issue I'm describing is objectively minor) but can we please add a FIXME or something?

Also, I believe we have similar issue in riscv013.

In general, I don't like how we handle state transitions (`poll`, etc.). I'm going to file an issue with a comprehensive overview, but it will take some time.

en-sc (changes requested):

src/target/riscv/riscv-011.c
en-sc: And sorry for taking so long to review this.

JanMatCodasip: Addressed.

> en-sc: I'm fine with merging this patch (the issue I'm describing is objectively minor) but can we please add a FIXME or something?

Thank you for your feedback. I have added the `TODO`. Please let me know if the wording looks clear enough.

> en-sc:  What about the following case [...]

I agree that when the hart is already halted on a semihosting magic sequence at the time OpenOCD connects to it, such a semihosting call should be processed normally.

It looks that the code in 0.11 should be re-arranged so that all semihosting handling (all calls to `riscv_semihosting()`) happen after the target examination is complete (after `target_set_examined()`). Not earlier than that.

JanMatCodasip: General note on semihosting in riscv_011:

It appears that the semihosting tests on 0.11 [have been broken](https://github.com/riscv-software-src/riscv-tests/blob/5e62d380c54524d8916573aa3397270f457e4901/debug/hifive1_excludes.yaml#L17-L18) for a long time. 

If anyone is interested in semihosting on 0.11, my recommendation would be to fix the tests first, and only then focus on special cases like @en-sc described [above](https://github.com/riscv-collab/riscv-openocd/pull/1228#discussion_r1993429266).

en-sc (approved):

PR #1229, opened 2025-02-14
CI: Run the "snapshot" action only on the main branch

JanMatCodasip:
Make sure that the "snapshot" action only runs when pushing changes to the main branch - "riscv". No need to run it for individual commits under side-branches.

Change-Id: I5877a3b5635fc6053af7718e38a4d5a093e02744

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -2,7 +2,10 @@
 
 # Copyright (C) 2020 by Tarek BOUCHKATI <tarek.bouchkati@gmail.com>
 
-on: push
+on:
+  push:
+    branches:
+      - riscv
 
 name: OpenOCD Snapshot
 


JanMatCodasip:
Note: This patch only prevents the "snapshot" workflow from getting executed on commits within a feature branch (as happened e.g. [here](https://github.com/riscv-collab/riscv-openocd/actions/runs/13330258839/)).

The "snapshot" workflow itself appears [broken](https://github.com/riscv-collab/riscv-openocd/actions/runs/13072454564/job/36476904853#step:10:2476) at the moment, and will require an update - <del>the jimtcl dependency will probably need to be built from source in the cross-build.sh script</del>.

EDIT: I can see the update of cross-build.sh has already been submitted in https://review.openocd.org/c/openocd/+/8751.

sobuch:
I like this change, the snapshot action is more of a "release" than a "test" job anyway, no need to run on every branch pushed.

aap-sc (approved):

PR #1235, opened 2025-03-04
target/riscv: implement abstract command cache

### LINKED ISSUE #1232: Implement generic abstract command support cache
The idea is similar to #928.

### The issue

Currently, there are many target properties derived from abstract command being supported or not:
https://github.com/riscv-collab/riscv-openocd/blob/f32f17831e94d9eba5138ecf2ef8fcab443d7f13/src/target/riscv/riscv-013.c#L178-L183

There are issues with the current approach, e.g.:
* Extensibility: Take `abstract_..._supported` for example. Current implementation assumes the support is the same across all the registers in a group.
*  Correctness: We should make sure `abstractcs.cmderr == not_supported`, not `abstractcs.cmderr != none` before asserting that the support is missing. This is not currently done in some places, e.g.
https://github.com/riscv-collab/riscv-openocd/blob/f32f17831e94d9eba5138ecf2ef8fcab443d7f13/src/target/riscv/riscv-013.c#L3791-L3807

### Proposed solution

* Store all non-supported commands in a cache (a sorted array should suffice).
* Query if the command is known to be not supported before executing a command (`bsearch()` in the array).
* Add the command after execution into the array iff `abstractcs.cmderr` is `not_supported`.

fkhaidari:
Implemented cache of unsupported abstract commands. It's purpose is to replace set of caching variables to one. So this commit provides one ac_not_supported_cache instead of abstract_read_csr_supported,
abstract_write_csr_supported, abstract_read_fpr_supported, abstract_write_fpr_supported, has_aampostincrement.

Dropped check for buggy aampostincrement

Fixes #1232

Change-Id: I75cae1481841f2cd0393d6cc80f0d913fbe34294

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -145,6 +145,66 @@ typedef struct {
 	struct target *target;
 } target_list_t;
 
+struct ac_cache {
+	uint32_t *commands;
+	size_t size;
+};
+
+static int ac_cache_elem_comparator(const void *p_lhs, const void *p_rhs)
+{
+	uint32_t lhs = *(const uint32_t *)p_lhs;
+	uint32_t rhs = *(const uint32_t *)p_rhs;
+	if (lhs < rhs)
+		return -1;
+	if (lhs > rhs)
+		return 1;
+	return 0;
+}
+
+static struct ac_cache ac_cache_construct(void)
+{
+	struct ac_cache cache = {
+		cache.commands = NULL,
+		cache.size = 0,
+	};
+	return cache;
+}
+
+static void ac_cache_free(struct ac_cache *cache)
+{
+	free(cache->commands);
+	cache->commands = NULL;
+	cache->size = 0;
+}
+
+static void ac_cache_insert(struct ac_cache *cache, uint32_t command)
+{
+	assert(cache);
+
+	size_t old_size = cache->size;
+	size_t new_size = old_size + 1;
+	size_t entry_size = sizeof(*cache->commands);
+
+	uint32_t *commands = realloc(cache->commands, new_size * entry_size);
+	if (!commands) {
+		LOG_ERROR("Reallocation to %zu bytes failed", new_size * entry_size);
+		return;
+	}
+
+	commands[old_size] = command;
+	cache->commands = commands;
+	cache->size = new_size;
+
+	qsort(cache->commands, cache->size, entry_size,
+			ac_cache_elem_comparator);
+}
+
+static bool ac_cache_contains(const struct ac_cache *cache, uint32_t command)
+{
+	return bsearch(&command, cache->commands, cache->size,
+			sizeof(*cache->commands), ac_cache_elem_comparator);
+}
+
 typedef struct {
 	/* The indexed used to address this hart in its DM. */
 	unsigned int index;
@@ -175,12 +235,7 @@ typedef struct {
 	 */
 	struct riscv_scan_delays learned_delays;
 
-	bool abstract_read_csr_supported;
-	bool abstract_write_csr_supported;
-	bool abstract_read_fpr_supported;
-	bool abstract_write_fpr_supported;
-
-	yes_no_maybe_t has_aampostincrement;
+	struct ac_cache ac_not_supported_cache;
 
 	/* Some fields from hartinfo. */
 	uint8_t datasize;
@@ -651,6 +706,30 @@ clear_cmderr:
 	return res;
 }
 
+enum riscv_debug_reg_ordinal get_cmdtype(uint32_t command)
+{
+	switch (get_field(command, DM_COMMAND_CMDTYPE)) {
+	case 0:
+		return AC_ACCESS_REGISTER_ORDINAL;
+	case 1:
+		return AC_QUICK_ACCESS_ORDINAL;
+	case 2:
+		return AC_ACCESS_MEMORY_ORDINAL;
+	default:
+		assert(false && "Unknown command type value");
+		return 0;
+	}
+}
+
+static void mark_command_as_unsupported(struct target *target, uint32_t command)
+{
+	LOG_TARGET_DEBUG(target, "Caching the abstract "
+			"command 0x%" PRIx32 " as not supported", command);
+	log_debug_reg(target, get_cmdtype(command),
+			command, __FILE__, __LINE__, __func__);
+	ac_cache_insert(&get_info(target)->ac_not_supported_cache, command);
+}
+
 int riscv013_execute_abstract_command(struct target *target, uint32_t command,
 		uint32_t *cmderr)
 {
@@ -684,6 +763,9 @@ int riscv013_execute_abstract_command(struct target *target, uint32_t command,
 
 	res = abstract_cmd_batch_check_and_clear_cmderr(target, batch,
 			abstractcs_read_key, cmderr);
+	if (res != ERROR_OK && *cmderr == CMDERR_NOT_SUPPORTED)
+		mark_command_as_unsupported(target, command);
+
 cleanup:
 	riscv_batch_free(batch);
 	return res;
@@ -829,38 +911,35 @@ uint32_t riscv013_access_register_command(struct target *target, uint32_t number
 	return command;
 }
 
+static bool is_command_unsupported(struct target *target, uint32_t command)
+{
+	bool unsupported = ac_cache_contains(&get_info(target)->ac_not_supported_cache, command);
+	if (!unsupported)
+		return false;
+
+	LOG_TARGET_DEBUG(target, "Abstract command 0x%"
+			PRIx32 " is cached as not supported", command);
+	log_debug_reg(target, get_cmdtype(command),
+			command, __FILE__, __LINE__, __func__);
+	return true;
+}
+
 static int register_read_abstract_with_size(struct target *target,
 		riscv_reg_t *value, enum gdb_regno number, unsigned int size)
 {
-	RISCV013_INFO(info);
-
-	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31 &&
-			!info->abstract_read_fpr_supported)
-		return ERROR_FAIL;
-	if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095 &&
-			!info->abstract_read_csr_supported)
-		return ERROR_FAIL;
 	/* The spec doesn't define abstract register numbers for vector registers. */
 	if (number >= GDB_REGNO_V0 && number <= GDB_REGNO_V31)
 		return ERROR_FAIL;
 
 	uint32_t command = riscv013_access_register_command(target, number, size,
 			AC_ACCESS_REGISTER_TRANSFER);
+	if (is_command_unsupported(target, command))
+		return ERROR_FAIL;
 
 	uint32_t cmderr;
 	int result = riscv013_execute_abstract_command(target, command, &cmderr);
-	if (result != ERROR_OK) {
-		if (cmderr == CMDERR_NOT_SUPPORTED) {
-			if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31) {
-				info->abstract_read_fpr_supported = false;
-				LOG_TARGET_INFO(target, "Disabling abstract command reads from FPRs.");
-			} else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095) {
-				info->abstract_read_csr_supported = false;
-				LOG_TARGET_INFO(target, "Disabling abstract command reads from CSRs.");
-			}
-		}
+	if (result != ERROR_OK)
 		return result;
-	}
 
 	if (value)
 		return read_abstract_arg(target, value, 0, size);
@@ -879,23 +958,17 @@ static int register_read_abstract(struct target *target, riscv_reg_t *value,
 static int register_write_abstract(struct target *target, enum gdb_regno number,
 		riscv_reg_t value)
 {
-	RISCV013_INFO(info);
-
 	dm013_info_t *dm = get_dm(target);
 	if (!dm)
-		return ERROR_FAIL;
-
-	if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31 &&
-			!info->abstract_write_fpr_supported)
-		return ERROR_FAIL;
-	if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095 &&
-			!info->abstract_write_csr_supported)
 		return ERROR_FAIL;
 
 	const unsigned int size_bits = register_size(target, number);
 	const uint32_t command = riscv013_access_register_command(target, number, size_bits,
 			AC_ACCESS_REGISTER_TRANSFER |
 			AC_ACCESS_REGISTER_WRITE);
+	if (is_command_unsupported(target, command))
+		return ERROR_FAIL;
+
 	LOG_DEBUG_REG(target, AC_ACCESS_REGISTER, command);
 	assert(size_bits % 32 == 0);
 	const unsigned int size_in_words = size_bits / 32;
@@ -915,18 +988,9 @@ static int register_write_abstract(struct target *target, enum gdb_regno number,
 	uint32_t cmderr;
 	res = abstract_cmd_batch_check_and_clear_cmderr(target, batch,
 			abstractcs_read_key, &cmderr);
+	if (res != ERROR_OK && cmderr == CMDERR_NOT_SUPPORTED)
+		mark_command_as_unsupported(target, command);
 
-	if (res != ERROR_OK) {
-		if (cmderr == CMDERR_NOT_SUPPORTED) {
-			if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31) {
-				info->abstract_write_fpr_supported = false;
-				LOG_TARGET_INFO(target, "Disabling abstract command writes to FPRs.");
-			} else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095) {
-				info->abstract_write_csr_supported = false;
-				LOG_TARGET_INFO(target, "Disabling abstract command writes to CSRs.");
-			}
-		}
-	}
 cleanup:
 	riscv_batch_free(batch);
 	return res;
@@ -1708,6 +1772,10 @@ static void deinit_target(struct target *target)
 	struct riscv_info *info = target->arch_info;
 	if (!info)
 		return;
+
+	riscv013_info_t *vsinfo = info->version_specific;
+	if (vsinfo)
+		ac_cache_free(&vsinfo->ac_not_supported_cache);
 
 	riscv013_dm_free(target);
 
@@ -2819,17 +2887,7 @@ static int init_target(struct command_context *cmd_ctx,
 	info->progbufsize = -1;
 	reset_learned_delays(target);
 
-	/* Assume all these abstract commands are supported until we learn
-	 * otherwise.
-	 * TODO: The spec allows eg. one CSR to be able to be accessed abstractly
-	 * while another one isn't. We don't track that this closely here, but in
-	 * the future we probably should. */
-	info->abstract_read_csr_supported = true;
-	info->abstract_write_csr_supported = true;
-	info->abstract_read_fpr_supported = true;
-	info->abstract_write_fpr_supported = true;
-
-	info->has_aampostincrement = YNM_MAYBE;
+	info->ac_not_supported_cache = ac_cache_construct();
 
 	return ERROR_OK;
 }
@@ -3719,16 +3777,22 @@ read_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
 {
 	assert(riscv_mem_access_is_read(args));
 
-	RISCV013_INFO(info);
-	bool use_aampostincrement = info->has_aampostincrement != YNM_NO;
-
 	memset(args.read_buffer, 0, args.count * args.size);
 
 	/* Convert the size (bytes) to width (bits) */
 	unsigned int width = args.size << 3;
 
-	/* Create the command (physical address, postincrement, read) */
-	uint32_t command = access_memory_command(target, false, width, use_aampostincrement, false);
+	uint32_t command = access_memory_command(target, /* virtual = */ false,
+			width, /* postincrement = */ true, /* is_write = */ false);
+	bool use_aampostincrement = !is_command_unsupported(target, command);
+	if (!use_aampostincrement)
+		/* It is already known that this abstract memory
+		 * access with aampostincrement=1 is not supported.
+		 * So try aampostincrement=0 right away.
+		 *
+		 * TODO: check if new command is supported  */
+		command = access_memory_command(target, /* virtual = */ false,
+				width, /* postincrement = */ false, /* is_write = */ false);
 
 	/* Execute the reads */
 	uint8_t *p = args.read_buffer;
@@ -3749,33 +3813,14 @@ read_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
 		/* Execute the command */
 		uint32_t cmderr;
 		result = riscv013_execute_abstract_command(target, command, &cmderr);
-
-		/* TODO: we need to modify error handling here. */
-		/* NOTE: in case of timeout cmderr is set to CMDERR_NONE */
-		if (info->has_aampostincrement == YNM_MAYBE) {
-			if (result == ERROR_OK) {
-				/* Safety: double-check that the address was really auto-incremented */
-				riscv_reg_t new_address;
-				result = read_abstract_arg(target, &new_address, 1, riscv_xlen(target));
-				if (result != ERROR_OK)
-					return mem_access_result(MEM_ACCESS_FAILED_DM_ACCESS_FAILED);
-
-				if (new_address == args.address + args.size) {
-					LOG_TARGET_DEBUG(target, "aampostincrement is supported on this target.");
-					info->has_aampostincrement = YNM_YES;
-				} else {
-					LOG_TARGET_WARNING(target, "Buggy aampostincrement! Address not incremented correctly.");
-					info->has_aampostincrement = YNM_NO;
-				}
-			} else {
-				/* Try the same access but with postincrement disabled. */
-				command = access_memory_command(target, false, width, false, false);
-				result = riscv013_execute_abstract_command(target, command, &cmderr);
-				if (result == ERROR_OK) {
-					LOG_TARGET_DEBUG(target, "aampostincrement is not supported on this target.");
-					info->has_aampostincrement = YNM_NO;
-				}
-			}
+		if (use_aampostincrement && result != ERROR_OK &&
+					cmderr == CMDERR_NOT_SUPPORTED) {
+			LOG_TARGET_DEBUG(target, "Trying the same abstract memory "
+					"read command, but without aampostincrement");
+			use_aampostincrement = false;
+			command = access_memory_command(target, /* virtual = */ false,
+					width, /* postincrement = */ false, /* is_write = */ false);
+			result = riscv013_execute_abstract_command(target, command, &cmderr);
 		}
 
 		/* TODO:
@@ -3791,7 +3836,7 @@ read_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
 			return mem_access_result(MEM_ACCESS_FAILED_DM_ACCESS_FAILED);
 		buf_set_u64(p, 0, 8 * args.size, value);
 
-		if (info->has_aampostincrement == YNM_YES)
+		if (use_aampostincrement)
 			updateaddr = false;
 		p += args.size;
 	}
@@ -3809,15 +3854,22 @@ write_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
 {
 	assert(riscv_mem_access_is_write(args));
 
-	RISCV013_INFO(info);
 	int result = ERROR_OK;
-	bool use_aampostincrement = info->has_aampostincrement != YNM_NO;
 
 	/* Convert the size (bytes) to width (bits) */
 	unsigned int width = args.size << 3;
 
-	/* Create the command (physical address, postincrement, write) */
-	uint32_t command = access_memory_command(target, false, width, use_aampostincrement, true);
+	uint32_t command = access_memory_command(target, /* virtual = */ false,
+			width, /* postincrement = */ true, /* is_write = */ true);
+	bool use_aampostincrement = !is_command_unsupported(target, command);
+	if (!use_aampostincrement)
+		/* It is already known that this abstract memory
+		 * access with aampostincrement=1 is not supported.
+		 * So try aampostincrement=0 right away.
+		 *
+		 * TODO: check if new command is supported  */
+		command = access_memory_command(target, /* virtual = */ false,
+				width, /* postincrement = */ false, /* is_write = */ true);
 
 	/* Execute the writes */
 	const uint8_t *p = args.write_buffer;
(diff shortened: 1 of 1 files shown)


fkhaidari:

src/target/riscv/riscv-013.c
```diff
@@ -685,6 +734,11 @@ int riscv013_execute_abstract_command(struct target *target, uint32_t command,
 	res = abstract_cmd_batch_check_and_clear_cmderr(target, batch,
 			abstractcs_read_key, cmderr);
 cleanup:
+	if (res != ERROR_OK && *cmderr == CMDERR_NOT_SUPPORTED) {
+		LOG_TARGET_DEBUG(target, "Command 0x%" PRIx32
+				"is cached as not supported, bailing out", command);
+		ac_cache_insert(&get_info(target)->ac_not_supported_cache, command);
+	}
 	riscv_batch_free(batch);
 	return res;
```
JanMatCodasip: 1) For code clarity, could you please move the `cleanup:` label just above the line 742. The reason is that if the batch_run_timeout() call on line 730 fails, it does not make sense to deal with the abstract command cache. Instead, we can go directly to riscv_batch_free().

2) The debug message should be adjusted to say `Caching the command ... as not supported` so that it is consistent with the same message on line 962 in register_write_abstract().

3) Detail: please don't forget a space between the command number and the subsequent text.

```c
	if (res != ERROR_OK)
		goto free_batch;

	res = abstract_cmd_batch_check_and_clear_cmderr(target, batch,
			abstractcs_read_key, cmderr);

	if (res != ERROR_OK && *cmderr == CMDERR_NOT_SUPPORTED) {
		LOG_TARGET_DEBUG(target, "Caching the command 0x%" PRIx32
				" as not supported", command);
		ac_cache_insert(&get_info(target)->ac_not_supported_cache, command);
	}

free_batch:
	riscv_batch_free(batch);
	return res;
```

src/target/riscv/riscv-013.c
```diff
@@ -916,16 +958,10 @@ static int register_write_abstract(struct target *target, enum gdb_regno number,
...
-			if (number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31) {
-				info->abstract_write_fpr_supported = false;
-				LOG_TARGET_INFO(target, "Disabling abstract command writes to FPRs.");
-			} else if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095) {
-				info->abstract_write_csr_supported = false;
-				LOG_TARGET_INFO(target, "Disabling abstract command writes to CSRs.");
-			}
-		}
+	if (res != ERROR_OK && cmderr == CMDERR_NOT_SUPPORTED) {
+		LOG_TARGET_DEBUG(target, "Caching the command as not supported");
```
JanMatCodasip: This print can contain the command number, same as above in `riscv013_execute_abstract_command()`:

```suggestion
		LOG_TARGET_DEBUG(target, "Caching the abstract command 0x% " PRIx32 " as not supported");
```

Even better, you can create small helper functions to mark/check if command is unsupported, and move the prints into these functions. That way, the prints will be located in a single place only:

```c
static void mark_command_as_unsupported(struct target* target, enum riscv_debug_reg_ordinal command_type, uint32_t command)
{
	LOG_TARGET_DEBUG(target, "Caching the command 0x% " PRIx32 " as not supported");
	log_debug_reg(target, command_type, command, __FILE__, __LINE__, __func__);
	RISCV013_INFO(info);
	ac_cache_insert(&info->ac_not_supported_cache, command);
}

static bool is_command_unsupported(struct target* target, uint32_t command)
{
	// similar
}
```

src/target/riscv/riscv-013.c
```diff
@@ -3720,15 +3750,19 @@ read_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
 
 	/* Create the command (physical address, postincrement, read) */
-	uint32_t command = access_memory_command(target, false, width, use_aampostincrement, false);
+	uint32_t command = access_memory_command(target, false, width,
+			/* postincrement = */ true, /* is_write= */ false);
+	bool use_aampostincrement =
+			!ac_cache_contains(&info->ac_not_supported_cache, command);
+	if (!use_aampostincrement)
+		command = access_memory_command(target, false, width,
+				/* postincrement = */ false, /* is_write= */ false);
```
JanMatCodasip: You can simplify this part of the code:

```suggestion
	bool use_aampostincrement =
			!ac_cache_contains(&info->ac_not_supported_cache, command);
	
	uint32_t command = access_memory_command(target, false, width,
			use_aampostincrement, /* is_write= */ false);	
```

src/target/riscv/riscv-013.c
```diff
@@ -3749,33 +3783,14 @@ read_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
-				command = access_memory_command(target, false, width, false, false);
-				result = riscv013_execute_abstract_command(target, command, &cmderr);
-				if (result == ERROR_OK) {
-					LOG_TARGET_DEBUG(target, "aampostincrement is not supported on this target.");
-					info->has_aampostincrement = YNM_NO;
-				}
-			}
+		if (use_aampostincrement && result != ERROR_OK &&
+					cmderr == CMDERR_NOT_SUPPORTED) {
+			LOG_TARGET_DEBUG(target, "aampostincrement is not supported on this target.");
```
JanMatCodasip: The command can result in cmderr==UNSUPPORTED due to other reasons than `command.aampostincrement` - for example due to `command.aamvirtual`.

For that reason, please adjust the message. For example:

```suggestion
			LOG_TARGET_DEBUG(target, "Trying the same abstract memory write command, but without aampostincrement");
```

src/target/riscv/riscv-013.c
```diff
@@ -3811,13 +3826,18 @@ write_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
 	int result = ERROR_OK;
-	bool use_aampostincrement = info->has_aampostincrement != YNM_NO;
 
 	/* Convert the size (bytes) to width (bits) */
 	unsigned int width = args.size << 3;
 
 	/* Create the command (physical address, postincrement, write) */
-	uint32_t command = access_memory_command(target, false, width, use_aampostincrement, true);
+	uint32_t command = access_memory_command(target, false, width,
+			/* postincrement = */ true, /* is_write = */ true);
```
JanMatCodasip: This part of code can be simplified (the `if` is redundant) - same as mentioned above.

src/target/riscv/riscv-013.c
```diff
@@ -3844,33 +3864,14 @@ write_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
-				command = access_memory_command(target, false, width, false, true);
-				result = riscv013_execute_abstract_command(target, command, &cmderr);
-				if (result == ERROR_OK) {
-					LOG_TARGET_DEBUG(target, "aampostincrement is not supported on this target.");
-					info->has_aampostincrement = YNM_NO;
-				}
-			}
+		if (use_aampostincrement && result != ERROR_OK &&
+					cmderr == CMDERR_NOT_SUPPORTED) {
+			LOG_TARGET_DEBUG(target, "aampostincrement is not supported on this target.");
```

```suggestion
			LOG_TARGET_DEBUG(target, "Trying the same abstract memory read command, but without aampostincrement");
```

src/target/riscv/riscv-013.c
```diff
@@ -885,17 +928,16 @@ static int register_write_abstract(struct target *target, enum gdb_regno number,
...
-	if (number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095 &&
-			!info->abstract_write_csr_supported)
-		return ERROR_FAIL;
-
 	const unsigned int size_bits = register_size(target, number);
 	const uint32_t command = riscv013_access_register_command(target, number, size_bits,
 			AC_ACCESS_REGISTER_TRANSFER |
 			AC_ACCESS_REGISTER_WRITE);
+	if (ac_cache_contains(&info->ac_not_supported_cache, command)) {
+		LOG_TARGET_DEBUG(target, "Command is cached as not supported, bailing out");
```
JanMatCodasip: In all the cache-related prints: instead of saying just `command`, could you please say `abstract command` (or even better would be `abstract command 0x% " PRIx32 "`).

The reason is that the word "command" alone is too generic and can mean many things (for example TCL command, GDB command, ...).

JanMatCodasip (changes requested):
Thank you for this improvement. Overall, the code looks all right. I have found several things to address, though. Could you please check my comments.

src/target/riscv/riscv-013.c
```diff
@@ -3811,13 +3826,18 @@ write_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
 
 	/* Create the command (physical address, postincrement, write) */
-	uint32_t command = access_memory_command(target, false, width, use_aampostincrement, true);
+	uint32_t command = access_memory_command(target, false, width,
+			/* postincrement = */ true, /* is_write = */ true);
+	bool use_aampostincrement =
+			!ac_cache_contains(&info->ac_not_supported_cache, command);
+	if (!use_aampostincrement)
+		command = access_memory_command(target, false, width,
+				/* postincrement = */ false, /* is_write= */ false);
```
JanMatCodasip: 1) There is a typo - `is_write` must be true.

2) Could you please check if abstract memory commands are in fact tested in the CI?

```suggestion
	if (!use_aampostincrement)
		command = access_memory_command(target, false, width,
				/* postincrement = */ false, /* is_write= */ true);
```

src/target/riscv/riscv-013.c
fkhaidari: Thanks! Addressed

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -145,6 +145,60 @@ typedef struct {
...
 
+struct ac_cache {
+	uint32_t *commands;
+	size_t size;
+};
+
+static int ac_cache_elem_comparator(const void *lhs, const void *rhs)
+{
+	return *(const uint32_t *)rhs - *(const uint32_t *)lhs;
+}
```
JanMatCodasip: There seems to be a hidden bug in this comparator:

- If `int` is 32 bits, this function will work as expected (but there is a hidden implicit conversion from uint32 to int that can change the sign of the result, and gcc would complain if `-Wconversion` is enabled).
- If `int` type is 64 bits wide, then this function would never return a negative value. (Also we would probably not get any `-Wconversion` notification.)

For that reason, please consider re-writing using if + comparison instead of subtraction:

```suggestion
static int ac_cache_elem_comparator(const void *p_lhs, const void *p_rhs)
{
	unint32_t lhs = *(const uint32_t *)p_lhs;
	unint32_t rhs = *(const uint32_t *)p_rhs;
	
	if (lhs < rhs)
		return -1;
	if (lhs > rhs)
		return 1;	
	return 0;
}
```

src/target/riscv/riscv-013.c
```diff
@@ -829,38 +902,35 @@ uint32_t riscv013_access_register_command(struct target *target, uint32_t number
...
+{
+	bool supported = !ac_cache_contains(&get_info(target)->ac_not_supported_cache, command);
+	if (supported)
+		return false;
+
+	LOG_TARGET_DEBUG(target, "Abstract command 0x%"
+			PRIx32 " is cached as not supported", command);
+	log_debug_reg(target, get_cmdtype(command),
+			command, __FILE__, __LINE__, __func__);
+	return supported;
```
JanMatCodasip: 1) There seems to be a bug - this function always returns `false`.

2) Please, don't name the variable `supported`. The fact that the command is not in the cache does not mean that it is supported. You can call it `already_seen_unsupported`, `is_unsupported`, `known_unsupported` or similar.

```suggestion
	bool already_seen_unsupported = ac_cache_contains(&get_info(target)->ac_not_supported_cache, command);
	if (!already_seen_unsupported)
		return false;

	LOG_TARGET_DEBUG(target, "Abstract command 0x%"
			PRIx32 " is cached as not supported", command);
	log_debug_reg(target, get_cmdtype(command),
			command, __FILE__, __LINE__, __func__);
	return true;
```

src/target/riscv/riscv-013.c
```diff
@@ -3749,33 +3799,15 @@ read_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
-				result = riscv013_execute_abstract_command(target, command, &cmderr);
-				if (result == ERROR_OK) {
-					LOG_TARGET_DEBUG(target, "aampostincrement is not supported on this target.");
-					info->has_aampostincrement = YNM_NO;
-				}
-			}
+		if (use_aampostincrement && result != ERROR_OK &&
+					cmderr == CMDERR_NOT_SUPPORTED) {
+			LOG_TARGET_DEBUG(target, "Trying the same abstract memory "
+					"write command, but without aampostincrement");
```
JanMatCodasip: Copy-paste typo

```suggestion
			LOG_TARGET_DEBUG(target, "Trying the same abstract memory "
					"read command, but without aampostincrement");
```

src/target/riscv/riscv-013.c
```diff
@@ -3719,16 +3768,17 @@ read_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
 	/* Convert the size (bytes) to width (bits) */
 	unsigned int width = args.size << 3;
 
 	/* Create the command (physical address, postincrement, read) */
-	uint32_t command = access_memory_command(target, false, width, use_aampostincrement, false);
+	uint32_t command = access_memory_command(target, false, width,
+			/* postincrement = */ true, /* is_write= */ false);
+	bool use_aampostincrement = !is_command_unsupported(target, command);
+	command = access_memory_command(target, false, width,
+			/* postincrement = */ use_aampostincrement, /* is_write = */ false);
```
JanMatCodasip: Why assign the variable `command` twice?

Please use just:

```c
bool use_aampostincrement = !is_command_unsupported(target, command);
uint32_t command = access_memory_command(target, false, width,
			/* postincrement = */ use_aampostincrement, /* is_write = */ false);
```

src/target/riscv/riscv-013.c
```diff
@@ -3809,15 +3841,17 @@ write_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
 	/* Convert the size (bytes) to width (bits) */
 	unsigned int width = args.size << 3;
 
 	/* Create the command (physical address, postincrement, write) */
-	uint32_t command = access_memory_command(target, false, width, use_aampostincrement, true);
+	uint32_t command = access_memory_command(target, false, width,
+			/* postincrement = */ true, /* is_write = */ true);
+	bool use_aampostincrement = !is_command_unsupported(target, command);
+	command = access_memory_command(target, false, width,
+			/* postincrement = */ use_aampostincrement, /* is_write = */ true);
```

Why assign the variable `command` twice?

src/target/riscv/riscv-013.c
```diff
@@ -145,6 +145,60 @@ typedef struct {
...
+{
+	assert(cache);
+
+	size_t old_size = cache->size;
+	size_t new_size = old_size + 1;
+	size_t entry_size = sizeof(*cache->commands);
+
+	uint32_t *commands = realloc(cache->commands, new_size * entry_size);
+	if (!commands) {
+		LOG_ERROR("Reallocation to %zu bytes failed", new_size);
```
```suggestion
		LOG_ERROR("Reallocation to %zu bytes failed", new_size * entry_size);
```

JanMatCodasip (changes requested):
@fk-sc thank you for addressing the findings so far.

I came across few more things - if you can please look at them.

src/target/riscv/riscv-013.c
fkhaidari: `is_command_unsupported` requires `command` as argument, it can't be defined after usage.

src/target/riscv/riscv-013.c
fkhaidari: `is_command_unsupported` requires `command` as argument, it can't be defined after usage.

src/target/riscv/riscv-013.c
JanMatCodasip: You're right, thank you for explanation, I missed it.

In that case the original code with the `if` actually looks more fitting. (Plus comments can help the reader - if you like).

I am sorry for this back-and-forth.

```c
	/* By default, use abstract memory access with aampostincrement=1 for improved performance. */
	uint32_t command = access_memory_command(target, false, width,
			/* postincrement = */ true, /* is_write= */ false);

	bool use_aampostincrement = !is_command_unsupported(target, command);
	if (!use_aampostincrement)
		/* It is already known that this abstract memory access with aampostincrement=1 is not supported.
		 * So try aampostincrement=0 right away. */
		command = access_memory_command(target, false, width, false, false);
```

src/target/riscv/riscv-013.c
JanMatCodasip: I mis-read this part (same as [above](https://github.com/riscv-collab/riscv-openocd/pull/1235/files#r2002556236)), sorry.

If you could please use the original code with the "if" (+ comments), that would be good. 

Thank you.

src/target/riscv/riscv-013.c
fkhaidari: Thanks for the catch! Addressed

src/target/riscv/riscv-013.c
fkhaidari: Thanks! Addressed

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
fkhaidari: Addressed. Sorry for silly mistakes from my side

src/target/riscv/riscv-013.c
```diff
@@ -145,6 +145,62 @@ typedef struct {
...
+struct ac_cache {
+	uint32_t *commands;
+	size_t size;
+};
+
+static int ac_cache_elem_comparator(const void *plhs, const void *prhs)
+{
+	const uint32_t lhs = *(const uint32_t *)plhs;
+	const uint32_t rhs = *(const uint32_t *)prhs;
+	return (lhs > rhs) - (lhs < rhs);
```
JanMatCodasip: Really a minor detail: p_lhs and p_rhs is bit more readable than plhs, prhs.

```suggestion
static int ac_cache_elem_comparator(const void *p_lhs, const void *p_rhs)
{
	const uint32_t lhs = *(const uint32_t *)p_lhs;
	const uint32_t rhs = *(const uint32_t *)p_rhs;
	return (lhs > rhs) - (lhs < rhs);
```

src/target/riscv/riscv-013.c
JanMatCodasip: No problem, thanks.

JanMatCodasip (commented):
Overall this MR looks good. I will do one last proper review tomorrow.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -3809,15 +3847,20 @@ write_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
-	uint32_t command = access_memory_command(target, false, width, use_aampostincrement, true);
+	uint32_t command = access_memory_command(target, /* virtual = */ false,
+			width, /* postincrement = */ true, /* is_write = */ true);
+	bool use_aampostincrement = !is_command_unsupported(target, command);
+	if (!use_aampostincrement)
+		/* It is already known that this abstract memory
+		 * access with aampostincrement=1 is not supported.
+		 * So try aampostincrement=0 right away. */
+		command = access_memory_command(target, /* virtual = */ false,
+				width, use_aampostincrement, /* is_write = */ true);
```
JanMatCodasip: Pedantic detail, but helps readability:  You can use `false` directly instead of `use_aampostincrement`, which makes it a bit easier for the code readers.

```suggestion
		command = access_memory_command(target, /* virtual = */ false,
				width, /* postincrement = */ false, /* is_write = */ true);
```

src/target/riscv/riscv-013.c
```diff
@@ -3844,33 +3887,16 @@ write_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
-			}
+		if (use_aampostincrement && result != ERROR_OK &&
+					cmderr == CMDERR_NOT_SUPPORTED) {
+			LOG_TARGET_DEBUG(target, "Trying the same abstract memory "
+					"write command, but without aampostincrement");
+			use_aampostincrement = false;
+
+			/* Try the same access but with postincrement disabled. */
+			command = access_memory_command(target, /* virtual = */ false,
+					width, use_aampostincrement, /* is_write = */ true);
```

```suggestion
			command = access_memory_command(target, /* virtual = */ false,
					width, /* postincrement = */ false, /* is_write = */ true);
```

src/target/riscv/riscv-013.c
```diff
@@ -3749,33 +3804,16 @@ read_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
-			}
+		if (use_aampostincrement && result != ERROR_OK &&
+					cmderr == CMDERR_NOT_SUPPORTED) {
+			LOG_TARGET_DEBUG(target, "Trying the same abstract memory "
+					"read command, but without aampostincrement");
+			use_aampostincrement = false;
+
+			/* Try the same access but with postincrement disabled. */
+			command = access_memory_command(target, /* virtual = */ false,
+					width, use_aampostincrement, /* is_write = */ false);
```
JanMatCodasip: (same)

```suggestion
			command = access_memory_command(target, /* virtual = */ false,
					width, /* postincrement = */ false, /* is_write = */ false);
```

src/target/riscv/riscv-013.c
```diff
@@ -3719,16 +3770,20 @@ read_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
-	uint32_t command = access_memory_command(target, false, width, use_aampostincrement, false);
+	uint32_t command = access_memory_command(target, /* virtual = */ false,
+			width, /* postincrement = */ true, /* is_write = */ false);
+	bool use_aampostincrement = !is_command_unsupported(target, command);
+	if (!use_aampostincrement)
+		/* It is already known that this abstract memory
+		 * access with aampostincrement=1 is not supported.
+		 * So try aampostincrement=0 right away. */
+		command = access_memory_command(target, /* virtual = */ false,
+				width, use_aampostincrement, /* is_write = */ false);
```
JanMatCodasip: (same)

```suggestion
		command = access_memory_command(target, /* virtual = */ false,
				width, /* postincrement = */ false, /* is_write = */ false);
```

src/target/riscv/riscv-013.c
```diff
@@ -145,6 +145,62 @@ typedef struct {
...
+struct ac_cache {
+	uint32_t *commands;
+	size_t size;
+};
+
+static int ac_cache_elem_comparator(const void *p_lhs, const void *p_rhs)
+{
+	const uint32_t lhs = *(const uint32_t *)p_lhs;
+	const uint32_t rhs = *(const uint32_t *)p_rhs;
+	return (lhs > rhs) - (lhs < rhs);
```
JanMatCodasip: This is a creative way how to return the proper value. Still, I am afraid that it makes the code harder to grasp for readers.

For that reason, and if you don't mind, I'd prefer the (more boring but very clear) version with the `if`s:

```suggestion
	if (lhs < rhs)
		return -1;		
	if (lhs > rhs)
		return 1;
	return 0;
```

src/target/riscv/riscv-013.c
```diff
@@ -145,6 +145,62 @@ typedef struct {
...
 
+struct ac_cache {
+	uint32_t *commands;
+	size_t size;
+};
+
+static int ac_cache_elem_comparator(const void *p_lhs, const void *p_rhs)
+{
+	const uint32_t lhs = *(const uint32_t *)p_lhs;
+	const uint32_t rhs = *(const uint32_t *)p_rhs;
```
JanMatCodasip: (Just a note - no change requested)

I am OK with the `const` in front of these variables, and would probably put it there myself without much thinking about it.

However, in upstream [I got a negative reaction on these](https://review.openocd.org/c/openocd/+/8750/comment/984b8414_f6f804e4/) - @tom-van is not fond of seeing the `const` in very simple cases like this.

No change needed from my point of view. Just wanted to point out that you may meet with different opinions.

src/target/riscv/riscv-013.c
```diff
@@ -3749,33 +3804,16 @@ read_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
-					info->has_aampostincrement = YNM_NO;
-				}
-			}
+		if (use_aampostincrement && result != ERROR_OK &&
+					cmderr == CMDERR_NOT_SUPPORTED) {
+			LOG_TARGET_DEBUG(target, "Trying the same abstract memory "
+					"read command, but without aampostincrement");
+			use_aampostincrement = false;
+
+			/* Try the same access but with postincrement disabled. */
```
JanMatCodasip: Pedantic detail: You can now remove this comment because the log print above already says the same thing.

```suggestion
```

src/target/riscv/riscv-013.c
```diff
@@ -3844,33 +3887,16 @@ write_memory_abstract(struct target *target, const riscv_mem_access_args_t args)
...
-					info->has_aampostincrement = YNM_NO;
-				}
-			}
+		if (use_aampostincrement && result != ERROR_OK &&
+					cmderr == CMDERR_NOT_SUPPORTED) {
+			LOG_TARGET_DEBUG(target, "Trying the same abstract memory "
+					"write command, but without aampostincrement");
+			use_aampostincrement = false;
+
+			/* Try the same access but with postincrement disabled. */
```

This comment can be removed because the log print right above says the same thing.

```suggestion
```

src/target/riscv/riscv-013.c
JanMatCodasip: It looks like the algorithm could be improved to also detect early unsupported memory read with aampostincrement=0.

No need to do it in this MR request, unless you want.

A simple TODO seems enough for now.

```suggestion
	if (!use_aampostincrement) {
		/* It is already known that this abstract memory
		 * access with aampostincrement=1 is not supported.
		 * So try aampostincrement=0 right away. */
		command = access_memory_command(target, /* virtual = */ false,
				width, /* postincrement = */ false, /* is_write = */ false);
		if (is_command_unsupported(target, command)) {
			LOG_TARGET_DEBUG(target, "Abstract memory read not supported (with or without aampostincrement)")
			return mem_access_result(...);
		 }
	}				
```

src/target/riscv/riscv-013.c

Unsupported write with aampostincrement=0 could be detected early, too.

You can either do it in this MR or leave a TODO.

```suggestion
	if (!use_aampostincrement) {
		/* It is already known that this abstract memory
		 * access with aampostincrement=1 is not supported.
		 * So try aampostincrement=0 right away. */
		command = access_memory_command(target, /* virtual = */ false,
				width, use_aampostincrement, /* is_write = */ true);
		if (is_command_unsupported(target, command)) {
			LOG_TARGET_DEBUG(target, "Abstract memory write not supported (with or without aampostincrement)")
			return mem_access_result(...);
		 }
	}
```

JanMatCodasip (commented):
Overall, this looks fine. I am sending my last batch of comments, mostly it is minor coding stuff.

I do not expect to have any more comments than that.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
fkhaidari: Decided to remove `const` qualifier

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
```diff
@@ -651,6 +706,27 @@ static int abstract_cmd_batch_check_and_clear_cmderr(struct target *target,
...
+{
+	enum riscv_debug_reg_ordinal val2cmdtype[4] = {
+		AC_ACCESS_REGISTER_ORDINAL,
+		AC_QUICK_ACCESS_ORDINAL,
+		AC_ACCESS_MEMORY_ORDINAL,
+		0,
+	};
+
+	return val2cmdtype[get_field(command, DM_COMMAND_CMDTYPE)];
+}
```
JanMatCodasip: Last detail:

1) Curious, why is the fourth item in the array (the zero) present? Is it in fact needed? 

2) For additional safety, could you please add an assertion to guard against cases when cmdtype is unexpected?

For example this way (or similar):

```c
switch (get_field(command, DM_COMMAND_CMDTYPE)) {
case 0:
	return AC_ACCESS_REGISTER_ORDINAL;
case 1:
	return AC_QUICK_ACCESS_ORDINAL;
case 2:
	return AC_ACCESS_MEMORY_ORDINAL;
default:
	assert(false);
	return 0;
}
```

JanMatCodasip (dismissed):
LGTM, thank you.

src/target/riscv/riscv-013.c
fkhaidari: Thanks. Addressed

JanMatCodasip (dismissed):

src/target/riscv/riscv-013.c
en-sc: @fk-sc, please, let's not do this in the same commit.
Current implementation returns different errors for the two consequtive runs of the same command in case abstract memory access is not supported:
`MEM_ACCESS_SKIPPED_ABSTRACT_ACCESS_CMDERR` vs `MEM_ACCESS_SKIPPED_AC_NOT_SUPPORTED`.

I believe this is confusing and should be addressed separately.
Please, add a `TODO` comment instead.

src/target/riscv/riscv-013.c
en-sc: Same as for read. Please, consider a `TODO` comment instead.

en-sc (changes requested):
@JanMatCodasip, @fk-sc, I've re-opened some of the discussions.
IMHO, the suggested improvement to abstract memory accesses should not be implemented in the same commit. Please, take a look.

src/target/riscv/riscv-013.c
JanMatCodasip: I do not feel strongly. If @en-sc prefers to make this improvement in a separate PR, that is fine with me.

src/target/riscv/riscv-013.c

JanMatCodasip (approved):

en-sc (approved):

src/target/riscv/riscv-013.c
```diff
@@ -145,6 +145,66 @@ typedef struct {
...
+	cache->commands = commands;
+	cache->size = new_size;
+
+	qsort(cache->commands, cache->size, entry_size,
+			ac_cache_elem_comparator);
+}
+
+static bool ac_cache_contains(const struct ac_cache *cache, uint32_t command)
+{
+	return bsearch(&command, cache->commands, cache->size,
```
aap-sc, line 204: this MR actually adds a subtle UB , according to ubsan. cache->commands may be NULL (when `cache>size == 0`)

aap-sc: MR with the fix: https://github.com/riscv-collab/riscv-openocd/pull/1297

PR #1240, opened 2025-03-14
target/riscv: avoid `config` modification on `jim_getopt_obj()` failure

en-sc:
Currently, `jim_getopt_obj()` only fails if `goi->argc` is zero, Link: https://github.com/riscv-collab/riscv-openocd/blob/41a225460c3b9a6c1f61a0777f101ff009f56007/src/helper/jim-nvp.c#L174-L185 so the check at the start of `jim_configure_ebreak()` Link: https://github.com/riscv-collab/riscv-openocd/blob/41a225460c3b9a6c1f61a0777f101ff009f56007/src/target/riscv/riscv.c#L526-L530 guarantees that the call will succeed.

However, the modification makes the code more robust and future-proof.

Change-Id: Ic8c2e057a285bf679d26e21bda138a1d2ae5d5ce

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -534,9 +534,12 @@ static int jim_configure_ebreak(struct riscv_private_config *config, struct jim_
 		/* Here a common "ebreak" action is processed, e.g:
 		 * "riscv.cpu configure -ebreak halt"
 		 */
+		int res = jim_getopt_obj(goi, NULL);
+		if (res != JIM_OK)
+			return res;
 		for (int ebreak_ctl_i = 0; ebreak_ctl_i < N_RISCV_MODE; ++ebreak_ctl_i)
 			config->dcsr_ebreak_fields[ebreak_ctl_i] = common_mode_nvp->value;
-		return jim_getopt_obj(goi, NULL);
+		return JIM_OK;
 	}
 
 	/* Here a "ebreak" action for a specific execution mode is processed, e.g:


src/target/riscv/riscv.c
```diff
@@ -534,9 +534,12 @@ static int jim_configure_ebreak(struct riscv_private_config *config, struct jim_
...
 		 * "riscv.cpu configure -ebreak halt"
 		 */
+		int res = jim_getopt_obj(goi, NULL);
+		if (res != JIM_OK)
+			return res;
 		for (int ebreak_ctl_i = 0; ebreak_ctl_i < N_RISCV_MODE; ++ebreak_ctl_i)
 			config->dcsr_ebreak_fields[ebreak_ctl_i] = common_mode_nvp->value;
-		return jim_getopt_obj(goi, NULL);
+		return JIM_OK;
 	}
```
JanMatCodasip, line 543: (No change requested - just a question for my understanding)

Do I understand correctly that `jim_getopt_obj(goi, ptr_to_Jim_Obj_or_NULL)` is supposed to pop the first argument in the argument list (goi), and put it into the Jim_Obj pointed by the second argument (if not NULL)?

Does it mean that if we wanted to strictly obey the JimTcl API, we would not be reading `goi->argv` directly, but instead we would get the values of arguments using jim_getopt_obj(), as done e.g. [here](https://github.com/riscv-collab/riscv-openocd/blob/41a225460c3b9a6c1f61a0777f101ff009f56007/src/target/arm_tpiu_swo.c#L971-L977)?

JanMatCodasip (approved):
LGTM.

src/target/riscv/riscv.c
en-sc: > Do I understand correctly that jim_getopt_obj(goi, ptr_to_Jim_Obj_or_NULL) is supposed to pop the first argument in the argument list (goi), and put it into the Jim_Obj pointed by the second argument (if not NULL)?

AFAIK, yes.

> Does it mean that if we wanted to strictly obey the JimTcl API, we would not be reading goi->argv directly, but instead we would get the values of arguments using jim_getopt_obj()...

There is a whole plethora of `jim_getopt_*()` functions in [`src/helper/jim-nvp.h`](https://github.com/riscv-collab/riscv-openocd/blob/41a225460c3b9a6c1f61a0777f101ff009f56007/src/helper/jim-nvp.h).

E.g. for the call you have mentioned, [`jim_getopt_string()`](https://github.com/riscv-collab/riscv-openocd/blob/41a225460c3b9a6c1f61a0777f101ff009f56007/src/helper/jim-nvp.h#L224) would be better suited IMHO.

However, sometimes we need to peek at the value without popping it (and there is no push in the API). This is why I access `goi->argv` directly [in this function](https://github.com/riscv-collab/riscv-openocd/blob/ece1827613d527d95aced1a1952e92ee86af56f9/src/target/riscv/riscv.c#L532).

JanMatCodasip: Ok, thanks for explanation. I don't mind the direct access to `goi->argv`, I was just curious.

PR #1241, opened 2025-03-18
target/riscv: add is_virtual parameter to memory access method

fkhaidari:
Added is_virtual parameter to memory access method. This change is required to support hw translation appropriately.

Updated repeat_read documentation according to new behaviour. It is provide physical access now

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -155,6 +155,13 @@ bool riscv_virt2phys_mode_is_sw(const struct target *target)
 	return r->virt2phys_mode == RISCV_VIRT2PHYS_MODE_SW;
 }
 
+bool riscv_virt2phys_mode_is_off(const struct target *target)
+{
+	assert(target);
+	RISCV_INFO(r);
+	return r->virt2phys_mode == RISCV_VIRT2PHYS_MODE_OFF;
+}
+
 const char *riscv_virt2phys_mode_to_str(riscv_virt2phys_mode_t mode)
 {
 	assert(mode == RISCV_VIRT2PHYS_MODE_OFF
@@ -3149,7 +3156,7 @@ static int riscv_address_translate(struct target *target,
 			.increment = 4,
 			.count = (1 << info->pte_shift) / 4,
 		};
-		int retval = r->access_memory(target, args);
+		int retval = r->access_memory(target, args, /* is_virtual */ false);
 		if (retval != ERROR_OK)
 			return ERROR_FAIL;
 
@@ -3382,6 +3389,14 @@ static int check_virt_memory_access(struct target *target, target_addr_t address
 	return ERROR_OK;
 }
 
+static int riscv_access_phys_memory(struct target *target,
+	const riscv_mem_access_args_t args)
+{
+	RISCV_INFO(r);
+	return r->access_memory(target, args, /* is_virtual */ false);
+}
+
+
 static int riscv_read_phys_memory(struct target *target, target_addr_t phys_address,
 			uint32_t size, uint32_t count, uint8_t *buffer)
 {
@@ -3392,8 +3407,7 @@ static int riscv_read_phys_memory(struct target *target, target_addr_t phys_addr
 		.count = count,
 		.increment = size,
 	};
-	RISCV_INFO(r);
-	return r->access_memory(target, args);
+	return riscv_access_phys_memory(target, args);
 }
 
 static int riscv_write_phys_memory(struct target *target, target_addr_t phys_address,
@@ -3406,12 +3420,11 @@ static int riscv_write_phys_memory(struct target *target, target_addr_t phys_add
 		.count = count,
 		.increment = size,
 	};
-
-	RISCV_INFO(r);
-	return r->access_memory(target, args);
+	return riscv_access_phys_memory(target, args);
 }
 
-static int riscv_rw_memory(struct target *target, const riscv_mem_access_args_t args)
+static int riscv_access_virt_memory(struct target *target,
+		const riscv_mem_access_args_t args)
 {
 	assert(riscv_mem_access_is_valid(args));
 
@@ -3423,8 +3436,11 @@ static int riscv_rw_memory(struct target *target, const riscv_mem_access_args_t 
 	}
 
 	RISCV_INFO(r);
+	if (riscv_virt2phys_mode_is_off(target))
+		return r->access_memory(target, args, /* is_virtual */ false);
+
 	if (riscv_virt2phys_mode_is_hw(target))
-		return r->access_memory(target, args);
+		return r->access_memory(target, args, /* is_virtual */ true);
 
 	int result = check_virt_memory_access(target, args.address,
 			args.size, args.count, is_write);
@@ -3456,7 +3472,8 @@ static int riscv_rw_memory(struct target *target, const riscv_mem_access_args_t 
 		else
 			current_access.read_buffer += current_count * args.size;
 
-		result = r->access_memory(target, current_access);
+		result = r->access_memory(target,
+				current_access, /* is_virtual */ false);
 		if (result != ERROR_OK)
 			return result;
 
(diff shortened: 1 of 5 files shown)


fkhaidari:
@JanMatCodasip, @MarekVCodasip could you please take a look?

fkhaidari:
In [this comment](https://github.com/riscv-collab/riscv-openocd/pull/1167#discussion_r1860733422), @JanMatCodasip suggested moving the `is_virtual` field out of the `riscv_mem_access_args_t` structure. However, while implementing `is_virtual` across progbuf, sysbus, and abstract memory access types, I’ve noticed the flag is becoming pervasive and required multiple places. Given this, should we reconsider keeping it in `riscv_mem_access_args_t` to avoid redundancy and centralize its handling?

JanMatCodasip:
> fk-sc: In [this comment](https://github.com/riscv-collab/riscv-openocd/pull/1167#discussion_r1860733422), JanMatCodasip suggested moving the is_virtual field out of the riscv_mem_access_args_t structure. [...]

I do not feel strongly about this one. Please feel free to use and submit whatever variant in your opinion leads to cleaner, more understandable code.

My concern was that `is_virtual` would be passed (as part of the structure) to functions for which it is not relevant. If those functions are in minority, maybe it is an acceptable compromise.

fkhaidari:
@JanMatCodasip, I’ve decided to keep `is_virtual` as a separate parameter for now. I can revisit integrating it into the riscv_mem_access_args_t structure once the progbuf, sysbus, and abstract memory subsystems are finalized

src/target/riscv/riscv.c
```diff
@@ -5228,7 +5229,9 @@ COMMAND_HANDLER(handle_repeat_read)
 		.count = count,
 		.increment = 0,
 	};
-	int result = r->access_memory(target, args);
+	/* TODO: Add a command parameter that enables
+		choosing between virtual and physical access */
```
en-sc: Please, fix the formatting

en-sc (changes requested):

src/target/riscv/riscv.c

src/target/riscv/riscv.c
```diff
@@ -3425,7 +3425,7 @@ static int riscv_rw_memory(struct target *target, const riscv_mem_access_args_t
 
 	RISCV_INFO(r);
 	if (!mmu_enabled)
-		return r->access_memory(target, args);
+		return r->access_memory(target, args, /* is_virtual */ true);
```
JanMatCodasip, line 3443: Could you please double-check this line.

If `mmu_enabled` is false, shouldn't `is_virtual` be false as well?

fkhaidari: Addressed

aap-sc: @fk-sc can we discuss this design once again?

1. I don't like the suggested approach because riscv_rw_memory should always operate on virtual addresses by design.  The decision on how the actual address translation is performed should be decided by lower level functions - that is I would expect for `r->access_memory()` to be called with `is_virutual = true`. The fact that in some execution context virtual address is equal to physical is irrelevant at this level of abstraction.

2. if there is a necessity to provide additional layer to simplify page-crossing translations - we can do it here and establish a contract that `r->access_memory` should access non-page crossing memory regions.

3. please rename `mmu_enabled` to `satp_translation_required`. `mmu_enabled` is too broad term that can confuse the reader (it definitely confuses me)

@JanMatCodasip sorry for the mess, most likely we @fk-sc , @en-sc  and me should have yet another round of internal discussion related to the above bulllets. Please consider this MR as a draft for now. You inputs are as always are very welcome, so please do share concerns/questions/suggestions regarding the matter (if you have any)

JanMatCodasip: No problem at all. Thank you for letting me know.

> aap-sc: [...] riscv_rw_memory should always operate on virtual addresses by design.

Sounds good to me. 

In that case, if you like, you can also make the code more clear by doing this:
- rename `riscv_rw_memory()` to `riscv_rw_virt_memory()`
- create `riscv_rw_phys_memory()` (a thin wrapper over `r->access_memory(args, /* is_virtual = */ false)`)

> aap-sc: if there is a necessity to provide additional layer to simplify page-crossing translations - we can do it here and establish a contract that r->access_memory should access non-page crossing memory regions.

- The algorithm for breaking the memory access into the page-sized chunks, which is already in riscv_rw_memory(), looks all right to me. Or do you see any gap or need for change there, which I might have missed?
- Do I understand correctly that the only case when we need to manually break the memory transfer to page-sized pieces is when `virt2phys_mode` is set to `SW`?

> aap-sc: please rename mmu_enabled to satp_translation_required. mmu_enabled is too broad term that can confuse the reader

I agree that `mmu_enabled` is not a clear name. It could also be called `manual_satp_translation_needed`. The word "manual" is IMO important here as it means that OpenOCD must do the translation work (as opposed to the hardware).

Furthermore, I am concerned about the `riscv_mmu()` itself. If `virt2phys` mode is set to `HW`, then `riscv_mmu()` returns False, which looks incorrect. As a result:
- The `riscv_virt2phys()` function (and the TCL command `virt2phys`) will return 1:1 mapping instead of correctly translated addresses.
- The `target_alloc_working_area_try()` will allocate the work area in the physical address range (`-work-area-phys`) instead of virtual (`-work-area-virt`).

Should the [initial check](https://github.com/riscv-collab/riscv-openocd/blob/2605cbd73c36998e76d32cded95a02b424870626/src/target/riscv/riscv.c#L3020) in riscv_mmu() be fixed this way?

```c
	if (riscv_virt2phys_mode_is_off(target))
		return ERROR_OK;
```

fkhaidari:
@JanMatCodasip, @aap-sc, @en-sc, I've revised the implementation of this PR. At first we need to merge #1256 (this commit fixes riscv_mmu). This PR includes mentioned one. Changes in updated version of PR:
* Renamed `riscv_rw_memory()` to `riscv_access_virt_memory()`
* Created `riscv_access_phys_memory()` (a thin wrapper over `r->access_memory(args, /* is_virtual = */ false)`)

JanMatCodasip:
@fk-sc If you don't mind, I will get back to this review at a later point - after we finish the discussion under #1256.

JanMatCodasip (commented):
IIRC, this was supposed to be an initial patch, and series of other changes around the memory access functions and virtual vs. physical addresses should follow.

@en-sc, @fkhaidari, can you please remind me what other changes you still intend to make on top of this?

Preferably, the patches should now go to review.openocd.org.

fkhaidari:
Hi @JanMatCodasip, thanks for getting back to this.

Yes, #1241 is the initial/foundational patch. On top of it we plan the following changes, roughly in order:

1. **Virtual access via program buffer.** Implement the `is_virtual` path for progbuf memory access.
2. **Virtual access via abstract command.** Same for the abstract-command access method.
3. **`riscv_mmu` clarification.** Rename the parameter `enabled` to `satp_translation_required` and move the translation-driver check out of `riscv_mmu`. This depends on #1256, which fixes `riscv_mmu`.
4. **virt2phys mode handling.** Handle the different virt2phys translation modes in the virtual memory access path.
5. **Force physical addressing before physical progbuf access** (YCAT-41723).
6. **Testsuite.** Update Spike and virtual-translation tests to cover the new behaviour.

We may also revisit folding `is_virtual` back into `riscv_mem_access_args_t`, as you originally suggested, once the progbuf/sysbus/abstract access methods are finalized. We'll only do that if it actually leads to cleaner code.

Moving the review to review.openocd.org sounds good to us.

JanMatCodasip:
Hi @fkhaidari,

thank you for your reply and the clear explanation of your further intentions around the virtual/physical memory access.

All looks fine to me - I do not have any concerns about any of the proposed steps.

> We may also revisit folding is_virtual back into riscv_mem_access_args_t, as you originally suggested, once the progbuf/sysbus/abstract access methods are finalized. We'll only do that if it actually leads to cleaner code.

I consider this a minor thing. Whatever solution that leads to clean code can be used. My only recommendation is to not pass the is_virtual flag in any form (in structure or as a standalone parameters) into functions that don't make any use of it. So for that reason, I am more inclined to leave `is_virtual` as a standalone parameter.

PR #1243, opened 2025-03-31
target/riscv: Add support for external triggers

lz-bro:
Add support for associating a halt/resume group with an external trigger via a newly exposed configuration option "riscv smp_add_ext_triggers".

Original merge request:
https://github.com/riscv-collab/riscv-openocd/pull/1179

Original author: Rob Bradford
rbradford@rivosinc.com
https://github.com/rbradford

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -11673,6 +11673,20 @@ The second argument configures how OpenOCD should use the selected trigger featu
 With no parameters, prints current trigger features configuration.
 @end deffn
 
+@deffn {Command} {riscv set_group} grouptype group ['trigger' triggernum]
+OpenOCD assigns a hart or a given external trigger into a RISC-V halt group.
+This command associates current target or supplied external trigger(s) with this halt group.
+
+@option{trigger} and @var{triggernum} Specify the index of the external trigger.
+When an external input trigger fires, the harts in the halt group will get halted.
+Similarly, all external output triggers will be notified when a hart in the halt group halts.
+
+The only supported value of @var{grouptype} is @option{halt_group}. RISC-V resume groups
+are not allowed because the concept of a target resuming due to a different reason than
+user's resume request is not supported by OpenOCD and GDB.
+@end deffn
+
+
 @subsection RISC-V Authentication Commands
 
 The following commands can be used to authenticate to a RISC-V system. Eg.  a
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -68,12 +68,8 @@ static int riscv013_access_memory(struct target *target, const riscv_mem_access_
 static bool riscv013_get_impebreak(const struct target *target);
 static unsigned int riscv013_get_progbufsize(const struct target *target);
 
-typedef enum {
-	HALT_GROUP,
-	RESUME_GROUP
-} grouptype_t;
 static int set_group(struct target *target, bool *supported, unsigned int group,
-		grouptype_t grouptype);
+		enum grouptype grouptype, bool is_trigger, unsigned int trigger_num);
 
 /**
  * Since almost everything can be accomplish by scanning the dbus register, all
@@ -138,6 +134,8 @@ typedef struct {
 	 * abstractcs.busy may have remained set. In that case we may need to
 	 * re-check the busy state before executing these operations. */
 	bool abstract_cmd_maybe_busy;
+
+	struct riscv_ext_trigger external_triggers[RISCV_MAX_EXTTRIGGERS];
 } dm013_info_t;
 
 typedef struct {
@@ -1732,7 +1730,7 @@ static int halt_set_dcsr_ebreak(struct target *target)
 
 	if (info->haltgroup_supported) {
 		bool supported;
-		if (set_group(target, &supported, 0, HALT_GROUP) != ERROR_OK)
+		if (set_group(target, &supported, 0, HALT_GROUP, false, 0) != ERROR_OK)
 			return ERROR_FAIL;
 		if (!supported)
 			LOG_TARGET_ERROR(target, "Couldn't place hart in halt group 0. "
@@ -1754,7 +1752,7 @@ static int halt_set_dcsr_ebreak(struct target *target)
 	/* Add it back to the halt group. */
 	if (info->haltgroup_supported) {
 		bool supported;
-		if (set_group(target, &supported, target->smp, HALT_GROUP) != ERROR_OK)
+		if (set_group(target, &supported, target->smp, HALT_GROUP, false, 0) != ERROR_OK)
 			return ERROR_FAIL;
 		if (!supported)
 			LOG_TARGET_ERROR(target, "Couldn't place hart back in halt group %d. "
@@ -1785,19 +1783,46 @@ static void deinit_target(struct target *target)
 }
 
 static int set_group(struct target *target, bool *supported, unsigned int group,
-		grouptype_t grouptype)
+		enum grouptype grouptype, bool is_trigger, unsigned int trigger_num)
 {
-	uint32_t write_val = DM_DMCS2_HGWRITE;
 	assert(group <= 31);
+	assert(trigger_num < 16);
+
+	if (!is_trigger && dm013_select_target(target) != ERROR_OK)
+		return ERROR_FAIL;
+
+	dm013_info_t *dm = get_dm(target);
+	if (!dm)
+		return ERROR_FAIL;
+	if (is_trigger && dm->external_triggers[trigger_num].haltgroup_was_set &&
+			dm->external_triggers[trigger_num].haltgroup_num == group) {
+		LOG_TARGET_WARNING(target, "External trigger %d (at address dbgbase=0x%" PRIx32 ") "
+				"for halt group %d has been set.", trigger_num, dm->base, group);
+		return ERROR_OK;
+	}
+
+	uint32_t write_val = DM_DMCS2_HGWRITE;
 	write_val = set_field(write_val, DM_DMCS2_GROUP, group);
 	write_val = set_field(write_val, DM_DMCS2_GROUPTYPE, (grouptype == HALT_GROUP) ? 0 : 1);
+	write_val = set_field(write_val, DM_DMCS2_DMEXTTRIGGER, trigger_num);
+	write_val = set_field(write_val, DM_DMCS2_HGSELECT,
+			    is_trigger ? DM_DMCS2_HGSELECT_TRIGGERS : DM_DMCS2_HGSELECT_HARTS);
 	if (dm_write(target, DM_DMCS2, write_val) != ERROR_OK)
 		return ERROR_FAIL;
 	uint32_t read_val;
 	if (dm_read(target, &read_val, DM_DMCS2) != ERROR_OK)
 		return ERROR_FAIL;
 	if (supported)
-		*supported = (get_field(read_val, DM_DMCS2_GROUP) == group);
+		*supported = (get_field(read_val, DM_DMCS2_GROUP) == group &&
+			get_field(read_val, DM_DMCS2_GROUPTYPE) == ((grouptype == HALT_GROUP) ? 0 : 1) &&
+			get_field(read_val, DM_DMCS2_HGSELECT) ==
+				(is_trigger ? DM_DMCS2_HGSELECT_TRIGGERS : DM_DMCS2_HGSELECT_HARTS) &&
+			get_field(read_val, DM_DMCS2_DMEXTTRIGGER) == trigger_num);
+	if (is_trigger && *supported) {
+		dm->external_triggers[trigger_num].haltgroup_was_set = true;
+		dm->external_triggers[trigger_num].haltgroup_num = group;
+	}
+
 	return ERROR_OK;
 }
 
@@ -2145,8 +2170,9 @@ static int examine(struct target *target)
 	}
 
 	if (target->smp) {
-		if (set_group(target, &info->haltgroup_supported, target->smp, HALT_GROUP) != ERROR_OK)
+		if (set_group(target, &info->haltgroup_supported, target->smp, HALT_GROUP, false, 0) != ERROR_OK)
 			return ERROR_FAIL;
+
 		if (info->haltgroup_supported)
 			LOG_TARGET_INFO(target, "Core %d made part of halt group %d.", info->index,
 					target->smp);
@@ -2880,6 +2906,8 @@ static int init_target(struct command_context *cmd_ctx,
 
 	generic_info->handle_became_unavailable = &handle_became_unavailable;
 	generic_info->tick = &tick;
+
+	generic_info->set_group = &set_group;
 
 	if (!generic_info->version_specific) {
 		generic_info->version_specific = calloc(1, sizeof(riscv013_info_t));
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -5588,6 +5588,76 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
 	return ERROR_OK;
 }
 
+COMMAND_HANDLER(riscv_set_group)
+{
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	if (CMD_ARGC < 2)
+		return ERROR_COMMAND_SYNTAX_ERROR;
+
+	enum grouptype grouptype;
+	if (!strcmp("halt_group", CMD_ARGV[0])) {
+		grouptype = HALT_GROUP;
+	} else {
+		/* Ext. triggers for resume groups are not supported: neither OpenOCD nor GDB
+		 * have a concept of hart resuming on its own due to an external reason, without
+		 * an explicit resume request. */
+		LOG_ERROR("%s is not a valid argument. "
+			"The only supported group type is 'halt_group'.", CMD_ARGV[1]);
+		return ERROR_COMMAND_ARGUMENT_INVALID;
+	}
+
+	unsigned int group;
+	COMMAND_PARSE_NUMBER(uint, CMD_ARGV[1], group);
+	if (group > 31) {
+		LOG_ERROR("%d is not a valid group number - expecting a number in range 0..31.",
+			group);
+		return ERROR_COMMAND_ARGUMENT_INVALID;
+	}
+
+	if (!r->set_group) {
+		LOG_TARGET_ERROR(target, "set_group is not implemented for this target.");
+		return ERROR_FAIL;
+	}
+
+	bool is_trigger = false;
+	unsigned int trigger_num = 0;
+	switch (CMD_ARGC) {
+	case 2:
+		break;
+	case 4:
+		if (strcmp("trigger", CMD_ARGV[2]))
+			return ERROR_COMMAND_SYNTAX_ERROR;
+		is_trigger = true;
+		COMMAND_PARSE_NUMBER(uint, CMD_ARGV[3], trigger_num);
+		if (trigger_num > 15) {
+			LOG_ERROR("%d is not a valid external trigger number - expecting a number in range 0..15.",
+				trigger_num);
+			return ERROR_COMMAND_ARGUMENT_INVALID;
+		}
+		break;
+	default:
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	bool supported;
+	if (r->set_group(target, &supported, group, grouptype, is_trigger, trigger_num) != ERROR_OK)
+		return ERROR_FAIL;
+
+	if (supported)
+		LOG_TARGET_INFO(target, "%s %d made part of halt group %d.",
+			(is_trigger ? "External trigger" : "Core"),
+			(is_trigger ? trigger_num : (unsigned int)target->coreid), group);
+	else
+		LOG_TARGET_WARNING(target, "%s %d could not be made part of halt group %d.",
+			(is_trigger ? "External trigger" : "Core"),
+			(is_trigger ? trigger_num : (unsigned int)target->coreid), group);
+
+	return ERROR_OK;
+}
+
+
 static const struct command_registration riscv_exec_command_handlers[] = {
 	{
 		.name = "dump_sample_buf",
@@ -5849,6 +5919,13 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 		.help = "When on (default), OpenOCD will automatically execute fence instructions in some situations. "
 			"When off, users need to take care of memory coherency themselves, for example by using "
 			"`riscv exec_progbuf` to execute fence or CMO instructions."
+	},
+	{
+		.name = "set_group",
+		.handler = riscv_set_group,
+		.mode = COMMAND_ANY,
+		.usage = "grouptype group ['trigger' triggernum]",
+		.help = "Set a hart or a given external trigger to the halt group."
 	},
 	COMMAND_REGISTRATION_DONE
 };
--- a/src/target/riscv/riscv.h
+++ b/src/target/riscv/riscv.h
@@ -21,6 +21,7 @@ struct riscv_program;
 #define RISCV_MAX_TRIGGERS 32
 #define RISCV_MAX_HWBPS 16
 #define RISCV_MAX_DMS 100
+#define RISCV_MAX_EXTTRIGGERS 16
 
 #define DEFAULT_COMMAND_TIMEOUT_SEC 5
 
@@ -122,6 +123,11 @@ typedef struct {
 	char *name;
 } range_list_t;
 
+enum grouptype {
+	HALT_GROUP,
+	RESUME_GROUP
+};
+
 #define DTM_DTMCS_VERSION_UNKNOWN ((unsigned int)-1)
 #define RISCV_TINFO_VERSION_UNKNOWN (-1)
 
@@ -143,6 +149,17 @@ typedef struct riscv_mem_access_args {
 	uint32_t count;
 	uint32_t increment;
 } riscv_mem_access_args_t;
+
+struct riscv_ext_trigger {
+	bool haltgroup_was_set;
+	unsigned int haltgroup_num;
+	/*
+	In future, this can be added:
+
+	bool resumegroup_was_set;
+	unsigned int resumegroup_num;
+	 */
+};
 
 static inline bool
 riscv_mem_access_is_valid(const riscv_mem_access_args_t args)
@@ -306,6 +323,9 @@ struct riscv_info {
 	int (*access_memory)(struct target *target, const riscv_mem_access_args_t args);
 
 	unsigned int (*data_bits)(struct target *target);
+
+	int (*set_group)(struct target *target, bool *supported, unsigned int group,
+		enum grouptype grouptype, bool is_trigger, unsigned int trigger_num);
 
 	COMMAND_HELPER((*print_info), struct target *target);
 


lz-bro:
I want to continue the implementation of the external triggers for halt/resume groups base on @rbradford‘s work，But current implementation has not yet completed the resume group for external triggers。If I want to implement it, what do I need to pay attention to? What I can confirm now is that a diable haltgroup is required before the step, and the haltgroup will be restored after the step. @JanMatCodasip, @aap-sc  could you please take a look and give some suggestions?

src/target/riscv/riscv.c
```diff
@@ -5588,6 +5588,46 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
...
+		} else if (!strcmp("resume_group", CMD_ARGV[i])) {
+			r->external_triggers[index].grouptype = RESUME_GROUP;
+		} else {
+			LOG_ERROR("%s is not a valid argument for command,"
+				"should be halt_group or resume_group.", CMD_ARGV[0]);
+			return ERROR_COMMAND_SYNTAX_ERROR;
+		}
+		int value = atoi(CMD_ARGV[i + 1]);
+		if (value <= 0 || value > 16) {
+			LOG_ERROR("%s is not a valid integer argument for command.", CMD_ARGV[i + 1]);
```
rbradford: This came up in the previous review - 0 is a valid external trigger. Also 16 is not a valid - so should this be `(value < 0 || value >= 16)`

rbradford:
Thank you for restarting this work.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
```diff
@@ -5850,6 +5890,13 @@ static const struct command_registration riscv_exec_command_handlers[] = {
 			"When off, users need to take care of memory coherency themselves, for example by using "
 			"`riscv exec_progbuf` to execute fence or CMO instructions."
 	},
+	{
+		.name = "smp_add_ext_triggers",
```
JanMatCodasip: This pull request adds this command to assign external triggers to halt and resume groups.

I am OK with halt groups but have strong reservations against resume groups:

- Input triggers in resume groups cause harts to resume without GDB/OpenOCD making the request. 
  - OpenOCD and GDB are not ready for this. They have no concept of process resuming "on its own".
  - Already mentioned [here](https://github.com/riscv-collab/riscv-openocd/pull/1179#issuecomment-2547878631) (and copied [here](https://github.com/riscv-collab/riscv-openocd/pull/1243/files#diff-701864e2be3210275085c84b8f6fa7f9eb6171a2a18ca71638be42a155c08c63R2184)).
  - So a feature would be added for for which the main components of the system are not ready.
- Also there is no command or other mechanism in OpenOCD to assign harts to resume groups. (Halt groups are used for SMP.)

@lz-bro Can you please explain your use case? Are resume groups in fact important for your use case? 

- If so, how do you intend to solve the principal issue with input triggers in resume groups, described above?
- If not, consider removing that part of the change, and let the user work with ext. triggers in halt groups only.

src/target/riscv/riscv-013.c
```diff
@@ -2153,6 +2180,20 @@ static int examine(struct target *target)
 		else
 			LOG_TARGET_INFO(target, "Core %d could not be made part of halt group %d.",
 					info->index, target->smp);
+
+		/* TODO: Resume groups with external input triggers look to be very problematic.
+		 * Neither OpenOCD nor GDB is ready for the case when a target suddenly
+		 * resumes (due to the trigger), without an explicit resume request and
+		 * at arbitrary moment in time. */
+
```
JanMatCodasip: As already mentioned in the other comment, this is a principal problem of the system (GDB + OpenOCD). Not something that can be fixed by adjusting a piece of code in OpenOCD.

So I am afraid that putting a TODO inside OpenOCD code is not enough.

src/target/riscv/riscv.c
lz-bro: @JanMatCodasip Thank you for your answer, resume groups is not very important to us at present and I have removed the part of resume groups.

src/target/riscv/riscv-013.c
```diff
@@ -1784,6 +1780,37 @@ static void deinit_target(struct target *target)
...
+	if (dm_read(target, &read_val, DM_DMCS2) != ERROR_OK)
+		return ERROR_FAIL;
+	if (get_field(read_val, DM_DMCS2_GROUP) == group &&
+		get_field(read_val, DM_DMCS2_DMEXTTRIGGER) == external_trigger.dmexttrigger &&
+		get_field(read_val, DM_DMCS2_HGSELECT) == 1) {
+		LOG_TARGET_INFO(target, "External trigger %d added to %s group %d", external_trigger.dmexttrigger,
+				external_trigger.grouptype ? "resume" : "halt", group);
+	} else {
+		LOG_TARGET_ERROR(target, "External trigger %d not supported %s group %d", external_trigger.dmexttrigger,
+				external_trigger.grouptype ? "resume" : "halt", group);
```
JanMatCodasip: Suggested reword for clarity:

```suggestion
		LOG_TARGET_ERROR(target, "Could not add external trigger %d to %s group %d", external_trigger.dmexttrigger,
				external_trigger.grouptype ? "resume" : "halt", group);
```

src/target/riscv/riscv-013.c
```diff
@@ -1784,6 +1780,37 @@ static void deinit_target(struct target *target)
 	info->version_specific = NULL;
 }
 
+static int smp_add_ext_trigger(struct target *target, unsigned int group,
```
JanMatCodasip: Since this function does not touch the SMP groups (`target->smp`) anyhow, please remove the word `smp` from its name.

It can be called for example `add_ext_trigger_to_group()` (or similar).

src/target/riscv/riscv.c
```diff
@@ -5588,6 +5588,48 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
...
+		LOG_ERROR("Command takes an even number of parameters.");
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	if (CMD_ARGC > 2 * RISCV_MAX_EXTTRIGGERS) {
+		LOG_ERROR("The number of parameters can not exceeds %d.", 2 * RISCV_MAX_EXTTRIGGERS);
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	unsigned int index = 0;
```
JanMatCodasip: The command is called "**add**_ext_triggers" so the user may think that it can be called multiple times and the parameters would merge. 

This is not the case - the function overwrites the values of the previous call. 

Please pick one of these two options:

- Rename the function to "set" 
- Or make sure that the previously set triggers are not overwritten. In this case the function can be simplified to take just two arguments, and the user can call it as many times as they need.

src/target/riscv/riscv.h
```diff
@@ -365,6 +377,8 @@ struct riscv_info {
 	bool wp_allow_ge_lt_trigger;
 
 	bool autofence;
+
+	riscv_ext_trigger_t external_triggers[RISCV_MAX_EXTTRIGGERS];
```
JanMatCodasip: External triggers are only part of 0.13/1.0 version of the spec. Furthermore, external triggers are a property of the debug module. For that reason, the information about the external triggers should be moved to `dm013_info_t`.

src/target/riscv/riscv.c
```diff
@@ -5588,6 +5588,48 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
...
+			r->external_triggers[index].grouptype = HALT_GROUP;
+		} else if (!strcmp("resume_group", CMD_ARGV[i])) {
+			r->external_triggers[index].grouptype = RESUME_GROUP;
+			LOG_ERROR("External triggers for resume groups is not yet supported.");
+			return ERROR_COMMAND_SYNTAX_ERROR;
+		} else {
+			LOG_ERROR("%s is not a valid argument for command,"
+				"should be halt_group or resume_group.", CMD_ARGV[0]);
+			return ERROR_COMMAND_SYNTAX_ERROR;
+		}
```
JanMatCodasip: I would suggest to accept only "halt_group" and do not indicate anything about "resume_group".

The information about unsupported resume groups, perhaps with further explanation, can be put to the documentation.

```suggestion
		if (!strcmp("halt_group", CMD_ARGV[i])) {
			r->external_triggers[index].grouptype = HALT_GROUP;
		} else {
			/* Ext. triggers for resume groups are not supported: neither OpenOCD nor GDB
			 * have a concept of hart resuming on its own due to an external reason, without
			 * an explicit resume request. */		
			LOG_ERROR("%s is not a valid argument. "
				"The only supported group type is 'halt_group'.", CMD_ARGV[0]);
			return ERROR_COMMAND_SYNTAX_ERROR;
		}
```

src/target/riscv/riscv-013.c
JanMatCodasip: I have suggested a shorter version of the comment in another comment. So this verbose comment can perhaps be removed.

```suggestion
```

src/target/riscv/riscv.c
JanMatCodasip: Ok, understood, thank you for your reply.

src/target/riscv/riscv.c
```diff
@@ -5588,6 +5588,48 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
 	return ERROR_OK;
 }
 
+COMMAND_HANDLER(riscv_smp_add_ext_triggers)
```
JanMatCodasip: If this command is used on a target that is not part of any SMP group, should an error be raised?

src/target/riscv/riscv.h
```diff
@@ -21,6 +21,7 @@ struct riscv_program;
 #define RISCV_MAX_TRIGGERS 32
 #define RISCV_MAX_HWBPS 16
 #define RISCV_MAX_DMS 100
+#define RISCV_MAX_EXTTRIGGERS 32
```
JanMatCodasip: In each debug module, there are max. 16 external triggers - numbered 0..15.

src/target/riscv/riscv.c
JanMatCodasip: It would be better to say:

`LOG_ERROR("%s is not a valid external trigger number - expecting a number in range 0..15.", CMD_ARGV[i + 1]);`

JanMatCodasip (changes requested):
I have managed a partial review.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv.c
lz-bro: Rename the function to "set", addressed.

src/target/riscv/riscv.c

src/target/riscv/riscv.c
lz-bro: Thanks, I have added the check as follows:
```
COMMAND_HANDLER(riscv_smp_set_ext_triggers)
{
       struct target *target = get_current_target(CMD_CTX);
       RISCV_INFO(r);

       if (!target->smp) {
               LOG_ERROR("Command cannot be used on a target that does not belong to any SMP group.");
       }
       ...
```

src/target/riscv/riscv.h
lz-bro: I also considered resume_group at first, so changed it to 16 now.

src/target/riscv/riscv.h
lz-bro: `riscv_command_handlers` is part of `riscv.c`. If the information about the external triggers is moved to `dm013_info_t`, it seems a bit weird if using `dm013_info_t` requires including `riscv-013.c` in `riscv.c`? Besides, put the `dm013_info_t` definition into riscv-013.h and include it, or create a new `riscv013_command_handlers` to include this `smp_set_ext_triggers` command? Do you have any suggestions?

doc/openocd.texi
```diff
@@ -11616,6 +11616,11 @@ The second argument configures how OpenOCD should use the selected trigger featu
 With no parameters, prints current trigger features configuration.
 @end deffn
 
+@deffn {Command} {riscv smp_add_ext_triggers} grouptype1 num1 [grouptype2 num2] ...
+Associate the supplied external trigger with the halt group for the harts. When
+the external trigger fires the harts in the halt group will be halted.
+@end deffn
```
JanMatCodasip: 1) The command was renamed to `riscv smp_set_ext_triggers`, so the documentation should be updated, too.

2) It would be good to stress that the only supported value for `grouptype` is `halt_group`.

3) It would be beneficial to explain why only halt groups are supported. 

4) The command can associate multiple triggers with the halt group, so plural should be used (trigger -> triggers).

5) The text spoke about `trigger firing -> harts halting`. The other direction should be mentioned, too: `hart halted -> trigger notified`.

For example:

```suggestion
@deffn {Command} {riscv smp_add_ext_triggers} grouptype1 num1 [grouptype2 num2] ...
OpenOCD assigns targets of a SMP group into a RISC-V halt group. This command associates the supplied external trigger(s) with this halt group.

When an external input trigger fires, the harts in the halt group will get halted. Similarly, all external output triggers will be notified when a hart in the halt group halts.

The only supported value of @var{grouptype} is @option{halt_group}. RISC-V resume groups
are not allowed because the concept of a target resuming due to a different reason than user's resume
request is not supported by OpenOCD and GDB.
@end deffn
```

To others: Please feel free to suggest even different wording, if you like.

src/target/riscv/riscv-013.c
```diff
@@ -1784,6 +1780,37 @@ static void deinit_target(struct target *target)
...
+		return ERROR_FAIL;
+
+	uint32_t read_val;
+	if (dm_read(target, &read_val, DM_DMCS2) != ERROR_OK)
+		return ERROR_FAIL;
+	if (get_field(read_val, DM_DMCS2_GROUP) == group &&
+		get_field(read_val, DM_DMCS2_DMEXTTRIGGER) == external_trigger.dmexttrigger &&
+		get_field(read_val, DM_DMCS2_HGSELECT) == 1) {
+		LOG_TARGET_INFO(target, "External trigger %d added to %s group %d", external_trigger.dmexttrigger,
+				external_trigger.grouptype ? "resume" : "halt", group);
```
JanMatCodasip: Please, don't rely on the fact that 1 stands for resume group and 0 for halt group. Instead, use the named enum items.

```suggestion
				external_trigger.grouptype == RESUME_GROUP ? "resume" : "halt", group);
```

src/target/riscv/riscv-013.c
```diff
@@ -1784,6 +1780,37 @@ static void deinit_target(struct target *target)
...
 }
 
+static int add_ext_trigger_to_group(struct target *target, unsigned int group,
+		riscv_ext_trigger_t external_trigger)
+{
+	uint32_t write_val = DM_DMCS2_HGSELECT;
+	assert(group <= 31);
+	assert(external_trigger.dmexttrigger < 16);
+	write_val = set_field(write_val, DM_DMCS2_GROUP, group);
+	write_val = set_field(write_val, DM_DMCS2_GROUPTYPE, external_trigger.grouptype);
```
JanMatCodasip: (same)

```suggestion
	write_val = set_field(write_val, DM_DMCS2_GROUPTYPE, external_trigger.grouptype == RESUME_GROUP ? 1 : 0);
```

src/target/riscv/riscv-013.c
```diff
@@ -1784,6 +1780,37 @@ static void deinit_target(struct target *target)
 	info->version_specific = NULL;
 }
 
+static int add_ext_trigger_to_group(struct target *target, unsigned int group,
+		riscv_ext_trigger_t external_trigger)
+{
+	uint32_t write_val = DM_DMCS2_HGSELECT;
+	assert(group <= 31);
+	assert(external_trigger.dmexttrigger < 16);
```
JanMatCodasip: (Optionally)

Detail for additional safety:

```suggestion
	assert(external_trigger.dmexttrigger < 16);
	assert(external_trigger.dmexttrigger.grouptype == HALT_GROUP || external_trigger.dmexttrigger.grouptype == RESUME_GROUP);	
```

src/target/riscv/riscv-013.c
```diff
@@ -1784,6 +1780,37 @@ static void deinit_target(struct target *target)
...
+	if (dm_read(target, &read_val, DM_DMCS2) != ERROR_OK)
+		return ERROR_FAIL;
+	if (get_field(read_val, DM_DMCS2_GROUP) == group &&
+		get_field(read_val, DM_DMCS2_DMEXTTRIGGER) == external_trigger.dmexttrigger &&
+		get_field(read_val, DM_DMCS2_HGSELECT) == 1) {
+		LOG_TARGET_INFO(target, "External trigger %d added to %s group %d", external_trigger.dmexttrigger,
+				external_trigger.grouptype ? "resume" : "halt", group);
+	} else {
+		LOG_TARGET_ERROR(target, "Could not add external trigger %d to %s group %d", external_trigger.dmexttrigger,
+				external_trigger.grouptype ? "resume" : "halt", group);
```
JanMatCodasip: (same)

```suggestion
				external_trigger.grouptype == RESUME_GROUP ? "resume" : "halt", group);
```

src/target/riscv/riscv-013.c
```diff
@@ -1784,6 +1780,37 @@ static void deinit_target(struct target *target)
...
+	write_val = set_field(write_val, DM_DMCS2_HGWRITE, 1);
+	if (dm_write(target, DM_DMCS2, write_val) != ERROR_OK)
+		return ERROR_FAIL;
+
+	uint32_t read_val;
+	if (dm_read(target, &read_val, DM_DMCS2) != ERROR_OK)
+		return ERROR_FAIL;
+	if (get_field(read_val, DM_DMCS2_GROUP) == group &&
+		get_field(read_val, DM_DMCS2_DMEXTTRIGGER) == external_trigger.dmexttrigger &&
+		get_field(read_val, DM_DMCS2_HGSELECT) == 1) {
```
JanMatCodasip: Should the value of `get_field(read_val, DM_DMCS2_GROUPTYPE)` be also checked here?

src/target/riscv/riscv.c
```diff
@@ -5588,6 +5588,52 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
...
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	if (CMD_ARGC % 2) {
+		LOG_ERROR("Command takes an even number of parameters.");
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+
+	if (CMD_ARGC > 2 * RISCV_MAX_EXTTRIGGERS) {
+		LOG_ERROR("The number of parameters can not exceeds %d.", 2 * RISCV_MAX_EXTTRIGGERS);
```
```suggestion
		LOG_ERROR("The number of parameters must not exceed %d.", 2 * RISCV_MAX_EXTTRIGGERS);
```

src/target/riscv/riscv.c
```diff
@@ -5588,6 +5588,52 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
...
 
+COMMAND_HANDLER(riscv_smp_set_ext_triggers)
+{
+	struct target *target = get_current_target(CMD_CTX);
+	RISCV_INFO(r);
+
+	if (!target->smp) {
+		LOG_ERROR("Command cannot be used on a target that does not belong to any SMP group.");
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
```
JanMatCodasip: 1) This is not a syntax error. Please use `return ERROR_FAIL` (or similar) instead of `return ERROR_COMMAND_SYNTAX_ERROR`.

2) Please move this check down - first check all syntax errors, then other errors.

3) Please use `LOG_TARGET_ERROR`. The message could be: `This target does not belong to any SMP group. Command 'riscv smp_set_ext_triggers' cannot be used`

src/target/riscv/riscv.c
```diff
@@ -5588,6 +5588,52 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
...
+			r->external_triggers[index].grouptype = HALT_GROUP;
+		} else {
+			/* Ext. triggers for resume groups are not supported: neither OpenOCD nor GDB
+			 * have a concept of hart resuming on its own due to an external reason, without
+			 * an explicit resume request. */
+			LOG_ERROR("%s is not a valid argument. "
+				"The only supported group type is 'halt_group'.", CMD_ARGV[0]);
+			return ERROR_COMMAND_SYNTAX_ERROR;
+		}
+		int value = atoi(CMD_ARGV[i + 1]);
```
JanMatCodasip: For code clarity, `value` can be renamed to `trigger_num` (or similar).

```suggestion
		int trigger_num = atoi(CMD_ARGV[i + 1]);
```

src/target/riscv/riscv.c
```diff
@@ -5588,6 +5588,52 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
...
+	for (unsigned int i = 0; i < CMD_ARGC - 1; i += 2) {
+		if (!strcmp("halt_group", CMD_ARGV[i])) {
+			r->external_triggers[index].grouptype = HALT_GROUP;
+		} else {
+			/* Ext. triggers for resume groups are not supported: neither OpenOCD nor GDB
+			 * have a concept of hart resuming on its own due to an external reason, without
+			 * an explicit resume request. */
+			LOG_ERROR("%s is not a valid argument. "
+				"The only supported group type is 'halt_group'.", CMD_ARGV[0]);
+			return ERROR_COMMAND_SYNTAX_ERROR;
```
JanMatCodasip, line 5631: This would be a more fitting return code:

```suggestion
			return ERROR_COMMAND_ARGUMENT_INVALID;
```

src/target/riscv/riscv.c
```diff
@@ -5588,6 +5588,52 @@ COMMAND_HANDLER(handle_riscv_virt2phys_mode)
...
+		}
+		int value = atoi(CMD_ARGV[i + 1]);
+		if (value < 0 || value > 15) {
+			LOG_ERROR("%s is not a valid integer argument for command.", CMD_ARGV[i + 1]);
+			return ERROR_FAIL;
```

```suggestion
			return ERROR_COMMAND_ARGUMENT_INVALID;
```

src/target/riscv/riscv.h
```diff
@@ -144,6 +145,17 @@ typedef struct riscv_mem_access_args {
...
+typedef enum {
+	HALT_GROUP,
+	RESUME_GROUP
+} grouptype_t;
+
+typedef struct {
+	grouptype_t grouptype;
+	uint32_t dmexttrigger;
+	bool is_set;
+} riscv_ext_trigger_t;
```
JanMatCodasip: 1) Please, don't introduce new typedefs for structures. The upstream code style (at openocd.org) no longer allows them. Use just `struct riscv_ext_trigger`.

2) The RISC-V debug spec says: "At any given time, each hart and each trigger is member of exactly one halt group and one resume group." Please consider re-arranging the structure to reflect it: 
- The array index (0..15) would mean the trigger number
- Each array item (struct riscv_ext_trigger) should say if halt group should be set, and what the halt group number is

```cpp
struct riscv_ext_trigger {
	bool is_haltgroup_set;
	unsigned int haltgroup_num;
	/*
	In future, this can be added:

	bool is_resumegroup_set;
	unsigned int resumegroup_num;
	 */
};
```

src/target/riscv/riscv.c
JanMatCodasip: I am afraid this is not fully addressed yet. If the user calls these two commands:

```
riscv smp_set_ext_triggers halt_group 2 halt_group 3 halt_group 4
riscv smp_set_ext_triggers halt_group 5
```

... the current result would be that only triggers 5, 3 and 4, would be assigned. It would not be very intuitive.

I would prefer if all the triggers (2, 3, 4 and 5) were assigned.

When the trigger structure (`struct riscv_ext_trigger`) is re-arranged as suggested in another comment, this should not be a problem anymore.

src/target/riscv/riscv-013.c
```diff
@@ -2153,6 +2180,15 @@ static int examine(struct target *target)
 		else
 			LOG_TARGET_INFO(target, "Core %d could not be made part of halt group %d.",
 					info->index, target->smp);
+
+		for (unsigned int i = 0; i < RISCV_MAX_EXTTRIGGERS; i++) {
+			if (r->external_triggers[i].is_set) {
+				if (add_ext_trigger_to_group(target, target->smp, r->external_triggers[i]) != ERROR_OK)
+					return ERROR_FAIL;
```
JanMatCodasip: It appears that `haltgroup_supported` from riscv013_info_t should be checked here. If it is false, then a warning should be displayed, and `add_ext_trigger_to_group()` not called.

src/target/riscv/riscv.h
JanMatCodasip: You can do it this way:
- Implement a new function in riscv013.c into `struct riscv_info`, called e.g. `riscv013_set_external_trigger`.
- Add it as a new callback into `riscv_info`, called e.g. `set_external_trigger`.
- Keep the command handler in riscv.c but check whether `r->set_external_trigger` is implemented.

As an example, you can take a look at `COMMAND_HANDLER(riscv_authdata_write)` which follows this approach.

JanMatCodasip (changes requested):

src/target/riscv/riscv.c
JanMatCodasip: Please, if this error message could be improved.

src/target/riscv/riscv.c
JanMatCodasip: It looks like you can use `COMMAND_PARSE_NUMBER(uint, CMD_ARGV[i + 1], trigger_num);` instead of `atoi()`.

src/target/riscv/riscv.h
lz-bro: I think `is_haltgroup_set` can be removed, because `haltgroup_num > 0` is equivalent to `is_haltgroup_set = true`. What do you think?

src/target/riscv/riscv.h
lz-bro: Thank you for your answer, but I encountered a new problem. The mode of this command is `COMMAND_CONFIG`,  `riscv smp_set_ext_triggers` command used before 'init' and `r->set_external_trigger` is not implemented before init target.

doc/openocd.texi
lz-bro: Nice description, thanks.

src/target/riscv/riscv-013.c
lz-bro: I think the following code is safe enough.
`write_val = set_field(write_val, DM_DMCS2_GROUPTYPE, (grouptype == HALT_GROUP) ? 0 : 1);`

src/target/riscv/riscv.h
lz-bro: Thank you for your suggestion, it is very helpful, and is `is_haltgroup_set` seems unnecessary?

src/target/riscv/riscv-013.c
lz-bro: Thanks, addressed.

src/target/riscv/riscv.c
lz-bro: Addressed.

src/target/riscv/riscv.c
lz-bro: Addressed.

src/target/riscv/riscv.c
lz-bro: Addressed.

src/target/riscv/riscv.c
lz-bro: Addressed.

src/target/riscv/riscv.h
lz-bro: @JanMatCodasip I want to introduce the `set_group` command instead of `smp_set_ext_triggers`, Create a  `set_group` command handler in riscv.c but check whether `r->set_group` is implemented, then called `set_group` in riscv013.c
```
       {
               .name = "set_group",
               .handler = riscv_set_group,
               .mode = COMMAND_ANY,
               .usage = "grouptype group ['triger' triggernum]",
               .help = "Set a hart or a given external trigger to the halt group."
       },
```

```
	{
		.name = "smp_set_ext_triggers",
		.handler = riscv_smp_set_ext_triggers,
		.mode = COMMAND_CONFIG,
		.usage = "num1 grouptype1 [num2 grouptype2]......",
		.help = "Set the given external triggers to the halt group."
	},
```

This way, users can also set hat group for harts and triggers.

JanMatCodasip:
Hi @lz-bro, I can see that this patch has been moved to https://review.openocd.org/c/openocd/+/9224. I have posted my review comments there.

To not have the same change open in two places (and to not confuse any newcomers), can you please close this PR? 

Thanks.

PR #1255, opened 2025-05-13
Add dcsr cetrig control

lz-bro:
Introduce RISC-V-sepecific `configure` parameter `-cetrig`


src/target/riscv/riscv-013.c
```diff
@@ -253,7 +253,7 @@ typedef struct {
 
 	/* When false, we need to set dcsr.ebreak*, halting the target if that's
 	 * necessary. */
-	bool dcsr_ebreak_is_set;
+	bool dcsr_configs_is_set;
```
en-sc: Please, adjust the comment to drop the mention of `ebreak`.
Also, this is a nitpick, but `dcsr_configs_is_set` does not feel right (plural with `is`). Maybe `dcsr_config_is_set`?

src/target/riscv/riscv-013.c
```diff
@@ -1694,14 +1694,15 @@ static int set_dcsr_ebreak(struct target *target, bool step)
...
 	if (dcsr != original_dcsr &&
 			riscv_reg_set(target, GDB_REGNO_DCSR, dcsr) != ERROR_OK)
 		return ERROR_FAIL;
-	info->dcsr_ebreak_is_set = true;
+	info->dcsr_configs_is_set = true;
 	return ERROR_OK;
 }
 
-static int halt_set_dcsr_ebreak(struct target *target)
+static int halt_set_dcsr_configs(struct target *target)
```
```suggestion
static int halt_set_dcsr_config(struct target *target)
```
For consistency with `dcsr_config_is_set` (if you decide to adopt the suggestion).

src/target/riscv/riscv-013.c
```diff
@@ -2939,13 +2940,13 @@ static int assert_reset(struct target *target)
 	return riscv013_invalidate_cached_progbuf(target);
 }
 
-static bool dcsr_ebreak_config_equals_reset_value(const struct target *target)
+static bool dcsr_configs_equals_reset_value(const struct target *target)
```
```suggestion
static bool dcsr_config_equals_reset_value(const struct target *target)
```
Same as above

src/target/riscv/riscv-013.c
```diff
@@ -5367,6 +5368,14 @@ static enum riscv_halt_reason riscv013_halt_reason(struct target *target)
 		return RISCV_HALT_INTERRUPT;
 	case CSR_DCSR_CAUSE_GROUP:
 		return RISCV_HALT_GROUP;
+	case CSR_DCSR_CAUSE_OTHER:
+		if (get_field(dcsr, CSR_DCSR_EXTCAUSE) == 0) {
+			LOG_TARGET_WARNING(target, "halted because of hart in a critical error state.");
+			return RISCV_HALT_ERROR;
```
en-sc: I believe this should be a separate entry in `enum riscv_halt_reason`.
Also, AFAIU, `RISCV_HALT_ERROR` is currently unused. Maybe a simple rename would suffice.

src/target/riscv/riscv-013.c
```diff
@@ -5367,6 +5368,14 @@ static enum riscv_halt_reason riscv013_halt_reason(struct target *target)
 		return RISCV_HALT_INTERRUPT;
 	case CSR_DCSR_CAUSE_GROUP:
 		return RISCV_HALT_GROUP;
+	case CSR_DCSR_CAUSE_OTHER:
+		if (get_field(dcsr, CSR_DCSR_EXTCAUSE) == 0) {
+			LOG_TARGET_WARNING(target, "halted because of hart in a critical error state.");
```
en-sc, line 5375:
```suggestion
			LOG_TARGET_INFO(target, "halted because of hart in a critical error state.");
```
I believe `INFO` is more appropriate here, since this is completely normal from the debugger's point of view.

src/target/riscv/riscv.c
```diff
@@ -525,6 +527,15 @@ static struct jim_nvp nvp_ebreak_mode_opts[] = {
 	{ .name = NULL, .value = RISCV_EBREAK_MODE_INVALID }
 };
 
+
+#define RISCV_CETRIG_INVALID -1
```
en-sc, line 531: I believe this can be inlined.

src/target/riscv/riscv.c
```diff
@@ -611,13 +622,43 @@ static int jim_report_ebreak_config(const struct riscv_private_config *config,
 	return JIM_OK;
 }
 
+static int jim_configure_cetrig(struct riscv_private_config *config,
+		struct jim_getopt_info *goi)
+{
+	if (goi->argc == 0) {
+		Jim_WrongNumArgs(goi->interp, 1, goi->argv - 1,
+				"?disable?or?enable?");
```
```suggestion
				"?disable|enable?");
```

src/target/riscv/riscv.c
```diff
@@ -611,13 +622,43 @@ static int jim_report_ebreak_config(const struct riscv_private_config *config,
...
+		return JIM_ERR;
+	}
+	config->dcsr_cetrig = opt_nvp->value;
+	return JIM_OK;
+}
+
+static int jim_report_cetrig_config(const struct riscv_private_config *config,
+		Jim_Interp *interp)
+{
+	const char * const cetrig_opt = jim_nvp_value2name_simple(nvp_cetrig_opts,
```
```suggestion
	const char *cetrig_opt = jim_nvp_value2name_simple(nvp_cetrig_opts,
```
I'd suggest dropping this `const`. AFAIU, the styleguide will be updated sometime to include this rule.
See https://review.openocd.org/c/openocd/+/8750/comment/984b8414_f6f804e4/

en-sc (commented):
Thank you for the patch!

Overall it looks great, the only concern I have is with the new return from `riscv013_halt_reason()`. I'm not sure `RISCV_HALT_ERROR` is the proper return code. I'd suggest adding a new status and handling it differently at the callsite.

Also, how did you test the changes? Is this at all possible to extend `riscv-tests` to include relevant tests?

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
lz-bro: Addressed

src/target/riscv/riscv.c
lz-bro: To be consistent with `RISCV_EBREAK_MODE_INVALID`, should this be kept?

src/target/riscv/riscv.c

src/target/riscv/riscv.c

lz-bro:
> Thank you for the patch!
> 
> Overall it looks great, the only concern I have is with the new return from `riscv013_halt_reason()`. I'm not sure `RISCV_HALT_ERROR` is the proper return code. I'd suggest adding a new status and handling it differently at the callsite.
> 
> Also, how did you test the changes? Is this at all possible to extend `riscv-tests` to include relevant tests?

1. Now, I have tested the changes on XiangShan-KunMingHu platform and spike by enabling smdbltrp extension. 
You can find the corresponding ISA extension from the XiangShan Github repository: https://github.com/OpenXiangShan/XiangShan/blob/master/src/main/scala/xiangshan/Parameters.scala
2. I don't have any good ideas yet to extend riscv-tests to include relevant tests.

src/target/riscv/riscv.h
```diff
@@ -75,6 +75,7 @@ enum riscv_halt_reason {
 	RISCV_HALT_TRIGGER,
 	RISCV_HALT_UNKNOWN,
 	RISCV_HALT_GROUP,
+	RISCV_HALT_CRITERR,
```
JanMatCodasip: If you like - this will be more clear to code readers:

```suggestion
	RISCV_HALT_CRITICAL_ERROR,
```

src/target/riscv/riscv-013.c
```diff
@@ -5367,6 +5368,14 @@ static enum riscv_halt_reason riscv013_halt_reason(struct target *target)
 		return RISCV_HALT_INTERRUPT;
 	case CSR_DCSR_CAUSE_GROUP:
 		return RISCV_HALT_GROUP;
+	case CSR_DCSR_CAUSE_OTHER:
+		if (get_field(dcsr, CSR_DCSR_EXTCAUSE) == 0) {
+			LOG_TARGET_INFO(target, "halted because of hart in a critical error state.");
```
JanMatCodasip: Since this is an error condition of the target, I would use at least warning level. Perhaps even an error?

```suggestion
			LOG_TARGET_WARNING(target, "halted because of hart in a critical error state.");
```

src/target/riscv/riscv-013.c
```diff
@@ -1694,14 +1694,15 @@ static int set_dcsr_ebreak(struct target *target, bool step)
 	dcsr = set_field(dcsr, CSR_DCSR_EBREAKU, config->dcsr_ebreak_fields[RISCV_MODE_U]);
 	dcsr = set_field(dcsr, CSR_DCSR_EBREAKVS, config->dcsr_ebreak_fields[RISCV_MODE_VS]);
 	dcsr = set_field(dcsr, CSR_DCSR_EBREAKVU, config->dcsr_ebreak_fields[RISCV_MODE_VU]);
+	dcsr = set_field(dcsr, CSR_DCSR_CETRIG, config->dcsr_cetrig);
 	if (dcsr != original_dcsr &&
 			riscv_reg_set(target, GDB_REGNO_DCSR, dcsr) != ERROR_OK)
 		return ERROR_FAIL;
-	info->dcsr_ebreak_is_set = true;
+	info->dcsr_config_is_set = true;
 	return ERROR_OK;
```
JanMatCodasip: Would it make sense to read DCSR back and see if these WARL bits were really set as the user wanted? If not, then a log info print could be made.

If you think this is a useful improvement, please open an issue or make a TODO comment into the code. I don't insist on making it in this PR. It'd better to do this separately.

src/target/riscv/riscv-013.c
```diff
@@ -5367,6 +5368,14 @@ static enum riscv_halt_reason riscv013_halt_reason(struct target *target)
...
 	case CSR_DCSR_CAUSE_GROUP:
 		return RISCV_HALT_GROUP;
+	case CSR_DCSR_CAUSE_OTHER:
+		if (get_field(dcsr, CSR_DCSR_EXTCAUSE) == 0) {
+			LOG_TARGET_INFO(target, "halted because of hart in a critical error state.");
+			return RISCV_HALT_CRITERR;
+		}
+		LOG_TARGET_ERROR(target, "Unknown DCSR extcause field: 0x%"
+				PRIx64, get_field(dcsr, CSR_DCSR_EXTCAUSE));
+		return RISCV_HALT_UNKNOWN;
```
JanMatCodasip: If you like, this could be a switch instead of if cascade.

This will encourage better coding style in future, when there are more EXTCAUSE values.

```suggestion
		switch (get_field(dcsr, CSR_DCSR_EXTCAUSE)) 
		{
		case 0: 
			/* ... */
			return RISCV_HALT_CRITICAL_ERROR;
		default:
			/* ... */
			return RISCV_HALT_UNKNOWN;
		}
```

src/target/riscv/riscv-013.c
```diff
@@ -251,9 +251,9 @@ typedef struct {
 	/* This target was selected using hasel. */
 	bool selected;
 
-	/* When false, we need to set dcsr.ebreak*, halting the target if that's
+	/* When false, we need to set dcsr config, halting the target if that's
 	 * necessary. */
-	bool dcsr_ebreak_is_set;
+	bool dcsr_config_is_set;
```
```suggestion
	bool dcsr_register_is_set;
```

src/target/riscv/riscv-013.c
```diff
@@ -251,9 +251,9 @@ typedef struct {
 	/* This target was selected using hasel. */
 	bool selected;
 
-	/* When false, we need to set dcsr.ebreak*, halting the target if that's
+	/* When false, we need to set dcsr config, halting the target if that's
 	 * necessary. */
```
```suggestion
	/* When false, we need to configure certain bits in the dcsr register. 
	 * To do that, we may momentarily halt the target, if necessary. */
```

JanMatCodasip (commented):
Overall this looks good. I have managed a partial review and am sending my comments so far.

src/target/riscv/riscv.h
lz-bro: Addressed

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c
lz-bro: It is useful, addressed, thanks.

src/target/riscv/riscv-013.c
lz-bro: Addressed, thanks.

src/target/riscv/riscv-013.c

src/target/riscv/riscv-013.c

doc/openocd.texi
```diff
@@ -11437,6 +11437,13 @@ action pairs.
 @end itemize
 @end itemize
 
+@itemize
+@item @code{-cetrig} @option{disable}|@option{enable} -- sets the @code{cetrig}
+field for dcsr. Defaults to @option{disable}.
+
+@item @code{cget} return a @code{cetrig} state.
+@end itemize
+
```
JanMatCodasip: It looks like this last bullet point should be indented by one more level.

Currently it looks like this:

image

Please don't forget to run `make html` and check the result.

JanMatCodasip: The sentence can also be reworded to make it easier to read and understand:

`@code{cget} returns the currently configured state for @code{cetrig}.`

doc/openocd.texi
```diff
@@ -11437,6 +11437,13 @@ action pairs.
 @end itemize
 @end itemize
 
+@itemize
+@item @code{-cetrig} @option{disable}|@option{enable} -- sets the @code{cetrig}
+field for dcsr. Defaults to @option{disable}.
```
JanMatCodasip: The sentence can be reworded like shown below. My point is that this option does not immediately write to the DCSR, but merely records what values OpenOCD should write to that register when the time comes.

```suggestion
@item @code{-cetrig} @option{disable}|@option{enable} -- determines how OpenOCD should configure the @code{cetrig}
field in the dcsr register. Defaults to @option{disable}.
```

src/target/riscv/riscv-013.c
JanMatCodasip: I am sorry but would like to disagree.

Even though this is not an error of OpenOCD or of the debug transport (JTAG adapter or similar), it is situation that is clearly exceptional and requires user attention. As such, it should be prominently displayed to the user. So a warning is in my opinion warranted here.

src/target/riscv/riscv.c
```diff
@@ -611,13 +622,43 @@ static int jim_report_ebreak_config(const struct riscv_private_config *config,
...
+static int jim_configure_cetrig(struct riscv_private_config *config,
+		struct jim_getopt_info *goi)
+{
+	if (goi->argc == 0) {
+		Jim_WrongNumArgs(goi->interp, 1, goi->argv - 1,
+				"?disable|enable?");
+		return JIM_ERR;
+	}
+
+	/* Here a dcsr field is processed, e.g: riscv.cpu configure -mprven enable" */
```
JanMatCodasip: This comment mentions `-mprven enable` but the function is dealing with `-cetrig`.

Perhaps it is a copy-paste mistake?

src/target/riscv/riscv-013.c
JanMatCodasip: I have also filed an issue for it so that the idea is not lost: https://github.com/riscv-collab/riscv-openocd/issues/1263

JanMatCodasip (commented):
I have few more comments, otherwise the changes look good.

doc/openocd.texi
lz-bro: Addressed and checked, thanks.

doc/openocd.texi
lz-bro: Addressed.

src/target/riscv/riscv.c
lz-bro: Thank for discovering this mistake.

JanMatCodasip (approved):
LGTM, thank you.

src/target/riscv/riscv-013.c
en-sc: @JanMatCodasip, I will insist)
1. IMHO a similar example will be an assertion failure. GDB does not warn about an assertion failure -- it just notifies you.
2. To be honest, this seems like an issue with the spec, but: what if the `Smdbltrp` extension is not implemented? The `cause` may still be `other` but this is not an exceptional situation.

JanMatCodasip:
I can see this change has been posted to the upstream OpenOCD: https://review.openocd.org/c/openocd/+/9223

@lz-bro, please consider closing this PR, then.

PR #1256, opened 2025-05-14
target/riscv: fix riscv_mmu behaviour

fkhaidari:
Fixed riscv_mmu behaviour: buggy check was removed.
As a result virt2phys command behaviour was fixed: now it
returns translated address even while virt2phys_mode is off.

Change-Id: Ie2e6d1057024ab794038d5ed3662ef49a4d71e70

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3017,9 +3017,6 @@ static int riscv_mmu(struct target *target, int *enabled)
 {
 	*enabled = 0;
 
-	if (!riscv_virt2phys_mode_is_sw(target))
-		return ERROR_OK;
-
 	/* Don't use MMU in explicit or effective M (machine) mode */
 	riscv_reg_t priv;
 	if (riscv_reg_get(target, &priv, GDB_REGNO_PRIV) != ERROR_OK) {
@@ -3425,16 +3422,11 @@ static int riscv_rw_memory(struct target *target, const riscv_mem_access_args_t 
 		return ERROR_OK;
 	}
 
-	int mmu_enabled;
-	int result = riscv_mmu(target, &mmu_enabled);
-	if (result != ERROR_OK)
-		return result;
-
 	RISCV_INFO(r);
-	if (!mmu_enabled)
+	if (riscv_virt2phys_mode_is_hw(target))
 		return r->access_memory(target, args);
 
-	result = check_virt_memory_access(target, args.address,
+	int result = check_virt_memory_access(target, args.address,
 			args.size, args.count, is_write);
 	if (result != ERROR_OK)
 		return result;


en-sc (dismissed):
LGTM (reviewed internally).

fkhaidari:

JanMatCodasip (dismissed):
LGTM. Thank you!

src/target/riscv/riscv.c
```diff
@@ -3017,9 +3017,6 @@ static int riscv_mmu(struct target *target, int *enabled)
 {
 	*enabled = 0;
 
-	if (!riscv_virt2phys_mode_is_sw(target))
-		return ERROR_OK;
-
```
lz-bro, line 3022: Access memory would fail while virt2phys_mode is hw.
```suggestion
	if (riscv_virt2phys_mode_is_hw(target))
		return ERROR_OK;
```

fkhaidari: With this check `virt2phys_mode` would return incorrect value: virtual address instead of physical

lz-bro: Access memory would fail without this check `virt2phys_mode` while virt2phys_mode is hw. It is correct to return a virtual address while virt2phys_mode is hw. Address translation should be done by hardware when set `dcsr.mprven`, I have tested it. When virt2phys_mode is hw, the virt2phys command returns the physical address, at least the current modification is incorrect, and both the virt2phys command and memory access will report errors.

fkhaidari: Sorry, I've mistaken. I meant that `virt2phys` would return incorrect value

fkhaidari: I think I understand you now. You are right, here we have double translation. But I think it has to be fixed in `riscv_rw_memory` function. Something like:
```c
if (riscv_virt2phys_mode_is_hw(target))
	return r->access_memory(target, args);
```
before translation.

Sorry for the mess. This change was a part of #1241. Looks like I have extracted this part incorrectly

lz-bro: When virt2phys_mode is hw, the virt2phys command still report errors. Translate page table accesses failed due to setting `dcsr.mprven`. I think maybe it could be fixed in `riscv_virt2phys` function like this:
``` 
        RISCV_INFO(r);
        int mode = r->virt2phys_mode;
        if (riscv_virt2phys_mode_is_hw(target))
                r->virt2phys_mode = RISCV_VIRT2PHYS_MODE_SW;

        int result = riscv_address_translate(target,
                        satp_info, get_field(satp_value, RISCV_SATP_PPN(xlen)),
                        NULL, 0,
                        virtual, physical);

        r->virt2phys_mode = mode;
        return result;
```

src/target/riscv/riscv.c
```diff
@@ -3425,16 +3422,11 @@ static int riscv_rw_memory(struct target *target, const riscv_mem_access_args_t
...
 
-	int mmu_enabled;
-	int result = riscv_mmu(target, &mmu_enabled);
-	if (result != ERROR_OK)
-		return result;
-
 	RISCV_INFO(r);
-	if (!mmu_enabled)
+	if (riscv_virt2phys_mode_is_hw(target))
 		return r->access_memory(target, args);
```
JanMatCodasip, line 3427: Do I understand correctly that when virt2phys is off, we should also directly call `r->access_memory`?

```suggestion
	if (riscv_virt2phys_mode_is_off(target))
		/* No address translation is performed. Virtual == physical is assumed. */
		return r->access_memory(target, args);
	if (riscv_virt2phys_mode_is_hw(target))
		/* Address translation virtual -> physical will be performed by the hardware. */	
		return r->access_memory(target, args);

	/* Otherwise address translation is performed by OpenOCD - for each 4 KiB page separately.
	 * The memory operation must be split to page-sized chunks.
	 */
	
	...
```

JanMatCodasip: @fk-sc It looks like you have independently came to the same conclusion [in this comment](https://github.com/riscv-collab/riscv-openocd/pull/1256/files#r2093662876).

src/target/riscv/riscv.c
JanMatCodasip: IIUC, riscv_address_translate() function should always perform physical accesses - it should work exclusively with physical addresses. @lz-bro, is this the issue that you are trying to point out?

If that's correct understanding, then we should probably create function `r->access_memory_phys` that will always perform physical accesses, regardless of `r->virt2phys_mode`. It looks to me as a cleaner solution than temporarily changing the `virt2phys_mode` flag.

lz-bro: @JanMatCodasip Yes, that's exactly what I meant.

en-sc: @lz-bro, great catch!

> If that's correct understanding, then we should probably create function r->access_memory_phys that will always perform physical accesses, regardless of r->virt2phys_mode. It looks to me as a cleaner solution than temporarily changing the virt2phys_mode flag.

@JanMatCodasip, this will not be enough. In case of 2-stage translation, the first PTE is addressed using a virtual address. I'd suggest bailing out with a message that this is not supported for now (AFAIU, `mstatus.MPP` will need to be modified/restored for proper support).

> ... create function r->access_memory_phys that will always perform physical accesses.

AFAIU, this is exactly the purpose of #1241.

en-sc:
@JanMatCodasip, @lz-bro, I'd suggest merging this one as-is. It makes the `virt2phys` work in some cases, so it's a step in the right direction. To finish this we will need #1241, and a follow-up that will actually consider ` is_virtual` parameter when accessing the memory, however IMHO it's easier to review them one-by-one.

src/target/riscv/riscv.c
fkhaidari: > AFAIU, this is exactly the purpose of https://github.com/riscv-collab/riscv-openocd/pull/1241

You are right, yeah

fkhaidari:
@JanMatCodasip, @lz-bro, can we continue merging? Maybe with some extra TODO's

fkhaidari:
Ping

fkhaidari:
@JanMatCodasip, @lz-bro, ping

fkhaidari:
@JanMatCodasip, @lz-bro, can we continue merging? Maybe with some extra TODO's

JanMatCodasip:
I'm sorry for being inactive around OpenOCD. IIUC, this merge request has not been integrated into riscv-openocd nor into vanilla OpenOCD.

@en-sc, @fkhaidari, are you interested in continuing with this patch? If so, my preference would be to move to review.openocd.org.

As for the patch itself:

IIUC, the primary purpose of riscv_mmu() (the `.mmu` callback in `struct target_type`) is to determine whether working area address should be taken from `target->working_area_phys` or from `target->working_area_virt`, see `target_alloc_working_area_try()`.

Any other use of riscv_mmu() within the riscv target code can be replaced by a different condition, if need be.

Hence, I think that riscv_mmu() should work this way:

- If virt2phys_mode is OFF, always return "disabled".
- If virt2phys_mode is SW or HW, return "enabled" if and only if we detect that address translation enabled.

Do you see it the same?

fkhaidari:
Thanks for picking this back up, @JanMatCodasip. No worries about the delay. Yes, we're still interested in landing it, and moving to review.openocd.org works for me; I'll push it there.

On the semantics: I agree with your model for `riscv_mmu()`. Tying it to "is address translation actually active" is the right contract for `target_alloc_working_area_try()`. If the user has set `virt2phys_mode` off, they're working with physical addresses, so the working area should come from `working_area_phys` and `.mmu` must report "disabled". So:

- `virt2phys_mode` OFF: always disabled
- `virt2phys_mode` SW/HW: enabled if translation is detected (not M-mode, satp.MODE != Bare)

One caveat, so we don't lose the original motivation of this PR. The reason I touched `riscv_mmu()` was to make the `virt2phys` command return the translated address even when `virt2phys_mode` is off. Under your model `riscv_mmu()` reports "disabled" in that case, so the fix would regress if the `virt2phys` path keeps relying on the `.mmu` callback.

The fix is to decouple the two: keep `riscv_mmu()` strictly for working-area selection per your two rules, and have the `virt2phys` command (and `riscv_rw_memory`) decide whether to translate from a direct `satp` check instead of reusing `.mmu`. That way `.mmu` stays correct for working areas and `virt2phys` can still translate when asked.

If that works for you, I'll rework the patch along these lines and push it to review.openocd.org.

JanMatCodasip:
Hi @fkhaidari,

thank you for your reply and explanation.

> fkhaidari: The fix is to decouple the two: keep riscv_mmu() strictly for working-area selection per your two rules, and have the virt2phys command (and riscv_rw_memory) decide whether to translate from a direct satp check instead of reusing .mmu. That way .mmu stays correct for working areas and virt2phys can still translate when asked.

Yes, this is exactly how I understand the situation. Please feel free to prepare the patch to review.openocd.org and ping me so that I can review it without unnecessary delays.

PR #1257, opened 2025-05-15
target/riscv: check nextdm address in abits range

zqb-all:


src/target/riscv/riscv-013.c
```diff
@@ -552,6 +553,11 @@ static bool check_dbgbase_exists(struct target *target)
 			LOG_TARGET_ERROR(target, "Reached the end of DM chain (detected %u DMs in total).", count);
 			break;
 		}
+		if (next_dm > (1U << info->abits) - 1) {
```
en-sc: [nitpick]
```suggestion
		if (next_dm >= 1U << info->abits) {
```
IMHO, it's a tiny bit simpler this way.

src/target/riscv/riscv-013.c
```diff
@@ -552,6 +553,11 @@ static bool check_dbgbase_exists(struct target *target)
...
+			LOG_TARGET_ERROR(target, "Supporting no more than %d bits address on this DMI bus. Aborting",
+					info->abits);
```
en-sc: The message seems to suggest that this is an OpenOCD limitation, however it's not. It's a HW issue. Please, take a look at my suggestion:
```suggestion
			LOG_TARGET_ERROR(target, "The address of the next Debug Module does not fit into %u bits, which is the width of the DMI bus address. This is a HW bug",
					info->abits);
```

In any case, please change the conversion specifier for `info->abits`, since it's an `unsigned int`, not `int`.

en-sc (commented):
@zqb-all, thank you for the patch!

Overall, it looks great. I have only two minor suggestions.

src/target/riscv/riscv-013.c
zqb-all: It's a HW issue. Addressed, thanks.

src/target/riscv/riscv-013.c
zqb-all: Thanks, I just thought that can use `if (next_dm >> info->abits)`

en-sc (dismissed):

src/target/riscv/riscv-013.c
```diff
@@ -552,6 +553,12 @@ static bool check_dbgbase_exists(struct target *target)
 			LOG_TARGET_ERROR(target, "Reached the end of DM chain (detected %u DMs in total).", count);
 			break;
 		}
+		if (next_dm >> info->abits) {
+			LOG_TARGET_ERROR(target, "The address of the next Debug Module does not fit into %u bits,"
+					"which is the width of the DMI bus address. This is a HW bug",
```
```suggestion
			LOG_TARGET_ERROR(target, "The address of the next Debug Module does not fit into %u bits, "
					"which is the width of the DMI bus address. This is a HW bug",
```
Detail: There is a missing space in the message.

MarekVCodasip (commented):
Hello, found a small detail, otherwise looks fine.

src/target/riscv/riscv-013.c
zqb-all: Nice catch! Addressed

JanMatCodasip (approved):
LGTM, thank you.

src/target/riscv/riscv-013.c
MarekVCodasip: thanks

MarekVCodasip (approved):

PR #1258, opened 2025-05-15
target/riscv: fix address translation in hypervisor mode

zqb-all:
address translation don't need to care hstatus.HU

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -3035,6 +3035,8 @@ static int riscv_mmu(struct target *target, int *enabled)
 	unsigned int xlen = riscv_xlen(target);
 
 	if (v_mode) {
+		/* In VU or VS mode, MMU is considered enabled when
+		 * either hgatp or vsatp mode is not OFF */
 		riscv_reg_t vsatp;
 		if (riscv_reg_get(target, &vsatp, GDB_REGNO_VSATP) != ERROR_OK) {
 			LOG_TARGET_ERROR(target, "Failed to read vsatp register; priv=0x%" PRIx64,
@@ -3042,7 +3044,7 @@ static int riscv_mmu(struct target *target, int *enabled)
 			return ERROR_FAIL;
 		}
 		/* vsatp is identical to satp, so we can use the satp macros. */
-		if (RISCV_SATP_MODE(xlen) != SATP_MODE_OFF) {
+		if (get_field(vsatp, RISCV_SATP_MODE(xlen)) != SATP_MODE_OFF) {
 			LOG_TARGET_DEBUG(target, "VS-stage translation is enabled.");
 			*enabled = 1;
 			return ERROR_OK;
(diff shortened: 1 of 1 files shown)


en-sc:
@zqb-all, can you please expand on the motivation for the patch?

I don't know much about virtualization, but AFAIU from [21.2.10. Hypervisor Guest Address Translation and Protection (hgatp) Register] of The RISC-V Instruction Set Manual: Volume II: Privileged Architecture version 20250508
> The hgatp register is considered active for the purposes of the address-translation algorithm unless the effective privilege mode is U and hstatus.HU=0.

The code you are deleting is correct.

What am I missing?

zqb-all:
> Field HU (Hypervisor in U-mode) controls whether the virtual-machine
> load/store instructions, HLV, HLVX, and HSV, can be used also in U-mode.
> When HU=1, these instructions can be executed in U-mode the same as in
> HS-mode. When HU=0, all hypervisor instructions cause an
> illegal-instruction exception in U-mode.

[hstatus.HU](https://github.com/riscv/riscv-isa-manual/blob/584390d9d872835b1bc45ab9ea946b56db3ad485/src/hypervisor.adoc?plain=1#L252-L256) only affects the behavior of the cpu when executing the virtual-machine load/store instructions `HLV`, `HLVX`, and `HSV`. When we do address translation in openocd, we should behave like a cpu doing normal load/store instructions

> The `hgatp` register is considered _active_ for the purposes of the
> address-translation algorithm _unless_ the effective privilege mode is U
> and `hstatus`.HU=0.

[active ](https://github.com/riscv/riscv-isa-manual/blob/584390d9d872835b1bc45ab9ea946b56db3ad485/src/hypervisor.adoc?plain=1#L952-L954) , means that the hardware mmu address translation **may** use this register. However, even if the register is not active, it does not mean that the address translation is not working.
 In U mode, most of the time address translation is done by reading `satp` registers to do `VA`(virtual address) -> `PA`(Physical Addresses) . Only when cpu executes instructions such as `HLV`,  register `hgatp` and `vsatp` are required (active) to do `GVA`(Guest Virtual Address)->`GPA`(Guest Physical Address)->`SPA`(Supervisor Physical Address) 

Back to this patch, when we go into this [if (vmode)](https://github.com/riscv-collab/riscv-openocd/blob/1aebdf8e3025e8a2ac65a1ebcdccd11448d9b46e/src/target/riscv/riscv.c#L3037), it must be `VU` or `VS`, not `U` mode. And `VU` and `VS` do not need to consider `hstatus`.HU. To ensure that the address translation in `VU` mode works properly, we need to delete these code.

en-sc (approved):
@zqb-all, I'm sorry for taking so long to review this one.

LGTM, thanks for the fix!

One minor thing -- could you please rebase the MR on top of the target branch to drop the first commit  (f780c27af43712c16f961a9fcf04c0241685f47a)? Merging as-is will not break anything, however with rebase the history will be a bit nicer.

@MarekVCodasip, could you please take a look at this one?

zqb-all:
> One minor thing -- could you please rebase the MR on top of the target branch to drop the first commit ([f780c27](https://github.com/riscv-collab/riscv-openocd/commit/f780c27af43712c16f961a9fcf04c0241685f47a))? Merging as-is will not break anything, however with rebase the history will be a bit nicer.

 Sure, I also prefer the rebased clean history. Thanks for the review.

aap-sc (approved):

MarekVCodasip (approved):
Sorry for taking so long. I am not deeply versed in RISC-V hypervisor, but the explanation given sounds fine to me.

PR #1259, opened 2025-05-19
riscv_openocd_poll(): fix premature termination of algorithms under SMP

Wren6991:
This code forced all harts to halt when they reported a mixture of halted and non-halted status. This breaks long-running algorithms run on a single core under SMP, because the first poll will force the core executing the algorithm to halt. For example, this killed long-running flash erase operations on RP2350, which are mask ROM routines called by OpenOCD algorithms, running on a single core while the other core remains halted.

Fix by adding an exception when all running cores are running under debugger control (target->status == 4). Do not force harts to halt in this case.

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -4057,6 +4057,7 @@ int riscv_openocd_poll(struct target *target)
 	unsigned int should_resume = 0;
 	unsigned int halted = 0;
 	unsigned int running = 0;
+	unsigned int running_under_debug = 0;
 	unsigned int cause_groups = 0;
 	struct target_list *entry;
 	foreach_smp_target(entry, targets) {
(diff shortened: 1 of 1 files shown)


MarekVCodasip (approved):
Looks good. Thanks for the fix.

I am just curious: Was this tested on another fork of OpenOCD? This fork does not support SWD for RISC-V targets, which is needed for RP2350.

en-sc (approved):
@Wren6991, thank you for the patch! Sorry for taking so long to review. LGTM

Wren6991:
> I am just curious: Was this tested on another fork of OpenOCD? This fork does not support SWD for RISC-V targets, which is needed for RP2350.

Yes this was tested on our downstream fork, but we've been working with an OpenOCD maintainer to get our patches upstream. Upstream now has support for accessing a Debug Module directly through an ADI DAP.

You can actually run this fork (`riscv-openocd`) against RP2350 with some slightly spicy probe firmware that emulates a JTAG-DTM and tunnels the writes through ADI, but that wouldn't be useful for a repro here, because you don't have the RP2350 flash driver. This could probably be repro'd on SMP spike -- there are already a few tests in `riscv-tests` that actually seem to be covering OpenOCD regressions.

JanMatCodasip:
Overall, this patch looks all right to me. Sorry for letting it sit here in this stale fork.

@Wren6991, would you like to submit it to the vanilla OpenOCD instead (review.openocd.org)?

PR #1261, opened 2025-05-28
target/riscv: extend `trigger` controls

lz-bro:
* Add control for action mode.
* Introduce `riscv mtrigger` command.
* Introduce `riscv *trigger list` command

Action identifiers 2-5 are reserved for trace actions in the RISC-V Debug Specification, I would like to use them in the future .

### DIFF
--- a/doc/openocd.texi
+++ b/doc/openocd.texi
@@ -11727,13 +11727,26 @@ the target supports, these commands let you
 set those triggers directly. (It's also possible to do so by writing the
 appropriate CSRs.)
 
-@deffn {Command} {riscv etrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] exception_codes
+@option{trigger_action} selects one of the configurable RISC-V trigger actions:
+
+@itemize @minus
+@item @option{exception}:  Raise a breakpoint exception
+@item @option{halt}: Enter Debug Mode
+@item @option{trace_on}:  Trace on
+@item @option{trace_off}: Trace off
+@item @option{trace_notify}: Trace notify
+@item @option{external0}: Send a signal to TM external trigger output 0
+@item @option{external1}: Send a signal to TM external trigger output 1
+@end itemize
+
+@deffn {Command} {riscv etrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] exception_codes [@option{trigger_action}]
 Set an exception trigger (type 5) on the current target, which halts the target when it
 fires.  @option{m}, @option{s}, @option{u}, @option{vs}, and @option{vu} control
 which execution modes the trigger fires in. @var{exception_codes} is a bit
 field, where each bit corresponds to an exception code in mcause (defined in the
 RISC-V Privileged Spec). The etrigger will fire on the exceptions whose bits are
-set in @var{exception_codes}.
+set in @var{exception_codes}. @option{trigger_action} control the desired behavior
+on the target when the trigger fires. Defaults to @option{halt}, which halts the target.
 
 For details on this trigger type, see the RISC-V Debug Specification.
 @end deffn
@@ -11742,7 +11755,11 @@ For details on this trigger type, see the RISC-V Debug Specification.
 Clear the type 5 trigger that was set using @command{riscv etrigger set}.
 @end deffn
 
-@deffn {Command} {riscv icount set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] [@option{pending}] count
+@deffn {Command} {riscv etrigger list}
+List the type 5 trigger that was set using @command{riscv etrigger set}.
+@end deffn
+
+@deffn {Command} {riscv icount set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] [@option{pending}] count [@option{trigger_action}]
 Set an instruction count
 trigger (type 3) on the current target, which halts the target when it fires.
 @option{m}, @option{s}, @option{u}, @option{vs}, and @option{vu} control which
@@ -11751,6 +11768,8 @@ pending bit is set, which is unlikely to be useful unless you're debugging the
 hardware implementation of this trigger.
 @var{count} sets the number of instructions to execute before the trigger is
 taken.
+@option{trigger_action} control the desired behavior on the target when the
+trigger fires. Defaults to @option{halt}, which halts the target.
 
 For details on this trigger type, see the RISC-V Debug Specification.
 @end deffn
@@ -11759,21 +11778,56 @@ For details on this trigger type, see the RISC-V Debug Specification.
 Clear the type 3 trigger that was set using @command{riscv icount set}.
 @end deffn
 
-@deffn {Command} {riscv itrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] [@option{nmi}] mie_bits
+@deffn {Command} {riscv icount list}
+List the type 3 trigger that was set using @command{riscv icount set}.
+@end deffn
+
+@deffn {Command} {riscv itrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] [@option{nmi}] mie_bits [@option{trigger_action}]
 Set an interrupt trigger (type 4) on the current target, which halts the target when it
 fires.  @option{m}, @option{s}, @option{u}, @option{vs}, and @option{vu} control
 which execution modes the trigger fires in.  If [@option{nmi}] is passed then
 the trigger will fire on non-maskable interrupts in those modes. @var{mie_bits}
 controls which interrupts the trigger fires on, using the same bit assignments
 as in the mie CSR (defined in the RISC-V Privileged Spec).
+@option{trigger_action} control the desired behavior on the target when the
+trigger fires. Defaults to @option{halt}, which halts the target.
 
 For details on this trigger type, see the RISC-V Debug Specification.
 @end deffn
 
+@deffn {Command} {riscv itrigger clear}
+Clear the type 4 trigger that was set using @command{riscv itrigger set}.
+@end deffn
+
+@deffn {Command} {riscv itrigger list}
+List the type 4 trigger that was set using @command{riscv itrigger set}.
+@end deffn
+
+@deffn {Command} {riscv mcontrol set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] address [@option{trigger_action}]
+Set a match trigger (type 2 or 6) on the current target, which sets the desired behavior on
+the target when it fires.  @option{m}, @option{s}, @option{u}, @option{vs}, and @option{vu}
+control which execution modes the trigger fires in. @option{execute}, @option{store},
+and @option{load} control the trigger how to fire.
+@option{trigger_action} control the desired behavior on the target when the
+trigger fires. Defaults to @option{halt}, which halts the target.
+
+For details on this trigger type, see the RISC-V Debug Specification.
+@end deffn
+
+@deffn {Command} {riscv mcontrol clear} [trigger_id]
+Clear the type 2 or 6 trigger that was set using @command{riscv mcontrol set}.
+@option{trigger_id} specifies the trigger id of a match trigger to clear.
+@end deffn
+
+@deffn {Command} {riscv mcontrol list}
+List the type 2 or 6 trigger that was set using @command{riscv mcontrol set}.
+@end deffn
+
 @deffn {Command} {riscv reserve_trigger} [index @option{on|off}]
 Manages the set of reserved triggers. Reserving a trigger results in OpenOCD
-not using it internally (e.g. skipping it when setting a watchpoint or a
-hardware breakpoint), so that the user or the application has unfettered
+not using it internally (e.g. skipping it when setting a watchpoint, a
+hardware breakpoint, an exception trigger, an instruction trigger or a
+match trigger), so that the user or the application has unfettered
 control over the trigger. By default there are no reserved triggers.
 
 @enumerate
@@ -11783,10 +11837,6 @@ control over the trigger. By default there are no reserved triggers.
 @item If called without parameters, returns indices of reserved triggers.
 @end enumerate
 
-@end deffn
-
-@deffn {Command} {riscv itrigger clear}
-Clear the type 4 trigger that was set using @command{riscv itrigger set}.
 @end deffn
 
 @subsection RISC-V Program Buffer Commands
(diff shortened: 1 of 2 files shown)


src/target/riscv/riscv.c
```diff
@@ -127,6 +127,8 @@ struct trigger {
 	uint64_t mask;
 	uint64_t value;
 	bool is_read, is_write, is_execute;
+	bool is_vs, is_vu, is_m, is_s, is_u;
+	uint32_t action;
```
MarekVCodasip: Please, consider introducing an enum `trigger_action` or similar instead of using `uint32_t`.

src/target/riscv/riscv.c
```diff
@@ -5009,6 +5023,88 @@ COMMAND_HELPER(riscv_clear_trigger, int trigger_id, const char *name)
...
+	if (riscv_reg_get(target, &tdata2, GDB_REGNO_TDATA2) != ERROR_OK)
+		return ERROR_FAIL;
+
+	static const char *const actions[] = {
+		"exception",
+		"halt",
+		"trace_on",
+		"trace_off",
+		"trace_notify",
+	};
```
MarekVCodasip: There are two more actions, 8 and 9 for sending a signal to trigger module external trigger output.

I would at least add a comment that they exist as we would need hardware which supports it to test it.

doc/openocd.texi
```diff
@@ -11789,6 +11806,31 @@ control over the trigger. By default there are no reserved triggers.
 Clear the type 4 trigger that was set using @command{riscv itrigger set}.
 @end deffn
 
+@deffn {Command} {riscv itrigger list}
+List the type 4 trigger that was set using @command{riscv itrigger set}.
+@end deffn
+
+@deffn {Command} {riscv mtrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}]
+[@option{execute}] [@option{store}] [@option{load}] [@option{exception}|@option{halt}|@option{trace_on}|
+@option{trace_off}|@option{trace_notify}] Set an match trigger (type 2 or 6) on the current target,
```
MarekVCodasip:
```suggestion
@option{trace_off}|@option{trace_notify}]
Set an match trigger (type 2 or 6) on the current target,
```
Detail: Newline seems to be missing

MarekVCodasip: I don't see you mention that one of the mandatory parameters is an address.

src/target/riscv/riscv.c
```diff
@@ -5059,16 +5166,19 @@ COMMAND_HANDLER(riscv_itrigger)
...
 
 	} else if (!strcmp(CMD_ARGV[0], "clear")) {
 		return riscv_clear_trigger(CMD, ITRIGGER_UNIQUE_ID, "itrigger");
 
+	} else if (!strcmp(CMD_ARGV[0], "list")) {
+		return riscv_list_trigger(CMD, ITRIGGER_UNIQUE_ID, "itrigger");
+
 	} else {
-		LOG_ERROR("First argument must be either 'set' or 'clear'.");
+		LOG_ERROR("First argument must be either 'set', 'clear' or 'list'.");
```
MarekVCodasip, line 5187: Please use LOT_TARGET_* wherever possible. This applies elsewhere as well.

doc/openocd.texi
```diff
@@ -11789,6 +11806,31 @@ control over the trigger. By default there are no reserved triggers.
...
+@option{vs}, and @option{vu} control which execution modes the trigger fires in. @option{execute},
+@option{store}, and @option{load} control the trigger how to fire.
+@option{exception}|@option{halt}|@option{trace_on}|@option{trace_off}|@option{trace_notify}
+control the desired behavior on the target when the trigger fires. Defaults to @option{halt},
+which halts the target.
+
+For details on this trigger type, see the RISC-V Debug Specification.
+@end deffn
+
+@deffn {Command} {riscv mtrigger clear}
```
MarekVCodasip: I don't think `mtrigger` is a good term to describe this. I recommend using `mcontrol` instead.

doc/openocd.texi
```diff
@@ -11728,12 +11728,15 @@ set those triggers directly. (It's also possible to do so by writing the
 appropriate CSRs.)
 
 @deffn {Command} {riscv etrigger set} [@option{m}] [@option{s}] [@option{u}] [@option{vs}] [@option{vu}] exception_codes
-Set an exception trigger (type 5) on the current target, which halts the target when it
-fires.  @option{m}, @option{s}, @option{u}, @option{vs}, and @option{vu} control
-which execution modes the trigger fires in. @var{exception_codes} is a bit
-field, where each bit corresponds to an exception code in mcause (defined in the
-RISC-V Privileged Spec). The etrigger will fire on the exceptions whose bits are
-set in @var{exception_codes}.
+[@option{exception}|@option{halt}|@option{trace_on}|@option{trace_off}|@option{trace_notify}]
```
MarekVCodasip:
```suggestion
[@option{exception}|@option{halt}|@option{trace_on}|@option{trace_off}|@option{trace_notify}]
```
```suggestion
[@option{trigger_action}]
```
You can simplify the syntax of the commands by saying `trigger_action`. Later, you can explain what are the valid values for the attribute. This does not have to be repeated for every trigger command, but can be mentioned just once.

This can maybe go to the top of the RISC-V Trigger Commands subsection. Before the individual trigger commands.

src/target/riscv/riscv.c
```diff
@@ -5186,16 +5321,163 @@ COMMAND_HANDLER(riscv_etrigger)
...
+				action = CSR_MCONTROL_ACTION_TRACE_OFF;
+			else if (!strcmp(CMD_ARGV[i], "trace_notify"))
+				action = CSR_MCONTROL_ACTION_TRACE_NOTIFY;
+			else
+				COMMAND_PARSE_ADDRESS(CMD_ARGV[i], address);
+		}
+		if (address == ULLONG_MAX) {
+			LOG_ERROR("Doesn't make sense to set match trigger with "
+				  "address=0xFFFFFFFFFFFFFFFF.");
+			return ERROR_FAIL;
```
MarekVCodasip: Is there a reason to check for this?

src/target/riscv/riscv.c
```diff
@@ -5791,22 +6073,33 @@ static const struct command_registration riscv_exec_command_handlers[] = {
...
+			"[('exception'|'halt'|'trace_on'|'trace_off'|'trace_notify')]|clear|list",
+		.help = "Set, clear or list a single interrupt trigger."
+	},
+	{
+		.name = "mtrigger",
+		.handler = riscv_mtrigger,
+		.mode = COMMAND_EXEC,
+		.usage = "set [vs] [vu] [m] [s] [u] [execute] [store] [load] <address> "
+			"[('exception'|'halt'|'trace_on'|'trace_off'|'trace_notify')] |clear [trigger_id]|list",
+		.help = "Set, clear or list a single match trigger."
```
MarekVCodasip: Is the goal to have multiple mtriggers? It looks to me that for itrigger, etrigger and icount, there can be at most one such trigger set, but for mtrigger, multiple are permitted.

For multi-trigger commands, the subcommands should be called `add`, `remove` and `list`. For single-trigger commands, the command names should be `set`, `clear` and `show` (or `get`).

src/target/riscv/riscv.c
```diff
@@ -5047,6 +5144,16 @@ COMMAND_HANDLER(riscv_itrigger)
...
+			else if (!strcmp(CMD_ARGV[i], "exception"))
+				action = CSR_ITRIGGER_ACTION_BREAKPOINT;
+			else if (!strcmp(CMD_ARGV[i], "halt"))
+				action = CSR_ITRIGGER_ACTION_DEBUG_MODE;
+			else if (!strcmp(CMD_ARGV[i], "trace_on"))
+				action = CSR_ITRIGGER_ACTION_TRACE_ON;
+			else if (!strcmp(CMD_ARGV[i], "trace_off"))
+				action = CSR_ITRIGGER_ACTION_TRACE_OFF;
+			else if (!strcmp(CMD_ARGV[i], "trace_notify"))
+				action = CSR_ITRIGGER_ACTION_TRACE_NOTIFY;
```
MarekVCodasip: It seems that this part of the code is repeated in many places, could try to refactor it?

It could be a function which takes a look on the last argument of the command and decides if its one of the valid trigger actions.

src/target/riscv/riscv.c
```diff
@@ -5009,6 +5023,88 @@ COMMAND_HELPER(riscv_clear_trigger, int trigger_id, const char *name)
 	return remove_trigger(target, trigger_id);
 }
 
+COMMAND_HELPER(riscv_list_trigger, int trigger_id, const char *name)
+{
+	struct target *target = get_current_target(CMD_CTX);
+	if (CMD_ARGC != 1)
+		return ERROR_COMMAND_SYNTAX_ERROR;
```
MarekVCodasip, line 5051: It seems to me that this can only display one trigger? If I understood correctly there can be multiple mtrigger ones.

src/target/riscv/riscv.c
```diff
@@ -5186,16 +5321,163 @@ COMMAND_HANDLER(riscv_etrigger)
...
 	} else {
-		LOG_ERROR("First argument must be either 'set' or 'clear'.");
+		LOG_ERROR("First argument must be either 'set', 'clear' or 'list'.");
+		return ERROR_COMMAND_SYNTAX_ERROR;
+	}
+	return ERROR_OK;
+}
+
+/* monotonic counter/id-number for match triggers */
+static int mtrigger_unique_id = -CSR_TDATA1_TYPE_MCONTROL6;
```
MarekVCodasip: I am not sure that CSR_TDATA1_TYPE_MCONTROL6 is a good way to base the IDs on.

I think it would be better to define two constants, `MCONTROL_MANUAL_TRIGGER_ID_START` and `MCONTROL_MANUAL_TRIGGER_ID_END` and use those.

Breakpoints and watchpoints placed by OpenOCD start at 0 and count up, so ensure the range for manual triggers is in negative numbers.

src/target/riscv/riscv.c
MarekVCodasip: Is there a reason for not matching on address ULLONG_MAX?

doc/openocd.texi
MarekVCodasip, line 1: Do I understand correctly that the reserved triggers (by `riscv reserve_trigger`) cannot be used by any of the commands listed in this subsection?

If yes, please make it clear in the documentation for `riscv reserve_trigger`.

MarekVCodasip (changes requested):
Hello, thanks for taking the effort for this patch. Here some things that I have found.

src/target/riscv/riscv.c
lz-bro: Addressed.

src/target/riscv/riscv.c
lz-bro: Addressed.

doc/openocd.texi
lz-bro: Addressed.

lz-bro: Addressed, thanks.

src/target/riscv/riscv.c
lz-bro: This is a a syntax error, Is it better to use LOG_ERROR?

doc/openocd.texi
lz-bro: Addressed.

doc/openocd.texi
lz-bro: Thanks, addressed.

src/target/riscv/riscv.c
lz-bro: `riscv_addr_t address = ULLONG_MAX;`

`address` is a mandatory parameter,  I choose to initialize the `address` to `ULLONG_MAX` instead of `0`, so it is necessary to check it here.

Same like `count` in `riscv_icount`, `exception_codes` in `riscv etrigger` and `interrupts` in `riscv itrigger`:
```
unsigned int count = 0;

if (count == 0) {
                        LOG_ERROR("Doesn't make sense to set icount trigger with "
                                 "count=0.");
                        return ERROR_FAIL;
}
...
```

src/target/riscv/riscv.c
lz-bro: Addressed.

src/target/riscv/riscv.c
lz-bro: Description as above.

doc/openocd.texi
lz-bro: Addressed.

src/target/riscv/riscv.c
lz-bro: For each physical trigger contains:
-1: the trigger is available
-3: The trigger is used by the icount command
-4: The trigger is used by the itrigger command
-5: The trigger is used by the etrigger command
\>= 0: unique_id of the breakpoint/watchpoint that is using it.

So, unique_id <= -6 could be used by the mcontrol command.
```
/* monotonic counter/id-number for match triggers */
static int mcontrol_unique_id = -CSR_TDATA1_TYPE_MCONTROL6;

unique_id = mcontrol_unique_id--;
```

src/target/riscv/riscv.c
lz-bro: I do not agresswith you, this can display all match triggers. You can see:

```
       } else if (!strcmp(CMD_ARGV[0], "list")) {
               for (unsigned int i = 0; i < r->trigger_count; i++) {
                       if (r->trigger_unique_id[i] <= -CSR_TDATA1_TYPE_MCONTROL6) {
                               if (riscv_list_trigger(CMD, r->trigger_unique_id[i], "match trigger") != ERROR_OK) {
                                       LOG_TARGET_ERROR(target, "Failed to list match trigger (unique_id=%" PRIi64 ").",
                                                       r->trigger_unique_id[i]);
                                       return ERROR_FAIL;
                               }
                       }
               }

```

It is only for `list` command syntax.
```
	if (CMD_ARGC != 1)
```

src/target/riscv/riscv.c
MarekVCodasip: Yes, LOG_TARGET_*, not LOT_TARGET_*. Typo on my part. Sorry.

src/target/riscv/riscv.c
MarekVCodasip: My mistake, you are right.

src/target/riscv/riscv.c
MarekVCodasip: Alright, in that case, please add a comment explaining this or change the log message to reflect that it is not set.

lz-bro: Addressed.

lz-bro:
@MarekVCodasip can we continue merging or any comments?

en-sc:
@lz-bro, I'd suggest you post the patch to mainline OpenOCD (https://review.openocd.org) -- the RISC-V part got updated there and now it is more recent then in this repo.
Moreover, it seems like @MarekVCodasip and @JanMatCodasip are no longer actively maintaining this repo and I no longer have write access, so continuing the development in the mainline OpenOCD seems like the only option.

lz-bro:
@en-sc Thank you! I'll push the patch to mainline OpenOCD.

JanMatCodasip:
> lz-bro: I'll push the patch to mainline OpenOCD.

Closing this PR.

PR #1265, opened 2025-06-13
target/riscv: Fix some timeout check order

lz-bro:
Reorder the timeout check and conditional judgment to ensure that timeout check is not wasted.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -534,7 +534,33 @@ static int dm_write(struct target *target, uint32_t address, uint32_t value)
 	return dmi_write(target, riscv013_get_dmi_address(target, address), value);
 }
 
-static bool check_dbgbase_exists(struct target *target)
+static int activate_dm(struct target *target, uint32_t dm_base_addr)
+{
+	uint32_t dmcontrol = 0;
+
+	LOG_TARGET_DEBUG(target, "Activating the DM with DMI base address (dbgbase) = 0x%x", dm_base_addr);
+	if (dmi_write(target, DM_DMCONTROL + dm_base_addr, DM_DMCONTROL_DMACTIVE) != ERROR_OK)
+		return ERROR_FAIL;
+
+	const time_t start = time(NULL);
+	LOG_TARGET_DEBUG(target, "Waiting for the DM to become active.");
+	while (1) {
+		if (dmi_read(target, &dmcontrol, DM_DMCONTROL + dm_base_addr) != ERROR_OK)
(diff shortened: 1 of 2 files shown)


MarekVCodasip (dismissed):
Thanks for adding the change elsewhere as well.

src/target/riscv/riscv-013.c
```diff
@@ -1887,19 +1887,19 @@ static int reset_dm(struct target *target)
...
 		const time_t start = time(NULL);
 		LOG_TARGET_DEBUG(target, "Waiting for the DM to acknowledge reset.");
-		do {
+		while (1) {
 			result = dm_read(target, &dmcontrol, DM_DMCONTROL);
 			if (result != ERROR_OK)
 				return result;
-
+			if (!get_field32(dmcontrol, DM_DMCONTROL_DMACTIVE))
+				break;
```

During the refactor, the debug message `DM reset initiated` got removed. I am not sure if this was intended or simply an oversight.

I would prefer to keep the message there. The message can also be slightly reworded for clarity.

```c
if (!get_field32(dmcontrol, DM_DMCONTROL_DMACTIVE)) {
	LOG_TARGET_DEBUG(target, "The DM has just become deactivated (dmcontrol.dmactive: 1 -> 0).");
	break;
}
```

JanMatCodasip (changes requested):

src/target/riscv/riscv-013.c
lz-bro: Addressed, thanks.

en-sc (commented):
@lz-bro, to be honest, I don't understand the need for this patch.

1. `dmi_write()` is time-bound by the same `riscv_get_command_timeout_sec()` (see https://github.com/riscv-collab/riscv-openocd/blob/8d5f2febf3d3780c0926ec4b4f0cd5e6c0ac7e74/src/target/riscv/riscv-013.c#L2531), therefore this is mostly a hypothetical issue.
2. IMHO, this is more straightforward when an operation that took longer then the timeout fails,  indicating that you need to increase the timeout.

JanMatCodasip:
@lz-bro, is this patch still relevant, please? @en-sc has indicated that he does not see any need for it.

If it is important for reasons that we do not see, please submit it to review.openocd.org with a proper description.

Thanks.

lz-bro:
> @lz-bro, is this patch still relevant, please? @en-sc has indicated that he does not see any need for it.
> 
> If it is important for reasons that we do not see, please submit it to review.openocd.org with a proper description.
> 
> Thanks.

No, thanks.

PR #1267, opened 2025-06-20
Merge up to afbd01b0a46f3a81fe6076c002ad66973dcfb64c from upstream

en-sc:


.github/workflows/snapshot.yml
```diff
@@ -11,7 +11,7 @@ name: OpenOCD Snapshot
 
 jobs:
   package:
-    runs-on: [ubuntu-20.04]
+    runs-on: [ubuntu-latest]
```
MarekVCodasip, line 14: Are you sure it is wise to set it to `ubuntu-latest` and not the latest LTS (24.04)?

MarekVCodasip (commented):
Just one thing I noticed, please take a look, otherwise looks fine.

.github/workflows/snapshot.yml
en-sc: First of all this CI is part of mainline OpenOCD's GitHub clone -- I believe it will be maintained.

Moreover, IMHO it's fine. Considering the fact that GitHub can drop particular containers (similar to what happened to `ubuntu-20.04`) I'm not sure what will be more reliable.

MarekVCodasip: Alright then, should be fine then.

MarekVCodasip (approved):

PR #1269, opened 2025-06-20
target/riscv: access registers via `reg->type`

en-sc:
* `int riscv_reg_get()` and `int riscv_reg_set()` are implemented in
  terms of `reg->type->get/set` instead of the other way around. This
  makes it easier to support custom behavior for some registers.
* Cacheability is determined by `reg->type` instead of
  `riscv_reg_impl_gdb_regno_cacheable()`.
* Issues with redirection of `priv` -> `dcsr` and `pc` -> `dpc` are
  addressed at the "topmost" level.
    - `priv` and `pc` are always invalid.
    - Fixed some issues, e.g. the first `pc` write printed-out an
      uninitialized value:
```
> reg pc 0
pc (/64): 0x000075da6b33db20
```

### DIFF
--- a/src/target/riscv/riscv-013_reg.c
+++ b/src/target/riscv/riscv-013_reg.c
@@ -13,73 +13,455 @@
 #include "debug_defines.h"
 #include <helper/time_support.h>
 
-static int riscv013_reg_get(struct reg *reg)
+/* Start of register fetch/send definitions.
+ * - "reg_fetcher" (named "fetch_*") -- loads the value from target into reg
+ *   cache entry. Modifies only "reg->value".
+ * - "reg_sender" (named "send_*") -- stores the value from "buf" to the
+ *   target. Does not modify "reg" at all.
+ */
+typedef int (*reg_fetcher)(struct reg *reg);
+typedef int (*reg_sender)(const struct reg *reg, const uint8_t *buf);
+
+/* TODO: Introduce separate "fetch_*" and "send_*" functions for register
+ * classes (xregs, fregs, csrs).
+ */
+static int fetch_riscv013_reg(struct reg *reg)
 {
-	struct target *target = riscv_reg_impl_get_target(reg);
+	riscv_reg_t value;
+	int res = riscv013_get_register(riscv_reg_impl_get_target(reg),
+			&value, reg->number);
+	if (res != ERROR_OK)
+		return res;
+	buf_set_u64(reg->value, 0, reg->size, value);
+	return res;
+}
 
-	/* TODO: Hack to deal with gdb that thinks these registers still exist. */
-	if (reg->number > GDB_REGNO_XPR15 && reg->number <= GDB_REGNO_XPR31 &&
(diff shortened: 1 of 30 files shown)


src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,455 @@
...
+			reg->valid ? "true" : "false",
+			reg->dirty ? "true" : "false");
+	return ERROR_OK;
+}
+
+static void access_epilogue(const struct reg *reg, bool write)
+{
+	assert(riscv_reg_impl_is_existing(reg));
+	if (debug_level < LOG_LVL_DEBUG)
+		return;
```
MarekVCodasip, line 125: I don't think this early return is needed.

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,455 @@
...
+{
+	return riscv013_get_register_buf(riscv_reg_impl_get_target(reg),
+			reg->value, reg->number);
+}
+
+static int send_vreg(const struct reg *reg, const uint8_t *buf)
+{
+	return riscv013_set_register_buf(riscv_reg_impl_get_target(reg),
+			reg->number, reg->value);
+}
```
MarekVCodasip, line 59: Aren't we ignoring the `buf` paramter here?

MarekVCodasip (commented):
These are the two things I have found so far, I will take another look later. So far I like the changes.

src/target/riscv/riscv-013_reg.c
```diff
@@ -13,73 +13,455 @@
...
 
-static int riscv013_reg_get(struct reg *reg)
+/* Start of register fetch/send definitions.
+ * - "reg_fetcher" (named "fetch_*") -- loads the value from target into reg
+ *   cache entry. Modifies only "reg->value".
+ * - "reg_sender" (named "send_*") -- stores the value from "buf" to the
+ *   target. Does not modify "reg" at all.
+ */
+typedef int (*reg_fetcher)(struct reg *reg);
+typedef int (*reg_sender)(const struct reg *reg, const uint8_t *buf);
```
MarekVCodasip, line 23: I have a feeling that the `send` keyword isn't the best. Maybe `reg_writer` or similar could be better. Up to your consideration.

MarekVCodasip (commented):
Took another looks, seems fine other than one nitpick.

JanMatCodasip:
This MR depends on the not-merged patch https://review.openocd.org/c/openocd/+/8070.

@en-sc, I think we should:

- Review and merge a temporary hotfix to immediately avoid the data loss issues associated with force-read, and have this part of OpenOCD v1.0.0 (my proposal: https://review.openocd.org/c/openocd/+/9737).
- Then draft a clean register access API including cache control, as Antonio suggests [here](https://review.openocd.org/c/openocd/+/8070/comments/44f61273_9cbb596c).
- Then work from there.

PR #1273, opened 2025-07-03
target/riscv/riscv.c: support auto-selection of software breakpoint size

AlexandraKulyatskaya:
Add auto-selection of breakpoint size for RISC-V targets. The new length depends on C extension. If the C extension is used, the auto-selected length is 2 bytes. Otherwise, the default length is 4 bytes.

Change-Id: Ie3473514beace15b76714aa6d5441122cd3262aa

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1609,6 +1609,16 @@ static int riscv_add_breakpoint(struct target *target, struct breakpoint *breakp
 	LOG_TARGET_DEBUG(target, "@0x%" TARGET_PRIxADDR, breakpoint->address);
 	assert(breakpoint);
 	if (breakpoint->type == BKPT_SOFT) {
+		if (breakpoint->length == 0) {
+			breakpoint->length = riscv_supports_extension(target, 'c') ? 2 : 4; 
+			uint8_t *instr_buffer = (uint8_t *)realloc (breakpoint->orig_instr, breakpoint->length);
+			if (!instr_buffer) {
+				LOG_ERROR("Fail to realloc memory with original instruction");
+				return ERROR_FAIL;
+			}
+			breakpoint->orig_instr = instr_buffer;
+		}
+
 		/** @todo check RVC for size/alignment */
 		if (!(breakpoint->length == 4 || breakpoint->length == 2)) {
 			LOG_TARGET_ERROR(target, "Invalid breakpoint length %d", breakpoint->length);


en-sc:
@AlexandraKulyatskaya, I'd suggest avoiding modifying the length of the breakpoint.

Consider instead introducing a new `target_type` method, something like `get_default_breakpoint_length` to be used in [handle_bp_command_set()](https://github.com/riscv-collab/riscv-openocd/blob/6f84e90d5cc0e98f895ea0fa31212326b50ab769/src/target/target.c#L3944) in case the user wants a default-length breakpoint.
IMHO, this will allow to avoid introducing special cases and simplify reasoning about breacpoint's lifecycle.

Please, note that in general `get_default_breakpoint_length` may fail if it requires communication with the target.

MarekVCodasip:
I don't think this will work anyway as support of c.ebreak isn't guaranteed when "C" extension is supported https://github.com/riscv-collab/riscv-openocd/pull/908#issuecomment-1771153677

TommyMurphyTM1234:
> I don't think this will work anyway as support of c.ebreak isn't guaranteed when "C" extension is supported [#908 (comment)](https://github.com/riscv-collab/riscv-openocd/pull/908#issuecomment-1771153677)

Hi @MarekVCodasip - I think that my comment that you linked to may have been based on incomplete or inaccurate information at the time (similar to previous comments linked from there etc.). If you look at the [latest ratified](https://riscv.atlassian.net/wiki/spaces/HOME/pages/16154769/RISC-V+Technical+Specifications#ISA-Specifications) [RISC-V Instruction Set Manual Volume I: Unprivileged ISA](https://drive.google.com/file/d/1uviu1nH-tScFfgrovvFCrj7Omv8tFtkp/view?usp=drive_link) then page 171 of chapter 27 has this as part of the `C` extension description:



So it seems that `c.break` IS part of the `C` extension after all?

MarekVCodasip:
> Hi @MarekVCodasip - I think that my comment that you linked to may have been based on incomplete or inaccurate information at the time (similar to previous comments linked from there etc.). If you look at the [latest ratified](https://riscv.atlassian.net/wiki/spaces/HOME/pages/16154769/RISC-V+Technical+Specifications#ISA-Specifications) [RISC-V Instruction Set Manual Volume I: Unprivileged ISA](https://drive.google.com/file/d/1uviu1nH-tScFfgrovvFCrj7Omv8tFtkp/view?usp=drive_link) then page 171 has this as part of the `C` extension description:
> 
> 
> So it seems that `c.break` IS part of the `C` extension after all?

Thanks for the info, I just remember that this was an issue back then as I tried to implement code to reject software breakpoints of size `2` and it had to be reverted.

TommyMurphyTM1234:
> Thanks for the info, I just remember that this was an issue back then as I tried to implement code to reject software breakpoints of size `2` and it had to be reverted.

I think that some of the confusion may have arisen from the `C` extension (or sub-extensions?) specifications omitting to mention `c.break` at some stage (maybe while still in draft form or an earlier ratified spec that was subsequently updated?) and it only being incorporated into the `C` extension specification and then incorporation of that extension specification into the main Unprivileged ISA spec? To be honest I am often confused by the way that RISC-V extension specification is done and how RISC-V International makes the relevant information available and changed resulting in a plethora of broken links... :-(

AlexandraKulyatskaya:
Reopened as https://github.com/riscv-collab/riscv-openocd/pull/1288

PR #1274, opened 2025-07-07
target/riscv: improve error messaging in case `sbasize` is zero

### LINKED ISSUE #1270: Memory Sampling support when sbasize is zero.
From the RISC-V Debug Specification (ratified February 2025), in the definition of the sbcs register, the sbasize field is defined (Page 44, Table on top) as:

`sbasize : Width of system bus addresses in bits. (0 indicates there is no bus access support.)`

But these [lines](https://github.com/riscv-collab/riscv-openocd/blob/6f84e90d5cc0e98f895ea0fa31212326b50ab769/src/target/riscv/riscv-013.c#L2613-#L2616) in riscv-013.c, where sbasize is examined to return status about implementation support for memory sampling, do not include the case when sbasize is zero.

`if (sbasize > 64) {`
`	LOG_TARGET_ERROR(target, "Memory sampling is only implemented for sbasize <= 64.");`
`	return ERROR_NOT_IMPLEMENTED;`
`}`	

This behaviour was observed when trying MemorySample tests from the RISC-V Debug Test suite on a design with only Program Buffer support.

The following patch fixes the issue:

`if (sbasize == 0 || sbasize > 64)`

Thanks.

en-sc:
From: Sriram Shanmuga <sriramharshalee@gmail.com>

RISC-V Debug Specification v1.0 [3.14.22. System Bus Access Control and Status (`sbcs`, at 0x38)] states in `sbasize` field description:
> Width of system bus addresses in bits. (0 indicates there is no bus
access support.)

Before the patch, the error message did not include the information about `sbcs.sbasize` being zero wich made it quite undescriptive:
```
[riscv.cpu] Turning off memory sampling because it failed.

```

Fixes #1270

Change-Id: I5402dd57dc9a81f65ee4c67d24e11c366006427c

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2610,8 +2610,8 @@ static int sample_memory_bus_v1(struct target *target,
 {
 	RISCV013_INFO(info);
 	unsigned int sbasize = get_field(info->sbcs, DM_SBCS_SBASIZE);
-	if (sbasize > 64) {
-		LOG_TARGET_ERROR(target, "Memory sampling is only implemented for sbasize <= 64.");
+	if (sbasize == 0 || sbasize > 64) {
+		LOG_TARGET_ERROR(target, "Memory sampling is only implemented for non-zero sbasize <= 64.");
 		return ERROR_NOT_IMPLEMENTED;
 	}
 


en-sc:
@RaamyPi, please take a look

RaamyPi:
Looks good, thank you!

aap-sc (approved):

MarekVCodasip (approved):

PR #1280, opened 2025-08-08
Autoerase for bin

Vishwajith26:


aap-sc:
@Vishwajith26 I'm sorry but as it is - there are lots of issues with the commit itself. Some of these changes should not even be published here, but should go directly to upstream openocd. riscv-specific code looks dubious too (since you hard-code breakpoint size).

But before we begin the discussion regarding the implementation itself, some general topics must be discussed.

1. what are 8-bytes breakpoints mentioned in this commit? Maybe this is a mistake on your part and these change set is not intended for publication? If it is not -  is there a public ally available  specification? Is it ratified by RISC-V International?
2. I guess that the MR needs a detailed description in any case.

PR #1282, opened 2025-08-16
switch snapshot job to 22.04

aap-sc:


en-sc:
@aap-sc, this file is part of mainline OpenOCD, and the proposed patch is already present here. Please, see https://github.com/riscv-collab/riscv-openocd/pull/1267.

PR #1295, opened 2025-10-01
Fix bug in scratch_reserve

fkhaidari:


fkhaidari:
@JanMatCodasip, @MarekVCodasip, could you take a look? This change is required to make this [Spike MR](https://github.com/riscv-software-src/riscv-isa-sim/pull/2086) pass the testing

aap-sc (approved):

fkhaidari:
Failure example: `gdbserver.py targets/RISC-V/spike32.py DebugChangeString` for target with datasize greater than 2

fkhaidari:
There's a logic error in the comparisons in `scratch_reserve()`. The conditions are backwards - they check if the required space is **greater than or equal to** the available space, when they should check if it **fits within** the available space.

If riscv_xlen == 32 then fpr_read_progbuf calls internal_register_read64_progbuf_scratch function that calls scratch_reserve. And here we have our comparisons:

Current code (wrong):
```c
// Option 1: data registers
if ((size_bytes + scratch->hart_address - info->dataaddr + 3) / 4 >=
		info->datasize) {
    // use data registers
}

// Option 2: progbuf  
if ((info->progbuf_writable == YNM_YES) &&
		((size_bytes + scratch->hart_address - info->progbuf_address + 3) / 4 >=
		info->progbufsize)) {
    // use progbuf
}
```

What happens:
- With `datasize=8` condition is **false**, so it skips using data registers even though there's plenty of space
- With `datasize=2` condition is **true**, so it uses data registers

The conditions should use `<=` instead of `>=`:

```c
// Option 1: data registers
if ((size_bytes + scratch->hart_address - info->dataaddr + 3) / 4 <=
		info->datasize) {
    // fine for datasize > 2
}

// Option 2: progbuf  
if ((info->progbuf_writable == YNM_YES) &&
		((size_bytes + scratch->hart_address - info->progbuf_address + 3) / 4 <=
		info->progbufsize)) {
    // fine for larger progbuf
}
```

TommyMurphyTM1234:
> The conditions should use `<=` instead of `>=`:

Was this issue (or at least part of it?) introduced in this PR?

- https://github.com/riscv-collab/riscv-openocd/pull/424

which includes this change:
```
	if ((info->progbuf_writable == YNM_YES) &&
			((size_bytes + scratch->hart_address - info->progbuf_address + 3) / 4 >=
			info->progbufsize)) {
```

fkhaidari:
@TommyMurphyTM1234, the bug was already there; #424 didn’t introduce it, it preserved the same `>=` check

MarekVCodasip (approved):
Changes look good to me, I wonder if there are more places where these macros could be used.

en-sc (approved):

PR #1296, opened 2025-10-07
fix error detection during trigger removal

aap-sc:
The error manifests on designs that allow CSR reads when hart is not halted. This results in `rwp` command to erroneously report "success"

### DIFF
--- a/src/target/riscv/riscv.c
+++ b/src/target/riscv/riscv.c
@@ -1669,14 +1669,23 @@ static int remove_trigger(struct target *target, int unique_id)
 
 	bool done = false;
 	for (unsigned int i = 0; i < r->trigger_count; i++) {
-		if (r->trigger_unique_id[i] == unique_id) {
-			riscv_reg_set(target, GDB_REGNO_TSELECT, i);
-			riscv_reg_set(target, GDB_REGNO_TDATA1, 0);
-			r->trigger_unique_id[i] = -1;
-			LOG_TARGET_DEBUG(target, "Stop using resource %d for bp %d",
-				i, unique_id);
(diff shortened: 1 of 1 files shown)


en-sc (approved):
LGTM (reviewed internally).

src/target/riscv/riscv.c
```diff
@@ -1669,22 +1669,38 @@ static int remove_trigger(struct target *target, int unique_id)
...
 
-	riscv_reg_set(target, GDB_REGNO_TSELECT, tselect);
+	result = riscv_reg_set(target, GDB_REGNO_TSELECT, tselect);
+	if (result != ERROR_OK) {
+		LOG_TARGET_ERROR(target,
+				"Could not restore tselect after bp %d removal. "
+				"Target is in a potentially unrecoverable state!",
+				unique_id);
+		return ERROR_FAIL;
+	}
```
JanMatCodasip, line 1703: @aap-sc Thank you for making this part of the code more robust.

I think this can be sent to review.openocd.org in its current form since it improves the situation.

Alternatively, if you like, you may consider an improvement: Try to clear **all** triggers with the given `unique_id` and remember any errors. Only after all triggers were visited, print the error messages:

```
// Pseudocode:

bool found = false;
bool error = false;
for (0 .. trigger_count - 1) {
    if (unique_id does not match)
        continue;

    found = true;

    if (riscv_set_reg(TDATA1, 0) != ERROR_OK) } {
        error = true;
    }
}

if (!found) {
    LOG_TARGET_ERROR(...);
}

if (error) {
    LOG_TARGET_ERROR(...);
}
```

PR #1297, opened 2025-10-08
target/riscv: fix potential UB reported by ubsan in ac cache lookup

aap-sc:
when running OpenOCD built with ubsan enabled I've encountered the following error message:

```
riscv-013.c:204:9: runtime error: null pointer passed as argument 2, which is declared to never be null
```

This is caused by a NULL pointer passed to bsearch function when abstract command cache is still empty. This behavor was introduced in ab97974d1b0d65bde3263596a27cfdbecda6c280


JanMatCodasip:
Thanks. LGTM and can be sent to review.openocd.org.

PR #1299, opened 2025-10-15
target/riscv: fix ub during instruction decode

aap-sc:
A left shift operation caused an implicit integer promotion, triggering the following UBSan error:

```
left shift of 254 by 24 places cannot be represented in type 'int'
```

NOTE: it seems that this code won't work correctly with BE targets, however this is a general problem of the whole implementation anyway.


en-sc (approved):
LGTM (reviewed internally).

JanMatCodasip:
@aap-sc, this looks all right and can be submitted to review.openocd.org, if you like. Thanks.

Issue #1264, opened 2025-06-12
Functions csr_read_progbuf() et al. should exit early if there is no program buffer


JanMatCodasip:
These functions currently do not exit early if program the program buffer in the target is insufficient:

- `csr_read_progbuf()`
- `fpr_read_progbuf()`
- `csr_write_progbuf()`
- `fpr_write_progbuf()`

Because of that, `riscv_program_insert()` is attempted and extra error messages get displayed which may be confusing to the user:

```
Error: [target_name] Unable to insert program into progbuf, capacity would be exceeded (progbufsize=0).
```

Steps to reproduce:

- Use a target with progbufsize == 0
- Make an attempt to access a non-existent CSR (this also happens out of the box during the examination - for example the read of VLENB CSR on targets that do not implement the vector extension).

Fix this by exiting early from the above functions if the program buffer is not large enough. Use function `has_sufficient_progbuf()`.

ssampath:
Hi,

I am interested in contributing to riscv-openocd.  Appreciate if anyone can guide me through the steps and I will document the steps that I followed to get productive so that it will be helpful for other 'newcomers' like myself.

Thank you

PR #1300, opened 2025-10-26
necessary tests for progbufsize | Skip CSR read if insufficient progbuf in RISC-V targets

Biancaa-R:
This PR resolves this issue https://github.com/riscv-collab/riscv-openocd/issues/1264
These functions currently do not exit early if program the program buffer in the target is insufficient:

    csr_read_progbuf()
    fpr_read_progbuf()
    csr_write_progbuf()
    fpr_write_progbuf()
 This pr resolves this issue using the pre existing function:
`if (!has_sufficient_progbuf(target, 2))`

Add a check in CSR read functions to skip the operation when the target's
progbuf is smaller than required. This prevents errors on targets with
limited progbuf size and ensures safer debugging operations.

- Added `has_sufficient_progbuf` checks to CSR read functions.
- Log debug message when skipping due to insufficient progbuf.
- Improves compatibility with smaller RISC-V targets.

image
* checks for the min size

Successful configuring:
image
Successful build ,after build the version was also tested to check working:
image
with `./src/openocd --version`

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -1385,6 +1385,11 @@ static int fpr_read_progbuf(struct target *target, uint64_t *value,
 {
 	assert(target->state == TARGET_HALTED);
 	assert(number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31);
+	if (!has_sufficient_progbuf(target, 2)) {
+		LOG_DEBUG("Skipping FPR read: insufficient progbuf (size=%d)",
+		          get_info(target)->progbufsize);
+		return ERROR_FAIL;
+	}
 
 	const unsigned int freg = number - GDB_REGNO_FPR0;
 
@@ -1417,6 +1422,11 @@ static int csr_read_progbuf(struct target *target, uint64_t *value,
 {
 	assert(target->state == TARGET_HALTED);
 	assert(number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095);
+	if (!has_sufficient_progbuf(target, 2)) {
+		LOG_DEBUG("Skipping CSR read: insufficient progbuf (size=%d)",
+		          get_info(target)->progbufsize);
+		return ERROR_FAIL;
+	}
 
 	if (riscv013_reg_save(target, GDB_REGNO_S0) != ERROR_OK)
 		return ERROR_FAIL;
@@ -1484,6 +1494,11 @@ static int fpr_write_progbuf(struct target *target, enum gdb_regno number,
 {
 	assert(target->state == TARGET_HALTED);
 	assert(number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31);
+	if (!has_sufficient_progbuf(target, 2)) {
+		LOG_DEBUG("Skipping FPR Write: insufficient progbuf (size=%d)",
+		          get_info(target)->progbufsize);
+		return ERROR_FAIL;
+	}
 	const unsigned int freg = number - GDB_REGNO_FPR0;
 
 	if (riscv013_reg_save(target, GDB_REGNO_S0) != ERROR_OK)
@@ -1559,6 +1574,11 @@ static int csr_write_progbuf(struct target *target, enum gdb_regno number,
 {
 	assert(target->state == TARGET_HALTED);
 	assert(number >= GDB_REGNO_CSR0 && number <= GDB_REGNO_CSR4095);
+	if (!has_sufficient_progbuf(target, 2)) {
+		LOG_DEBUG("Skipping CSR write: insufficient progbuf (size=%d)",
+		          get_info(target)->progbufsize);
+		return ERROR_FAIL;
+	}
 
 	if (riscv013_reg_save(target, GDB_REGNO_S0) != ERROR_OK)
 		return ERROR_FAIL;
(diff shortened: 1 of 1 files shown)


Biancaa-R:
This code built and ran without issues ,In case of any needed modification of functionality ,please let me know ,Ill do it :)

src/target/riscv/riscv-013.c
```diff
@@ -1385,6 +1385,11 @@ static int fpr_read_progbuf(struct target *target, uint64_t *value,
 {
 	assert(target->state == TARGET_HALTED);
 	assert(number >= GDB_REGNO_FPR0 && number <= GDB_REGNO_FPR31);
+	if (!has_sufficient_progbuf(target, 2)) {
+		LOG_DEBUG("Skipping FPR read: insufficient progbuf (size=%d)",
```
en-sc: [here and around] Please, use `LOG_TARGET_DEBUG(...)` to give a bit more context:
```suggestion
		LOG_TARGET_DEBUG(target, "Skipping FPR read: insufficient progbuf (size=%d)",
```
Also `progbufsize` is unsigned, so the correct format specifier is `%u` (though this is undoubtedly a nitpick).

en-sc (changes requested):
@Biancaa-R, thank you for submitting the patch!

Please, take a look at Checkpatch failures.
The first `ERROR:COMMIT_MESSAGE:` can be quite cryptic -- the commit message is split into the title and the description with an empty line in-between.
Please see https://git-scm.com/docs/git-commit#_discussion for more.

The other errors seem descriptive enough.

You can run the script locally with:
```
> tools/scripts/checkpatch.pl -g HEAD
```

Also, what about `vtype_write_progbuf()`/`vl_write_progbuf()`?

en-sc:
Also, if you have the time please check #1301 -- I believe it is a better approach to the issue.
It will allow to drop the magic constant from the calls to `has_sufficient_progbuf()` and to replace them with something like `has_progbuf()`.
In any case this patch is an improvement by itself and can be merged without #1301.

Biancaa-R:
> @Biancaa-R, thank you for submitting the patch!
> 
> Please, take a look at Checkpatch failures. The first `ERROR:COMMIT_MESSAGE:` can be quite cryptic -- the commit message is split into the title and the description with an empty line in-between. Please see https://git-scm.com/docs/git-commit#_discussion for more.
> 
> The other errors seem descriptive enough.
> 
> You can run the script locally with:
> 
> ```
> > tools/scripts/checkpatch.pl -g HEAD
> ```
> 
> Also, what about `vtype_write_progbuf()`/`vl_write_progbuf()`?

Sure I'll check for those errors. In the issue only csr_read_progbuf()
fpr_read_progbuf()
csr_write_progbuf()
fpr_write_progbuf()

Were mentioned, sure I'll add those 2 as well.

src/target/riscv/riscv-013.c
Biancaa-R: Changed it in the updated version.

src/target/riscv/riscv-013.c
```diff
@@ -1536,6 +1554,11 @@ static int vtype_write_progbuf(struct target *target, riscv_reg_t value)
 static int vl_write_progbuf(struct target *target, riscv_reg_t value)
 {
 	assert(target->state == TARGET_HALTED);
+	/* Ensure program buffer is large enough for 2 instructions */
+	if (!has_sufficient_progbuf(target, 2)) {
```
```suggestion
	if (!has_sufficient_progbuf(target, 3)) {
```

The sequence is:
```
csrr ...
vsetvl ...
ebreak
```

src/target/riscv/riscv-013.c
```diff
@@ -1515,6 +1528,11 @@ static int fpr_write_progbuf(struct target *target, enum gdb_regno number,
 static int vtype_write_progbuf(struct target *target, riscv_reg_t value)
 {
 	assert(target->state == TARGET_HALTED);
+	/* Ensure program buffer is large enough for 2 instructions */
+	if (!has_sufficient_progbuf(target, 2)) {
```
```suggestion
	if (!has_sufficient_progbuf(target, 3)) {
```
(Same as for `vl`.)

en-sc (changes requested):

src/target/riscv/riscv-013.c
Biancaa-R: Oh as For vector register (v1) or vector type (vtype) operations we need to see 3 instructions ,right !
Ill fix it .Thanks @en-sc

Biancaa-R:
Sorry about the indentation for comments, I used 
image
To fix indentation issues. I hope now its in the way you have expected .
and ofc it runs:
image

src/target/riscv/riscv-013.c
```diff
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 
 /*
- * Support for RISC-V, debug version 0.13, which is currently (2/4/17) the
- * latest draft.
- */
+	* Support for RISC-V, debug version 0.13, which is currently (2/4/17) the
```
en-sc: This is not needed, it's okay to have a space here.

Biancaa-R:
Made changes to fix checkpatch warnings.

src/target/riscv/riscv-013.c
en-sc: Now the space is missing)

Biancaa-R: > Now the space is missing)

Ok Ill add it back 🦅

JanMatCodasip:
@Biancaa-R thank you for the patch.

If you are interested in continuing with this change (or even https://github.com/riscv-collab/riscv-openocd/issues/1301, as pointed out by @en-sc), please submit a patch to the vanilla OpenOCD: review.openocd.org.

Also your current code contains a lot of unrelated whitespace changes. Please be sure to remove those before sumitting. Thanks.

Issue #1284, opened 2025-08-18
Memory leak in `riscv_openocd_step_impl()`


en-sc:
As reported by Marc Schink in https://review.openocd.org/c/openocd/+/8893/comments/4fc4226b_3a8c92bb there is a leak in `riscv_openocd_step_impl()`:

The leak was introduced in commit 5a8697b3cf5c4e4abde77ac1726a76877592dfa9 ("target/riscv: manage triggers available to OpenOCD for internal use"). I don't quite understand how we have missed it during the review of #1111

The RISC-V Debug Tests should have caught it when run under Valgrind/Address Sanitizer.

The task is twofold:
1. Implement a fix.
2. Investigate why the bug was missed during testing. Adjust the testsuite if needed.

en-sc:
There is a patch proposed to the mainline OpenOCD that should fix this. See https://review.openocd.org/c/openocd/+/9163

PR #1290, opened 2025-09-12
Merge up to 88b9bd396d5a1f83c49fa1b28745fd7feaca2b2e from  upstream

en-sc:
Conflicts:

* `configure.ac`: commit bd5649dfbe10a79f72bee2ee48cf24c10de7b5d8 ("Support cJTAG JScan3 mode") and commit 7d0e125896291e7683f76aa1ad10532bb192692b ("jtag/drivers: Add support for CH347-based JTAG adapters")
* `src/jtag/drivers/ftdi.c`: commit f61098c2258c2ca81fd7c0ed943c36f6a82c5d49 ("Rename ftdi_oscan1 to ftdi_cjtag") and commit 867611d3e19502d3affa780a38de3b41201ee712 ("jtag: drivers: align switch and case statements") -- resolved by aligning the statements.
* `src/rtos/freertos.c`: commit 50a5971be27ff11061776afc8014bbff22870b7c ("RISC-V Freertos support (#582)") and commit commit 3954896d6e41298efd74e8bb7af60ec49efae809 ("rtos/FreeRTOS: fix next pointer member offset in FreeRTOS lists") -- dropped the upstream change since the offsets are dynamically computed by `freertos_compute_offsets()`.
* `src/rtos/freertos.c`:  due to commit b3b790e4e01331f79b196f8312193ae0130d2394 ("rtos: rework rtos_create()") -- the change is already adopted.
* `src/rtos/hwthread.c`: commit 95cb368095695fc233282264532704c7141d26ed ("Merge up to 1f3f635693a1ddc85f362dc324cb49c3e7b75f27 from upstream") and commit c545b9c4ab32c5566091dfe05e7f1765c39ed069 ("rtos/hwthread: use printf format specifier") -- resolved in favour of upstream version.
* `src/rtos/rtos.c`: commit 737f013ada404fdd620c2c0d4f1cb318a0ff64b2 ("Support RV32/RV64 mainline/metal stackings (#586)") and commit b3b790e4e01331f79b196f8312193ae0130d2394 ("rtos: rework rtos_create()") -- adjusted the handlers, adopted the error checks.
[...]

PR #1303, opened 2025-10-27
Fix: Memory leak in riscv_openocd_step_impl() 

Biancaa-R:
cleared wps_to_enable in _exit after there is no further reference to it .
The memory allocated at :
image

```
_exit:
	if (enable_watchpoints(target, wps_to_enable) != ERROR_OK) {
		success = false;
		LOG_TARGET_ERROR(target, "Failed to re-enable watchpoints "
				"after single-step.");
	}
	free(wps_to_enable);
	//no longer needed to be kept
```

Fixes the issue: https://github.com/riscv-collab/riscv-openocd/issues/1284

* Working on the Valgrind part ,will add soon .

### DIFF
--- a/.github/workflows/spike-openocd-tests.yml
+++ b/.github/workflows/spike-openocd-tests.yml
@@ -105,7 +105,15 @@ jobs:
           git clone "$RISCV_TESTS_REPO"
           cd riscv-tests
           git checkout "$RISCV_TESTS_REV"
-          git submodule update --init --recursive
+          git submodule update --init --recursive      
+          
+      - name: Run basic debug test under Valgrind (memory leak check)
+        run: |
+          sudo apt-get install -y valgrind
+          echo "Running RISC-V debug test under Valgrind for memory leak detection..."
+          # Run a minimal configuration to detect leaks in riscv_openocd_step_impl()
+          valgrind --leak-check=full --error-exitcode=1 \
+            ./src/openocd -f tcl/interface/jtag_dummy.cfg -f tcl/target/riscv.cfg -c "init; shutdown"
 
       - name: Run Spike32 Tests
         id: spike32-tests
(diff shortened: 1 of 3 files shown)


en-sc:
@Biancaa-R, thank you for the patch.

Unfortunately, there is already a patch with a fix. I should have mentioned it in the issue, sorry.
See https://review.openocd.org/c/openocd/+/9163

en-sc:
> Working on the Valgrind part ,will add soon.

Are you referring to introducing test runs under Valgrind to the CI?

Biancaa-R:
> > Working on the Valgrind part ,will add soon.
> 
> Are you referring to introducing test runs under Valgrind to the CI?

Yes I am ,to ensure it is caught .

Biancaa-R:
> @Biancaa-R, thank you for the patch.
> 
> Unfortunately, there is already a patch with a fix. I should have mentioned it in the issue, sorry. See https://review.openocd.org/c/openocd/+/9163

Oh no worries :)

Biancaa-R:
Ok let me check and get back.

Biancaa-R:
If this approach also doesnt work ,I would probably :
Run Valgrind only through the internal TCL interpreter

If our goal is just to check core memory management (no targets at all):

valgrind --leak-check=full --error-exitcode=1 \
  ./src/openocd -c "shutdown"

Or use a dummy config file additionally.

en-sc:
@Biancaa-R, I'd suggest not to spend too much time on this one right now. We are currently merging RISC-V OpenOCD with the mainline and after this work is done, I'd like to stop the development in this repo, posting patches to https://review.openocd.org instead.
So, the workflow will need to be adjusted (mainline OpenOCD uses Gerrit and Jenkins).

As for what can be run under valgrind -- there is the `make check` and the tests from `testing/tcl_commands` that are already part of the mainline but are not yet merged into the fork. See 1b2a2b818573269e9704e1e117424210941de3c3 and #1290

JanMatCodasip:
> en-sc: I'd suggest not to spend too much time on this one right now. (...)

@Biancaa-R, @en-sc, should this PR be closed, please?

en-sc:
I think it's okay to drop this PR unless @Biancaa-R objects.

Biancaa-R:
> I think it's okay to drop this PR unless @Biancaa-R objects.

Yes , you can as there is already a fix :)

PR #1304, opened 2025-10-28
target/riscv: fix progbuf memory writes in case last write is busy

en-sc:
Restarting the program buffer memory write pipeline when the write of the last element resulted in the busy response triggers an extra memory wrtite, that is cought by an assertion:
```
src/target/riscv/riscv-013.c:5048: write_memory_progbuf_inner: Assertion
`next_addr_on_target - args.address <= (target_addr_t)args.size *
args.count' failed.
```

Change-Id: I0f27145cad24686cf539aebfea7f6578b7cd78ab

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -4873,7 +4873,8 @@ static int write_memory_progbuf_teardown(struct target *target)
  * failed write.
  */
 static int write_memory_progbuf_handle_busy(struct target *target,
-		target_addr_t *address_p, uint32_t size, const uint8_t *buffer)
+		target_addr_t *address_p, target_addr_t end_address, uint32_t size,
+		const uint8_t *buffer)
 {
 	int res = riscv013_clear_abstract_error(target);
 	if (res != ERROR_OK)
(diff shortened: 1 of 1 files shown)


tom-van:
@en-sc Fantastic speed of fixing!
```
gd32vf103:
:::::::::::::::::::::::::::[ ran 73 tests in 142s ]:::::::::::::::::::::::::::
25 tests returned not_applicable
48 tests returned pass
```
I confirm the fix works. Thanks!

src/target/riscv/riscv-013.c
```diff
@@ -4889,8 +4890,12 @@ static int write_memory_progbuf_handle_busy(struct target *target,
 	if (register_read_direct(target, &address_on_target, GDB_REGNO_S0) != ERROR_OK)
 		return ERROR_FAIL;
 	const uint8_t * const curr_buff = buffer + (address_on_target - *address_p);
```
aap-sc, line 4892:
```
buffer + (address_on_target - *address_p)
```

While we are at it - could you please make some validation of the address read from target? C language has some rules with respect to how pointer arithmetic works (the offset should be inside allocated memory region + 1, I believe) - it would not hurt to comply.

en-sc: I'd like to make it a separate PR since the same issue is present for memory read via PB. See #1311

aap-sc (approved):

en-sc:
Mainline patch: https://review.openocd.org/c/openocd/+/9233

Issue #1308, opened 2025-11-05
No need for busy delays after a DMI NOP.


en-sc:
When running a NOP, there is no need to add extra RTI cycles -- according to the spec it should not cause a DMI busy status. See [6.1.5. Debug Module Interface Access (dmi, at 0x11)], `op` field description:
> 0 (nop): Ignore data and address. Don’t send anything over the DMI during Update-DR. This operation should never affect DMI busy or error status.
The address and data reported in the following CaptureDR are undefined.

Therefore, instead of:
```
Debug: 147 2 batch.c:291 riscv_batch_run_from(): [riscv.cpu] Running batch of scans [0, 2)
Debug: 148 3 batch.c:255 log_batch(): 41b w 00000001 @10 -> + 00000000 @00; 3i
Debug: 149 3 batch.c:199 log_dmi_decoded(): write: dmcontrol=1 {dmactive=active}
Debug: 150 3 batch.c:255 log_batch(): 41b - 00000000 @00 -> + 00000001 @10; 3i
```
the sequence of:
```
Debug: 147 2 batch.c:291 riscv_batch_run_from(): [riscv.cpu] Running batch of scans [0, 2)
Debug: 148 3 batch.c:255 log_batch(): 41b w 00000001 @10 -> + 00000000 @00; 3i
Debug: 149 3 batch.c:199 log_dmi_decoded(): write: dmcontrol=1 {dmactive=active}
Debug: 150 3 batch.c:255 log_batch(): 41b - 00000000 @00 -> + 00000001 @10; 0i
```
can be issued.

This will result in a speed-up of `rti / ((5 + 32 + 2 + abits + rti) * 2)` (with `abits = 7` and `rti = 5` it's about 5%).

However, this should be done with caution, considering:
* An abstract command can be running.
* The delay derivation algorithm can require adjustments.

Biancaa-R:
@en-sc  I had added a simple fix ,but what abstract command should I look into ?

en-sc:
@Biancaa-R, thanks! I've answered in the PR to keep the discussion in one place.

PR #1312, opened 2025-11-07
riscv: add batch.c and batch.h for skip-rti-nop sequence handling

Biancaa-R:
This patch introduces batch.c and batch.h to implement the skip-rti-nop mechanism for RISC-V DTM transactions. The new logic optimizes DMI command batching by skipping redundant RTI and NOP sequences, improving overall debug interface efficiency.

Files added:
- src/target/riscv/batch.c
- src/target/riscv/batch.h

Fixes the issue : https://github.com/riscv-collab/riscv-openocd/issues/1308

Reading the value of batch->fields->outvalue before setting delay : 
image

Also runs without issue:
image

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -11,7 +11,7 @@
 #include "field_helpers.h"
 
 // TODO: DTM_DMI_MAX_ADDRESS_LENGTH should be reduced to 32 (per the debug spec)
-#define DTM_DMI_MAX_ADDRESS_LENGTH	((1<<DTM_DTMCS_ABITS_LENGTH)-1)
+#define DTM_DMI_MAX_ADDRESS_LENGTH	((1 << DTM_DTMCS_ABITS_LENGTH) - 1)
 #define DMI_SCAN_MAX_BIT_LENGTH (DTM_DMI_MAX_ADDRESS_LENGTH + DTM_DMI_DATA_LENGTH + DTM_DMI_OP_LENGTH)
 
 #define DMI_SCAN_BUF_SIZE (DIV_ROUND_UP(DMI_SCAN_MAX_BIT_LENGTH, 8))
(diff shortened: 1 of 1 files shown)


src/target/riscv/batch.c
```diff
@@ -295,14 +295,24 @@ int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
...
 
-		delay = get_delay(batch, i, delays, resets_delays,
-				reset_delays_after);
-		if (delay > 0)
-			jtag_add_runtest(delay, TAP_IDLE);
+		//conditional execution
+		const unsigned int out_op = buf_get_u32(batch->fields->out_value,
+					DTM_DMI_OP_OFFSET, DTM_DMI_OP_LENGTH);
+
+		if (out_op == DTM_DMI_OP_NOP) {
```
en-sc, line 306: I'd suggest to keep the logic around delay times inside `get_delay()`. The main reason -- here the operations are scheduled, but the results are reported separately, in the `log_batch()` function. With the current implementation the delay will be logged incorrectly.

src/target/riscv/batch.c
```diff
@@ -295,14 +295,24 @@ int riscv_batch_run_from(struct riscv_batch *batch, size_t start_idx,
 	for (size_t i = start_idx; i < batch->used_scans; ++i) {
 		if (bscan_tunnel_ir_width != 0)
 			riscv_add_bscan_tunneled_scan(batch->target->tap, batch->fields + i,
-							batch->bscan_ctxt + i);
+					batch->bscan_ctxt + i);
```
en-sc, line 298: This does not seem like a correct formatting fix -- AFAIK, throughout the codebase the prevailing pattern is +2 TABs when inside the braces of a function call, e.g.:
https://github.com/Biancaa-R/riscv-openocd/blob/f9ef41bf7622d65a14066056eda3876a0b9d29cc/src/target/target.c#L738-L739

src/target/riscv/batch.c
```diff
@@ -11,7 +11,7 @@
 #include "field_helpers.h"
 
 // TODO: DTM_DMI_MAX_ADDRESS_LENGTH should be reduced to 32 (per the debug spec)
-#define DTM_DMI_MAX_ADDRESS_LENGTH	((1<<DTM_DTMCS_ABITS_LENGTH)-1)
+#define DTM_DMI_MAX_ADDRESS_LENGTH	((1 << DTM_DTMCS_ABITS_LENGTH) - 1)
```
en-sc, line 14: Let's separate formatting fixes from functional once for the sake of a sane `git blame`.

en-sc (changes requested):
Thanks for the patch!

Regarding the part about Abstract Commands -- i've mistaken, it's irrelevant here -- a NOP can not start an Abstract Command.

src/target/riscv/batch.c
Biancaa-R: @en-sc so you want me to add this same logic but inside get_delay() right? so the timing logs and scheduling stay consistent.

en-sc: Yes

JanMatCodasip:
The overall improvement looks correct to me from the logic point of view.

The actual coding needs to be fixed, as @en-sc suggests.

@Biancaa-R, in case you would like to continue with this change, I suggest to send a patch to review.openocd.org instead.

Issue #1309, opened 2025-11-05
Assertion failure when force-reading a dirty register


en-sc:
Steps to reproduce:
1. Run Spike:
```
> spike --rbb-port=7777 ~/openocd/boot_spike64.elf
```
2. Run OpenOCD:
```
> cat spike.cfg
adapter driver remote_bitbang
remote_bitbang port 7777
remote_bitbang host 127.0.0.1

set _CHIPNAME riscv
jtag newtap $_CHIPNAME cpu -irlen 5

set _TARGETNAME $_CHIPNAME.cpu
target create $_TARGETNAME riscv -chain-position $_TARGETNAME -rtos hwthread
> openocd \
  -f spike.cfg \
  -c init \
  -c halt \
  -c 'reg a0 0' \
  -c 'reg a0 force' \
  -c shutdown
Open On-Chip Debugger 0.12.0+dev-02029-geb01c632a (2025-11-05-14:51)
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
Warn : DEPRECATED: auto-selecting transport "jtag". Use 'transport select jtag' to suppress this message.
Info : [riscv.cpu] Hardware thread awareness created
Info : Initializing remote_bitbang driver
Info : Connecting to 127.0.0.1:7777
Info : remote_bitbang driver initialized
Info : Note: The adapter "remote_bitbang" doesn't support configurable speed
Info : JTAG tap: riscv.cpu tap/device found: 0xdeadbeef (mfg: 0x777 (Fabric of Truth Inc), part: 0xeadb, ver: 0xd)
Info : [riscv.cpu] datacount=2 progbufsize=2
Info : [riscv.cpu] Examined RISC-V core
Info : [riscv.cpu]  XLEN=64, misa=0x800000000014112d
[riscv.cpu] Target successfully examined.
Info : [riscv.cpu] Examination succeed
Info : [riscv.cpu] starting gdb server on 3333
Info : Listening on port 3333 for gdb connections
riscv.cpu halted due to undefined.
a0 (/64): 0x0000000000000000
openocd: ocd_sources/src/target/riscv/riscv_reg_impl.h:30: riscv_reg_impl_is_initialized: Assertion `reg->valid || !reg->dirty' failed.
Aborted                    (core dumped) build/Release/src/openocd -f spike.cfg -c init -c halt -c 'reg a0 0' -c 'reg a0 force' -c shutdown
```

[...]

PR #1313, opened 2025-11-08
Update riscv_reg_impl.h to add warning before the assert.

Biancaa-R:
* Solves the issue : https://github.com/riscv-collab/riscv-openocd/issues/1309
* Added a warning message incase of force writing a dirty register.
* Helps to see which registers are inconsistent.
* Can later filter by this message if you suspect cache corruption.
* A soft indication if regs are forced before initializaton.

### DIFF
--- a/src/target/riscv/riscv_reg_impl.h
+++ b/src/target/riscv/riscv_reg_impl.h
@@ -27,6 +27,11 @@ static inline bool riscv_reg_impl_is_initialized(const struct reg *reg)
 	assert(reg->arch_info);
 	assert(((riscv_reg_info_t *)reg->arch_info)->target);
 	assert((!reg->exist && !reg->value) || (reg->exist && reg->value));
+	if (!(reg->valid || !reg->dirty)) {
+        LOG_WARNING("riscv_reg_impl_is_initialized: reg cache invalid or dirty (%s)",
+                    reg->name ? reg->name : "unknown");
+        return false;
+    }
 	assert(reg->valid || !reg->dirty);
 	return true;
 }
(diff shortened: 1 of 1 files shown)


Biancaa-R:
So as for the issue raised by @en-sc 

The reg a0 0 / reg a0 force commands expect a fully functional RISC-V GDB server or real hardware.

Spike + remote_bitbang driver in OpenOCD is very minimal — it supports halt/resume, memory access, and basic register reads, but not all “force register write” operations reliably.

Assertion is a bug in OpenOCD, triggered by doing a force write before the register cache is valid.
* So this patch aims to send a warning in case of such an issue ,instead of abruptly stopping riscv -openocd.

Biancaa-R:
Fix errors pointed by CI :
image

en-sc:
@Biancaa-R, I believe this patch does not fix the problem, just hides the symptom.

I should have added a detailed description of the issue when creating it.
Please, take a look https://github.com/riscv-collab/riscv-openocd/issues/1309#issuecomment-3516299224

Biancaa-R:
> @Biancaa-R, I believe this patch does not fix the problem, just hides the symptom.
> 
> I should have added a detailed description of the issue when creating it. Please, take a look [#1309 (comment)](https://github.com/riscv-collab/riscv-openocd/issues/1309#issuecomment-3516299224)

Sure , I'll look into it and get back .

JanMatCodasip:
> en-sc: I believe this patch does not fix the problem, just hides the symptom.

Hi @en-sc, I agree.

I recall I tried to address the root cause of the issue in https://review.openocd.org/c/openocd/+/8070, which waited long time for a review, and then I did not have time to work on that anymore. We can continue any further discussion under https://review.openocd.org/c/openocd/+/8070, if needed.

Let me please close this PR, hope that @Biancaa-R does not mind.

Issue #587, opened 2021-03-27
Loading programs with program buffer expects autoexec support (optional feature according to debug-spec 0.13)


sujay-pandit:
Simple fix in "write_memory_progbuf" in riscv-013.c
A check can be added to confirm if the support exists (by writing then reading back 0x18) and then based on that setting variable "setup_needed" as False in the first if-condition in the for-loop can be removed.

JanMatCodasip:
Hi Sujay,

Thank you for filing this ticket.

I can confirm this is a valid issue. 

All bits in "abstractauto" DM registers are optional per the RISC-V Debug specification, and targets need not support it. However, the current OpenOCD code for both memory read and write via program buffer relies on the autoexecdata[0] bit - without checking if it is really supported.

Side note: Most targets appear to have autoexecdata[0] bit implemented, which is also recommended for performance reasons. That fact perhaps masked this issue in OpenOCD.

sujay-pandit:
Thanks Jan, yes I agree. We have added autoexec support now.

TommyMurphyTM1234:
I take it that this issue is still extant and awaiting a fix?

* https://github.com/riscv/riscv-openocd/blob/a01bd76e5c10eba8bc3481eb7552fb8feab79f91/src/target/riscv/riscv-013.c#L3990

en-sc:
> I take it that this issue is still extant and awaiting a fix?

Yes, it is still relevant. Though IMHO it's a low priority one, since most hardware does support `autoexec` AFAIK.

nadime15:
Is this still an issue? I usually work with the upstream version of OpenOCD, so I am not sure whether it also affects the downstream version.

> Though IMHO it's a low priority one, since most hardware does support autoexec AFAIK.

This might be true, but at the end of the day it is still an optional feature and should not be assumed to exist. It literally took me a full day to figure out why my GDB sessions were spitting out random data. Most people will probably support it, but OpenOCD should still check whether abstractauto is writable before using it. I am sure most people will test features incrementally rather than implementing autoexec right away.

TommyMurphyTM1234:
> OpenOCD should still check whether abstractauto is writable before using it.

Patches and PRs are always welcome. Feel free to submit a PR to address this issue.

nadime15:
Yeah I will take a look!

Issue #653, opened 2021-10-18
abstractauto is used without checking if it is implemented


allexoll:
Abstract auto is used for memory writes without stoping when it is not implemented. the register is not actively read back, but even if it comes back from JTAG, no warning/stopping is done when neither `autoexecdata` or `autoexecprog` is implemented. 

this log (with anotation) shows the current behavior:

```
Debug: 2093 7506 riscv-013.c:1495 register_read_direct(): {0} s1 = 0x4078504
Debug: 2094 7507 program.c:35 riscv_program_write(): debug_buffer[00] = DASM(0x00942023)     // sw s1, 0(s0)
Debug: 2095 7514 riscv-013.c:397 scan(): 66b w 00942023 @20 -> + 04078504 @04; 0i
Debug: 2096 7524 riscv-013.c:397 scan(): 66b - 00000000 @20 -> + 00942023 @20; 0i
Debug: 2097 7524 program.c:35 riscv_program_write(): debug_buffer[01] = DASM(0x00440413)        // addi s0, s0, 4
Debug: 2098 7534 riscv-013.c:397 scan(): 66b w 00440413 @21 -> + 00942023 @20; 0i
Debug: 2099 7542 riscv-013.c:397 scan(): 66b - 00000000 @21 -> + 00440413 @21; 0i
Debug: 2100 7542 riscv-013.c:3436 write_memory_progbuf(): writing until final address 0x000000008000106c
Debug: 2101 7542 riscv-013.c:3439 write_memory_progbuf(): transferring burst starting at address 0x0000000080000000
Debug: 2102 7542 riscv-013.c:2259 log_memory_access(): M[0x80000000] writes 0x800000b7
Debug: 2114 7635 riscv-013.c:787 execute_abstract_command(): command=0x271009; access register, size=32, postexec=1, transfer=1, write=1, regno=0x1009
...
Debug: 2119 7663 riscv-013.c:397 scan(): 66b w 00000001 @18 -> + 02000801 @16; 0i     // writing autoexecdata = 1
Debug: 2120 7663 riscv-013.c:407 scan():  ->  progbufsize=2 datacount=1
Debug: 2121 7671 riscv-013.c:397 scan(): 66b - 00000000 @18 -> + 00000000 @18; 0i      // abstractauto comes back with == 0, telling the debugger the functionality is not implemented
Debug: 2122 7671 riscv-013.c:2259 log_memory_access(): M[0x80000004] writes 0x00808067
[...]

PR #1314, opened 2025-11-12
Add support for batch read/write handling for targets without abstractauto

nadime15:
This is currently just a draft, but this potential PR aims to remove the use of the optional abstractauto register, which enables efficient batch reads and writes, and attempts to address the issue described in https://github.com/riscv-collab/riscv-openocd/issues/587.

I would appreciate any feedback on my approach, as well as suggestions for alternative or improved ways to handle this.

The part I am uncertain about is the handling of dirty registers. I am not entirely sure if my current approach requires modification. While testing, GDB occasionally reported that S0 and S1 were dirty.

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -257,6 +257,12 @@ typedef struct {
 
 	/* This hart was placed into a halt group in examine(). */
 	bool haltgroup_supported;
+
+    /* Tracks which abstractauto.autoexecdata bits are supported by hardware.
+	 * When set, reading/writing the corresponding data register re-executes
+	 * the last abstract command, enabling efficient batch memory operations.
+	 */
+    bool autoexecdata[12];
 } riscv013_info_t;
 
 static OOCD_LIST_HEAD(dm_list);
@@ -2103,6 +2109,33 @@ static int examine(struct target *target)
 
 	LOG_TARGET_INFO(target, "datacount=%d progbufsize=%d",
 			info->datacount, info->progbufsize);
+
+	/* Check which abstractauto.autoexecdata bits are supported by the hardware */
+	uint32_t abstractauto;
+	if (dm_write(target, DM_ABSTRACTAUTO,
+		DM_ABSTRACTAUTO_AUTOEXECDATA << DM_ABSTRACTAUTO_AUTOEXECDATA_OFFSET) == ERROR_OK) {
+		if (dm_read(target, &abstractauto, DM_ABSTRACTAUTO) == ERROR_OK) {
+			uint32_t autoexecdata_mask = get_field(abstractauto, DM_ABSTRACTAUTO_AUTOEXECDATA);
+			// Check each bit individually
+			for(unsigned int i = 0; i < info->datacount; i++){
+				info->autoexecdata[i] = (autoexecdata_mask & (1 << i)) != 0;
+				LOG_TARGET_DEBUG(target, "abstractauto.autoexecdata[%d] = %s",
+                    i, info->autoexecdata[i] ? "supported" : "not supported");
+			}
+
+			dm_write(target, DM_ABSTRACTAUTO, 0);
+
+			if (info->autoexecdata[0] & info->autoexecdata[1])
+				LOG_TARGET_INFO(target, "abstractauto.autoexecdata supported for efficient memory access");
+			else
+				LOG_TARGET_WARNING(target, "abstractauto.autoexecdata not supported - "
+					"program buffer memory access may not work or will be slower");
+		}
+		else
+			LOG_TARGET_DEBUG(target, "Unable to write abstractauto. Register may not be present on this target.");
+	}
+	else
+		LOG_TARGET_DEBUG(target, "Unable to write abstractauto. Register may not be present on this target.");
 
 	info->impebreak = get_field(dmstatus, DM_DMSTATUS_IMPEBREAK);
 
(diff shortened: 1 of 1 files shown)


en-sc:
@nadime15, thanks for the patch!

I'd suggest trying to keep using the batched access. Without batching (one batch per element, like in the current implementation) the resulting speed will be much slower (this is related to USB communication delays).

The idea is to add writes of the `command` register into the batch (see [`read_memory_progbuf_inner_fill_batch()`](
https://github.com/nadime15/riscv-openocd/blob/b38b1b30e1ae5773b613740d4664362a4260a589/src/target/riscv/riscv-013.c#L4232)) -- currently it adds one or two DMI reads of `data*` registers per element. Without `abstractauto` an extra write of `command` is needed. Note that the read of `data0` triggers the next command execution, so it will be logical to add the write of `command` after this read, not before.

This will allow for a better performance and should not be too hard to implement.

Are you using some simulator to test this or is this tested on HW? Please, consider adding an argument to control `abstractauto` support in Spike, so that the patch can be tested.

src/target/riscv/riscv-013.c
```diff
@@ -2104,6 +2110,33 @@ static int examine(struct target *target)
...
 			info->datacount, info->progbufsize);
 
+	/* Check which abstractauto.autoexecdata bits are supported by the hardware */
+	uint32_t abstractauto;
+	if (dm_write(target, DM_ABSTRACTAUTO,
+		DM_ABSTRACTAUTO_AUTOEXECDATA << DM_ABSTRACTAUTO_AUTOEXECDATA_OFFSET) == ERROR_OK) {
+		if (dm_read(target, &abstractauto, DM_ABSTRACTAUTO) == ERROR_OK) {
+			uint32_t autoexecdata_mask = get_field(abstractauto, DM_ABSTRACTAUTO_AUTOEXECDATA);
+			// Check each bit individually
+			for(unsigned int i = 0; i < info->datacount; i++){
```
en-sc, line 2120: ~This seems like quite a bit of work. I'd suggest using `yes_no_maybe` and evaluating the results lazily.
See [`examine_progbuf()`](https://github.com/nadime15/riscv-openocd/blob/b38b1b30e1ae5773b613740d4664362a4260a589/src/target/riscv/riscv-013.c#L1051) for an example.~
Oh, I'm sorry I didn't read the code carefully.

en-sc: Oh, I'm sorry I didn't read the code carefully. Please, disregard my first comment.

I've looked at the spec and from [3.14.8. Abstract Command Autoexec (abstractauto, at 0x18)]:
> If this register is implemented then bits corresponding to implemented progbuf and data registers
must be writable. Other bits must be hard-wired to 0.

It seems like there are only two options -- either all the bits are supported or none of the bits are supported.
Perhaps we can drop the array then?

nadime15: I dont think this is what it means. I think it means that the abstractauto.autoexecdata (and autoexecprogbuf) bits are writable for the data* and progbuf* registers that are implemented, while the others are hardwired to 0.

For example, if an implementation has 4 data registers (data0 through data3), these would be writable (bits 3:0), but the rest (bits 11:4) would be hardwired to 0.

Furthermore, it says:

> When a bit in this field is 1, read or write accesses to the
> corresponding data word cause the DM to act as if the
> current value in command was written there again after
> the access to data completes.

The phrase “When a bit in this field is 1...” implies that each bit can be either 1 or 0. Besides that, I think it would be misleading to have, for example, all 12 bits for autoexecdata supported (writable to 1) when the debug module only has 4 data registers.

If this needs more clarification, I could open an issue on the debug specification repo and ask there as well.

nadime15:
Thanks @en-sc  for the feedback. I will take a look at `read_memory_progbuf_inner_fill_batch()` is there a similar function for writes?

> Are you using some simulator to test this or is this tested on HW? Please, consider adding an argument to control abstractauto support in Spike, so that the patch can be tested.

I am working on debug module for the [Sail RISC-V Model](https://github.com/riscv/sail-riscv). I have not opened a PR yet but I tested it using my [own](https://github.com/nadime15/sail-riscv/tree/debug_all) implementation (which still needs quite a bit of work). So far, it works correctly on my setup, I have tested it under different scenarios, such as when the JTAG module is much faster than the hart (to trigger retries due to the debug module being busy) and vice versa.

But I agree I also would feel more comfortable adding an option to Spike to test it. I open a PR shortly!

en-sc:
> Thanks @en-sc for the feedback. I will take a look at `read_memory_progbuf_inner_fill_batch()` is there a similar function for writes?

Yes, please look at `write_memory_progbuf_fill_batch()`.

> I am working on debug module for the [Sail RISC-V Model](https://github.com/riscv/sail-riscv). I have not opened a PR yet but I tested it using my [own](https://github.com/nadime15/sail-riscv/tree/debug_all) implementation (which still needs quite a bit of work).

This is very cool! Perhaps you can point me at some documentation/examples I can use to try it out?

JanMatCodasip:
@nadime15

Overall, I think it is good idea to add support for memory read/write through program buffer on targets that do not have the support of `abstractauto`. This feature has been missing for long time. Thanks for attempting a patch.

Just in case you would like to continue with the changes:

1) I would suggest to create a patch at review.openocd.org instead. This repo has become stale, and I would like to make steps towards archiving this repo soon.

2) I think we don't need to keep track of all twelve abstractauto.autoexecdata bits. It should be enough to keep track of just whether abstractauto.autoexecdata[0] is supported. The other autoexecdata bits should not really be needed. The algorithms can be always written in such a way that they read or populate the `data0` register as the last one, thereby auto-starting a new abstract command.

nadime15:
@JanMatCodasip It's been quite a while since I last worked with openocd.. I may come back to it at some point in the future, but as of today I no longer remember the details well enough to continue effectively. If anyone would like to take this over, please feel free to do so.

JanMatCodasip:
Hi @nadime15, 

thank you for your reply.

> nadime15: I may come back to it at some point in the future (...) If anyone would like to take this over, please feel free to do so.

Overall, I think it would be good for OpenOCD to support memory access via the program buffer without abstractauto. At the same time, there does not seem to be much demand for it, since no one had implemented it before your PR.

So to summarize:

- I would leave it up to the demand from the community whether to implement this and when.
- If anyone wishes to continue this work, please submit your patches to review.openocd.org instead.

nadime15:
@JanMatCodasip What I vividly remember is that my implementation at the time did not support abstractauto, since it is optional. Because of that, certain GDB commands (e.g., loading an ELF) did not work, as OpenOCD relies on them without warning the user.

I think this is a problematic approach and took me days to actually debug. A tool should not depend on optional features by default. Otherwise, what is the point of having them optional in the first place if they are effectively required?

TommyMurphyTM1234:
> Because of that, certain GDB commands (e.g., loading an ELF) did not work, as OpenOCD relies on them without warning the user.

Maybe these previous issues are relevant? 

- https://github.com/riscv-collab/riscv-openocd/issues/653
- https://github.com/riscv-collab/riscv-openocd/issues/587

JanMatCodasip:
@nadime15, thanks for your reply.

> I think this is a problematic approach and took me days to actually debug. A tool should not depend on optional features by default. 

I agree with your statement.

However, because OpenOCD is community-driven, features are often developed only when someone needs them and is willing to invest the time. So support for a concrete device configuration is not necessarily guaranteed. Contributions are welcome, and anyone can submit a patch to help improve things.

PR #1321, opened 2026-03-16
Align LOG_* messages with upstream style - remove trailing punctuation (#1212)

### LINKED ISSUE #1212: Align LOG_* messages with the upstream style - remove trailing punctuation
The upstream OpenOCD style asks that log messages (printed by LOG_* and LOG_TARGET_* macros) do **not** end with any punctuation mark like `.` or `!`.

See the [C Style Guide](https://openocd.org/doc/doxygen/html/stylec.html) which says: `Do not use a period or exclamation mark at the end of a message.`

The task is to:
- Remove the trailing punctuation in all log messages in `src/target/riscv/*`,
- except for messages that consist of multiple sentences.

habeelali:
Fixes #1212. Removes trailing period/exclamation from log messages in src/target/riscv/* per OpenOCD C style guide. Multi-sentence messages were left unchanged.

### DIFF
--- a/src/target/riscv/batch.c
+++ b/src/target/riscv/batch.c
@@ -55,34 +55,34 @@ struct riscv_batch *riscv_batch_alloc(struct target *target, size_t scans)
 	 * the real DR scan length on the given target) */
 	out->data_out = malloc(sizeof(*out->data_out) * scans * DMI_SCAN_BUF_SIZE);
 	if (!out->data_out) {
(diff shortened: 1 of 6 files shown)


JanMatCodasip:
Hi @habeelali, thanks for your patch.

I assume this stale repository will get archived in near term, so you may implement your improvement in the OpenOCD upstream instead: review.openocd.org.

PR #1324, opened 2026-05-25
target/riscv: handle unavailable harts during reset halt

jerryzj:
Teach the reset-halt path to tolerate harts that temporarily report
unavailable instead of treating them as hard reset failures. Mark
those harts unavailable, defer examination when they cannot be fully
examined, and make the Tcl reset flow skip unavailable or deferred
targets while it waits for halted state.

Keep unavailable state transitions consistent by invalidating register
state, avoiding register flushes on unexamined targets during deinit,
and preserving enough target state for harts to be examined again when
they return.

Update DCSR setup to use progbuf plus DSCRATCH when the debug module
provides scratch storage and enough program buffer space. This avoids
extra abstract register access while still falling back to the existing
direct register path on smaller implementations.

Co-developed-by: cgsfv <cgsfv@users.noreply.github.com>
Signed-off-by: Jerry Zhang Jian <jerry.zhangjian@sifive.com>

### DIFF
--- a/src/target/riscv/riscv-013.c
+++ b/src/target/riscv/riscv-013.c
@@ -2150,19 +2150,19 @@ static int examine(struct target *target)
 	/* Don't call any riscv_* functions until after we've counted the number of
 	 * cores and initialized registers. */
 
-	enum riscv_hart_state state_at_examine_start;
-	if (riscv_get_hart_state(target, &state_at_examine_start) != ERROR_OK)
+	enum riscv_hart_state riscv_state_at_examine_start;
+	if (riscv_get_hart_state(target, &riscv_state_at_examine_start) != ERROR_OK)
 		return ERROR_FAIL;
 
 	/* Skip full examination and reporting of hart if it is currently unavailable */
-	const bool hart_unavailable_at_examine_start = state_at_examine_start == RISCV_STATE_UNAVAILABLE;
+	const bool hart_unavailable_at_examine_start = riscv_state_at_examine_start == RISCV_STATE_UNAVAILABLE;
 	if (hart_unavailable_at_examine_start) {
 		LOG_TARGET_DEBUG(target, "Did not fully examine hart %d as it was currently unavailable, deferring examine.", info->index);
 		target->state = TARGET_UNAVAILABLE;
 		target->defer_examine = true;
 		return ERROR_OK;
 	}
-	const bool hart_halted_at_examine_start = state_at_examine_start == RISCV_STATE_HALTED;
+	const bool hart_halted_at_examine_start = riscv_state_at_examine_start == RISCV_STATE_HALTED;
 	if (!hart_halted_at_examine_start) {
 		if (riscv013_halt_target(target) != ERROR_OK) {
 			LOG_TARGET_ERROR(target, "Fatal: Hart %d failed to halt during %s",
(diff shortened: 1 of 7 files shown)


src/target/riscv/riscv-013.c
```diff
@@ -2954,6 +3012,101 @@ static bool dcsr_ebreak_config_equals_reset_value(const struct target *target)
...
+
+		if (time(NULL) - start > riscv_get_command_timeout_sec()) {
+			riscv013_log_reset_wait_timeout(target, condition, *dmstatus);
+			result = ERROR_TIMEOUT_REACHED;
+			break;
+		}
+
+		wait_done = riscv013_reset_wait_done(condition, *dmstatus);
+		if (!wait_done)
+			usleep(10);
```
en-sc, line 3101: Could you please explain the point of this `usleep()`?

You have the call to `dmstatus_read()`. Internally, it fills the queues a bunch of operations, sends them to execution and then waits for completion. It can either busy-wait or block on a syscall, but anyway it waits.
If there is a point where yielding control from this thread via a `usleep()` could be useful, this point is in the inner loop, internal to `dmstatus_read()` (that is probably absent, since the call is blocking).

So please, remove this `usleep()`. It just confuses the reader without any positive impact on performance.

src/target/riscv/riscv.c
```diff
@@ -724,7 +724,9 @@ static void riscv_deinit_target(struct target *target)
 	if (!tt)
 		LOG_TARGET_ERROR(target, "Could not identify target type.");
 
-	if (riscv_reg_flush_all(target) != ERROR_OK)
+	/* Unexamined targets may still own per-target allocations, but they won't
+	 * have dirty register state to flush. */
+	if (target_was_examined(target) && riscv_reg_flush_all(target) != ERROR_OK)
```
en-sc, line 729: I don't quite get what are you trying to fix here. If there are no dirty register in the cache the `riscv_reg_flush_all()` will not do anything.

If the register cache state is broken, the issue is in the code that breaks the state, not in the code that operates on this broken state.

src/target/riscv/riscv-013.c
```diff
@@ -1688,29 +1691,68 @@ static int set_dcsr_ebreak(struct target *target, bool step)
...
+		riscv_program_insert(&program, csrrs(ZERO, S0, CSR_DCSR));
+		riscv_program_insert(&program, lui(S0, clear_ebreak_bits));
+		riscv_program_insert(&program, csrrc(ZERO, S0, CSR_DCSR));
+		if (step)
+			riscv_program_insert(&program, csrsi(CSR_DCSR, 0x4));
+		else
+			riscv_program_insert(&program, csrci(CSR_DCSR, 0x4));
+		riscv_program_insert(&program, csrr(S0, CSR_DSCRATCH0));
+		if (riscv_program_exec(&program, target) != ERROR_OK)
+			return ERROR_FAIL;
```
en-sc, line 1733: Please consider making this a separate patch where we can discuss whether avoiding the extra read does speed up any real-world scenario (the register is cached on reads).

In general, I'd suggest to separate functional changes from optimizations and style-changes to make the review faster.

en-sc (commented):
Thank you for taking the time to contribute!

1. Please, consider submitting it to the mainline OpenOCD (https://review.openocd.org/). This fork is stale and is not actively maintained. Moreover the RISC-V support is more up-to-date in the mainline.
2. Please, update the commit description (at least for the second commit -- you are talking about the change in poll interaction with reset, but it seems to be mostly cosmetic changes like `state -> riscv_state` rename).
3. Please make an effort to split the patches as small as possible (cosmetics/functional/optimization). From my experience, review in the mainline is very long, and posting simpler smaller patches will help you go through the process quicker.

jerryzj:
> Thank you for taking the time to contribute!
> 
> 1. Please, consider submitting it to the mainline OpenOCD ([review.openocd.org](https://review.openocd.org/)). This fork is stale and is not actively maintained. Moreover the RISC-V support is more up-to-date in the mainline.
> 2. Please, update the commit description (at least for the second commit -- you are talking about the change in poll interaction with reset, but it seems to be mostly cosmetic changes like `state -> riscv_state` rename).
> 3. Please make an effort to split the patches as small as possible (cosmetics/functional/optimization). From my experience, review in the mainline is very long, and posting simpler smaller patches will help you go through the process quicker.

Noted. Thanks for the reply & review.

src/target/riscv/riscv-013.c
JanMatCodasip, line 1733: @jerryzj Thanks for your patch.

Why do you please introduce a second code path (program buffer access) here?

Instead, I would just suggest to:

1. Read the current value using riscv_reg_get().
1. Make the modifications - set and clear bits as needed.
1. Write the value back using riscv_reg_write().

In any case, as @en-sc already mentioned, please submit your changes to the vanilla OpenOCD repository instead: https://review.openocd.org.

PR #1326, opened 2026-06-20
target/riscv: remove RISCV_SCAN_DELAY_MAX

### LINKED ISSUE #1137: Update `RISCV_SCAN_DELAY_MAX` to `UINT_MAX`
https://github.com/riscv-collab/riscv-openocd/blob/3bed4c8015b319b97e6651a383db9ae001f11aa9/src/target/riscv/batch.h#L48-L55

The reason for making `RISCV_SCAN_DELAY_MAX` `INT_MAX / 2` no longer holds --  `jtag_add_runtest()` accepts an `unsigned int` since 0847a4d7fb9809a9bb36ba0965f43cb9d43ca2f3.
https://github.com/riscv-collab/riscv-openocd/blob/3bed4c8015b319b97e6651a383db9ae001f11aa9/src/jtag/core.c#L592

This makes it possible to increase `RISCV_SCAN_DELAY_MAX` and remove/simplify checks.

OpinionatedDeer:
jtag_add_runtest() accepts unsigned int since commit 0847a4d,
so the INT_MAX/2 overflow guard is no longer necessary.
Removed the RISCV_SCAN_DELAY_MAX macro, the assert in
riscv_scan_set_delay(), and the overflow check in
riscv_scan_increase_delay(). The latter is replaced with
a wrapping check to catch genuine unsigned integer overflow.
Closes #1137.

### DIFF
--- a/src/target/riscv/batch.h
+++ b/src/target/riscv/batch.h
@@ -46,13 +46,8 @@ riscv_scan_delay_class_name(enum riscv_scan_delay_class delay_class)
 }
 
 /* The scan delay values are passed to "jtag_add_runtest()", which accepts an
- * "int".  Therefore, the passed value should be no greater than "INT_MAX".
- *
- * Since the resulting delay value can be a sum of two individual delays,
- * individual delays are limited to "INT_MAX / 2" to prevent overflow of the
- * final sum.
+ * "unsigned int". 
  */
-#define RISCV_SCAN_DELAY_MAX (INT_MAX / 2)
 
 struct riscv_scan_delays {
 	unsigned int base_delay;
@@ -82,9 +77,6 @@ riscv_scan_get_delay(const struct riscv_scan_delays *delays,
 static inline void riscv_scan_set_delay(struct riscv_scan_delays *delays,
 		enum riscv_scan_delay_class delay_class, unsigned int delay)
 {
-	assert(delay <= RISCV_SCAN_DELAY_MAX);
-	LOG_DEBUG("%s delay is set to %u.",
-			riscv_scan_delay_class_name(delay_class), delay);
 	switch (delay_class) {
 	case RISCV_DELAY_BASE:
 		delays->base_delay = delay;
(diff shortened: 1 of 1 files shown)


src/target/riscv/batch.h
```diff
@@ -82,9 +77,6 @@ riscv_scan_get_delay(const struct riscv_scan_delays *delays,
 static inline void riscv_scan_set_delay(struct riscv_scan_delays *delays,
 		enum riscv_scan_delay_class delay_class, unsigned int delay)
 {
-	assert(delay <= RISCV_SCAN_DELAY_MAX);
-	LOG_DEBUG("%s delay is set to %u.",
```
JanMatCodasip, line 86: Why have you please removed this LOG_DEBUG() print?

I think it is useful to have the changes of the delays recorded in the debug log for troubleshooting. So unless there is a strong reason, I'd keep that print.

JanMatCodasip:
Hi @OpinionatedDeer,

thank you for your contribution!

Few remarks:

1) **Please submit your improvement to https://review.openocd.org instead**, where the latest RISC-V support is being developed now. I expect to archive this stale riscv-openocd fork soon.

   - Instructions how to submit a patch to upstream OpenOCD can be found at: https://openocd.org/doc/doxygen/html/patchguide.html

2) Overall the code looks all right to me. I only have a single comment about the removed LOG_DEBUG print.

3) Don't forget to check and fix the coding style (using the `checkpatch` script) so that your patch can be accepted upstream.

4) I assume that you have run into a situation where the maximum delay was really reached. Out of curiosity, can you describe how that happened?

OpinionatedDeer:
1. Then i will close the patch here and after changes send it to upstream.
2. Thank you, I will revert the LOG_DEBUG print, there was no actual reason i just forgot to write it back after deletion.
4. No actually, I just made a PR to understand how to work on OpenSource and get some experience with bigger codebases , I am just a UG student and decided to also try for LFX mentorship. I am quite interested in hardware/software boundary so i just decided to commit few simple PR's.
----
If there is anything pending that I could work on please let me know, i think it is good time for me to gain some experience.

